PR #31, opened 2018-12-04
Vagov 641 paragraph migrate poc

beeyayjay:
Migrates content into paragraphs (currently only starred hr and wysiwyg)


andyhawks:
Wow! tremendous work here @beeyayjay , great class files, very well done !

PR #45, opened 2018-12-13
VAGOV-870: Add wildcard proxy DNS to Lando for temporary staging URL.

ElijahLynn:


ElijahLynn:
Merged this one because I need it now to test a deployment.

PR #46, opened 2018-12-13
VAGOV-834 Add Docker files for ECS image containers

mbenziane:


docker/config/certs/lndo.site.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
ElijahLynn: What is the private key doing here?

PR #53, opened 2018-12-20
VAGOV-788: React widget paragraph config

acabouet:


mbenziane:
Can one of the admins verify this patch?

PR #55, opened 2018-12-20
VAGOV-000: Re-exported landing page teaser display config.

beeyayjay:
This fixes a minor glitch that left config/sync/core.entity_view_display.node.landing_page.teaser.yml in an 'update' state even after it was successfully imported using drush cim. To see the problem, run drush cim. It will import successfully. But when you run drush cim again, core.entity_view_display.node.landing_page.teaser.yml still shows up as needing to update.

This patch shouldn't change the landing page teaser at all (not that that matters). It just makes cim recognize that it's in sync.


mbenziane:
Can one of the admins verify this patch?

PR #56, opened 2018-12-21
VAGOV-885 Add script to download STG DB and sync into Lando.

ElijahLynn:
This is an interim solution until we have a systemd timer (cron)
automate backups to a central location where we can sanitize and pull
from.

### DIFF
--- a/scripts/sync-db.sh
+++ b/scripts/sync-db.sh
@@ -1,24 +1,22 @@
 #!/usr/bin/env bash
 
+# Run outside of Lando, from within /scripts directory, e.g. ./sync-db.sh
+
 # Exit immediately if a command fails with a non-zero status code
 set -e
 
 # Setup
 source sync-db-var-config
 mkdir --parents .sql-dumps
-# @todo 4 check if Lando stack is running
+# @todo check if Lando stack is running, if not return error message or just start it
 
-echo "Downloading latest database"
+echo "Downloading latest STG database"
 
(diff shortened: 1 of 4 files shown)


scripts/sync-db.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+# Exit immediately if a command fails with a non-zero status code
+set -e
+
+# Setup
+source sync-db-var-config
```
ElijahLynn: Need to check if file exists and return error message if not.

ElijahLynn:
Interesting build failure though
> Thin Pool has 10077 free data blocks which is less than minimum required 10116 free data blocks. Create more free space in thin pool or use dm.min_free_space option to change behavior

ElijahLynn:
Demo recording of this here > https://zoom.us/recording/play/wBeI7LxmmzXHAYekEYPw9sBS-u2aVUeDSi02qc2HCvzPvJtMkZg_GxGwmx0Q4rxN?continueMode=true

scripts/sync-db.sh
ElijahLynn: Actually, the preceding `set -e` takes care of this well enough.

ElijahLynn:
@mouncifb K, ready for final review.

mbenziane:
Created an environment test-vagovcms#pr-56<br><br> [button_visit-site](http://pr-56.va.agile6.com)

PR #94, opened 2019-01-17
VAGOV-1248

ethanteague:


andyhawks:
@ethanteague on preliminary code review it's worth pointing out a lot was updated in composer.lock without any version changes to composer.json, including drupal core (thx for doing that!), lightning, and behat. 

@beeyayjay or @schiavo if one of you have time, can one of you do a quick run through locally on lando to ensure that all the composer updates haven't introduced any new issues? I  reviewed on the PR site but won't have time to review locally until later today. thx!

ethanteague:
Yep, @andyhawks. I blew out composer.lock when I was testing a drupal-spec-tool add-on that I didn't end up using. We should probably refresh the lock file out more frequently than we do.

PR #102, opened 2019-01-22
VAGOV-1324

ethanteague:


kevwalsh:
the files with content_creator should not have been deleted, IMO.

kevwalsh:
Never mind my last comment. i forgot it was renamed to content_editor.

PR #109, opened 2019-01-24
Vagov 1310 migrate as published

beeyayjay:
Steps to test:
1. Run config sync, if on local
2. Migrate "Import hub pages and paragraphs from va.gov" and "Import basic pages and paragraphs from va.gov"
3. Go to Content page and verify that Moderation State for all nodes is "published".


mbenziane:
Created an environment test-vagovcms#pr-109<br><br> [button_visit-site](http://pr-109.va.agile6.com)

PR #116, opened 2019-01-30
VAGOV-1494 Add test to confirm that roles are associated with the expected permissions

schiavo:


mbenziane:
Created an environment test-vagovcms#pr-116<br><br> [button_visit-site](http://pr-116.va.agile6.com)

PR #136, opened 2019-02-12
VAGOV-1214: va-socks-proxy trigger builds from Drupal

ElijahLynn:
This PR has a change to .lando.yml that will require the ENTIRE engineering team to upgrade to Lando 3.0.0-rc9 or higher, we must coordinate before merging. 

As long as the correct ENV variables are exported, a `lando start|rebuild` will now bring up a fully working SOCKS proxy and will support triggering builds locally. See /README.md for more information.


ElijahLynn:
This was closed because I put a private_key file in here and was trying to kill the PR to buy some time while we removed from our systems (which was done shortly after) and wouldn't let me re-open due to me force pushing the changes (which didn't work anyways). 

New PR is #137

PR #143, opened 2019-02-12
VAGOV-1513: Migrate React Widget paragraphs.

beeyayjay:
There should be a react widget with "Call To Action Widget" checked and only Widget Type of 'rx' on 'VA Prescription Refill and Tracking'.

There should be a react widget with "Call To Action Widget" unchecked and all other fields populated on "How to Apply for VA Health Care".


mbenziane:
Created an environment test-vagovcms#pr-143<br><br> [button_visit-site](http://pr-143.va.agile6.com)

PR #146, opened 2019-02-14
VAGOV-000: Add scheme to internal uri's and format to rich text fields

beeyayjay:
This fixes the problem with internal urls on the site. They still won't link to anything yet, but they won't break the page now. Also fixed a couple of places where the text format wasn't set to rich text. It looked fine in drupal, but graphql was encoding the html.


mbenziane:
Created an environment test-vagovcms#pr-146<br><br> [button_visit-site](http://pr-146.va.agile6.com)

ElijahLynn:
#147 Has updates to this that cause it to pass.

PR #181, opened 2019-03-06
VAGOV-1739 Add test1 preview env, temporary.

ElijahLynn:
This should be done with settings.php and ENV variables etc. but we need
to get this working right away (yesterday) and this works for now.

https://vetsgov-pr-9681.herokuapp.com/preview?nodeId=123 works back to test1, this PR gets us from origin to preview.


mbenziane:
Created an environment test-vagovcms#pr-181<br><br> [button_visit-site](http://pr-181.va.agile6.com)

PR #183, opened 2019-03-07
VAGOV-1681 Create files directory for PRs

mbenziane:
we need to create files dir since docker entry point script will expect it to be present in the case we are not using EFS


mbenziane:
Created an environment test-vagovcms#pr-183<br><br> [button_visit-site](http://pr-183.va.agile6.com)

ElijahLynn:
:+1:

PR #184, opened 2019-03-08
Vagov 1654 metatags

beeyayjay:
Metatag settings.
Note that 'canonical_url' will be set in the metalsmith template so that it uses the domain the site's served from instead of the drupal domain. 
All other metatags should appear in the header section of regular page views in drupal.
J Balboni will update image urls in metatags on the metalsmith side as well, to point to local images.
Metatags for listing pages, e.g. facility news stories, will be done on the metalsmith side. All others are configurable in drupal.


mbenziane:
Created an environment test-vagovcms#pr-184<br><br> [button_visit-site](http://pr-184.va.agile6.com)

PR #185, opened 2019-03-08
VAGOV-1757: VAGOV-1758: VAGOV-1760: VAGOV-1761: Caching / Logging / UI / Fast 404. 

andyhawks:
https://va-gov.atlassian.net/browse/VAGOV-1757
https://va-gov.atlassian.net/browse/VAGOV-1758
https://va-gov.atlassian.net/browse/VAGOV-1760
https://va-gov.atlassian.net/browse/VAGOV-1761

PR #186, opened 2019-03-09
VAGOV-1757 Set Dockerfile-appserver to copy settings.lando.php to new path.

ElijahLynn:
Follow up to #185


mbenziane:
Created an environment test-vagovcms#pr-186<br><br> [button_visit-site](http://pr-186.va.agile6.com)

PR #200, opened 2019-03-13
WIP: VAGOV-1757 Add trusted_host_pattern for ELB IPs.

ElijahLynn:
This was causing the target group to list the instance as "unhealthy"
because the ELB healthcheck was recieving a 400 response and _sometimes_
would restart the container and be in a "draining" mode. Not sure why
this didn't exhibit immediately after deployment but one thing is for
sure, that when an ELB is unhealthy, it just sends traffic to all
containers.


mbenziane:
Created an environment test-vagovcms#pr-200<br><br> [button_visit-site](http://pr-200.va.agile6.com)

andyhawks:
Merge after Sprint Demo on 3/13.

PR #227, opened 2019-03-20
VAGOV-320 Fix fast404 include and add to DEV and TEST/PR

ElijahLynn:
VAEC deploys were breaking with failed includes on STG and PROD but not DEV. Not sure why this wasn't showing up in A6 STG deploy. Also added to TEST/PR ENVs so we can keep functionality the same across ENVs and catch stuff like this earlier in the process.

PR #229, opened 2019-03-20
VAGOV-320 Fix fast404 include path.

ElijahLynn:
Follow up to https://github.com/department-of-veterans-affairs/va.gov-cms/pull/227


ElijahLynn:
Just added to local as well, this wasn't failing on local and made tracking this down hard. I was able to reproduce locally now. Is there a specific case to not have this on local?

PR #247, opened 2019-03-23
VAGOV-2041: Create error message when unwrapping spans loses text.

beeyayjay:
This doesn't actually fix anything, but it reports to dblog when removing bad spans results in a loss of text.

Other than that, there should be no change to the migration.

There's no particular rush to deploy this. We'll want it for making edits post-migration.


andyhawks:
build

PR #259, opened 2019-04-05
VAGOV-000: Remove starter dbs.

kevwalsh:
I don't think we need these db's anymore? They don't appear in any documentation.


andyhawks:
rebuild

ElijahLynn:
rebuild

PR #263, opened 2019-04-08
VAGOV-1814 megamenu poc

acabouet:
This PR adds a custom block type - megamenu - and two paragraph types to manage the site megamenu in Drupal. 

It also installs and configures the corresponding entity reference module to sync promo blocks between megamenu menu items and their corresponding hub landing pages.


ElijahLynn:
rebuild

PR #253, opened 2019-04-03
WIP: VAGOV-1811: converted ParagraphLocalFacilityHealthService -> NodeHealthCareLocalHealthService

ahay-agile6:
+ added patch for GraphQL query entity to return null results

How to test:
Content editors can add local health care service descriptions within regional descriptions

PR #268, opened 2019-04-12
VAGOV-1811

ethanteague:


ethanteague:
Replaces https://github.com/department-of-veterans-affairs/va.gov-cms/pull/253

PR #275, opened 2019-04-16
VAGOV-000 Update permissions in PhpUnit test

schiavo:
Test update to account for changes in permissions.

No Jira issue.


ElijahLynn:
Thanks, this will fail right now until we fix something else we are working on now, I'll rebuild once that is resolved.

ElijahLynn:
This failed because of `503  Service Unavailable [IP: 130.89.148.14 80]"` while building database image, that is a Debian server. We have been seeing these lately, we believe it is because of the TIC and it is intermittent. So we need to build in some redundancy to these image builds. cc @mbenziane

PR #279, opened 2019-04-22
VAGOV-2193 migration GitHub api

beeyayjay:
To test, run the va_benefits_records migration. Expected result: 4 records are imported, all with updated date of Jan 1, 1970.

Note that the migration now depends on 'va_cms_bot_github_auth_token' having a value in settings.php, which in turn depends on there being an environment variable, VA_CMS_BOT_GITHUB_AUTH_TOKEN. 

va_cms_bot_github_auth_token is also used in BuildTriggerForm, so I assume it works, but haven't tested (used local override instead).


andyhawks:
rebuild

ElijahLynn:
Story title needs to be in format of VAGOV-XXX for Jira integration to pick it up.

andyhawks:
rebuild

mbenziane:
rebuild

PR #286, opened 2019-04-28
[WIP] VAGOV-3079 Fix sync-db, sync-files to use private bucket.

ElijahLynn:


ElijahLynn:
#287, accidentally used `hub` twice.

PR #290, opened 2019-04-30
VAGOV-2973: Updating outreach asset config to add field_office.

ethanteague:
This pr adds the existing field_office field to the Outreach asset content type. No changes outside of config are introduced, with one exception: a new path introduced for consumer source to fix broken token from github to populate Drupal user guide.

JIRA: https://va-gov.atlassian.net/browse/VAGOV-2960
There is no qa associated with this task.


ethanteague:
> @ethanteague are the package-lock.json changes to restrict versions of many libraries to specific versions rather than a range intentional here?

Apologies, @andyhawks - that was sloppy. I'm gonna remove the changes I unintentionally made to this file.

PR #296, opened 2019-05-03
VAGOV-3247 remove vagov custom theme

ethanteague:
Github is alerting that the package.json files for this theme are using insecure versions of various packages. We aren't using, so let's just remove.

There is no qa associated with this task.


andyhawks:
rebuild

andyhawks:
@ethanteague  when you're back looks like we need to split your VAGOV-3247 into two PRs, one config PR to deactivate the theme first and a second to uninstall

ethanteague:
This pr is replaced by https://github.com/department-of-veterans-affairs/va.gov-cms/pull/303 and https://github.com/department-of-veterans-affairs/va.gov-cms/pull/304

PR #332, opened 2019-05-21
VAGOV-0000 Update permissions security phpunit test.

schiavo:
Updated roles/permissions configuration test to match current configuration.


andyhawks:
rebuild

ElijahLynn:
This is going to fail on Drupal permissions again.

PR #334, opened 2019-05-22
VAGOV-3561: Exporting config from production.

jonpugh:
It's important to keep an up to date export of all config from production. I'm new here, but it looks like the config/sync folder does not have the latest config export from production.

This PR is only for config export, taken from a site running master branch and production SQL snapshot provided by scripts/sync-db.sh

@ElijahLynn and I discussed this a little bit, there's some confusion because he is doing config exports and getting much smaller diffs.


andyhawks:
rebuild

jonpugh:
Updated the config from a direct export (downloaded tar.gz via production UI)

jonpugh:
@ElijahLynn Seeing a Bad Gateway: http://internal-test-vagovcms-3668-1387755238.us-gov-west-1.elb.amazonaws.com/

kevwalsh:
@jonpugh  there may be another slack or other conversation about this, but if not: the develop branch config has moved very far since the last release to prod. this PR looks like it would undo all that work.

andyhawks:
@jonpugh just seeing the scope of this PR - I thought the scope was only going to be for the manual changes on prod, not the entire config .yml dir. @kevwalsh is correct that merging all of prod's config would undo a lot of work I think we can close this and I don't think we need a PR for the manual config overrides on PR because I believe the latest-and-greatest of that is on stage now as of last stage push, but please confirm @kevwalsh

andyhawks:
closing per discussion at https://dsva.slack.com/archives/CDJQ0JCE4/p1558646180012900

PR #347, opened 2019-05-24
VAGOV-000 records menu migration

beeyayjay:
It would be great if this could get in before the next deploy to PROD, but it's not necessary. If it doesn't make it, we can do manual edits post-migration (note that that includes a config change).

To test:
1. Go to Migrate Benefits Hub (/admin/structure/migrate/manage/va_new_benefits/migrations)
2. Run 'Migrate Records benefits pages from VA.gov' and 'Migrate records sidebar menu'.
3. Go to Content (/admin/content/benefits?title=&moderation_state=editorial-draft&type_1=page). You should see 4 nodes with Owner of 'Records benefits hub', all updated 4/18/2019, and all unpublished.
4. Click on the name of one to view. The breadcrumbs on that page should be identical to the ones on the live site with one exception: the drupal breadcrumbs have an additional 'Get records' crumb with no link (this will be removed by the template).
5. Edit the page. The right sidebar should show a menu entry with Menu link title and Parent item the same as the live site.
6. Go to the Records Benefits Hub menu (/admin/structure/menu/manage/records-benefits-hub). The menu items should match the left nav menu on the live site.


andyhawks:
rebuild

PR #355, opened 2019-05-31
VAGOV-000 Update counts in migration count test.

schiavo:
Update phpunit counts in migration count test.


ElijahLynn:
rebuild

failed because we reached max containers for ec2 instances available
>  was unable to place a task because no container instance met all of its requirements

PR #364, opened 2019-06-04
VAGOV-3888: admin menu improvements. (Sprint 16)

kevwalsh:
QA notes: https://va-gov.atlassian.net/browse/VAGOV-3888


kevwalsh:
rebuild

andyhawks:
rebuild

ElijahLynn:
Failed because the composer.lock file was not updated. Best to always run `composer require ...` to add packages.

ElijahLynn:
Sorry @kevwalsh merged in some composer changes before this. I'll fix the conflict for you on Monday.

kevwalsh:
@ElijahLynn is this stil on your radar for today? if not i'll try to get to it in the morning.

tests/phpunit/SecurityRolesPermissionsTest.php
```diff
@@ -465,8 +465,8 @@ public function expectedPerms() {
           'manage password reset',
           'use moderation dashboard',
           'use text format rich_text',
-          'view sections in toolbar',
           'view any moderation dashboard',
+          'view sections in toolbar',
```
ElijahLynn: Can we clarify this here causing tests to pass? I am seeing that the ordering matters here or am I just not seeing the change here besides ordering?

kevwalsh: @ElijahLynn  seems like it's just an ordering problem yeah.

kevwalsh: i thought i had the same order originally, but i was comparing manually. the comparison is between an array that has line breaks, and one that does not.

ElijahLynn: thanks

PR #367, opened 2019-06-04
VAGOV-4033: Add 2 new environment variables so the CMS can control the devops repo

jonpugh:
Right now you have to go into the jenkins job to configure what branch of the devops repo is used.

This PR simply adds environment variables that we can source in the jenkins job so that the CMS repo can control what version of the devops repo to use.

This would remove the need to add a github tag called "devops", for the main build job could source this file and just use that repo.  (git url and ref supported.)

### DIFF
--- a/docker/env_files/common.env
+++ b/docker/env_files/common.env
@@ -3,3 +3,5 @@ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/var/www/.comp
 LANDO_WEBROOT=/app/docroot
 LANDO_MOUNT=/app
 LANDO_APP_NAME=vagovcms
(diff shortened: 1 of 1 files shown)


ElijahLynn:
So, after testing this in the other PRs, if we use the same branch name in both PRs we can use the branchname from the PR here. And we already have the repo so we can technically close this.

jonpugh:
@ElijahLynn bear in mind this was cherry picked to https://github.com/department-of-veterans-affairs/va.gov-cms/blob/develop/docker/env_files/common.env

I think we need to talk to the team about this, at the very least add to documentation before we call this "closed". 

Right now the way to work on the devops pipeline is tribal knowledge. By having explicit environment variables, it's always clear what is being used. 

Additionally, this variable should really be tags or SHA. Branches are always moving.

PR #381, opened 2019-06-11
VAGOV-4180: Basic behat test.

jonpugh:
This PR is just to get behat PRs running.

- [x] Move "tests/behat" folder to "tests/drupal_spec_tool".
- [x] Add a passing test that doesn't need drupal or drush apis.
- [x] Add base behat config.
- [ ] Setup environment variables in DevOps repo for BEHAT_PARAM so we can set environment URL, path, and drush alias.
- [ ] Add a `smoke-test.feature` file with basic checks for login and admin and content pages.
- [ ] Merge once passing. Do not over-complicated this PR. Additional Behat test will be added.


tests.static.yml
```diff
@@ -0,0 +1,35 @@
...
+
+# va/static/composer_lock:
+#   description: Ensure the composer lock file is up to date
+
+# va/static/composer_update:
+#   description: Ensure the composer libraries are up to date
+
+va/static/behat:
+  description: Behat tests that can be run against code alone.
+  command: behat -dl
```
ElijahLynn: Do these short options have long options? Can we use the long options instead?

jonpugh: Yeah, sure. This is just a PoC but might as well be more descriptive.

ElijahLynn: Thanks, I like long options for all code as they help with self-documentation.

ElijahLynn:
> * [ ]  Setup environment variables in DevOps repo for BEHAT_PARAM so we can set environment URL, path, and drush alias.
> * [ ]  Add a `smoke-test.feature` file with basic checks for login and admin and content pages.

Thanks Jon, are you saying these two above still need to be done before merge?

jonpugh:
Not sure, actually. If this works now, might want to merge it tonight. 

Figuring out the right BEHAT_PARAMs to set and setting it correctly can be tricky, especially with multiple environments, especially since it can't be tested until the web services are up, which takes a really long time right now.

ElijahLynn:
The behat tests shows as va/tests/behat. 
image

Is that intentional and should we rename the rest to that pattern too?

jonpugh:
> Is that intentional and should we rename the rest to that pattern too?

Yes, I like it better and would like to change the rest.

Agree?

jonpugh:
Was this our first automated behat test pass? :D

```
21:03:48   
21:03:48   
21:03:48       1 scenario
21:03:48       (1 passed)
21:03:48   
21:03:48   
21:03:48       3 steps
21:03:48       (3 passed)
21:03:48   
21:03:48   
21:03:48       0m0.75s (14.17Mb)
21:03:48   
21:03:48       Process Succeeded in 1.442s Output: /path/to/file
21:03:48       GitHub Status: va/tests/behat: success
```

jonpugh:
@ElijahLynn Instead of pushing a name change to this PR can we just merge and I'll rename them as we finalize the test suite? I don't see any impact. We aren't configuring per-test forced checks, which is all it is used for beyond human readable names.

ElijahLynn:
Agreed, let's do this in the next PR.

PR #388, opened 2019-06-12
VAGOV-2753: Content locking

kevwalsh:
QA notes: https://va-gov.atlassian.net/browse/VAGOV-2753

### DIFF
--- a/config/sync/content_lock_timeout.settings.yml
+++ b/config/sync/content_lock_timeout.settings.yml
@@ -1,0 +1,4 @@
+content_lock_timeout_minutes: '180'
+content_lock_timeout_on_edit: 1
+_core:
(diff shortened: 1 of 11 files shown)


config/sync/content_lock_timeout.settings.yml
```diff
@@ -0,0 +1,4 @@
+content_lock_timeout_minutes: '10080'
```
andyhawks: 7 days seems problematic for this and likely to introduce editors having to chase people down and break locks, but, ok with merging as-is.

andyhawks:
rebild

config/sync/content_lock_timeout.settings.yml
kevwalsh: i think you're right. i just reduced it to 3 hours, amended the config commit. And fixed the composer.lock conflict

PR #392, opened 2019-06-12
VAGOV-3122: Create staff profile paragraph type, and enable on Detail Page.

kevwalsh:
QA notes for backend config: https://va-gov.atlassian.net/browse/VAGOV-3122

This user story (3122) still needs front end work, including a new fragment for this paragraph type, and a feature flag etc... i'm not sure this can be merged safely without that FE work. It's purely additive in nature to the content model.


andyhawks:
@acabouet or @ahay-agile6 - can one of you review this based on Kev's comment in the description around uncertainty if merging this backend PR will break the FE even though it's additive work? Thanks!

mbenziane:
rebuild

PR #331, opened 2019-05-21
VAGOV-1555 HTTPS: Add private key and certificate paths to Apache config.

ElijahLynn:
Ansible lays these down for COPY. 

Corresponding DevOps PR https://github.com/department-of-veterans-affairs/va.gov-cms-devops/pull/69

PR #401, opened 2019-06-17
VAGOV-3930: SimpleSAMLPHP Configuration. 

andyhawks:
QA instructions are at https://va-gov.atlassian.net/browse/VAGOV-4201.

In reviewing you can ignore everything under /simplesamlphp-1.17.2 and /saml2, those are third-party library files unchanged except for /simplesamlphp-1.1.7.2/config/config.php.

SimpleSAMLPHP 1.17.2 is installed manually outside of composer at project root because: a) our Vendor dir being under docroot is problematic (simplesamlphp needs to be outside of docroot), b) should not be installed via composer because we need symlinks, so the directory needs to be permanent in the codebase and not an artifact.


andyhawks:
@ElijahLynn going to the /simplesaml URL results in ```Fatal error: Uncaught Exception: Unable to load Composer autoloader in /app/simplesamlphp-1.17.2/lib/_autoload.php on line 19 Exception: Unable to load Composer autoloader in /app/simplesamlphp-1.17.2/lib/_autoload.php on line 19 Call Stack: 0.0001 398344 1. {main}() /app/simplesamlphp-1.17.2/www/index.php:0 0.0001 398736 2. require_once('/app/simplesamlphp-1.17.2/www/_include.php') /app/simplesamlphp-1.17.2/www/index.php:3 0.0001 398984 3. require_once('/app/simplesamlphp-1.17.2/lib/_autoload.php') /app/simplesamlphp-1.17.2/www/_include.php:4```. The /app/simplesamlphp-1.17.2/vendor/autoload.php file exists in the codebase and should be found, this tests fine for me locally.  I'm thinking this could potentially be because the /simplesamlphp-1.17.2/vendor dir is getting deleted on site build because it's a vendor dir maybe?

mbenziane:
rebuild

ElijahLynn:
> /app/simplesamlphp-1.17.2/vendor/autoload.php

K, just checked out this branch and don't see in the code. I do see in .gitignore 
> # Ignore Composer-managed dependencies.
> vendor/

Which probably left it out of the commit and would explain why it works locally. I think maybe the above should be changed to be `docroot/vendor`, then the saml/vendor would get picked up.

andyhawks:
rebuild

ElijahLynn:
image

ElijahLynn:
I just merged and pushed https://github.com/department-of-veterans-affairs/va.gov-cms/pull/331/files into here for a manual HTTPS deployment on `test1`. I cancelled the build as it will fail on standard builds till the devops HTTPS code is merged in. Don't run `build` on this right now.

ElijahLynn:
> Need to update the email address to [cms-admin@va.gov](mailto:cms-admin@va.gov)

Or make sure we change it by time it gets to PROD.

ElijahLynn:
> > Need to update the email address to [cms-admin@va.gov](mailto:cms-admin@va.gov)
> 
> Or make sure we change it by time it gets to PROD.

@andyhawks Can we make this email change and then merge this? I think it is good to merge as this is all working on DEV SSOi environment now.

andyhawks:
rebuild

ElijahLynn:
rebuild

ElijahLynn:
Oh that is right, unless the key is laid down by my other devops PR then this will just hang because Apache won't boot with a missing key. We have to pause on merging this and it may be we merge just after migration. The path to the key will change and I'll update it here in this PR when that happens, then we can merge this.

ElijahLynn:
> Oh that is right, unless the key is laid down by my other devops PR then this will just hang because Apache won't boot with a missing key. We have to pause on merging this and it may be we merge just after migration. The path to the key will change and I'll update it here in this PR when that happens, then we can merge this.

Migration part 1 is complete and now we need to migrate to the new CI system, which doesn't use these docker files. That should happen tomorrow.

ElijahLynn:
We need two changes now that migration is done so that we get a passing CI build here:

1. Update cert paths to:

>   SSLCertificateFile "/certs/cms.va.gov.public.crt"
>   SSLCertificateKeyFile "/certs/cms.va.gov.private.key"
>   SSLCertificateChainFile "/certs/cms.va.gov.chain.crt"

2. Revert cert paths at va.gov-cms/docker/config/services/config/php/httpd-ssl.conf to, this part will get CI passing (using Lando containers still):

>     SSLCertificateFile "/certs/cert.crt"
>     SSLCertificateKeyFile "/certs/cert.key"

simplesamlphp-1.17.2/config/authsources.php
```diff
@@ -37,8 +37,8 @@
         'assertion.encryption' => true,
         'sign.logout'          => true,
 
-        'privatekey'           => 'cms.va.gov-private-key.pem',
-        'certificate'          => 'cms.va.gov-public-ca-chain-certificate.crt',
+        'privatekey'           => 'cms.va.gov.private.key',
+        'certificate'          => 'cms.va.gov.chain.crt',
```
ElijahLynn: So, the last time we deployed, the chain was actually incorrect and it was just the certificate, 3 times, and not a real chain. So let's change this to the public.crt, I am pretty sure that is what it needs to be.

PR #409, opened 2019-06-20
VAGOV-4117: Add optional config var to set server (defaults to va.gov)

beeyayjay:
It's not possible to test this yet. No migration scripts have been merged that use the new functionality. 
Tested locally.


andyhawks:
rebuild

PR #429, opened 2019-07-09
VAGOV-4688: Configure new email links on facility pages.

kevwalsh:
QA: 
* Edit node 318 (Pittsburgh Health Care). 
* all fields should be available to support the "Get updates from us" block, and the AX should be decent.


kevwalsh:
I'm going to push another commit to this branch momentarily

PR #431, opened 2019-07-09
VAGOV-000: Make alert to single value, and update landing page AX.

kevwalsh:
QA: 
* Go /node/add/landing_page
* The alert field should be single value, and with a select list (not autocomplete).


andyhawks:
@kevwalsh there are some additional field deletions here, is that intentional?

PR #436, opened 2019-07-10
VAGOV-4770 Update Jenkins job build trigger URL.

ElijahLynn:
Ended up refactoring a bit because it was too confusing before. Simplified it a bit so my brain hurt less.


ElijahLynn:
I'm still testing this by mocking ENV vars when the ENV is ready. Wait for thumbs up to merge.

ElijahLynn:
> I'm still testing this by mocking ENV vars when the ENV is ready. Wait for thumbs up to merge.

K, works for DEV. _should_ be fine for STAGING and PROD too. I can test those too but maybe only do that during business hours, don't want to trigger any alarms. 
image

PR #438, opened 2019-07-10
VAGOV-4891 Update bin paths for .lando.yml, they were moved.

ElijahLynn:
A previous commit moved the bin path and missed lando.


ElijahLynn:
@andyhawks can we merge this now?

PR #447, opened 2019-07-16
VAGOV-4965: Updating office config.

ethanteague:
This mr removes event and publication listing description fields from office, and adds administration field to said content type.


kevwalsh:
QA pass.

PR #453, opened 2019-07-19
VAGOV-5013: Inline images and files.

kevwalsh:
@andyhawks can you review but not merge this? i want to get @ahay-agile6 's take on it too, because of the front end lift coming in Sprint 20. 

QA: 
* Go to /node/add/page and /node/add/health_care_region_detail_page
* On each page, click on "Add content block". Validate that you have options for adding Embedded image or Downloadable file, and they are stored together in a multimedia section of the Paragraphs browser.  
* Add several of each to your node, and try everything you can to break it. There are a _lot_ of options, so it's very hard to write QA steps for every scenario. 

For both content block types:
* the AX should be pretty good. Not perfect, but good. 
* the media browsers are  intuitive. 
* the in-drupal preview (on the node view page) should provide a basic overview of the content. 

Things we definitely want to cover for embeddable image:
* You should see a checkbox that will determine (in the FE) whether the image is clickable or not
* you shouldn't be able to add video or documents, just existing or new images

For Downloadable file:
* You should be able to upload an image or a document, or choose an existing image or document.
* The table-based view display for document in the media browser is not amazing but we will iterate on it.


kevwalsh:
I've updated this to use Media Library instead of Entity browser, which avoids that error, and is more in line with where Drupal is going. 

@andyhawks Do you have any ideas about when we update this site to Drupal 8.7.x? We'll see major improvements with Media Library there, including better support for remote URLs (and video).

PR #458, opened 2019-07-21
5212 alert ax improvements

kevwalsh:
QA: 
* Go to /block/19 on PR environment and compare to /block/19 on any other environment
* the fields should be better organized into sensical field groups
* the help text should make sense
* the _label_ changes to the field settings for `field_is_this_a_header_alert_` should not have not cause any front end build issues! but let's make sure. A second set of eyes will help. i did not change the keys.


kevwalsh:
Rebuild

ElijahLynn:
rebuild

PR #464, opened 2019-07-24
[DO NOT MERGE] Update settings for BRD migration.

ElijahLynn:
This is for ongoing work for the VSP BRD migration.


ElijahLynn:
Branch needed to not be a fork.

PR #469, opened 2019-07-26
VAGOV-4967: Improve TOC help text.

kevwalsh:
QA:
* Go to /node/add/health_care_region_detail_page 
* Check the TOC boolean field, the help text should be good.


kevwalsh:
Thanks for that fix Andy!

PR #471, opened 2019-07-26
VAGOV-000: Adding file download field to staff profile content type.

ethanteague:
**ABOUT**
Relates to https://va-gov.atlassian.net/browse/VAGOV-5022
I thought I could fake the bio by grabbing text and generating pdf with js library. Decided too many unknowns. This approach allows for a user to upload a pdf to drupal file system, which we can path to on frontend with graphql.


kevwalsh:
Yeah this definitely needs to be boring download PDF field.

kevwalsh:
rebuild

andyhawks:
rebuild

kevwalsh:
Looks great to me! Thanks for taking the time to configure the form.

PR #476, opened 2019-07-30
VAGOV-000: Adding missed biography field.

ethanteague:
This adds a bio download field to staff profile type. Missed during initial config.


kevwalsh:
@ethanteague i just reviewed #471 and i saw that download field there?

ethanteague:
> @ethanteague i just reviewed #471 and i saw that download field there?

Thanks @kevwalsh - forgot about that one!

PR #492, opened 2019-08-02
VAGOV-4863: Dismissable alerts.

kevwalsh:
QA:
* Go to /admin/content then click on the blocks tab. It should appear as a table. 
* Edit a block, or add a new one.
* For banner alerts (formerly known as "full width alerts"), you should be able to specify whether or not they are dismissable. 
* The frequency field's help text should be updated to explain its relationship to dismissable.


config/sync/field.field.block_content.alert.field_alert_title.yml
```diff
@@ -10,7 +10,7 @@ field_name: field_alert_title
 entity_type: block_content
 bundle: alert
 label: 'Alert title'
-description: ''
+description: 'This will appear as an h2 a the top of the alert.'
 required: true
```
ethanteague: Got a typo here, @kevwalsh

kevwalsh: Good catch, thanks!

PR #391, opened 2019-06-12
VAGOV-2937: Unify CMS and WEB build processes. DEPRECATED

jonpugh:
## DEPRECATED

Good changes in this PR are being moved to other PRs:

- [ ]  #493

## ORIGINAL ISSUE

There is a way to integrate node/npm projects with Composer.

This PR builds the front-end code as well as Drupal on `composer install`. The front end right now goes into `./docroot/vendor/va/app` and the `post-install-cmd` sets up a symlink to `./app`.  

It uses the `mouf/nodejs-installer` package to actually download `node` and `npm` executables into the bin directory of the project on `composer install`

Then, you can add any `npm` command you want to run after `composer install` by adding it to the `scripts.post-install-cmd` section of `composer.json`.

This will allow us to create unified development environments where a CMS on a branch and a APP instance on another branch can live together.

Versions can be pinned in composer.json, so when you stamp a release of the front-end, you can then pin that version in the CMS codebase, and then with one shot, release the new version of CMS+FE.

This is a great example of how Node and PHP are really starting to play along.

I need assistance getting it fully up and running, and there are some design decisions to make, like what should the front-end app be called? what directory should it go in? Should we consider the option of designing our deployment system to be able to host both?

When I presented on this at DrupalNYC meetup, an audience member said they tweaked the Lando Node container to run as a host, not just as a CLI.

Food for thought!

WIP: This should block builds by standard job?

## UPDATE

This PR now includes code from https://github.com/department-of-veterans-affairs/va.gov-cms/pull/404 so if this is merged, that one will automatically be merged as well.

## Alignment

Here is the most accurate documentation on how to get the project working: https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started

This page is very clear on how to set things up. We should aim to make things as easy as this.

# Update 2:

Determined command to generate HTML, so we can quickly render the front end inside the Drupal site, at `/web` for example:

https://dsva.slack.com/archives/CJ34D8TPB/p1561742139334600?thread_ts=1561741663.333000&cid=CJ34D8TPB

```
npm --no-color run build -- --buildtype=${envName} --asset-source=${assetSource} --drupal-address=${drupalAddress} ${drupalMode}
```

From @jbalboni:

[...]

PR #493, opened 2019-08-02
VAGOV 2937: WEB in CMS

jonpugh:
Simple PR from changes in https://github.com/department-of-veterans-affairs/va.gov-cms/pull/391

- [x] Add `va-gov/web` project.
- [x] Add web build commands to composer install.
- [x] Add `/app` symlink to `va-gov/web` for source code and runnable web application.
- [x] ~~Add `/docroot/web` symlink to the generated assets.
- [x] Add 'composer va:web:build` and customize to use the current site to build the static files and dump into `/docroot/web`. Provide a link to the generated files for ease of use.


jonpugh:
Added new yaml tests locally. 

About to test with github token in `~/.env` file.

jonpugh:
YES

jonpugh:
Behat test steps pass in Lando, @ElijahLynn 
Screenshot from 2019-08-13 11-14-55

ElijahLynn:
The use of the term /app here is confusing because Lando uses that too and the fact that we all call it WEB. Can we change this to use /web instead of /app? I'd rather do that now than change it later.

jonpugh:
Sure, but we need a folder for the static build too.

- /web -> source of `department-of-veterans-affairs/vets-website`
- /static -> rendered HTML.

composer.json
```diff
@@ -82,14 +82,16 @@
...
+        "mouf/nodejs-installer": "^1.0",
+        "npm-asset/yarn": "1.12.3",
         "phpunit/phpunit": "^6",
         "provision-ops/yaml-tests": "dev-composer-argg",
         "querypath/querypath": "^3.0",
         "simplesamlphp/simplesamlphp": "^1.17",
         "symfony/config": "^3.4",
         "symfony/finder": "^3",
         "symfony/phpunit-bridge": "^4.3",
-        "twig/twig": "1.37.1",
```
ElijahLynn: Was twig/twig supposed to be removed?

jonpugh: It's not necessary to define it here. A few other packages require it. It should not be fixed to that version, either, unless there was a specific reason?


I did it because I was trying to `composer install acquia/headless_lightning` and it failed because twig/twig version is incompatible.

Not

ElijahLynn: K, it is specified in `develop` right now. Did you get success with `headless_lightning` install? Is it still needed to remove in this PR?

ElijahLynn:
K, this is all passing for me on local dev now (Lando), the main issue I had was a dirty node/yarn cache (lando ssh; rm -rf /var/www/.cache/yarn/v4/.tmp/0d668c9a6bff1e118fcee948031e888c`. And I also had to wipe out the `web/` folder and `lando composer install` again.

ElijahLynn:
The one thing I think will be confusing to users though is the need to `cp .env-local .env` as tests won't pass locally until they do that. Can we put some help text in that testing step telling them to do that? Then we are good to merge.

ElijahLynn:
The last failure was from me, I was messing with vhosts live on the CI server. Fixed now.

PR #495, opened 2019-08-02
VAGOV-5323 - Fix PeformanceLoginTest failure, decrease threshold from 5 > 2.

ElijahLynn:
Test was failing because SAML local_users login setting excluded the
default authenticated_user role.

image

Added a login before the test to prime the opcache and Drupal cache to
test a real-world login scenario, this means we can test for under 2
seconds now.


schiavo:
@swirtSJW This looks good

PR #497, opened 2019-08-04
VAGOV-1955 - Workflow Participant notifications

ethanteague:
## What this does
Adds `Workflow participants` tab and form to nodes for assigning users that will track workflow changes.
Sends email alerts to users who have been added to node workflow participants forms.
Alerts are only sent if a workflow state change has occurred, and this is triggered by node save.

## Testing Performed
 - Visual
 - Xdebug Profiling
 - Email sending via SMTP (locally)
 - Full functionality test performed on testing instance: http://internal-test-vagovcms-3879-1510337173.us-gov-west-1.elb.amazonaws.com/ 

## Screenshots
workflow-form
confirmation
email

## Definition of done
 - Workflow participants form appears here: `node/{NID}/workflow-participants`
 - Adding a user and saving is successful.
 - Changing a workflow status sends workflow participant an email like the one in the screenshot above.

## To test
 - As an admin, edit the email of the account you are using for testing to an email account you check.
 - As an admin, add the above account to any published node workflow participant form, here:  `node/{NID}/workflow-participants`
 - As an admin, change the workflow state of the above node, and save.
 - Visually verify that you receive an email that is similar in appearance to the screenshot, above.

### DIFF
--- a/docroot/modules/custom/va_gov_notifications/README.md
+++ b/docroot/modules/custom/va_gov_notifications/README.md
@@ -1,0 +1,27 @@
+# VA Notifications
+
+## VA.gov Notifications
+This module sends email alerts to users who have been added to node workflow participants forms.
+Alerts are only sent if a workflow state change has occurred, and this is triggered by node save.
+
+### Workflow Participants form
+An alter has been added to remove the `Editors` fieldset - it is redundant,
+as workflow participants module doesn't pull from va roles to autopopulate fields,
+and doesn't assign any extra perms to users entered in `Editors` or `Reviewers` fields.
+This form can be found on these paths: `node/{NID}/workflow-participants`
+Once a user is added to this form, any time a workflow state has changed, an email
+will be sent to them on node save.
+Once a user is removed from the workflow participants form, they will no longer receive emails when workflow state changes.
+
+### Email data
+```Content title: Make an appointment
+Link: http://va.gov/pittsburgh-health-care/make-an-appointment
+Status Change From: review
+Status Change To: draft
+Log Message: This is a thing.
+Change made by: some.user
+Time changed: 2019-08-04 11:59:04
+```
+
+### Error Reporting
+Email send errors will be reported to Recent Log Messages (`/admin/reports/dblog`)
--- a/docroot/modules/custom/va_gov_notifications/src/WorkflowChangeEmail.php
+++ b/docroot/modules/custom/va_gov_notifications/src/WorkflowChangeEmail.php
@@ -1,0 +1,127 @@
+<?php
+
+namespace Drupal\va_gov_notifications;
+
+use Drupal\Core\Render\Markup;
+
+/**
+ * For sending workflow status change emails.
+ */
+class WorkflowChangeEmail {
+
+  /**
+   * Creates and sends workflow change email.
+   *
+   * @param array $mail_load
+   *   Contains node, recipient and revision author data for creating email.
+   *   Sample array:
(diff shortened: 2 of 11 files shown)


docroot/modules/custom/va_gov_notifications/README.md
```diff
@@ -0,0 +1,27 @@
+# VA Notifications
```
swirtSJW: @ethanteague  Nice Readme.

docroot/modules/custom/va_gov_notifications/src/WorkflowChangeEmail.php
```diff
@@ -0,0 +1,127 @@
...
+      [
+        ':title' => $mail_load['node_title'],
+        ':break' => "\r\n",
+      ]);
+    // Link to the above title.
+    $html .= t(
+      'Link: :url :break',
+      [
+        ':url' => $mail_load['url'],
+        ':break' => "\r\n",
```
swirtSJW: minor:  Since this string is repeated so often, it would be nice to have that as a var.

docroot/modules/custom/va_gov_notifications/src/WorkflowChangeEmail.php
```diff
@@ -0,0 +1,127 @@
...
+    // The name of the user who revised the node.
+    $html .= t(
+      'Change made by: :name :break', [
+        ':name' => $mail_load['change_author']['name'],
+        ':break' => "\r\n",
+      ]
+    );
+    // Now loop through workflow participants.
+    foreach ($mail_load['users'] as $mail) {
+      // Get their language for appropriate translation.
```
swirtSJW: Nice!

docroot/modules/custom/va_gov_notifications/va_gov_notifications.module
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Implements hook_mail().
+ */
+function va_gov_notifications_mail($key, &$message, $params) {
+  // This allows us to translate to users selected lalnguage.
+  $options = [
+    'langcode' => $message['langcode'],
+  ];
+  switch ($key) {
```
swirtSJW: NIce building in extendability.

docroot/modules/custom/va_gov_notifications/src/WorkflowChangeEmail.php
ethanteague: Yeah, when I was writing this I thought about messing around with a template, and realized that didn't make much sense. Then I made a method, and started plugging that in, but it was a big ol thing - I couldn't seem to make it very concise. What you are seeing here is my resignation, lol.

ethanteague:
> This all looks real good Ethan. If we had more time it would've been nice to make it contributable.

Thanks, @swirtSJW !

ElijahLynn:
Somehow the workflow_participants config-import installation caused this error:

> Next Drupal\Core\Database\DatabaseExceptionWrapper: SQLSTATE[HY000]: 
> ┃ General error: 1419 You do not have the SUPER privilege and binary 
> ┃ logging is enabled (you *might* want to use the less safe 
> ┃ log_bin_trust_function_creators variable): DROP TRIGGER IF EXISTS 
> ┃ workflow_participants_der_update; Array 

Created upstream issue here https://www.drupal.org/project/workflow_participants/issues/3073354 with fix of setting `log_bin_trust_function_creators=1` in RDS. This unblocks deploys for now.

PR #511, opened 2019-08-14
VAGOV-5472: Reenable views bulk edit.

kevwalsh:
QA
* as admin, go to /admin/content/bulk
* From the Action dropdown, you should see Modify field values.


kevwalsh:
This is a regression, the module was [inadvertently disabled](https://github.com/department-of-veterans-affairs/va.gov-cms/commit/a1b4810eaa35096ef0a7f9b17719ecd38b60c9e6#diff-9576c5e8cbad8dc78b15a60047d94d91) last week after being enabled the week before.

PR #515, opened 2019-08-15
VAGOV-5520: Adding comments to local patches.

ethanteague:
## What this does
 - Adds comments to local patches explaining that they are not suitable for contrib, as they are project specific.

## QA
 - Visit /patches, and confirm that all three files have a comment above diff explaining why they are local.


ElijahLynn:
>   field.storage.block_content.field_help_text                                         �[31;40m�[1mdelete�[0m
> Import the listed configuration changes? (y/n): y
> 
> Drupal\Core\Config\ConfigImporterException: config_importer is       �[31;40m�[1m[error]�[0m
> already importing in Drupal\Core\Config\ConfigImporter->initialize()
> (line 525 of

@jonpugh hrm, "already importing".

patches/moderationSidebar.patch
```diff
@@ -1,3 +1,4 @@
+Patch is project specific, and not a candidate for drupal contribution.
```
swirtSJW: @ethanteague  These comments are a good start, but it might be helpful to have them describe what the patch does and why it is necessary.   Normally drupal patches match up by number to a drupal issue so it is easy to see the connection to the details of why they exist.  In this case the patches lack any reference to what and more importantly why they exist.   Any chance you can add these details before the reasons are lost to brain fog or personnel change?

ethanteague: Yep - these patches don’t correspond to any Drupal issues, but I can add a more detailed explanation for each of them, @swirtSJW

ethanteague: Okay - I updated the comments on two patches, and decided to make an issue on DO for the third and changed the path in our composer file to point to it.

swirtSJW: Yah that is a goofy situation, but I think you handled it correctly.  Nice work.

PR #517, opened 2019-08-15
VAGOV-5519 behat: migration

swirtSJW:
@jonpugh this is putting your behat test back in that I removed before our deployment to staging.

### DIFF
--- a/tests.yml
+++ b/tests.yml
@@ -6,7 +6,7 @@ va/debug/env:
   description: Debug Environment
   command: env
 
-# TODO: These can be removed once the DevShop deploy task runs the install task before it runs the deploy hooks. 
+# TODO: These can be removed once the DevShop deploy task runs the install task before it runs the deploy hooks.
 va/deploy/0-composer:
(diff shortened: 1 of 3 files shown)


jonpugh:
Nice! Make sure it has the latest `develop` @swirtSJW and that the right command is in tests.yml.

jonpugh:
Roping in @beeyayjay 

This PR can be used to add a basic migration test!

swirtSJW:
This test is actually running correctly now.  Only the test itself seems to be legitimately failing or at least hitting a false positive. 
  ` The text "ERROR" appears in the text of this page, but it should not.`

swirtSJW:
@jonpugh I got your test to pass, but I am not sure of:
a) the implication of changing api_driver from drush to drupal.
b) a more precise way to validate that errors are not present on the page.  The method you had seems to be failing  on 'error' which is in present in the menu.
(I am not having any success running this scenario locally so I can only test by pushing it up.)

PR #524, opened 2019-08-17
VAGOV-5582 Makes Behat DST work again

ethanteague:
## What this does
- Moves `/tests/drupal-spec-tool` to `/tests/behat/drupal-spec-tool` - DST are behat tests, so this is where they should live.
- Autoloads CustomDrupal namespace necessary for custom behat steps
- Updates behat.yml to function properly, autoload DST classes, and look inside `/drupal-spec-tool` directory

## QA
 - cd `/tests`
 - `lando behat --tags=@access` (this runs a DST test)
 - Confirm test runs (see screenshot)
 - `lando behat --tags=@errors` (this runs a behat test in `/features` - note that a step is missing, but this is expected)
 - Confirm test runs (see screenshot)

## Screenshots
Screenshot_2019-08-17_11-39-34
Screenshot_2019-08-17_11-43-01


ethanteague:
> I haven't tested this yet, but my main comment would be that I think we should differentiate the "behat tests" (Behavior testing) from the "Drupal Spec Tests" (tool for verifying content architecture).
> 
> Do you think it would be good to add a new test section to `tests.yml` called `va/tests/spec`?

I hear what you are saying about behavior vs architecture, but DST uses behat, and are in fact behat tests. I think they should live in behat. I could see adding a behavior testing dir inside of behat for non-spec tests, but moving DST to a folder outside of behat is gonna be a pain point and will require multiple behat.yml files. @jonpugh @swirtSJW

jonpugh:
Great points. :)

You could, however, still make two separate "test sets" in tests.yml, even if the tests are in the same folder by adding those options you mention.

Something like this?
```
va/tests/behat:
  command: cd tests/behat && behat --tag=spec
va/tests/spec:
  command: cd tests/behat && behat --skip-tag=spec
```

FYI I started work on adding the name kind of "filter" option to `yaml-tests` so that users could run commands against a tests.yml file above, like:

`composer y behat` or `composer y spec`

swirtSJW:
Closed at Ethan's request.

PR #512, opened 2019-08-14
VAGOV-5481: Canary Build: Stabilize CMS with node edit/web rebuild/node view test.

jonpugh:
# Stabilize CMS: The Canary Build

This Pull Request contains the test that will confirm that Drupal edits get propagated to the `vets-website` front end.

Once this build passes, we know that all of the issues surrounding WEB generation from our desired Drupal site are solved.

PR #526, opened 2019-08-19
VAGOV-5389: Add test to confirm web build failed and fail the CMS test.

jonpugh:
This PR just adds a GREP on the web build output so it will trigger failure if the WEB app was unable to build from Drupal content.

I believe every bad situation spits out this error: 

> Failed to pipe Drupal content into Metalsmith!

So I am checking against that.


jonpugh:
@ElijahLynn @ethanteague @jonbot We should merge this ASAP and get it into the other branches so all other PR environments have this "web removal" script.

jonpugh:
I'm going to destroy this environment and let it be recreated to be really sure it's working right.

I also want to add the test to do the editing.

tests.yml
```diff
@@ -26,7 +26,7 @@ va/deploy/3-config:
 
 va/web/build:
   description: Build VA.gov Front-end
-  command: composer va:web:build | grep  "Failed to pipe Drupal content into Metalsmith!" -C4 && echo "Metalsmith Drupal error detected." && exit 1 || echo "Front end site was built! Check $DRUPAL_ADDRESS/web for raw output!"
+  command: composer va:web:build | grep  "Failed to pipe Drupal content into Metalsmith!" -B1000 -C8 && echo "Metalsmith error detected in npm build process." && exit 1 || echo "Front end site was built! Check $DRUPAL_ADDRESS/web for raw output!"
```
ElijahLynn: Especially in this case `--before 1000 --context 8` would make this more self-documenting/readable`. Although `--context` is before and after so this likely should just be `--after 8` since you are already doing the `--before`.

jonpugh: @ElijahLynn This is just a shim until the web build fails properly with a good error message.

If you think I should still change it let me know.

ElijahLynn: Ahh, no need then.

jonpugh:
@jonbot We now treat "WEB build" as a test: You can see `va/web/build` under the "Checks" in the PR. 

Right now, WEB builds still pass (exit with 0 code) even if the Drupal site content import fails.

This PR looks for the error message in the WEB build output and fails the test if it is found.

jonpugh:
Tests are passing, I don't think there are any more blockers to merge?

jonpugh:
Closing, see https://github.com/department-of-veterans-affairs/va.gov-cms/pull/512

PR #528, opened 2019-08-21
VAGOV-3930 Auto-assign content_editor roles for all staging SSOi login.

ElijahLynn:
For SSOi testers.


ElijahLynn:
Target was master

PR #529, opened 2019-08-21
VAGOV-3930 Auto-assign content_editor roles for all staging SSOi login.

ElijahLynn:
For SSOi testers.


ElijahLynn:
Anything against `master` will fail right now. We need things merged from `develop` for these checks to work. Disabling the checks for now @swirtSJW.

swirtSJW:
@ElijahLynn   This looks good.  Thanks for the solution.  
a)  I am assuming you tested this on stage to make sure it works  (i don't have a way to test locally (I don't think I do anyway) Can yo confirm where this was tested?

b) Why are we bypassing develop and going straight to master with this?  Assuming we are going to be deploying a tagged develop to stage.  (I'd rather keep develop and master from diverging too much unless there is good reason)

ElijahLynn:
> @ElijahLynn This looks good. Thanks for the solution.
> a) I am assuming you tested this on stage to make sure it works (i don't have a way to test locally (I don't think I do anyway) Can yo confirm where this was tested?

Yes, this is currently on STAGING now as a manual change, in fact, I actually exported the STAGING config for this commit. 

> b) Why are we bypassing develop and going straight to master with this? Assuming we are going to be deploying a tagged develop to stage. (I'd rather keep develop and master from diverging too much unless there is good reason)

It was in case we did another hotfix to master that this would get in and not require a full merge of develop which may or may not be desired to go to STAGING. I can make it against develop, but it won't have any effect on DEV. As long as our next STAGING release has develop merged in then I can make this against develop.

ElijahLynn:
I just changed the base from `master` to `develop`.

PR #535, opened 2019-08-22
VAGOV-5690: Globally load environment variables.

jonpugh:
Adds EnvironmentHandler.php file that has simple logic for setting .env vars.

1. If `.env` exists, load it.
2. If $_SERVER['LANDO'], load `.env.lando`.
3. If env file was found, make DRUPAL_ADDRESS and BEHAT_PARAMS required.

Now, all PHP that uses the main autoloader should have the proper environment variables set, FORCED from `.env` or `.env.lando` file.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,6 @@
     },
     "require": {
         "acquia/headless_lightning": "^1.4.0-beta2",
-        "bangpound/composer-dotenv": "^1.0@dev",
         "cweagans/composer-patches": "^1.6.0",
         "devshop/behat-drupal-extension": "dev-master",
         "drupal-composer/drupal-scaffold": "^2.0.0",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -104,7 +103,10 @@
         "psr-4": {
             "Traits\\": "tests/traits",
             "CustomDrupal\\": "tests/behat/Drupal"
-        }
+        },
+        "files": [
+            "scripts/composer/EnvironmentHandler.php"
+        ]
```
ethanteague: Should this be inside the ps-4 object, @jonpugh ? Really don't know, just wanna make sure that autoloading files without associating it with a spec is okay.

jonpugh: We can try another way, if you like. 

I tried PSR-4 and had problems getting it to run, but I think it was something else I was doing wrong. Might be able to change it now to PSR-4 since it is now the right code.

Would you mind testing it? You have to `composer dump-autoload` when this changes.

PR #536, opened 2019-08-22
VAGOV-000: Example of Automatable WEB->CMS PR: Just change REF

jonpugh:


composer.lock
```diff
@@ -18124,9 +18136,6 @@
                         "status": "covered",
                         "message": "Covered by Drupal's security advisory policy"
                     }
-                },
-                "patches_applied": {
-                    "Include end date / time in ICS file.": "https://www.drupal.org/files/issues/2018-10-06/ics_field-2852239-11.patch"
```
ElijahLynn: Was this patch supposed to be removed?

composer.lock
```diff
@@ -10725,7 +10737,7 @@
                     }
                 },
                 "patches_applied": {
-                    "Fix uncaught type error by extending class.": "patches/workflowParticipantsUncaughtType.patch"
+                    "Fix revision uncaught type error by extending class.": "https://www.drupal.org/files/issues/2019-08-16/workflow_participants-revision-error-3075426-1.patch"
```
ElijahLynn: Was this updated intentionally?

jonpugh: I ran `composer update --lock` ...

composer.lock
```diff
@@ -17371,10 +17383,10 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/department-of-veterans-affairs/vets-website",
-                "reference": "origin/VAGOV-2937-unity"
+                "reference": "bf54229be12badf4078abab5ae156f30ce6908f9"
```
jonpugh: Looks like composer update is working on va-gov/web after all :|

composer.lock
ElijahLynn: Okay, so that means the previous composer.json was out of sync. Fair enough.

jonpugh: Ohh ok... I just added a new PR against develop with `composer update --lock` ran against it.

jonpugh:
See https://github.com/department-of-veterans-affairs/va.gov-cms/pull/537 PR with JUST composer.lock file changes.

swirtSJW:
If this was just an example, we should consider closing this as I moved the example to https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/content/cms/authoring-experience/engineering/development.md#building-cms-locally-with-a-specific-fe-build

ElijahLynn:
It is an example, but we actually also need to get this merged soon, actually we need to merge in master to the upstream web PR and use that hash too.

PR #538, opened 2019-08-23
Remove `env` dump from tests.yml, sensitive output.

ElijahLynn:


jonpugh:
@ElijahLynn it occurs to me we could use the ENV test to confirm the presence of the critical Env vars in different parts of the system?

Now that it works we don't need to see them all, but seeing DRUPAL_ADDRESS and BEHAT_PARAMS would be nice.

PR #547, opened 2019-08-26
VAGOV-5767: Add WEB tests to CMS test suite.

jonpugh:
This PR adds the new test `va/web/unit`.

It kicks off the WEB unit tests after the WEB build and Behat tests run.

this PR is only going to add this test. Getting the full WEB test suite to pass requires further work.

### Background

Environment variables and CLI options affect the build:
NODE_ENV=prod
 `--buildtype=VAGOVPROD` will cause the build to fail if there are bad links.

So... Do we **want** the build to fail if the link checker fails? because right now that would block a merge!  Who would this be a question for, @swirtSJW ?


ElijahLynn:
I changed the base branch from `develop` to `master` as we switched our CD to be based on the GitHub Flow model (as opposed to Git Flow) and `develop` is now deprecated (https://github.com/department-of-veterans-affairs/va.gov-cms/pull/617). It may be best instead of fixing conflicts to rebase/replay your commits on top of current master:

`git fetch --all`
`git rebase origin/master`
`git push --force`

jonpugh:
I want to bring this back once we have the Rebuild Web uI.

swirtSJW:
Blowing the dust off this.  @jonpugh  Does this do something we aren't already doing or can we close it?

swirtSJW:
Closing this as unresponsive.  Re-open if needed.

PR #548, opened 2019-08-27
VAGOV-5687: Create separate Test Set for "behat" and "spec"

jonpugh:
## Problem

Behat tests (behavoral testing) are very different from the Drupal Spec Tool, even if they are built using the same thing.

1. No differentiator when a Spec test fails vs a Behat test fails. Devs must dig through the error logs to figure out what problem they have.
2. Devs can't run Behavioral tests without running spec tests too. The current setup allows Spec tests to be run by themselves with `behat --tags=dst` but there's no way to run all tests that _don't_ have that tag.
3. When a Behavioral test errors, the entire output of the spec tests are printed out too, making the github error comments very long. For example: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/547#commitcomment-34840945  https://github.com/department-of-veterans-affairs/va.gov-cms/commit/6f291c434f953e8d4eac35fa7f1e4ff7a9c8359f
4. Drupal Spec Tool does not test any behavior at all. It simply loads the list of expected entities from the .feature files and uses Drupal API code to confirm that they exist. Every view, node form, node view page could be broken, and the Drupal Spec tests would pass. 

## Solution

With `behat.yml`, you can filter out a certain scenario tag using behat.yml.

This PR adds the DST tag to the filtered scenarios, so they are not called with `bin/behat`, but they are called with `bin/behat --tags=dst`.

This allows us to create two different test sets: one for "behat" and one for "spec", and allows two distinct commands to run spec vs behat tests.

## Usage:

- `behat` or `lando behat` or `composer yaml-tests va/tests/behat`: Runs all tests except dst.
- `behat --tags=dst` or `lando spec` or `composer yaml-tests va/tests/spec`: Runs only DST tests.


swirtSJW:
There was a lot of discussion here.
https://dsva.slack.com/archives/CJ34D8TPB/p1566912488355500  
about this.  I opened it up there for other devs to weigh in.

ElijahLynn:
I changed the base branch from `develop` to `master` as we switched our CD to be based on the GitHub Flow model (as opposed to Git Flow) and `develop` is now deprecated (https://github.com/department-of-veterans-affairs/va.gov-cms/pull/617). It may be best instead of fixing conflicts to rebase/replay your commits on top of current master:

`git fetch --all`
`git rebase origin/master`
`git push --force`

PR #559, opened 2019-08-31
VAGOV-5843 Entity clone

ethanteague:
## What this does
 - Installs and enables entity_clone module

## Use case / User story
 - As a site user, I would like a way to easily duplicate content

## Screenshot
Screenshot from 2019-08-31 17-08-49

## QA
 - As an admin, visit `admin/content/bulk` and choose `Clone` from the `Operations` button for any node in the table.
 - Complete the subsequent form, visually verifying that clone operation is successful

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,7 @@
         "drupal/content_lock": "^1.0@alpha",
         "drupal/cshs": "^1.0@beta",
         "drupal/devel": "^2.1",
+        "drupal/entity_clone": "^1.0@beta",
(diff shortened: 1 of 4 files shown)


swirtSJW:
@ethanteague  Should we have any perms setup associated with this?

ethanteague:
> @ethanteague Should we have any perms setup associated with this?

It's a poc, so I don't _think_ so @swirtSJW - @jefflbrauer mentioned that this would be a nice to have for users. I think if someone can create content, it's okay to allow them access to this.

PR #561, opened 2019-09-04
VAGOV-5857: Adding all flags to config.

ethanteague:
Adding config to mirror prod version of flags.


ElijahLynn:
Let's keep this in mind for when we override these flags since config overrides don't show in the UI and is very confusing. 

https://www.drupal.org/project/config_override_warn (Dave Reid Module) (see https://www.drupal.org/project/drupal/issues/2408549 for context)

PR #562, opened 2019-09-05
VAGOV-5827 Use PROD SSO metadata and endpoint.

ElijahLynn:


ElijahLynn:
This needs one more commit, working on removing from config_split for staging.

ElijahLynn:
right now it is enabled on staging, disabled on prod, which we are going to keep disabled on prod but don't want it on staging because endpoint is not correct anymore.
I will be making a follow up PR to keep this enabled on DEV, STAGING, PROD but that is on the devops side.

Elijah Lynn  < 1 minute ago
For now it will just work on PROD.

Elijah Lynn  < 1 minute ago
Because SAML PHP settings cannot be in config (config_split)

ElijahLynn:
Steve Wirt  25 minutes ago
Hmmm Has an issue with changing config from split to global.

Steve Wirt  13 minutes ago
I don't think we can have config set on sync but not have the module enabled.

Steve Wirt  10 minutes ago
Unless I am missing something (entirely possible) you can only have config on the environments where the module should be enabled.

Elijah Lynn  10 minutes ago
Okay, we can enable it on PROD then, and then just disable it manually

Elijah Lynn  8 minutes ago
Does that work for you?

Elijah Lynn  8 minutes ago
pushed

Elijah Lynn  4 minutes ago
pushed for real now, the upstream merge blocked the push before

Steve Wirt  < 1 minute ago
Yah that should work.

ElijahLynn:
Had to close to blow away ENV because of:

> [RuntimeException]                                                                                                                 
>   Failed to clone https://github.com/department-of-veterans-affairs/vets-website via https, ssh protocols, aborting.                 
>                                                                                                                                      
>   - https://github.com/department-of-veterans-affairs/vets-website                                                                   
>     fatal: destination path '/var/aegir/projects/cms/pr562/docroot/vendor/va-gov/web' already exists and is not an empty directory.  
>                                                                                                                                      
>   - git@github.com:department-of-veterans-affairs/vets-website                                                                       
>     Cloning into '/var/aegir/projects/cms/pr562/docroot/vendor/va-gov/web'...                                                        
>     done.                                                                                                                            
>     fatal: remote composer already exists.

Reopening now to build anew.

PR #565, opened 2019-09-07
VAGOV-5892 Adjust SSO settings.

swirtSJW:
…ons.


config/sync/config_split.config_split.prod.yml
```diff
@@ -1,6 +1,6 @@
 uuid: e20d7e7e-1568-451c-b4f5-b3939607203b
 langcode: en
-status: true
+status: false
```
swirtSJW: I don't think this should have flipped to false.   I need to look into this.

docroot/sites/default/settings.php
```diff
@@ -137,43 +137,45 @@
...
-    // User info and syncing.
-    // `unique_id` is specified in Transient format, otherwise this should be `UPN`
-    // Please talk to your SSO adminsitrators about which format you should be using.
-    'unique_id'               => 'VAUID',
-    'user_name'               => 'adUPN',
-    'mail_attr'               => 'adUPN',
-    'sync' => [
-        'mail'      => TRUE,
-        'user_name' => TRUE,
+  // Basic settings.
```
ElijahLynn: With the addition of config_ignore, we want to move from settings to config right?

swirtSJW: @ElijahLynn No config_ignore makes config not get saved or used for anything related to simplesaml_php   This is relying 100% on settings so that we hit Robbie's concern of not allowing someone to disable settings that might make it less secure.

swirtSJW: It is basically keeping it operating like it was, without the added confusion of having config that did not match what was actually in place from settings.php.

ElijahLynn: K, I was thinking of it as we keep it only in config and confi_ignore just makes it so it can't get saved in the UI. 

Thanks

ElijahLynn:
> ```shell
> FetchError: maximum redirect reached at: http://pr565.ci.cms.va.gov/user/16
> ```

docroot/sites/default/settings.php
ElijahLynn: We need to remove the settings from config too then if we are going to ignore them with config_ignore.

swirtSJW: Gone.

ElijahLynn: I am feeling like we should go back to config and config_split for this. It is getting kinda complex without much gain here since the site will be forced HTTPS. Your call but that is my vote.

ElijahLynn:
I forked from this PR and made some more commits on top. So once this is merged then #576 will be a smaller diff.

docroot/sites/default/settings.php
swirtSJW: Config and config split would allow for UI override and hardening to not allow UI override is something Robbie had asked for.  If it is config split then we need separate configs for each environment.  This method only needs an override for prod.

swirtSJW:
Tests are passing... just Unity is not reporting it to gitlab for some reason
image

composer.json
```diff
@@ -24,6 +24,7 @@
         "drupal/coder": "~8.2.12",
         "drupal/components": "^1.0",
         "drupal/config_ignore": "^2.1",
+        "drupal/config_override_warn": "^1.0",
```
ElijahLynn: Do we still want to keep this one?

config/prod/simplesamlphp_auth.settings.yml
```diff
@@ -1,38 +1,34 @@
...
     admnistrator_users: admnistrator_users
     administrator: administrator
-    redirect_administrator: redirect_administrator
-    authenticated: '0'
-    documentation_editor: '0'
-  default_login_users: '1'
+  default_login_users: '1,16'
 logout_goto_url: ''
-user_register_original: admin_only
+unique_id: VAUID
```
ElijahLynn: This needs to be `adUPN` for it to work the way it is now.

config/dev/simplesamlphp_auth.settings.yml
```diff
@@ -0,0 +1,34 @@
...
+    content_editor: content_editor
+    content_reviewer: content_reviewer
+    content_publisher: content_publisher
+    content_admin: content_admin
+    redirect_administrator: redirect_administrator
+    admnistrator_users: admnistrator_users
+    administrator: administrator
+  default_login_users: '1,16'
+logout_goto_url: ''
+unique_id: VAUID
```
ElijahLynn: This needs to be `adUPN` to work as is.

ElijahLynn: We added hook_simplesamlphp_auth_existing_user() so we can keep VAUID as is.

config/prod/simplesamlphp_auth.settings.yml
ElijahLynn: We added hook_simplesamlphp_auth_existing_user() so we can keep VAUID as is.

config/sync/config_split.config_split.prod.yml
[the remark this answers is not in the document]
ElijahLynn: Thanks

config/sync/config_split.config_split.stg.yml
```diff
@@ -1,6 +1,6 @@
 uuid: 8b4816ce-c4d3-4621-9868-8fed1e85ebaa
 langcode: en
-status: true
+status: false
```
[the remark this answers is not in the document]
ElijahLynn: thanks

ElijahLynn:
@swirtSJW K, got it working now with https://github.com/department-of-veterans-affairs/va.gov-cms/pull/565/commits/2121b2f34c75a62f2973a920a3e32bf7b639ee8c. The module was not enabled AND was using the wrong value. 

Worth noting that this does enable some old functionality that we had when email sending wasn't working, which, is the same state now, so it actually should be enabled until we get email sending working again.

composer.json
swirtSJW: Yes we  do. I think it is good to have in place to warn of other config discrepancies should they occur.

PR #566, opened 2019-09-09
VAGOV-5908: Config ignore menu breadcrumbs settings.

kevwalsh:
Menu breadcrumbs settings are changing in a way that's similar to menu items themselves. They should be treated more like content than config. This PR adds the config_ignore module, so that this configuration can be updated on prod by admins.  It also contains the "latest" menu_breadcrumbs settings from prod, but it will be ignored :) 

To QA: 

1. Go to admin/config/user-interface/menu-breadcrumb and make a bunch of changes random changes, like to the order. 
2.Go to admin/config/development/configuration and note that:
- [ ] There are no changes to import
- [ ] menu_breadcrumb.settings is listed in the Config Ignore section.


config/sync/core.extension.yml
```diff
@@ -34,7 +35,6 @@ module:
   embed: 0
   entity_block: 0
   entity_browser: 0
-  entity_clone: 0
```
swirtSJW: @kevwalsh   Was entity_clone supposed to get clobbered by this?

kevwalsh: No. Thanks for catching this.

kevwalsh: Resubmitted!

swirtSJW:
Seems like there are some failures of tests that are unrelated to this PR.  Will need to trigger another build to see if it resolves.

PR #568, opened 2019-09-09
VAGOV-5884: Content API Consumer Menu Permission

jonbot:
• Adds 'Administer Menus' permission to Content API Consumer role. (Permission is mitigated by Workbench menu access.)
• Adds Workbench Menu Access settings to several menus.


swirtSJW:
Looks like the phpunit test needs to be updated to include this change or perms.

PR #569, opened 2019-09-09
VAGOV-5865: Bottom Rail Footer Config

jonbot:
• Adds bottom rail footer menu.
• Adds an existing menu field to that menu ("Open link in new tab?").
• Adds workbench menu access permission to that menu.


swirtSJW:
I triggered a rebuild on devshop and got another failure and then triggered a gain and got a success.
image

PR #572, opened 2019-09-10
VAGOV-5776: Add content admin role.

kevwalsh:
The Content_admin role will be given out to a small team of users (for now, 5-7) from the VA's Global UX team. These users can bypass content permissions (workbench access, workbench menu access, taxonomy, media), and can control sensitive content such as the homepage entity queues. In the future, this role may expand to include certain SEO tasks, or to manage other microcopy on the site. 

To test:
- [ ]  As an admin, review the permissions from /admin/people/permissions
- [ ]  Note that they are now sorted so the admin role is on the far right (previously it was a bit random).
- [ ]  Note that Redirect administrator has been renamed to Redirect admin, for brevity, and consistency with the new Content admin role. 
- [ ]  Scan the list of permissions, and compare the content_publisher role to the content_admin role. You'll see that they have several new permissions, including, notably, these ones, which are not available to content_publisher 

  - 'administer menu'
  - 'administer nodes'
  - 'administer taxonomy'
  - 'break content lock'
  - 'bypass node access'
  - 'bypass workbench access'
  - 'clone block entity'
  - 'clone node entity'
  - 'manipulate all entityqueues'
  - 'manipulate entityqueues'

- [ ]  Create a user with content_admin role (say, "Content Admin"), then log in as Content Admin. Even without any workbench access, you should be able to edit any 
  - [ ]  node
  - [ ]  block
  - [ ]  taxonomy term
  - [ ]  menu, or 
  - [ ]  media entity. 

You should also be able to manipulate the homepage entity queues.


swirtSJW:
@kevwalsh It looks like there are some phpunit tests that are failing that need some updating.

- tests/phpunit/SecurityRolesPermissionsTest.php

kevwalsh:
@swirtSJW  I fixed test and rebased.

PR #573, opened 2019-09-10
VAGOV-5913: Adding cac role test for menu access.

ethanteague:
## What this does
 - Adds a simple test to ensure `content_api_consumer` user cannot alter existing menus
 - Tests for the presence of edit links on menu admin page = if edit buttons are there, test fails.

## QA
 - `lando behat --tags="@cac_menus"`


ethanteague:
> Cool solution to this!
> 
> I hope we get rid of all those dumb "View json" operations links in the future, no one will actually use that, AFAIK. But we can update the test then.

Yeah, it's probably enough just to check for the presence of the edit buttons and stop there.

PR #575, opened 2019-09-10
VAGOV-5766: Changing node type meta tag order.

ethanteague:
## What this does
Sets metatags for these content types
- Press release
- News Story
- Event

To this pattern: [node:field_office] | [node:title] | [site:name]


config/sync/metatag.metatag_defaults.node__event.yml
```diff
@@ -5,9 +5,9 @@ dependencies: {  }
 id: node__event
 label: 'Content: Event'
 tags:
-  title: '[node:title] | [node:field_office] | [site:name]'
-  og_image_width: '600'
+  title: '[node:field_office] | [node:title] | [site:name]'
   og_image: '[node:field_media:entity:image:entity:url]'
+  og_image_width: '600'
```
kevwalsh: was this intentional?

kevwalsh: has this been tested for whether or not it messes with the aspect ratio? i didn't really expect og_image_height to be specified either, on the next line, but i worry about specifying both.

ethanteague: Updated and removed, @kevwalsh

config/sync/metatag.metatag_defaults.node__event.yml
```diff
@@ -5,8 +5,7 @@ dependencies: {  }
 id: node__event
 label: 'Content: Event'
 tags:
-  title: '[node:title] | [node:field_office] | [site:name]'
-  og_image_width: '600'
```
swirtSJW: @ethanteague  Was this supposed to go away?  It seems unrelated to the ticket.

ethanteague: Yep. It was added in inadvertently.

PR #576, opened 2019-09-11
VAGOV-5873 SSO improvements

ElijahLynn:
Forked from #565, adds to it.


ElijahLynn:
We cherry-picked into #565.

PR #580, opened 2019-09-17
VAGOV-5952: Stopcode and health service export.

kevwalsh:
QA:
As admin
- [x] Go to /vha-health-services
- [x] Edit a couple taxonomy terms with the edit link, and add stop codes, help text should be enough to help you.

As anon
- [x] Go to /vha-health-services
- [x] Click the button at the top and download the CSV
- [x] The CSV should contain all fields except the edit field

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -78,6 +78,7 @@
         "drupal/uswds": "^1.0@beta",
         "drupal/views_bulk_edit": "^2.3",
         "drupal/views_bulk_operations": "^2.5",
+        "drupal/views_data_export": "^1.0@beta",
         "drupal/views_tree": "^2.0@alpha",
         "drupal/workbench_access": "^1.0@beta",
         "drupal/workbench_menu_access": "^1.0",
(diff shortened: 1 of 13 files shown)


swirtSJW:
Looks like we have a composer conflict.

composer.json
```diff
@@ -77,6 +77,7 @@
         "drupal/uswds": "^1.0@beta",
         "drupal/views_bulk_edit": "^2.3",
         "drupal/views_bulk_operations": "^2.5",
+        "drupal/views_data_export": "^1.0@beta",
         "drupal/views_tree": "^2.0@alpha",
```
swirtSJW: Did drupal/csv_serialization also need to get added here?  It is in the .lock, but not the require ??

swirtSJW:
Got a conflict that needs resolving here.

swirtSJW:
I resolved conflict and force pushed.  Waiting on tests.

PR #590, opened 2019-09-19
VAGOV-6018: disable devel and remove test

swirtSJW:
Expected behaviour:
1. behat menu test will pass on all environments
2. devel, devel_generate and devel_kint will not be enabled on any environment.
3. If devel, devel_generate or devel_kint are enabled locally and config is exported, any settings for them will not be exported.

Testing instructions are in the ticket https://va-gov.atlassian.net/browse/VAGOV-6018


config/sync/lightning_core.versions.yml
```diff
@@ -114,7 +114,7 @@ ui_patterns_library: 1.0.0
 xmlsitemap: 1.0.0-alpha3
 xmlsitemap_engines: 1.0.0-alpha3
 menu_export: 1.2.0
-devel: 1.2.0
+devel: 2.1.0
```
swirtSJW: Verified:  This is the current version of devel.

.gitignore
```diff
@@ -18,6 +18,13 @@ node_modules/
 
 # Ignore local dev files
 /docroot/sites/default/settings/settings.local.php
+# Prevent Devel from appearing anywhere.
```
swirtSJW: Prevents them from being accidentally committed and causing havoc on all environments.

config/sync/config_ignore.settings.yml
```diff
@@ -1,4 +1,10 @@
 ignored_config_entities:
-  - menu_breadcrumb.settings
+  0: devel.settings
```
swirtSJW: In reality these exclusions do not have any effect unless devel is actually enabled, but it can not be enabled, so it is an odd conundrum.

PR #591, opened 2019-09-20
VAGOV-5645 More behat tests

ethanteague:
## What this does
 - tests for presence in build of header menu items
 - tests for presence in build of footer menu items
 - tests for absence in build of unpublished node
 - tests for presence in build of published node

## To test
 - menus: `behat --tags=@frontend_menus`
 - unpublished content: `behat --tags=@unpublished_content`
 - published content: `behat --tags=@published_content`


tests/behat/features/menus.feature
```diff
@@ -0,0 +1,183 @@
...
+      @TODO: When header and footer integration work is complete,
+      test changing of menu items in cms being mapped to website build.
+      """
+    In order to confirm menu item existence
+    I need to build front end
+    And check for presence of all menu items
+
+  @frontend_menus
+  Scenario: Build front end and check for presence of all menu items.
+    When I run "cd .. && composer va:web:build"
```
swirtSJW: This step takes a long time to run and already runs twice in features/content.feature   I'm concerned about running it three additional times in these tests.   
Can we either make it so we are not testing the front end directly... or consolidate these tests into the ones that are already running so that we do not have to run va:web:build more times?

tests/behat/features/decoupled.feature
```diff
@@ -0,0 +1,214 @@
...
+    And I should see "Twitter"
+    And I should see "Flickr"
+    And I should see "YouTube"
+    And I should see "All VA social media"
+    And I should see "Veterans Crisis Line"
+    And I should see "Find a VA location"
+    And I should see "Ask a question"
+    And I should see "844-698-2311"
+
+  Scenario: The homepage is the log in form and the site title is as intended.
```
swirtSJW: Minor:  This one is unrelated to the decoupled portion.   I think it may be better to move it here
behat/features/user.feature

PR #609, opened 2019-09-26
VAGOV-000 BSD `rm` needs the -f in the beginning or it fails if file not found.

ElijahLynn:


ElijahLynn:
> Approved
> Strange. The -f at the end works locally

You may have GNU `rm`, e.g. homebrew or Linux. It works for me locally too, but I'm on Arch Linux.

PR #611, opened 2019-09-30
VAGOV-5963: Custom Updater

jonpugh:
This PR adds a hook to `composer update` that manually scans and sets the `va-gov/web` package.

This makes it so the pseudopackage gets updated to the latest master branch SHA automatically.

### DIFF
--- a/scripts/composer/UpdateHandler.php
+++ b/scripts/composer/UpdateHandler.php
@@ -51,6 +51,7 @@ class UpdateHandler {
 
       self::printLine("Found package " . $package->getPrettyString());
       self::printLine("Source Ref " . $package->getSourceReference());
+      self::printLine("Dist Ref " . $package->getDistReference());
 
       // Load origin/master SHA.
       self::printLine("Looking up SHA of vets-website... ");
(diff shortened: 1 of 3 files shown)


scripts/composer/UpdateHandler.php
```diff
@@ -0,0 +1,101 @@
...
+    const REPOSITORY_URL = 'https://github.com/department-of-veterans-affairs/vets-website';
+    const REPOSITORY_BRANCH = 'master';
+    const PACKAGE_NAME = 'va-gov/web';
+    const WEB_PATH = 'web';
+
+    const CLI_PREFIX = 'UpdateHandler.php | ';
+
+    /**
+     * Runs before composer update does dependency lookup: This forces
+     * va-gov/web to use the SHA we detect from the desired branch.
```
ElijahLynn: Should we just call this va-gov/vets-website, it isn't that much longer and is semantic then.

ElijahLynn: this is trivial btw, I don't care if this is discussed before merge.

PR #612, opened 2019-10-01
VAGOV-6106: Add script to test, reproduce, and maybe create workaround for GraphQL Error. 

jonpugh:
Recently the GraphQL endpoint started failing during the web build, but only inside the node edit test.

@ElijahLynn has dug in, and with the help of a curl script, reproduced the error and created a workaround.

As he demoed this to @jonpugh he wrote the steps into the tests.yml file so we can reproduce and have a regression test for this.


ElijahLynn:
I fixed the graphql error, it was a memory exhaustion error, and was extra confusing because it wasn't showing in the logs (error_log wasn't set), and it was a `200` response code. I think the composer command is good to keep in here and am thinking we should keep it commented out in tests.yml because of how much time it adds to the test suite. Thoughts?

jonpugh:
It would be nice to add a curl test to the GraphQL however. 

A badly configured server in the future could  cause this problem again.

PR #617, opened 2019-10-03
Merge develop to master in prep for CD off of master

ElijahLynn:
We are switching to GitHub Flow style CD, where we branch off of master. 

This PR from develop to master should be the last work we do on develop.


ElijahLynn:
Closed to delete PR ENV as it was borked when I changed base branches.

PR #623, opened 2019-10-03
VAGV-000 Remove commit-msg hook check, check branch instead, we squash & merge now so the branch name is all that matters.

ElijahLynn:
Since we are squashing and merging now, all the commits will become part of the single merge commit (not a real merge commit fwiw). We were not checking the branch name before, this PR removes the commit message check and checks the branch name instead. The branch name will be associated with Jira tickets and that should lead to PRs that have the branch name in it. 

This PR:
1. Deletes the commit-msg hook from git and after a composer install. 
1. Checks the branch name in the pre-commit hook to match vagov-*
1. Gives instructions on how to rename the branch if named incorrectly
1. Keeps the phpcs code check


hooks/pre-commit
```diff
@@ -1,2 +1,14 @@
-# Just run codesniffer for now pre-commit.
-composer va:test:cs
```
ElijahLynn: Moved this to the bottom.

PR #631, opened 2019-10-10
VAGOV-5333: Add fields and logic for facility API sourced data.

ethanteague:
## What this does
Sucks in these fields (and disables user editing) on facility nodes in a presave hook:
- Title
- Address
- Hours
- Phone

## Screenshot
Screenshot from 2019-10-11 13-00-25

## Path forward
Migration to update facilities on daily(ish) interval

### DIFF
--- a/config/sync/field.field.node.health_care_local_facility.field_address.yml
+++ b/config/sync/field.field.node.health_care_local_facility.field_address.yml
@@ -12,7 +12,7 @@ field_name: field_address
 entity_type: node
 bundle: health_care_local_facility
 label: Address
-description: 'This field is populated by Facility API and is not intended to be changed by editors.'
+description: 'The physical address (populated by Facility API)'
 required: true
 translatable: true
 default_value: {  }
--- a/config/sync/field.field.node.health_care_local_facility.field_address.yml
+++ b/config/sync/field.field.node.health_care_local_facility.field_address.yml
@@ -1,0 +1,44 @@
+uuid: 2988497e-c800-417e-a246-d3deb36241c5
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.node.field_address
+    - node.type.health_care_local_facility
+  module:
+    - address
+id: node.health_care_local_facility.field_address
+field_name: field_address
+entity_type: node
+bundle: health_care_local_facility
+label: Address
+description: 'This field is populated by Facility API and is not intended to be changed by editors.'
+required: true
+translatable: true
+default_value: {  }
+default_value_callback: ''
+settings:
+  available_countries:
+    US: US
+  langcode_override: ''
+  field_overrides:
+    givenName:
+      override: hidden
+    familyName:
+      override: hidden
+    organization:
+      override: hidden
+    addressLine1:
+      override: optional
+    addressLine2:
+      override: hidden
+    postalCode:
+      override: optional
+    sortingCode:
+      override: hidden
+    locality:
+      override: optional
+    administrativeArea:
+      override: optional
+  fields: {  }
+field_type: address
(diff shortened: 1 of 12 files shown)


config/sync/field.field.node.health_care_local_facility.field_address.yml
```diff
@@ -0,0 +1,44 @@
...
+    - field.storage.node.field_address
+    - node.type.health_care_local_facility
+  module:
+    - address
+id: node.health_care_local_facility.field_address
+field_name: field_address
+entity_type: node
+bundle: health_care_local_facility
+label: Address
+description: 'This field is populated by Facility API and is not intended to be changed by editors.'
```
swirtSJW: I think we want to have the field description reflect what the field is, but then disable the field on the node edit form so that it can't be changed by hand.

config/sync/field.field.node.health_care_local_facility.field_facility_hours.yml
```diff
@@ -0,0 +1,58 @@
...
+    - field.storage.node.field_facility_hours
+    - node.type.health_care_local_facility
+  module:
+    - tablefield
+id: node.health_care_local_facility.field_facility_hours
+field_name: field_facility_hours
+entity_type: node
+bundle: health_care_local_facility
+label: Hours
+description: 'This field is populated by Facility API and is not intended to be changed by editors. '
```
swirtSJW: Same as others.  Description should describe the field.   Also trailing space.

config/sync/field.field.node.health_care_local_facility.field_phone_number.yml
```diff
@@ -0,0 +1,21 @@
...
+    - field.storage.node.field_phone_number
+    - node.type.health_care_local_facility
+  module:
+    - telephone
+id: node.health_care_local_facility.field_phone_number
+field_name: field_phone_number
+entity_type: node
+bundle: health_care_local_facility
+label: 'Phone Number'
+description: 'This field is populated by Facility API and is not intended to be changed by editors.'
```
swirtSJW: Same

docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
```diff
@@ -32,3 +35,94 @@ function _va_gov_consumers_content() {
...
   return $consumer->contentRender($source, $attribution);
 }
+
+/**
+ * Implements hook_entity_presave().
+ */
+function va_gov_consumers_entity_presave(EntityInterface $entity) {
+  switch ($entity->bundle()) {
+    // Populate title, address, phone, and hours via facility api.
+    case 'health_care_local_facility':
```
swirtSJW: I have some concerns about the architecture of this.
Updating the data on save seems heavy.   It requires that every node be loaded and saved every time the update process is run.

If this were done as a migration, the updates would only happen if the data changed, which it does not change that often.  The changes are tracked by migration and would not require each node to be loaded.

It would also allow for thousands of VMACS to get built as part of the migration, not requiring them to be built by hand initially.

config/sync/field.field.node.health_care_local_facility.field_address.yml
ethanteague: :+1:

config/sync/field.field.node.health_care_local_facility.field_facility_hours.yml
ethanteague: :+1:

config/sync/field.field.node.health_care_local_facility.field_phone_number.yml
ethanteague: :+1:

docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
ethanteague: Adding a note that we covered this in our call with Kev / plan for MIGRATING ALL THE THINGS!!!!! :smile:

PR #632, opened 2019-10-10
VAGOV-6125 Add hook to update git display on cache clear.

swirtSJW:


jonpugh:
Works great, but might be usedful to show the branch and the SHA?

Screenshot from 2019-10-11 12-56-45

The main build already has this library, because yaml-tests requires it: 

You can use:

```php
use TQ\Git\Repository\Repository;
$git = Repository::open('/path/to/your/repository');
$git->getCurrentBranch();
```

PR #637, opened 2019-10-15
VAGOV-5333

ethanteague:
## Description:
Replaces facilities api fully open endpoint with apikey required version.

### DIFF
--- a/docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
+++ b/docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
@@ -19,6 +19,7 @@ class FacilityConsumer {
    */
   public function contentRender($facility_id) {
     if (!empty(getenv('CMS_FACILITY_API'))) {
+      $data = '';
       $client = new Client();
       // Make sure we get a good response before heavy lifting.
       try {
@@ -38,10 +39,9 @@ class FacilityConsumer {
       catch (GuzzleException $e) {
         watchdog_exception('Facility_content', $e->getMessage());
       }
-
-      return $data;
-
     }
+    return $data;
+
   }
 
 }
--- a/docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
+++ b/docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
@@ -18,12 +18,13 @@ class FacilityConsumer {
    *   The facility id.
    */
   public function contentRender($facility_id) {
+    $data = '';
+
     if (!empty(getenv('CMS_FACILITY_API'))) {
-      $data = '';
       $client = new Client();
       // Make sure we get a good response before heavy lifting.
       try {
-        $data_source = 'https://api.va.gov/services/va_facilities/v0/facilities/' . $facility_id;
+        $data_source = 'https://dev-api.va.gov/services/va_facilities/v0/facilities/' . $facility_id;
         $response = $client->get($data_source, [
           'headers' => [
             'apikey' => getenv('CMS_FACILITY_API'),
--- a/docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
+++ b/docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
@@ -22,7 +22,7 @@ class FacilityConsumer {
       $client = new Client();
       // Make sure we get a good response before heavy lifting.
       try {
-        $data_source = 'https://dev-api.va.gov/services/va_facilities/v0/facilities/' . $facility_id;
+        $data_source = 'https://api.va.gov/services/va_facilities/v0/facilities/' . $facility_id;
         $response = $client->get($data_source, [
           'headers' => [
             'apikey' => getenv('CMS_FACILITY_API'),
--- a/docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
+++ b/docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
@@ -18,25 +18,30 @@ class FacilityConsumer {
    *   The facility id.
    */
   public function contentRender($facility_id) {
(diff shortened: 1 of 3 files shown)


ElijahLynn:
API Key in credstash has been merged, https://github.com/department-of-veterans-affairs/devops/pull/5714.

docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
```diff
@@ -18,25 +18,30 @@ class FacilityConsumer {
    *   The facility id.
    */
   public function contentRender($facility_id) {
+    if (!empty(getenv('FACILITY_API'))) {
```
ElijahLynn: This should be `CMS_FACILITY_API` right?

docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
```diff
@@ -18,25 +18,30 @@ class FacilityConsumer {
...
+      $client = new Client();
+      // Make sure we get a good response before heavy lifting.
+      try {
+        $data_source = 'https://dev-api.va.gov/services/va_facilities/v0/facilities/' . $facility_id;
```
ElijahLynn: Should this be api.va.gov or dev-api.va.gov?

docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
```diff
@@ -18,25 +18,30 @@ class FacilityConsumer {
...
+        $response = $client->get($data_source, [
+          'headers' => [
+            'apikey' => getenv('CMS_FACILITY_API'),
+          ],
+        ]);
+
+        if ($response->getStatusCode() === 200) {
+          $data = Json::decode($response->getBody());
```
swirtSJW: Can we define $data as FALSE higher up.   So in case we don't get the 200, $data is still defined for the return.

docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
ethanteague: Thanks for catching that, @ElijahLynn - updated.

docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
ethanteague: The latter, @ElijahLynn - I incorrectly told you the former.

docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
```diff
@@ -18,25 +18,30 @@ class FacilityConsumer {
...
       }
 
-    }
-    // Record any trouble to watchdog.
-    catch (GuzzleException $e) {
-      watchdog_exception('Facility_content', $e->getMessage());
-    }
-
-    return $data;
+      return $data;
```
swirtSJW: Can we move this return out of the `if (!empty(getenv('FACILITY_API')` so that the call to contentRender($facility_id) always returns something, even if it is only FALSE.   We want predictable behavior to this call.

docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
```diff
@@ -76,6 +89,9 @@ function va_gov_consumers_entity_presave(EntityInterface $entity) {
           }
           $entity->field_facility_hours = $days_array;
         }
+        else {
+          drupal_set_message(t('Facility api response is empty. CMS_FACILITY_API env var may not be set.'), 'warning');
```
swirtSJW: minor:  depricated https://api.drupal.org/api/drupal/core%21includes%21bootstrap.inc/function/drupal_set_message/8.5.x 

but we are going to remove this as soon as the migration is in place, so nothing we need to address now.

docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
ElijahLynn: Thanks, can you add a comment as to why we are using the dev-api.va.gov instead of api.va.gov endpoint.

docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
ElijahLynn: Our codesniffer checks should catch deprecated function calls. I am thinking we just have regular codesniffer setup and not the Drupal 8 rules loaded as well.

docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
ethanteague: :+1: updated, @swirtSJW

ethanteague: :+1: updated, @swirtSJW

docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
ethanteague: :+1: updated, @swirtSJW

docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
```diff
@@ -18,23 +18,28 @@ class FacilityConsumer {
    *   The facility id.
    */
   public function contentRender($facility_id) {
+    if (!empty(getenv('CMS_FACILITY_API'))) {
+      $data = '';
```
swirtSJW: Its got to go one level higher.  Otherwise it may be undefined when it hits the return

swirtSJW:
Now that the kill PR has been merged, this will need to be rebased with some conflict resolution.  Sorry Ethan.

swirtSJW:
Closing this in favor of PR-644

PR #644, opened 2019-10-17
VAGOV-5333 Facility API

ethanteague:
## What this does
 - Adds facility api consumer class

## How this differs from last attempt
 - Takes into account new prod api key (set as env var on prod)
  -- Prod has unique key, all other environments use the original dev key (per api team instruction)
 - Env var has same naming convention in all environments, so no env switch necessary in code.

### DIFF
--- a/docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
+++ b/docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
@@ -19,15 +19,16 @@ class FacilityConsumer {
    */
   public function contentRender($facility_id) {
     $data = '';
-
-    if (!empty(getenv('CMS_FACILITY_API_KEY'))) {
+    $cms_facility_api_key = getenv('CMS_FACILITY_API_KEY');
+    $cms_facility_api_url = getenv('CMS_FACILITY_API_URL');
+    if ((!empty($cms_facility_api_key)) && (!empty($cms_facility_api_url))) {
       $client = new Client();
       // Make sure we get a good response before heavy lifting.
       try {
-        $data_source = getenv('CMS_FACILITY_API_URL') . '/services/va_facilities/v0/facilities/' . $facility_id;
+        $data_source = $cms_facility_api_url . '/services/va_facilities/v0/facilities/' . $facility_id;
         $response = $client->get($data_source, [
           'headers' => [
-            'apikey' => getenv('CMS_FACILITY_API_KEY'),
+            'apikey' => $cms_facility_api_key,
           ],
         ]);
 
--- a/docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
+++ b/docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\va_gov_consumers;
 
+use Drupal\Component\Serialization\Json;
 use GuzzleHttp\Client;
 use GuzzleHttp\Exception as GuzzleException;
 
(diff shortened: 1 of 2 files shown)


docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
```diff
@@ -76,6 +89,9 @@ function va_gov_consumers_entity_presave(EntityInterface $entity) {
           }
           $entity->field_facility_hours = $days_array;
         }
+        else {
+          drupal_set_message(t('Facility api response is empty. CMS_FACILITY_API env var may not be set.'), 'warning');
```
ElijahLynn:
```suggestion
          drupal_set_message(t('Facility API response is empty. CMS_FACILITY_API_KEY environment variable may not be set.'), 'warning');
```

ethanteague: Updated, @ElijahLynn :+1:

swirtSJW: Medium: It would be good to output the facility id number here for debugging.   Watchdog would be  be a good idea too.

ElijahLynn:
Deprecates #637.

ElijahLynn:
I don't think this is a blocker but I think all our ENV vars should be mapped inside settings*.php files to a $conf variable so that we can easily track all of them down in one place. Then in the application code we call a $conf variable instead. 

This way we can list non-secrets side by side with other vars. 

Not a blocker, just communicating the intent in the future. We can have another task to move all getenv()'s to settings*.php files.

ElijahLynn:
K, https://github.com/department-of-veterans-affairs/devops/pull/5756 is ready to be merged to compliment this. 

Still need to get it to CI, @jonpugh can you handle that task for now?

docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
```diff
@@ -46,6 +46,19 @@ function va_gov_consumers_entity_presave(EntityInterface $entity) {
       $facility_id_raw = $entity->get('field_facility_locator_api_id')->getValue();
       $facility_id = !empty($facility_id_raw) ? reset($facility_id_raw) : '';
 
+      // Make sure we have something to prevent empty field errors.
+      if (empty($entity->title)) {
+        $entity->title = 'TBD';
+      }
+      if (empty($entity->field_address['address_line1'])) {
```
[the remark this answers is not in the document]
ethanteague: Ha! Just realized I have this set to optional in the content model, @swirtSJW, so little bit less to worry about here :+1:

docroot/modules/custom/va_gov_consumers/src/FacilityConsumer.php
```diff
@@ -17,19 +18,31 @@ class FacilityConsumer {
...
-    $client = new Client();
-    // Make sure we get a good response before heavy lifting.
-    try {
-      // @TODO  Restore the logic here when the env-API call is finalized.
-    }
-    // Record any trouble to watchdog.
-    catch (GuzzleException $e) {
-      watchdog_exception('Facility_content', $e->getMessage());
-    }
+    if (!empty(getenv('CMS_FACILITY_API_KEY'))) {
```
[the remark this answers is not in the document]
ethanteague: Updated @swirtSJW :+1:

docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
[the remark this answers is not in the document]
ethanteague: Updated @swirtSJW :+1:

docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
ethanteague: Updated @swirtSJW :+1:

swirtSJW:
This looks good.  Will test on Unity as soon as it finishes building.

swirtSJW:
I tested this on Unity   It failed smoothly as expected since Unity does not have access to the API.
image

PR #645, opened 2019-10-17
Remove static assets before building them again to prevent straggler html

jonpugh:
Please merge me ASAP :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -348,6 +348,7 @@
         ],
         "va:web:build": [
             "drush config-set va_gov.build web.build.pending 1 -y",
+            "rm -rf vendor/va-gov/web/build/localhost/",
             "cd web && npm run build -- --pull-drupal",
             "drush config-set va_gov.build web.build.pending 0 -y"
         ],


composer.json
```diff
@@ -348,6 +348,7 @@
         ],
         "va:web:build": [
             "drush config-set va_gov.build web.build.pending 1 -y",
+            "rm -rf vendor/va-gov/web/build/localhost/",
```
ElijahLynn: Is this always localhost? Even on BRD?

jonpugh: We don't run va:web:build in BRD

jonpugh: I'm very open to fixing this, but it's hard coded by the symlink at /docroot/static to this path as well, so I think it's ok for now.

jonpugh: Plus in BRD, this would never run twice on the same system, right?

This is really only used by systems that have to run va:web:build more than once, like behat tests, CMS-CI front-end builds, and localhost.

ElijahLynn: va:web:build is run by lando tests on brd in DEV/STAGING. But yeah, we don't really use that just yet, so this is good.

PR #649, opened 2019-10-18
VAGOV-000 Add configuration for Landos use of xdebug.

swirtSJW:


ElijahLynn:
Continues to work with IntelliJ/PHPStorm!

If you are feeling up to it, this would be good to get in too, a `lando xdebug-on` and `lando xdebug-off` option so we don't have to deal with uncommenting and rebuilding at all. https://github.com/lando/lando/issues/1668#issuecomment-507191275 (second option, apache). 

If not, this is good to merge!

swirtSJW:
@ElijahLynn  I like that suggestion.   I am just not clear on two parts of it that I will need your guidance on.  We can chat on Mon.

swirtSJW:
@ElijahLynn  Disregard my needing to discuss.  I figured it out.  The toggle is working for me and vscode.   No need to 'lando rebuild'.

PR #654, opened 2019-10-23
VAGOV-3950: Disable xmlsitemap module.

kevwalsh:
Do not merge until after #653 is released to prod, per https://va-gov.atlassian.net/browse/VAGOV-3950


swirtSJW:
I think we also want to remove xmlsitemap from composer json and rebuild the .lock

https://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/composer.json#L87

kevwalsh:
> I think we also want to remove xmlsitemap from composer json and rebuild the .lock
> 
> https://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/composer.json#L87

Indeed. i mistakenly assumed it was part of the distro.

PR #667, opened 2019-10-28
VAGOV-6185: Sidebars for Drupal node view pages with menus and metadata.

kevwalsh:


ElijahLynn:
The failure here was cut off due to the GH max comment length limit. 

Actual error is here: http://devshop.cms.va.gov/node/24668
 

>   │  stdClass Object
>       │  (
>       │      [wid] => 357230
>       │      [uid] => 1367
>       │      [type] => php
>       │      [message] => %type: @message in %function (line %line of %file) @backtrace_string.
>       │      [variables] => Array
>       │          (
>       │              [%type] => Warning
>       │              [@message] => Drupal\Core\Render\Markup Object
>       │                  (
>       │                      [string:protected] => assert(): Tokens need to be valid Twig variables. failed
>       │                  )

config/sync/views.view.right_sidebar_latest_revision.yml
```diff
@@ -1277,7 +1277,7 @@ display:
       displays:
         block_1: block_1
       display_description: ''
-      enabled: true
+      enabled: false
```
kevwalsh: i disabled to just disable this view for now, which was causing some twig php problems. i'm pretty sure it's related to this drupal core issue https://www.drupal.org/project/drupal/issues/2831233. this view display is not critical at all, so i've hidden it and re-surfaced those fields in the node display, instead of the sidebar, in one of the other commmits.

kevwalsh:
@swirtSJW i've added a couple new commits that need review, it's passing tests now.

composer.json
```diff
@@ -199,7 +199,7 @@
                 "Incompatibility with inline entity form": "https://www.drupal.org/files/issues/2018-10-30/field_group-inline_entity_form_compatibility-2986704-3.patch"
             },
             "drupal/path_redirect_import": {
-                "Add an option to import non-existent paths": "https://www.drupal.org/files/issues/2018-08-28/path_redirect_import-2995543-2.patch"
+                "Add an option to import non-existent paths": "https://www.drupal.org/files/issues/2019-09-17/path_redirect_import-2995543-4.patch"
```
swirtSJW: Thanks for catching this.

PR #681, opened 2019-11-01
VAGOV-6227: Create menus for 9 VAMC systems.

kevwalsh:
QA

- [ ] As admin, go to Manage > Content > Menus. You should see menus are now categorized by type, to help keep this manageable. The parent menus that are not links (`route:<nolink>`) should have the same padding and size as the the other menu items.
   - [ ] Look in the VAMC system menus, you should see 9 menus, one for each system in VISN 9
- [ ] Edit a couple of those menus. You should see that it has the Workbench menu access set (eg, /admin/structure/menu/manage/va-coatesville-health-care/access is set to use the corresponding Coatesville VA section).
- [ ] Go to /admin/structure/block, and scroll to the bottom. All VAMC system menus should appear, with similar config (start with 2nd level, and url-based visibility based on their system's respective URL structure)
- [ ] All VAMC-eligible content types should be enabled to have menu items added to the 9 VISN 4 VAMC system menus. You can check out a few by editing a couple VAMC content types and checking out their Menu settings (eg /admin/structure/types/manage/health_care_local_facility)


tests/behat/drupal-spec-tool/menus.feature
```diff
@@ -8,26 +8,34 @@ Feature: Menus
      Scenario: Menus
        Then exactly the following menus should exist
        | Name | Machine name | Description |
-| Administration | admin | Administrative task links |
```
kevwalsh: i sorted this by machine name, but really there are only 8 new rows.

swirtSJW: Thank you for sorting this.  It creates a little confusion now, but it is so much better long-term.

PR #685, opened 2019-11-06
VAGOV-000 Test config:import imports cleanly.

ElijahLynn:
This does add about 6 seconds to tests locally. It prevents unimportable config from being merged.

### DIFF
--- a/tests.yml
+++ b/tests.yml
@@ -14,15 +14,16 @@ va/deploy/0-composer:
 
 va/deploy/1-cache:
   description: Rebuild Drupal caches
-  command: drush $DRUSH_ALIAS cr
+  command: drush $DRUSH_ALIAS cache:rebuild
 
 va/deploy/2-update:
   description: Run Drupal Update Hooks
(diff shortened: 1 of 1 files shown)


ElijahLynn:
Hrm, it should actually fail on the va/deploy/3-config step right now because we have config that isn't importing cleanly in master now. Once this fails, we can fix that file, va.gov-cms/config/sync/node_revisions_autoclean.settings.yml, and run tests again. Perfect scenario actually!

ElijahLynn:
@swirtSJW Just realized that I had tested the check for drush config:import cleanly was using Drush 9 (needed for Tome)and there is no drush:status command in Drush 8. https://github.com/department-of-veterans-affairs/va.gov-cms/pull/685

I think that is why it passed when it should have failed, because the grep didn't find that string in "The drush command 'config:status' could not be found.". Will have to figure out a way to do this same thing with Drush 8 for right now (we are locked to Drush 8 because DevShop doesn't support Drush 9 yet).

PR #705, opened 2019-11-13
VAGOV-000 Ignore samlsessiondb.sq3 for all users on composer install hook.

ElijahLynn:
This keeps samlsessiondb.sq3 from always showing up in git status.


ElijahLynn:
@todo Before merge: need to remove this from https://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/READMES/getting-started.md too. Can't do it now.

PR #698, opened 2019-11-11
WIP: VAGOV-6257: Computed field for VAMC banner alerts.

kevwalsh:

PR #712, opened 2019-11-16
VAGOV-6257 computed field

ethanteague:
## What this does

- Leverages computed field config in https://github.com/department-of-veterans-affairs/va.gov-cms/pull/698
- Stores json array to field_banner_alert_computdvalues table containing the following values
-- Name of VAMC
-- URL of VAMC
-- URL of Operating status page
-- Granicus topic id
- Requires `field_link_facility_emerg_list` & `field_operating_status` to be set on Vamc system node

## Screenshots
image

## Questions
@kevwalsh I changed this field from a multivalue text field to a single longtext field - this makes it a lot easier to work with / output in graphql json. Is this okay with you?


docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -206,3 +207,44 @@ function va_gov_backend_rebuild() {
 
   \Drupal::state()->set("environment_indicator.current_release", $indicator);
 }
+
+/**
+ * Implements hook_entity_presave().
+ */
+function va_gov_backend_entity_presave(EntityInterface $entity) {
+
```
swirtSJW: This hook contains a lot of logic that also needs to leave the door open for using the same hook to  do other things.  It would be nice if the logic you added resided in a single function.   So this hook would simply call `_va_gov_backend_build_alert_computed_field_data($entity);

Advantages:
1. The function would have a docblock that would describe its purpose.
2. The logic is contained better so it does not become interwoven with other stuff added to hook presave.
3) sets a pattern for others to follow in adding stuff to a hook.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -206,3 +207,44 @@ function va_gov_backend_rebuild() {
...
+    $query = $database->select('node__field_operating_status', 'n');
+    $query->fields('n', ['entity_id']);
+    $query->condition('n.field_operating_status_uri', $vamcs_ids, 'IN');
+    $results = $query->execute()->fetchCol();
+
+    $vamcs_loaded = \Drupal::entityTypeManager()->getStorage('node')->loadMultiple($results);
+    $computed_return = [];
+    foreach ($vamcs_loaded as $key => $vamc_single) {
+      // Pull the vha id out of the path string.
+      $vamcs_topic_id = explode('?topic_id=', $vamc_single->get('field_link_facility_emerg_list')->getString());
```
swirtSJW: Concern:  This assumes that topic_id= is going to be there AND that it is the first AND only thing.  That makes it a bit brittle.   You might want to pass the $vamc_single->get('field_link_facility_emerg_list')->getString() to parse_url() and then from there use parse_string.   Something like 
```
$query = parse_url($vamc_single->get('field_link_facility_emerg_list')->getString(), PHP_URL_QUERY);
$params = [];
parse_str($query, $params);
$topic_id = (!empty($params['topic_id'])) ? $params['topic_id'] : FALSE;
```

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -206,3 +207,44 @@ function va_gov_backend_rebuild() {
...
+      // Get our op status uri out of nested array for building path alias.
+      $vamcs_op_status_raw = $vamc_single->get('field_operating_status')->getValue();
+      $vamcs_op_status_reset = !empty($vamcs_op_status_raw) ? reset($vamcs_op_status_raw) : '';
```
swirtSJW: Minor: "reset" tells you about the process of how it got there, but tells nothing of what the variable represents.   I suggest, `$vamcs_op_status_entry`   Tells what it is without the baggage of how it got there.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: Great point, @swirtSJW :+1: - doing that.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: You're right, @swirtSJW - gonna do that :+1:

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: Works for me, @swirtSJW :+1:

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -206,3 +207,56 @@ function va_gov_backend_rebuild() {
...
+    foreach ($vamc_op_nodes as $vamc_op_node) {
+      $computed_return['vamc_op_status_path'][] = \Drupal::service('path.alias_manager')->getAliasByPath('/node/' . $vamc_op_node->id());
+      $vamc_office_values[] = $vamc_op_node->get('field_office')->getString();
+    }
+
+    // Grab what we need from our vamcs.
+    $vamc_system_nodes = $node_storage->loadMultiple($vamc_office_values);
+    foreach ($vamc_system_nodes as $key => $vamc_system_node) {
+      // Pull the vha id out of the path string.
+      $query = parse_url($vamc_system_node->get('field_link_facility_emerg_list')->getString(), PHP_URL_QUERY);
```
swirtSJW: I like the improved logic, but still seems like a brittle data model.  Checking with Kev on why we can't have this value in a field.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -206,3 +207,56 @@ function va_gov_backend_rebuild() {
...
+      $params = [];
+      parse_str($query, $params);
+      $computed_return['vamc_topic_id'][] = (!empty($params['topic_id'])) ? $params['topic_id'] : FALSE;
```
swirtSJW: Seems odd that these are going to have multiple values, each disconnected from each other.  If one of them happens to be NULL you might end up with data that no longer aligns with itself.
Seems like this would group them together better:
```suggestion
      $computed_return[$key]['vamc_topic_id'] = (!empty($params['topic_id'])) ? $params['topic_id'] : FALSE;
```

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -206,3 +207,56 @@ function va_gov_backend_rebuild() {
...
+      $computed_return['vamc_title'][] = $vamc_system_node->getTitle();
```
swirtSJW:
```suggestion
      $computed_return[$key]['vamc_title'] = $vamc_system_node->getTitle();
```

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -206,3 +207,56 @@ function va_gov_backend_rebuild() {
...
+      $computed_return['vamc_path'][] = $vamc_system_node->toUrl()->toString();
```
swirtSJW:
```suggestion
       $computed_return[$key]['vamc_path'] = $vamc_system_node->toUrl()->toString();
```

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -206,3 +207,56 @@ function va_gov_backend_rebuild() {
...
+  _va_gov_backend_build_alert_computed_field_data($entity);
+}
+
+/**
+ * Populates computed field on banner alert nodes.
+ *
+ * @param object $entity
+ *   The node object.
+ */
+function _va_gov_backend_build_alert_computed_field_data($entity) {
```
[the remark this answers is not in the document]
ethanteague: I've wondered about this in the past, @swirtSJW - is this a readability thing, or does indicating the type pass in some extra magic?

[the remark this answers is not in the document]
ethanteague: :cry: doesn't pass code sniffer test, @swirtSJW - gonna leave out for now.

swirtSJW:
We've got failing behat and config.  :(

docroot/modules/custom/va_gov_backend/va_gov_backend.module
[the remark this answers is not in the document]
swirtSJW: Type casting sets the contract for what the function is expecting.  It is a good practice for coding.
It doesn't pass code sniffer because if you declare the type in the function you have to also declare it in the docblock.

composer.json
```diff
@@ -23,6 +23,7 @@
         "drupal/cer": "^4.0@alpha",
         "drupal/coder": "~8.2.12",
         "drupal/components": "^1.0",
+        "drupal/computed_field": "^2.0",
```
swirtSJW: This is added, but I don't see an accompanying composer.lock change

docroot/modules/custom/va_gov_backend/va_gov_backend.module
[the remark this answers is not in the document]
ethanteague: Ah, my bad @swirtSJW - that got it :+1:

config/sync/field.field.node.full_width_banner_alert.field_banner_alert_computdvalues.yml
```diff
@@ -0,0 +1,24 @@
...
+label: 'Computed values for alerts'
+description: 'The name of affected VAMC system, its path, its Granicus topic ID, and the URL of its Operating Status page. '
+required: false
+translatable: false
+default_value:
+  -
+    value: ''
+default_value_callback: ''
+settings:
+  code: '$value = '''';'
```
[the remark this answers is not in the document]
ethanteague: This is tying the computed field to our banner content type, @swirtSJW - seems legit to me. Do you have a different worry about this / something about the contents of this config file?

composer.json
ethanteague: Yep - I messed up the rebase for said file, @swirtSJW

config/sync/field.field.node.full_width_banner_alert.field_banner_alert_computdvalues.yml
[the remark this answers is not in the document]
ElijahLynn: This file/field is named ...computdvalues.yml, should we name the field something more semantic about what it actually is? If not we should rename it to be "computed_values" at the least.

ElijahLynn:
This file/field is named `field_banner_alert_computdvalues`, should we name the field something more semantic about what it actually is? If not we should rename it to be "computed_values" at the least (or is this a max field name limit)?

composer.lock
```diff
@@ -5689,8 +5757,8 @@
                     "dev-3.x": "3.x-dev"
                 },
                 "drupal": {
-                    "version": "8.x-3.0-rc1+18-dev",
-                    "datestamp": "1572476284",
+                    "version": "8.x-3.0-rc1+15-dev",
```
swirtSJW: Why the downgrade?

swirtSJW:
Code review looks good.  I just need to test it in CI.

swirtSJW:
I tested adding and updating the full width banner on http://pr712.ci.cms.va.gov/node/add/full_width_banner_alert and then checked the log for any related warnings.  It was clear.

swirtSJW:
Nice work Ethan,  I know it was a bit of a long road.  Good job seeing it all through.

PR #727, opened 2019-11-22
VAGOV-000 Add `composer clear-cache` to `composer nuke` command.

ElijahLynn:
To test run `hub pr checkout 727` then `composer nuke` and you should see some composer directories emptied.

```                                                                                           
❯ lando composer nuke                                                                                                                            
> rm -rf bin node_modules docroot/libraries docroot/core docroot/modules/contrib docroot/themes/contrib docroot/profiles/contrib/ docroot/vendor 
> composer clear-cache                                                                                                                           
Cache directory does not exist (cache-vcs-dir):                                                                                                  
Clearing cache (cache-repo-dir): /var/www/.composer/cache/repo                                                                                   
Clearing cache (cache-files-dir): /var/www/.composer/cache/files                                                                                 
Clearing cache (cache-dir): /var/www/.composer/cache                                                                                             
All caches cleared.       
```


composer.json
```diff
@@ -36,6 +36,7 @@
         "drupal/fast_404": "^1.0@alpha",
         "drupal/feature_toggle": "^1.2",
         "drupal/field_group": "dev-3.x#b245b85f99c48221edf5fb60344dfe53737ae0f9",
+        "drupal/govdelivery_bulletins": "1.x-dev",
```
swirtSJW: Unexpected.  Should this be here.

composer.json
```diff
@@ -99,7 +100,7 @@
         "symfony/finder": "^3",
         "symfony/phpunit-bridge": "^4.3",
         "twig/twig": "~1.37.1",
-        "va-gov/web": "dev-master",
+        "va-gov/web": "dev-VAGOV-5991-banners-et",
```
swirtSJW: Unexpected.  We don't want this change to go out do we?

composer.json
ElijahLynn: Nice catch, nope shouldn't be here.

composer.json
ElijahLynn: Nice catch, nope shouldn't be here.

composer.json
ElijahLynn: Force pushed change

composer.json
ElijahLynn: Force pushed change

ElijahLynn:
@swirtSJW This is ready for merge now, next week.

PR #731, opened 2019-11-26
VAGOV-6306 Add deploy permission to content_publisher.

swirtSJW:


config/sync/user.role.content_publisher.yml
```diff
@@ -157,6 +157,7 @@ permissions:
   - 'use moderation sidebar'
   - 'use text format rich_text'
   - 'use workbench access'
+  - 'va gov deploy content build'
```
jefflbrauer: Looks good (assuming that permission name is accurate) I didn't review the previous PR in which you named the permission.

PR #732, opened 2019-11-26
VAGOV-6252 Add GovDelivery banner alert processing. Part 1

swirtSJW:
This has a lot of items affected and does not fully get the queue working but it does the following:

1. Removes workbench moderation from the full_width_banner_alert content type.
1. Disables the send email checkbox upon node being published
image

2. Utilizes the  "publish" checkbox 
image

3. removes the publish checkbox from situation update paragraph.
image

4. Adds the logic to decide when to queue  status alert bulletins and when to queue a situation update bulletin.
5.  Logging of adding items to the queue.
image

6. Adds all current features of govdelivery_bulletins (perms, queue adding, configuration killswitches, endpoint triggering)
https://www.drupal.org/project/issues/govdelivery_bulletins?status=7
7. Can visit /api/govdelivery_bulletins/queue?EndTime=1234567899 to get a 200 response if the EndTime timestamp is valid.
**Valid**
image

**Not valid**
image

**Not present**
image

What is does NOT do.
1 Actually process the queue (no sending, no clearing out of the queue.
https://www.drupal.org/project/issues/govdelivery_bulletins?
2. Formatting content if the email message with twig templates.
3.  automated testing.

### DIFF
--- a/.lando/xdebug/php.ini
+++ b/.lando/xdebug/php.ini
@@ -8,5 +8,5 @@ xdebug.collect_params = 0
 xdebug.remote_enable = 1
 xdebug.remote_autostart = 1
 xdebug.remote_host = ${LANDO_HOST_IP}
-; xdebug.remote_connect_back = 1
+xdebug.remote_connect_back = 1
 xdebug.remote_log = /tmp/xdebug.log
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -5,7 +5,6 @@
  * Contains va_gov_backend.module.
  */
 
-use Drupal\Core\Entity\EntityInterface;
 use Drupal\node\NodeInterface;
 use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Field\WidgetBase;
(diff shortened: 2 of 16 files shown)


docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -89,12 +88,7 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
       [':date1' => $date1, ':date2' => $date2]);
     }
   }
-  if ($form_id === 'node_full_width_banner_alert_edit_form') {
```
swirtSJW: Moved all this to va_gov_govdelivery module.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -207,52 +201,3 @@ function va_gov_backend_rebuild() {
 
   \Drupal::state()->set("environment_indicator.current_release", $indicator);
 }
-
```
swirtSJW: Moved all this to va_gov_govdelivery module.

.lando/xdebug/php.ini
```diff
@@ -8,5 +8,5 @@ xdebug.collect_params = 0
 xdebug.remote_enable = 1
 xdebug.remote_autostart = 1
 xdebug.remote_host = ${LANDO_HOST_IP}
-; xdebug.remote_connect_back = 1
+xdebug.remote_connect_back = 1
```
ElijahLynn: Should this be in this PR?

swirtSJW: Yes it only has an effect if somebody runs `lando xdebug-on` which can't be run on production or any CI environs.

swirtSJW:
Treating two partial approvals (one of code, one of config) as a full approval.  Tests are passing.

PR #738, opened 2019-11-28
VAGOV-6313: Change user admin role name and improve user admin view.

kevwalsh:
User story: https://va-gov.atlassian.net/browse/VAGOV-6313 

People___Veterans_Affairs


kevwalsh:
I can't make much sense of the errors at http://devshop.cms.va.gov/node/26225

PR #747, opened 2019-12-02
VAGOV-6319: Webform for CMS request form.

kevwalsh:


config/sync/webform.webform_options.marital_status.yml
```diff
@@ -0,0 +1,14 @@
+uuid: 8e267815-77e7-4b87-a7b2-c9c2bb35535a
+langcode: en
+status: true
+dependencies:
+  enforced:
+    module:
+      - webform
+_core:
+  default_config_hash: cjVDbAE5BlopGrY-tfxsxMoyEibwBBNgdQwDdxOSkow
+id: marital_status
```
swirtSJW: ?

kevwalsh: i used an existing field-collection type thing that ships with webform, but then disabled most of the fields. this must be one of the disabled ones.

PR #753, opened 2019-12-04
VAGOV-6328: Only content_admin can edit URLs.

kevwalsh:
This moves one permission from content_publisher to content_admin


kevwalsh:
@ElijahLynn Looks like a bad db again (serialization failures etc on the phpunit tests). In cases like this, what should i do?

PR #754, opened 2019-12-04
Disable automatic STAGING deploys temporarily

ElijahLynn:


ElijahLynn:
When we merge this we need to keep a close eye on the autodeploy job and the STAGING deploy job, the STAGING deploy job should not kick off and if it does it _must_ be cancelled immediately.

ElijahLynn:
Nice, this worked great, after the build it just triggered DEV, not STAGING too. 

image

We need to revert this commit next week when we put STAGING back to normal.

PR #757, opened 2019-12-06
VAGOV-6175 Fix address field to preserve data.

swirtSJW:


ElijahLynn:
> This is an annoying failure on php unit.
> image

Yeah, performance testing is a bit tricky, to do it right they ideally need a bare metal box that has zero other load on it to guarantee comparable resources. If the server is under load it could cause this failure OR application code regression could cause it too. For now, if it happens again then it likely is application code regression.

PR #768, opened 2019-12-10
VAGOV-6253: Adding patch to address false 0 = empty error.

ethanteague:
## What this does
 - When a "0" is entered in a table cell, tablefield validation function considers 0 string to be an empty value. We need to allow 0's in cells, and this patch modifies validation function to account for this.

## QA
 - Visit `node/745/edit` and open a table in the `Main Content` section. Enter a "0" in one table cell, and save, visually verifying that form errors **do not** prevent the form from submitting.
 - In same form, enter an empty value in a table cell, and save, visually verifying that form errors **do** prevent the form from submitting.


ethanteague:
> This patch seems to be doing a lot more short circuiting than just allowing integer 0 values.
> https://www.drupal.org/files/issues/2019-12-10/0-value-throwing-empty-error.patch
> I think there may be some unintented consequences to the reduction in logic.
> 
> Seems to me that changing
> ` if (empty($cell)) {`
> to
> `if ($cell === "") {`
> Would be all that is needed. Did the count need to be killed off?
That would be enough, but the count is always returning empty (I checked multiple scenarios) - it's akin to having a check like if `1==1`. Glad to leave it in if you think better safe than sorry, @swirtSJW , but seems to me like this is a holdover from a different path the developer was on in a previous iteration.

swirtSJW:
The count returns empty even if there is content in the field?   That seems odd.

PR #769, opened 2019-12-10
VAGOV-6252 Add template and processing for bulletin body.

swirtSJW:


docroot/modules/custom/va_gov_govdelivery/templates/va-gov-body-alert.html.twig
```diff
@@ -0,0 +1,135 @@
...
+	}
+	.rss_description {
+		color: #333333;
+		font-family: myriad pro, calibri, arial, sans serif;
+		font-size: 19px;
+	}--></style>
+<table border="0" width="700" cellspacing="0" cellpadding="0" align="center" bgcolor="#0482BE">
+	<tbody>
+		<tr>
+			<td colspan="3" bgcolor="#C4262E">&nbsp;</td>
```
ethanteague: Should we be using old school html attributes for styling, @swirtSJW ?

docroot/modules/custom/va_gov_govdelivery/src/Service/ProcessStatusBulletin.php
```diff
@@ -31,52 +31,62 @@ public function processNode(NodeInterface $node) {
...
       // Pull the data from the alert fields.
-      $body = $node->get('field_body')->getString();
-      $type = $node->get("field_alert_type")->getString();
-      $header = '';
-      $footer = '';
-      $subject = $node->get('title')->getString();
-      // @TODO  get the time the alert was saved.
+      $template_variables['message'] = $node->get('field_body')->value;
+      $template_variables['situation_update'] = FALSE;
+      // "Alert - White River Junction VA Medical Center".
```
ethanteague: Did you leave this comment to give an output example, or was this inadvertent, @swirtSJW ?

docroot/modules/custom/va_gov_govdelivery/templates/va-gov-body-alert.html.twig
swirtSJW: No.  We are trying to preserve as much of the original formatting as we can since presumably a bunch of testing went into the format of the send... and we are just doing as much as a lift and shift as possible.   We don't want to take on a bunch of email client testing.

docroot/modules/custom/va_gov_govdelivery/src/Service/ProcessStatusBulletin.php
swirtSJW: I wanted to leave it in there to preserve what it should look like.... but there is no good reason for that.  I'll remove it.

docroot/modules/custom/va_gov_govdelivery/templates/va-gov-body-alert.html.twig
swirtSJW: Sorry I answered your question backwards... yes we should be using the old method, because we assume the old method was tested.  Updating to improved method can wait until it is a feature improvement.

ethanteague: Gotcha, @swirtSJW - makes sense :+1:

docroot/modules/custom/va_gov_govdelivery/src/Service/ProcessStatusBulletin.php
```diff
@@ -31,52 +31,61 @@ public function processNode(NodeInterface $node) {
...
         // Add the item to queue.
         \Drupal::service('govdelivery_bulletins.add_bulletin_to_queue')
           ->setFlag('dedupe', TRUE)
           ->setQueueUid($queue_id)
-          // @TODO run body through a twig template.
           ->setBody($body)
-          ->setFooter($footer)
+          ->setFooter(NULL)
           ->setFromAddress('us@example.org')
           ->setGovDeliveryID('some_unique_id')
```
swirtSJW:
```suggestion

```

PR #777, opened 2019-12-11
VAGOV-6065: Adding user_history module and patch.

ethanteague:
## What this does
 - Adds a user history tab to user pages for tracking user creation and editing events
 - Adds a user history list page: `/user_history/list`

## QA / Installation
 - Go to `/user_history/initialise` or visit the status report and click on the link.
 - Visit user edit page, look for history tab.
 - Visually verify that creation / editing events are being logged.
 - Go to `user_history/list` and visually verify that all users and account history appears

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,6 +79,7 @@
         "drupal/toolbar_menu_clean": "^1.0",
         "drupal/tour_ui": "^1.0@beta",
         "drupal/ui_patterns": "^1.0",
+        "drupal/user_history": "^1.0@alpha",
         "drupal/uswds": "^1.0@beta",
         "drupal/views_bulk_edit": "^2.3",
         "drupal/views_bulk_operations": "^2.5",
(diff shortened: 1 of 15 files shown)


kevwalsh:
@ElijahLynn can you look at this phpunit fail? looks like a blip to me.

ElijahLynn:
It is accurate it appears. Need to enable module in config. 

> The module user_history does not exist.

composer.json
```diff
@@ -240,6 +241,9 @@
             "drupal/textfield_counter": {
                 "Prevent form submission when character limit exceeded option does not work": "https://www.drupal.org/files/issues/2018-10-04/textfield_counter-prevent_form_submission_does_not_work-3004457-2.patch"
             },
+            "drupal/user_history": {
+                "Invalid argument in foreach when saving a user": "https://www.drupal.org/files/issues/2019-12-11/invalid-argument-on-save_0.patch"
```
swirtSJW: I checked out the Patch.   Looks good.

config/sync/core.extension.yml
```diff
@@ -152,6 +152,7 @@ module:
   toolbar_menu: 0
   update: 0
   user: 0
+  user_history: 0
```
swirtSJW: Shouldn't user_created_by show up here too?

composer.json
```diff
@@ -79,6 +79,7 @@
         "drupal/toolbar_menu_clean": "^1.0",
         "drupal/tour_ui": "^1.0@beta",
         "drupal/ui_patterns": "^1.0",
+        "drupal/user_history": "^1.0@alpha",
```
swirtSJW: Should user_created_by be here too?

config/sync/core.extension.yml
ethanteague: user_created_by shouldn't be here - leftover from a previous project usage attempt. Removed, @swirtSJW

composer.lock
```diff
@@ -19217,6 +19212,65 @@
                 "source": "https://git.drupalcode.org/project/media_entity_generic"
             }
         },
+        {
+            "name": "drupal/user_created_by",
```
swirtSJW: I think this needs to go away.

ethanteague: Nice catch, @swirtSJW :+1: - removed

PR #779, opened 2019-12-11
VAGOV-6347 Add dependabot config.yml for auto-updates to vets-website daily w/auto merge

ElijahLynn:
VAGOV-6347

https://dependabot.com/docs/config-file/

* Security updates
* va.gov/web (vets-website) update, automerged if status checks pass
* Daily


ElijahLynn:
This doesn't appear to be being picked up by Dependabot and is not updating the `va.gov/web` dependency. I wonder if there is a repo validator.

PR #784, opened 2019-12-12
VAGOV-6350 Add basic auth to govdelivery queue endpoint.

ElijahLynn:
VAGOV-6350

### DIFF
--- a/docroot/modules/custom/va_gov_govdelivery/src/Routing/RouteSubscriber.php
+++ b/docroot/modules/custom/va_gov_govdelivery/src/Routing/RouteSubscriber.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace Drupal\va_gov_govdelivery\Routing;
+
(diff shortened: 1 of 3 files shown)


docroot/modules/custom/va_gov_govdelivery/src/Routing/RouteSubscriber.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+class RouteSubscriber extends RouteSubscriberBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function alterRoutes(RouteCollection $collection) {
+
+    $route_provider = \Drupal::service('router.route_provider');
+    $graphql_routes = $route_provider->getRoutesByPattern('api/govdelivery_bulletins/queue');
```
swirtSJW: These need to be named more correctly. ... but we may move them into the contrib module.... so will call it good for now.

PR #794, opened 2019-12-17
VAGOV-6354: Fixing node form layout display issue.

ethanteague:
## What this does
Fixes display issue on nodeforms when textfields have lots of characters

## Screenshot
image


docroot/themes/custom/vagovadmin/assets/scss/styles.scss
```diff
@@ -88,6 +88,7 @@
   cursor: pointer;
 }
 
-
-
-
+.layout-region-node-main,
+.layout-region-node-footer {
+  display: grid !important;
```
swirtSJW: I don't like forcing this with important unless we have to ;) .   Is there anything unrelated that you can think of that this might break?

ethanteague: Yeah, you're right, @swirtSJW - I took out the important and replaced with more specific targetting.

PR #797, opened 2019-12-17
VAGOV-6290: Removing deprecated vamc system fields.

ethanteague:
## What this does
- Removes legacy fields from vamc system
- Depends on https://github.com/department-of-veterans-affairs/vets-website/pull/11338 first being merged and deployed to prod


swirtSJW:
Looks like we have a conflict here.

config/sync/core.entity_form_display.node.health_care_region_page.default.yml
```diff
@@ -573,14 +567,4 @@ content:
...
-  created: true
-  field_email_subscription: true
-  field_email_subscription_links: true
-  field_link_facility_emerg_list: true
-  field_link_facility_news_list: true
-  field_links: true
-  field_sign_up_for_emergency_emai: true
-  promote: true
-  sticky: true
-  uid: true
```
swirtSJW: I think these as well as 'created' got caught in the crossfire.

ethanteague: Yeah? I'll admit I'm confused - I had a conflict I resolved where codebase didn't know if these fields should be hidden or visible, so I defaulted to visible.....if you say they should be visible, I defer to you, @swirtSJW - can you confirm?

swirtSJW: Yah I think Kev wants those to remain hidden.

ethanteague: Cool - thanks for catching that, @swirtSJW ! Just updated.

config/sync/core.entity_form_display.node.health_care_region_page.default.yml
```diff
@@ -567,4 +567,14 @@ content:
     region: content
     settings: {  }
     third_party_settings: {  }
-hidden: {  }
+hidden:
+  created: true
+  field_email_subscription: true
+  field_email_subscription_links: true
+  field_link_facility_emerg_list: true
+  field_link_facility_news_list: true
```
swirtSJW: Aren't these supposed to be going away?  I think ony created, promote, sticky, and promote and uid should remain.

ethanteague: Yep, @swirtSJW - updated

config/sync/core.entity_form_display.node.health_care_region_page.default.yml
kevwalsh: yep, hidden. i made this change recently.

PR #799, opened 2019-12-18
VAGOV-6252 va bulletins  updated

swirtSJW:


config/sync/govdelivery_bulletins.settings.yml
```diff
@@ -1,3 +1,6 @@
 api_queue_trigger_enabled: 1
 enable_bulletin_queuing: 1
-enable_bulletin_queue_sends_to_govdelivery: 0
+enable_bulletin_queue_sends_to_govdelivery: 1
+govdelivery_endpoint: 'https://stage-api.govdelivery.com/api/account/USVHA/bulletins/send_now'
+govdelivery_username: jeffrey.barnes4@va.gov
```
swirtSJW: Hmm these should not be showing up here.   They need to be config ignored.

swirtSJW:
The pr environment for this seemed completely broken so I just destroyed it.  Waiting for it to recreate itself.

PR #778, opened 2019-12-11
VAGOV-000 `lando composer update va-gov/web`

ElijahLynn:

PR #801, opened 2019-12-20
VAGOV-5133 Unhide password reset tab, email sending works now.

ElijahLynn:
VAGOV-5133

To test on BRD environments (CI doesn't work just yet) you must set the users email to something, as they are all empty by default on DEV/STAGING due to DB sanitization. PROD should work out of the box now.

The sitename in this PR changes it from "Veterans Affairs" to "VA.gov CMS".

### DIFF
--- a/config/sync/system.site.yml
+++ b/config/sync/system.site.yml
@@ -1,7 +1,7 @@
 uuid: 45814d2f-3e31-4878-aa12-733ea60ff326
-name: 'Veterans Affairs'
+name: 'VA.gov CMS'
 mail: vacmssupport@va.gov
-slogan: 'VA.gov CMS'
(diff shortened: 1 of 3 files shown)


ElijahLynn:
Needs https://github.com/department-of-veterans-affairs/va.gov-cms/pull/778/files merged for this to have a chance at passing.

ElijahLynn:
Legit failure in Behat tests because I changed the sitename in this PR so email sending would use it and make more sense in the emails. 

> And I should see "Veterans Affairs" in the "title" element
>                                                  # Drupal\DrupalExtension\Context\MinkContext::assertElementContainsText()
> 
>       The text "Veterans Affairs" was not found in the text of the element matching css "title". (Behat\Mink\Exception\ElementTextException)
>

PR #803, opened 2019-12-23
VAGOV-6252 GovDelivery updates.

swirtSJW:


docroot/modules/custom/va_gov_govdelivery/va_gov_govdelivery.services.yml
```diff
@@ -2,7 +2,3 @@ services:
   va_gov_govdelivery.process_status_bulletin:
     class: Drupal\va_gov_govdelivery\Service\ProcessStatusBulletin
     arguments: []
-  va_gov_govdelivery.route_subscriber:
-    class: Drupal\va_gov_govdelivery\Routing\RouteSubscriber
-    tags:
-      - { name: event_subscriber }
```
swirtSJW: This was all moved into the contrib module govdelivery_bulletins.

docroot/modules/custom/va_gov_govdelivery/src/Routing/RouteSubscriber.php
```diff
@@ -1,30 +0,0 @@
-<?php
```
swirtSJW: All moved to contrib

PR #804, opened 2019-12-24
VAGOV-0000 Bump va-gov/web to the latest commit.

swirtSJW:


swirtSJW:
Tests passing.  Moving this along.

PR #805, opened 2019-12-29
VAGOV-6268 GovDelivery Tests

swirtSJW:


docroot/modules/custom/va_gov_govdelivery/src/Service/ProcessStatusBulletin.php
```diff
@@ -75,7 +75,15 @@ public function processNode(NodeInterface $node) {
...
         $template_variables['ops_page_url'] = $vmac->vamc_op_status_path;
-        $body = (string) twig_render_template(drupal_get_path('module', 'va_gov_govdelivery') . '/templates/va-gov-body-alert.html.twig', $template_variables);
+        if (PHP_SAPI === 'cli') {
+          // The function twig_render_template throws a 'context not set'
+          // exception when run from CLI.
+          // Set a limited mock body for phpUnit Tests.
+          $body = "This is a phpUnit body that should never be sent to GovDelivery.";
+        }
+        else {
+          $body = (string) twig_render_template(drupal_get_path('module', 'va_gov_govdelivery') . '/templates/va-gov-body-alert.html.twig', $template_variables);
```
ethanteague: This seems kind of peculiar, @swirtSJW - twig_render_template is supposed to return a string, so casting it as a string should be unnecessary.

tests/phpunit/BulletinQueueTest.php
```diff
@@ -0,0 +1,138 @@
...
+    $this->deleteQueue();
+
+    // Save an unpublished node that should not create a bulletin.
+    $node = $this->createNode(['type' => 'full_width_banner_alert']);
+    $node->set('title', 'PHPUnit Test Alert');
+    $node->set('uid', '1');
+    $node->set('field_operating_status_sendemail', '1');
+    $node->set('field_alert_type', 'warning');
+    $node->set('field_body', 'This is a test created by phpUnit.  Please disregard.');
+    $node->field_banner_alert_vamcs->target_id = 1010;
```
ethanteague: Seems kind of brittle to hardcode a value, but understand there might not be a way around this, @swirtSJW

docroot/modules/custom/va_gov_govdelivery/src/Service/ProcessStatusBulletin.php
swirtSJW: @ethanteague  Yah the function is weird, it returns a markupinterface which needs to be cast to a string.
image

tests/phpunit/BulletinQueueTest.php
swirtSJW: It is a bit brittle... but otherwise it has to look one up.... which is more intensive... plus if we lose Pittsburgh, we've probably been fired ;)

swirtSJW: We would have to look it up by name, which would be just as brittle.

docroot/modules/custom/va_gov_govdelivery/src/Service/ProcessStatusBulletin.php
ethanteague: :+1: @swirtSJW

tests/phpunit/BulletinQueueTest.php
ethanteague: Lol, makes sense to me, @swirtSJW :+1:

PR #806, opened 2020-01-02
VAGOV-6367: Add VAMC system official name and update form.

kevwalsh:
Update the top few fields for VAMC system, and remove some of the unnecessary fields in right sidebar. 

image


config/sync/field.field.node.health_care_region_page.field_vamc_system_official_name.yml
```diff
@@ -0,0 +1,19 @@
...
+dependencies:
+  config:
+    - field.storage.node.field_vamc_system_official_name
+    - node.type.health_care_region_page
+id: node.health_care_region_page.field_vamc_system_official_name
+field_name: field_vamc_system_official_name
+entity_type: node
+bundle: health_care_region_page
+label: 'VAMC system official name'
+description: 'Eg: "VA Pittsburgh Healthcare System". This official name is used sparingly on some page templates. If this VAMC system does not have an official name, those templates will just use the VAMC system name above. '
```
swirtSJW: Inconsistent use of 'healthcare' with the one below

config/sync/node.type.health_care_region_page.yml
```diff
@@ -19,7 +19,7 @@ third_party_settings:
       - va-wilmington-health-care
     parent: 'pittsburgh-health-care:'
   node_title_help_text:
-    title_help: 'Eg "Pittsburgh VA health care"'
+    title_help: 'Eg "VA Pittsburgh health care". This name should follow the convention: "VA [geographic region or city]  health care".'
```
swirtSJW: 'health care' vs 'healthcare'  used above.   Not sure which is right.

config/sync/field.field.node.health_care_region_page.field_vamc_system_official_name.yml
kevwalsh: This is on purpose.   "healthcare" is the official name, "health care" is the plain language name.

swirtSJW: I under stand... or understand.  ;)

kevwalsh:
I've had to revert one commit, which breaks web builds. We'll have to do some FE investigation before deleting that one field.

PR #809, opened 2020-01-08
VAGOV-6372: Adding & enabling password_strength and mass_pwreset.

ethanteague:
## What this does
 - Adds and enables password_strength module - password_policy module remains enabled, as the former lists the latter as a dependency. Decision to augment password hardening process with password_strength is predicated by the enhanced security, and "it just works." From the readme:

> Instead of checking on strict rules, Password Strength classifies the expected brute-force time for the summed entropy of common underlying patterns in the password. Patterns that can be detected in passwords include:
> 
>     Words that are found in a dictionary of common words, common first and last names, or common passwords.
>     Words that are found in the dictionary, but with common "1337" or "leet" substitutions, such as 4 or @ for a, and 5 for s.
>     Common sequences of letters (abcde), numbers (12345), or characters spatially near each other on common keyboards (qwerty).
>     Three or more of the same characters, such as "aaa" or "8888".
>     Dates or years, such as "1921" or "19-11-1978."

 - Adds and enables mass_pwreset project for resetting all user passwords on live (also sends out an email using the default drupal reset password form).

## Implementation
**This will be handled by @howardbrande is https://va-gov.atlassian.net/browse/VAGOV-6373**
1. After this pr is deployed to prod, change password change email verbiage (or don't - it's probably fine, as is): `admin/content-models/users` click on the `Password recovery` vertical tab, and save.
2. Go to `/admin/people/mass-pwreset` check the ` Check all users` box in `Authenticated Role` fieldset, then check `Notify active users of password reset via email` box in `Notify Users` fieldset, then press `Reset Passwords` button


composer.lock
```diff
@@ -5785,8 +5806,8 @@
                     "dev-3.x": "3.x-dev"
                 },
                 "drupal": {
-                    "version": "8.x-3.0-rc1+20-dev",
-                    "datestamp": "1573591272",
+                    "version": "8.x-3.0-rc1+15-dev",
```
swirtSJW: Seems like this is going backwards.

composer.lock
```diff
@@ -18729,11 +18842,11 @@
                 "webflo/drupal-finder": "^1.1"
             },
             "require-dev": {
-                "behat/mink-selenium2-driver": "1.4.0 | 1.3.1 | 1.3.x-dev",
+                "behat/mink-selenium2-driver": "1.3.x-dev",
```
swirtSJW: Seems like an unexpected reduction.

composer.lock
```diff
@@ -18729,11 +18842,11 @@
...
                 "composer/installers": "^1.2",
                 "dmore/chrome-mink-driver": "^2.6",
                 "drupal-composer/drupal-scaffold": "2.3.0",
-                "drupal/core": "^8.8@alpha",
+                "drupal/core": "^8.7@alpha",
```
swirtSJW: Odd reduction.

ethanteague:
@swirtSJW - thanks for catching these. After doing a `lando composer clear-cache` then `lando composer install` the versions aren't cranked back up. Do you think I should just change manually, or something else?

swirtSJW:
CI seems to be stuck on this one, even after a couple of merges to bring it up-to-date

swirtSJW:
This is still showing the tests as not running.  I went and kicked them off in devshop.  They passed.
image

PR #824, opened 2020-01-14
VAGOV-6411: Adding clone perms for content creators.

ethanteague:
## What this does
Allows content creators to use the clone tab - said role(s) will now have a clone tab available on content

## Implementation / Testing
Stan Gardener and his folks will need this. After updating perms, I logged in as his user (1270), and successfully cloned a couple of content items in my local.


swirtSJW:
@ethanteague  I think the phpunit pems need updating to match so they will pass.

ethanteague:
> @ethanteague I think the phpunit pems need updating to match so they will pass.

Yep - updating now, @swirtSJW

Issue #834, opened 2020-01-17
Meta tag bugs related to VA.gov CMS site name


kevwalsh:
**Describe the bug**
VA.gov CMS is still showing up in some meta fields. 

**To Reproduce**
1. Go to view-source:https://www.va.gov/pittsburgh-health-care/

**Expected behavior**
site name should always be Veterans Affairs

**Screenshots**
![view-source_https___www_va_gov_pittsburgh-health-care_](https://user-images.githubusercontent.com/643678/72583782-8f24a180-38b5-11ea-8da0-daef8d521f21.jpg)

PR #835, opened 2020-01-17
VACMS-834: Replace all instances of [site:name] with Veterans Affairs in content metatags.

kevwalsh:
See https://github.com/department-of-veterans-affairs/va.gov-cms/issues/834

### DIFF
--- a/config/sync/metatag.metatag_defaults.node.yml
+++ b/config/sync/metatag.metatag_defaults.node.yml
@@ -7,12 +7,13 @@ _core:
 id: node
 label: Content
 tags:
-  title: '[node:title] | [site:name]'
   description: '[node:field_description:value]'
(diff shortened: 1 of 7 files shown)


config/sync/metatag.metatag_defaults.node.yml
```diff
@@ -7,12 +7,13 @@ _core:
 id: node
 label: Content
 tags:
-  title: '[node:title] | [site:name]'
   description: '[node:field_description:value]'
   image_src: '[node:field_media:entity:image:entity:url]'
-  og_image: '[node:field_media:entity:image:entity:url]'
```
swirtSJW: Seems like an unrelated loss.

kevwalsh: look a few lines down.

PR #837, opened 2020-01-17
VACMS-836: Specify VACMS instead of VAGOV for branch naming.

kevwalsh:
Don't merge this yet! I think we should merge this sometime next week when we officially make the change to github for issues.


swirtSJW:
Closed in preference to  #843

PR #843, opened 2020-01-17
VACMS-836 Fix commit hook to allow VACMS- notation.

swirtSJW:
This will allow both branch names based on both formats
Zenhub: VACMS-###  
Jira:  VAGOV-###


kevwalsh:
Awesome

PR #876, opened 2020-02-03
VAGOV-1738: Adding excluded type logic for previews.

ethanteague:
## What this does
  - Removes "Preview" button from node view page for 5 node types.

## QA
  - Go to `admin/content`
  - Visit the node view page and confirm "Preview" button does not appear for 1 of each type:

1. outreach_asset
2. person_profile
3. support_service
4. regional_health_care_service_des
5. health_care_local_health_service

### DIFF
--- a/docroot/modules/custom/va_gov_build_trigger/va_gov_build_trigger.module
+++ b/docroot/modules/custom/va_gov_build_trigger/va_gov_build_trigger.module
@@ -87,9 +87,17 @@ function va_gov_build_trigger_preprocess_page(&$variables) {
     // It's a node.
     $path = \Drupal::request()->getpathInfo();
     $arg = explode('/', $path);
-
-    // Make sure we aren't on the node form.
-    if (empty($arg[1]) || empty($arg[2]) || $arg[1] !== 'node' && !is_numeric($arg[2])) {
+    $exclusion_types = [
+      'outreach_asset',
+      'person_profile',
+      'support_service',
+      'regional_health_care_service_des',
+      'health_care_local_health_service',
+    ];
+    // Make sure we aren't on the node form or an excluded type.
+    if ((empty($arg[1]) || empty($arg[2]) || $arg[1] !== 'node') &&
+     (!is_numeric($arg[2])) &&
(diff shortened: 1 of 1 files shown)


docroot/modules/custom/va_gov_build_trigger/va_gov_build_trigger.module
```diff
@@ -87,9 +87,17 @@ function va_gov_build_trigger_preprocess_page(&$variables) {
...
-    if (empty($arg[1]) || empty($arg[2]) || $arg[1] !== 'node' && !is_numeric($arg[2])) {
+    $exclusion_types = [
+      'outreach_asset',
+      'person_profile',
+      'support_service',
+      'regional_health_care_service_des',
+      'health_care_local_health_service',
+    ];
+    // Make sure we aren't on the node form or an excluded type.
+    if ((empty($arg[1]) || empty($arg[2]) || $arg[1] !== 'node') &&
```
swirtSJW: This is kind of an old school method for determining the route.  I think we want to use a routname here instead

```
$route_name = \Drupal::routeMatch()->getRouteName();
if ($route_name !== 'entity.node.edit_form') {

}
```

ethanteague: That's a lot nicer, @swirtSJW :+1: - updated.

ElijahLynn:
> Step 18 start: Create header and footer
> 
> /var/aegir/projects/cms/pr876/docroot/vendor/va-gov/web/src/site/stages/build/index.js:245
>     if (err) throw err;
>              ^
> 
> TypeError: Cannot read property 'data' of undefined
>     at formatHeaderData (/var/aegir/projects/cms/pr876/docroot/vendor/va-gov/web/src/site/stages/build/drupal/menus.js:267:3

We should `composer update va-gov/web` I think. We are supposed to have this updated automatically with dependabot but it is broken right now due to an upstream issue here https://github.com/dependabot/feedback/issues/803.

swirtSJW:
Passed php unit Locally
image


Passed behat locally 
image

Can no longer wait on devshop, merging based on local tests passing.

PR #880, opened 2020-02-03
Create a way to disable "Remote Build Triggers".

jonpugh:
TODO:

- [x] Add a setting to disable front-end build triggers.
- [ ] Figure out where to set that setting in tests: Behat "When I run drush" or PHPUnit?


ElijahLynn:
Closed in favor of #886 for now. Seeing if that gets us moving again first, may reopen this.

PR #888, opened 2020-02-03
VAGOV-0000 Update va-gov/web.

swirtSJW:


composer.lock
```diff
@@ -10689,7 +10689,8 @@
                     }
                 },
                 "patches_applied": {
-                    "0 string value in cell throws empty error": "https://www.drupal.org/files/issues/2019-12-10/0-value-throwing-empty-error.patch"
+                    "0 string value in cell throws empty error": "https://www.drupal.org/files/issues/2019-12-10/0-value-throwing-empty-error.patch",
+                    "Make tables draggable": "https://www.drupal.org/files/issues/2019-04-25/tablefield-tabledrag_support-3040358-4-D8.patch"
```
swirtSJW: This extra bit was applied in another PR in composer.json but did not get added to composer.lock, so it is getting added now.

swirtSJW:
phpUnit passing locally 
image

Issue #889, opened 2020-02-04
Node history tab is broken


kevwalsh:
**Describe the bug**
The node history tab hasn't worked for awhile, and is 

**To Reproduce**
1. Go to any node and click on history (eg https://prod.cms.va.gov/node/318/moderation-history)

**Expected behavior**
Either the tab should work, or be removed altogether.  For now, let's remove it. 

**Screenshots**
![Moderation_History___Veterans_Affairs](https://user-images.githubusercontent.com/643678/73707257-0ab18b80-46c9-11ea-98ca-946b9c2aa4d8.jpg)

swirtSJW:
Merged https://github.com/department-of-veterans-affairs/va.gov-cms/pull/908

PR #908, opened 2020-02-04
VAGOV-889: Remove the history tab from node.

ethanteague:
## What this does
 Removes `History` node tab, per #889


swirtSJW:
Merging this based on tests passing previously. and rebase being unrelated code. 
image

PR #920, opened 2020-02-07
VAGOV-911: Adding better perm ui experience.

ethanteague:
## What this does
 - Adds a new tab to perms place for better ui filtering / experieince

## QA
 - Go to `admin/people/pfm-permissions`
 - Click on the PFM permissions tab
 - Spot check filtering by various modules and roles

## Screenshot
image


kevwalsh:
@swirtSJW  I just renamed this PR to VACMS-911, it was VAGOV-911. how much do we care to fix the commit history?

PR #945, opened 2020-02-10
VAGOV-938: Adding entity_clone patch to prevent child node ref recursion

### LINKED ISSUE #938: VAMC editors cannot clone detail pages.
**Describe the defect**
Clone button on detail pages 

**To Reproduce**
As content_publisher (but true for all content types)
* Go to /pittsburgh-health-care/internships-fellowships/forms
* Click clone
* see 500 error

**Expected behavior**
* i should be able to clone the node.

**Screenshots**

ethanteague:
## Description
Fixes #938 :
Prevents recursive references to child nodes when cloning content. This behavior is not needed for VA, and results in time outs.

## Testing done
Visual / manual

## Screenshots
image

## QA steps

As user with any role
- [ ] Go to `node/296/edit` and click the clone tab
- [ ] Visually verify the site doesn't error or time out
- [ ] Select as many items as you like, then complete the cloning operation, confirming it was successful
- [ ] Repeat the above steps for nodes 285, 290, and 738

## Definition of Done
- [ ] Entity cloning just works

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -204,6 +204,9 @@
                 "2815221 - Add quickedit to the latest-revision route": "https://www.drupal.org/files/issues/2019-03-05/2815221-116.patch",
                 "Allow menu items which link to unpublished nodes to be selected in the parent item selector": "https://www.drupal.org/files/issues/drupal-core-allow-menu-items-to-link-to-unpublished-2807629-7-8.3.x.patch"
             },
+            "drupal/entity_clone": {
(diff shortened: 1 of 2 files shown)


ethanteague:
> Seems to be missing the addition to composer.lock showing that it was applied. I think you might need to run `lando composer update --lock`

Updated, @swirtSJW :+1:

Issue #930, opened 2020-02-07
Stricter permissions for adding content to products


kevwalsh:
### User Story or Problem Statement

We have no way of preventing cross-publishing of content between "business units" or orgs within VA (field_office).

![Create_Story___VA_gov_CMS](https://user-images.githubusercontent.com/643678/74045819-6e72d780-499b-11ea-90b8-aefc335cb4ca.jpg)

### Affected users and stakeholders

* CMS editors of VAMC systems and Outreach hub, for now, but this will come up for other content that has field_office 

## Acceptance Criteria
* Ryan Stubblebine can only add content to VAMC system
* Tim Hudak can only add content to Outreach hub
* Louis Scavnicky can only add content to VISN 4 VAMC systems
* If a user can only add to one thing, field_office is prepopulated with that one thing, just like field_administration (owner)
   * For MVP the two fields should not be related/interconnected.  Just add the logic to offer the user the choices available to that user.  The complexity of the business logic to make selecting one select the other should be left out of this until it has a more established need.

## Impementation details
  
- [ ] A drupal service that can be called to get user's "membership" (getMemberships()) and the memberships parents (getMembershipsParents()) This will also be needed for #92

ethanteague:
Service pr is here: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/1018 (only provides service, not form checking - this will be in future pr)

PR #1018, opened 2020-02-20
VACMS-930: Adding new service for user perms.

ethanteague:
## Description

See #930 

## Testing done
Kint output dump to view results with various user and entity id combos

## Screenshots
image

## QA
 - [x] Open two browsers, one as user with admin privileges, and one as user 1277
 - [x] As admin, visit `/user/1277/workbench_access` and take note of checked sections
 - [x] As admin, visit `/node/2753/edit` and set Governance -> `Owner` to `VA Pittsburgh health care` and save
 - [x] As user 1277, visit `/node/2753/edit` and click the `Event Listing` dropdown
 - [x] Visually verify that items (and child items) not associated with user 1277 sections do not appear
 - [x] As admin, go back to `/user/1277/workbench_access` page, and deselect all items, then check `--- VISN 4`
 - [x] As user 1277, go back to `/node/2753/edit` and click the `Event Listing` dropdown and visually verify that all VAMC systems appear, but only headers appear for other items


config/sync/core.extension.yml
```diff
@@ -170,6 +171,7 @@ module:
   va_gov_migrate: 0
   va_gov_notifications: 0
   va_gov_preview: 0
+  va_gov_services: 0
```
swirtSJW: I'm thinking we don't need the redundancy of having two modules that do nothing by themselves.  I think all of this could reside in va_gov_backend as that module has the description of 'Tools to improve backend user experience for VA.gov'

config/sync/lightning_core.versions.yml
```diff
@@ -204,6 +204,7 @@ ui_patterns_library: 1.0.0
 update: 8.6.2
 user: 8.6.2
 user_history: 1.0.0-alpha3
+va_gov_access: 0.0.0
```
swirtSJW: Curious that this file is even in play anymore now that we are off lightning.  Though I am gessing this was from cex before d8.8 was merged in.

docroot/modules/custom/va_gov_access/va_gov_access.module
```diff
@@ -0,0 +1,29 @@
...
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Implements hook_form_alter().
+ */
+function va_gov_access_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  $perms = \Drupal::service('va_gov_services.user_perms');
+  $targets = ['field_office'];
+  foreach ($targets as $target) {
+    if (in_array('field_office', $form)) {
```
swirtSJW: Shouldn't the hard coded 'field_office' be  $target?

docroot/modules/custom/va_gov_access/va_gov_access.module
```diff
@@ -0,0 +1,29 @@
...
+      foreach ($form[$target]['widget']['#options'] as $header_key => $option_header) {
+        foreach ($option_header as $option_key => $option_item) {
```
swirtSJW: Do we need to check that we for sure that $option_header is an array?

docroot/modules/custom/va_gov_services/src/UserPermsService.php
```diff
@@ -0,0 +1,99 @@
...
+   * @param string $entity_id
+   *   The entity id.
+   * @param string $entity_type
+   *   E.g. node, block.
+   * @param string $user_id
+   *   The user id.
+   * @param string $op
+   *   E.g., create, update, delete.
+   *
+   * @return string
```
swirtSJW: @return   type ought to be bool

docroot/modules/custom/va_gov_access/va_gov_access.module
```diff
@@ -0,0 +1,29 @@
...
+          $status = $perms->userAccess($option_key, 'node');
```
swirtSJW: userAccess returns a Bool so no need to do a check for === FALSE  Could use it directly.   $status is kind of misleading in this context.  I think what you really want is $access.

```
$access = $perms->userAccess($option_key, 'node');
if ($access) {
```

docroot/modules/custom/va_gov_access/va_gov_access.module
```diff
@@ -0,0 +1,29 @@
...
+ * Contains va_gov_access.module.
+ */
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Implements hook_form_alter().
+ */
+function va_gov_access_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  $perms = \Drupal::service('va_gov_services.user_perms');
```
swirtSJW: If this entire thing is moved to va_gov_backend  it would be good to put this all in a function and then just call it in   va_gov_backend_form_alter.   form_alters can get pretty bloated so I like to see their logic compartmentalized into individual function calls.

docroot/modules/custom/va_gov_services/src/UserPermsService.php
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Returns a Drupal user's roles.
+   */
+  public function userRoles() {
+    return $this->currentUser->getRoles();
+  }
+
+  /**
+   * Returns true if user has perm.
```
swirtSJW: Needs @param and @return defined.

docroot/modules/custom/va_gov_services/src/UserPermsService.php
```diff
@@ -0,0 +1,99 @@
...
+   * UserPermsService constructor.
+   */
+  public function __construct(AccountInterface $currentUser) {
+    $this->currentUser = $currentUser;
+  }
+
+  /**
+   * Returns a Drupal user ID.
+   */
+  public function userId() {
```
swirtSJW: Do all these methods need to be public?   There might be public use for some of them, but most seem like they are just called internally.

swirtSJW: It would be nice to follow established standards so that anything that is a getter, is actually named as such.  
Example 
 `function userId()`  should really be `function getUserId()`

docroot/modules/custom/va_gov_services/src/UserPermsService.php
```diff
@@ -0,0 +1,99 @@
...
+    return $this->currentUser->id();
+  }
+
+  /**
+   * Returns a Drupal username.
+   */
+  public function userName() {
```
swirtSJW: getter

docroot/modules/custom/va_gov_services/src/UserPermsService.php
```diff
@@ -0,0 +1,99 @@
...
+    return $this->currentUser->getDisplayName();
+  }
+
+  /**
+   * Returns a Drupal user's roles.
+   */
+  public function userRoles() {
```
swirtSJW: getter

swirtSJW: A getter that just calls another getter?

docroot/modules/custom/va_gov_services/src/UserPermsService.php
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * Return an access status string for entity.
+   *
+   * @param string $entity_id
+   *   The entity id.
+   * @param string $entity_type
+   *   E.g. node, block.
+   * @param string $user_id
```
swirtSJW: int ?

docroot/modules/custom/va_gov_services/src/UserPermsService.php
```diff
@@ -0,0 +1,99 @@
...
+   *
+   * @return string
+   *   Access will be TRUE or FALSE.
+   */
+  public function userAccess($entity_id, $entity_type, $user_id = NULL, $op = NULL) {
+
+    $account = $this->currentUser;
+    // If uid passed, use it.
+    if (!empty($user_id)) {
+      $account = User::load($user_id);
```
swirtSJW: Here we are loading the user if we didn't have it... should his call a setter to pass the newly loaded user to $this->currentUser ?

docroot/modules/custom/va_gov_services/src/UserPermsService.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+    $entity = \Drupal::entityTypeManager()->getStorage($entity_type)->load($entity_id);
+
+    // If op is passed, use it.
+    if (empty($op)) {
+      $op = 'create';
+      if (strpos($entity->bundle(), '_listing') !== FALSE) {
+        $op = 'view';
```
swirtSJW: a code comment here explaining why this case the op should be view, would be helpful.  It is not obvious from the code, so we need to explain the reason.

docroot/modules/custom/va_gov_services/src/UserPermsService.php
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+    return array_reduce(\Drupal::entityTypeManager()->getStorage('access_scheme')->loadMultiple(), function (AccessResult $carry, AccessSchemeInterface $scheme) use ($entity, $op, $account) {
```
swirtSJW: There is some odd mojo going on here.   I need a download on this.

docroot/modules/custom/va_gov_services/src/UserPermsService.php
```diff
@@ -0,0 +1,99 @@
...
+   *   Access will be TRUE or FALSE.
+   */
+  public function userAccess($entity_id, $entity_type, $user_id = NULL, $op = NULL) {
+
+    $account = $this->currentUser;
+    // If uid passed, use it.
+    if (!empty($user_id)) {
+      $account = User::load($user_id);
+    }
+    $entity = \Drupal::entityTypeManager()->getStorage($entity_type)->load($entity_id);
```
swirtSJW: This should probably passed into the constructor by declaring it as a argument in the service.

config/sync/core.extension.yml
ethanteague: Yeah, I'm down with that.

config/sync/lightning_core.versions.yml
ethanteague: :+1:

docroot/modules/custom/va_gov_access/va_gov_access.module
ethanteague: Yep - thanks for catching that, @swirtSJW

docroot/modules/custom/va_gov_access/va_gov_access.module
ethanteague: Couldn't hurt :+1:

docroot/modules/custom/va_gov_services/src/UserPermsService.php
ethanteague: Yep :+1:

docroot/modules/custom/va_gov_access/va_gov_access.module
ethanteague: Ah....that's much nicer :+1:

docroot/modules/custom/va_gov_access/va_gov_access.module
ethanteague: :+1:

docroot/modules/custom/va_gov_services/src/UserPermsService.php
ethanteague: I shouldn't have included this - it isn't a workbench_access method thing. Gonna dump.

docroot/modules/custom/va_gov_services/src/UserPermsService.php
ethanteague: They really aren't needed, at all - I was just thinking we might leverage them in the future. I'm gonna get rid of them (for now).

ethanteague: Per above comment - dumped it.

docroot/modules/custom/va_gov_services/src/UserPermsService.php
ethanteague: Per above comment - dumped

docroot/modules/custom/va_gov_services/src/UserPermsService.php
ethanteague: We don't need this for the workbench access stuff: Per above comment - dumped

docroot/modules/custom/va_gov_services/src/UserPermsService.php
ethanteague: Yep - good eye.

docroot/modules/custom/va_gov_services/src/UserPermsService.php
ethanteague: Yeah - I refactored this a bit and added a comment. It's a special snowflake situation.

docroot/modules/custom/va_gov_services/src/UserPermsService.php
ethanteague: Can you show me what that would look like? Sorry...glad to do, but not sure I understand how to do it efficiently, @swirtSJW

docroot/modules/custom/va_gov_services/src/UserPermsService.php
ethanteague: Lol, yeah. I grabbed this pattern from `docroot/modules/contrib/workbench_access/workbench_access.module` `workbench_access_entity_access` and whittled it down a bit, @swirtSJW

docroot/modules/custom/va_gov_services/src/UserPermsService.php
ethanteague: If you have a minute or two, can you help me out with this, @swirtSJW ? I'm not sure I'm following you on this one, e.g.: I don't understand why we want an entity passed in the service.

docroot/modules/custom/va_gov_services/src/UserPermsService.php
swirtSJW: Sorry, My recommendation of getter setter combo was a little misguided.   A traditional  combo getter setter would look like this https://git.drupalcode.org/project/node_link_report/blob/8.x-1.x/src/Plugin/Block/NodeLinkReport.php#L563   So that the first time it is called, it sets it, then gets it.

I mistook yours to be a similar condition, but yours is like a getter with alternate back up.  More like this

```
/**
* Getter setter for a, user, defaults to current user.
@param int $user_id 
  A user id to lookup the user. (optional)

@return AccountInterface
  The user specified by $user_id or the current user
*/
private  function getUser($user_id = NULL) {
  if ( !empy($user_id)) {
    // Not available, so load it. 
    return User::load($user_id);
  } 
  else {
 return  $this->currentUser;
  }
}

```

It just compartmentalizes the logic in one call, instead of a bunch of nested if's hanging out in your userAccess method.

docroot/modules/custom/va_gov_services/src/UserPermsService.php
swirtSJW: You would not want the entity passed in, but you would want the entityTypeManager passed in (injected) as opposed to calling /Drupal.   The same way you pass in the accountInterface.

ethanteague: Got it! Ty, @swirtSJW :+1:

docroot/modules/custom/va_gov_services/src/UserPermsService.php
ethanteague: :+1: @swirtSJW

ElijahLynn:
>  Test Failed: `va/tests/phpunit`
> phpunit --exclude-group disabled tests/phpunit --colors=always
> ```
> PHPUnit 6.5.14 by Sebastian Bergmann and contributors.
> 
> Testing tests/phpunit
> ........E..................                                       27 / 27 (100%)
> 
> Time: 17.95 seconds, Memory: 134.00MB
> 
> There was 1 error:
> 
> 1) tests\phpunit\LoginPerformance::testLoginPerformance with data set #0 (2)
> Behat\Mink\Exception\ElementNotFoundException: Form field with id|name|label|value "name" not found.
> 
> /var/aegir/projects/cms/pr1018/docroot/vendor/behat/mink/src/WebAssert.php:638
> /var/aegir/projects/cms/pr1018/docroot/core/tests/Drupal/Tests/UiHelperTrait.php:271
> /var/aegir/projects/cms/pr1018/docroot/core/tests/Drupal/Tests/UiHelperTrait.php:243
> /var/aegir/projects/cms/pr1018/tests/phpunit/PerformanceLoginTest.php:36
> 
> ERRORS!
> Tests: 27, Assertions: 63, Errors: 1.
> 
> Operation took 0.096 seconds compared to the benchmark of 2 seconds.
> 
> Operation took 0.078 seconds compared to the benchmark of 2 seconds.
> 
> Operation took 0.139 seconds compared to the benchmark of 5 seconds.
> 
> Operation took 0.027 seconds compared to the benchmark of 5 seconds for type page.
> 
> Operation took 0.036 seconds compared to the benchmark of 5 seconds for type landing_page.
> ```
> 
> * **On:** devshop.cms.va.gov
> * **In:** 18.015s

@ethanteague Is this legit/happening locally?

ethanteague:
> > Test Failed: `va/tests/phpunit`
> > phpunit --exclude-group disabled tests/phpunit --colors=always
> > ```
> > PHPUnit 6.5.14 by Sebastian Bergmann and contributors.
> > 
> > Testing tests/phpunit
> > ........E..................                                       27 / 27 (100%)
> > 
> > Time: 17.95 seconds, Memory: 134.00MB
> > 
> > There was 1 error:
> > 
> > 1) tests\phpunit\LoginPerformance::testLoginPerformance with data set #0 (2)
> > Behat\Mink\Exception\ElementNotFoundException: Form field with id|name|label|value "name" not found.
> > 
> > /var/aegir/projects/cms/pr1018/docroot/vendor/behat/mink/src/WebAssert.php:638
> > /var/aegir/projects/cms/pr1018/docroot/core/tests/Drupal/Tests/UiHelperTrait.php:271
> > /var/aegir/projects/cms/pr1018/docroot/core/tests/Drupal/Tests/UiHelperTrait.php:243
> > /var/aegir/projects/cms/pr1018/tests/phpunit/PerformanceLoginTest.php:36
> > 
> > ERRORS!
> > Tests: 27, Assertions: 63, Errors: 1.
> > 
> > Operation took 0.096 seconds compared to the benchmark of 2 seconds.
> > 
> > Operation took 0.078 seconds compared to the benchmark of 2 seconds.
> > 
> > Operation took 0.139 seconds compared to the benchmark of 5 seconds.
> > 
> > Operation took 0.027 seconds compared to the benchmark of 5 seconds for type page.
> > 
> > Operation took 0.036 seconds compared to the benchmark of 5 seconds for type landing_page.
> > ```
> > 
> > 
> > 
> > * **On:** devshop.cms.va.gov
> > * **In:** 18.015s
> 
> @ethanteague Is this legit/happening locally?

No, all passing:
`ethan @ ~/workspace/va-et
VACMS-930-perms-et└─ $ lando phpunit

Yaml Tests Initialized
======================

 ! No GitHub token found. forcing --dry-run

 Git Remote: https://github.com/ElijahLynn/va.gov-cms
 Local Git Branch: VACMS-930-perms-et
 Composer working directory: /app
 Git Repository directory: /app
 Git Commit: 66b6b2e5805b5227eb6a5bc6e8d769a9af593b3f
 Tests File: /app/tests.yml
 -------------------------- ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Tests found in tests.yml                                                                                                                                                                                                                                                                                                        
 -------------------------- ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  va/deploy/0-composer       rm -rf ./docroot/vendor/va-gov/web && composer install --dev --no-interaction --no-progress --ansi | tee /dev/stderr | grep 'Could not apply patch' && exit 1 || true                                                                                                                                
  va/deploy/1-cache          drush $DRUSH_ALIAS cr                                                                                                                                                                                                                                                                                
  va/deploy/2-update         drush $DRUSH_ALIAS updatedb --yes                                                                                                                                                                                                                                                                    
  va/deploy/3-config         drush $DRUSH_ALIAS cim --yes                                                                                                                                                                                                                                                                         
                             drush $DRUSH_ALIAS cr                                                                                                                                                                                                                                                                                
  va/tests/phpcs             composer va:test:cs                                                                                                                                                                                                                                                                                  
  va/tests/phplint           composer va:test:lint                                                                                                                                                                                                                                                                                
  va/tests/phpunit           phpunit --exclude-group disabled tests/phpunit --colors=always                                                                                                                                                                                                                                       
  va/tests/accessibility     composer va:test:accessibility                                                                                                                                                                                                                                                                       
  va/web/build               composer va:web:build | tee /dev/stderr | grep "Failed to pipe Drupal content into Metalsmith!" -B1000 -C8 && echo "tests.yml | composer va:web:build included the Drupal/Metalsmith error." && exit 1 || echo "tests.yml | Front end site was built! Check $DRUPAL_ADDRESS/static for raw output!"  
  va/tests/behat             cd tests/behat                                                                                                                                                                                                                                                                                       
                             behat --colors                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                  
 -------------------------- ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 

 ------------------------------------------------- ---------------------------------------------------------------- 
  Tests to run based on filter 'va/tests/phpunit'                                                                   
 ------------------------------------------------- ---------------------------------------------------------------- 
  va/tests/phpunit                                  phpunit --exclude-group disabled tests/phpunit --colors=always  
 ------------------------------------------------- ---------------------------------------------------------------- 

 ! Skipping commit status posting, dry-run enabled.

Running test va/tests/phpunit: PHPUnit
--------------------------------------

 $ phpunit --exclude-group disabled tests/phpunit --colors=always Output:/path/to/filePHPUnit 6.5.14 by Sebastian Bergmann and contributors.


Testing tests/phpunit

.

Operation took 0.099 seconds compared to the benchmark of 2 seconds.

.

Operation took 0.081 seconds compared to the benchmark of 2 seconds.

.
.
.

Operation took 0.168 seconds compared to the benchmark of 5 seconds.

.

Operation took 0.032 seconds compared to the benchmark of 5 seconds for type page.

.

Operation took 0.041 seconds compared to the benchmark of 5 seconds for type landing_page.

.

Operation took 0.844 seconds compared to the benchmark of 2 seconds.

.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.                                       27 / 27 (100%)


Time: 19.04 seconds, Memory: 136.25MB


OK (27 tests, 66 assertions)


 ✔ Process Succeeded in 19.101s Output: /path/to/file

Executed all tests
==================

 ------------------ ---------------------------------------------------------------- ---------- 
  Test Results                                                                                  
 ------------------ ---------------------------------------------------------------- ---------- 
  va/tests/phpunit   phpunit --exclude-group disabled tests/phpunit --colors=always   ✔ Passed  
 ------------------ ---------------------------------------------------------------- ---------- `

ElijahLynn:
Thanks. There is some network related flickering going on and I wonder if that is somehow causing that above error. I retriggered the test run.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,111 @@
...
+      // Not available, so load it.
+      return User::load($user_id);
+    }
+    else {
+      return $this->currentUser;
+    }
+  }
+
+  /**
+   * Return an access status string for entity.
```
swirtSJW:
```suggestion
   * Check to see if a user has access.
```

swirtSJW: Let the @return declare what is being returned.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,111 @@
...
+   *   The entity id.
+   * @param string $entity_type
+   *   E.g. node, block.
+   * @param int $user_id
+   *   The user id.
+   * @param string $op
+   *   E.g., create, update, delete.
+   *
+   * @return bool
+   *   Access will be TRUE or FALSE.
```
swirtSJW:
```suggestion
   *   TRUE if the user has access, FALSE otherwise.
```

swirtSJW: Specificity is good.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,111 @@
...
+   */
+  public function userAccess($entity_id, $entity_type, $user_id = NULL, $op = NULL) {
+
+    $account = $this->currentUser;
+    // If uid passed, use it.
+    if (!empty($user_id)) {
+      $account = $this->getUser($user_id);
+    }
```
swirtSJW: You compartmentalized this in a method already, time to use it.
```
$account = $this->getUser($user_id);
```

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,111 @@
...
+    $entity = $this->entityInterface->getStorage($entity_type)->load($entity_id);
+
+    // If op is passed, use it.
+    if (empty($op)) {
+      $op = 'create';
```
swirtSJW: Not clear on why the default op is `create` It would be good to have that in a comment.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,111 @@
...
+    // scope of workbench_access.
+    $database = \Drupal::database();
+    $query = $database->select('section_association__user_id', 's');
+    $query->condition('s.user_id_target_id', $account->id());
+    $query->condition('s.entity_id', 4);
+    $query->fields('s', ['entity_id']);
+    $results = $query->execute()->fetchAll();
+    if (count($results) > 0) {
+      return TRUE;
+    }
```
swirtSJW: This db query is going to get called a bunch of times due to `$perms->userAccess($option_key, 'node');` being called inside a foreach.  Lets hop on a call to see if we can come up with a way to reduce this.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,111 @@
...
+
+    // Compare user sections against subject section to determine access.
+    return array_reduce(\Drupal::entityTypeManager()->getStorage('access_scheme')->loadMultiple(), function (AccessResult $carry, AccessSchemeInterface $scheme) use ($entity, $op, $account) {
```
swirtSJW: This is also going to get called a bunch of times. ... lets have think on this to see if we can reduce the calls... maybe through keeping state.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -120,6 +123,32 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
   }
 }
 
+/**
+ * For determining dropdown field access on nodeforms.
+ *
+ * @param array $form
+ *   The node form.
+ */
+function _va_gov_backend_access(array &$form) {
```
swirtSJW: Minor:  Your description is very specific, which is good, but  the function name is very generic and does not convey much.  It would be good if the function name matched the description.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -120,6 +123,32 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
...
+  // Limit options on forms with field_office field.
+  $perms = \Drupal::service('va_gov_backend.user_perms');
+  $targets = ['field_office'];
```
swirtSJW: Good job planning for the future.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: Arghhh.....missed that. :+1:

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: :+1: good point

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: Can't argue with that :+1:

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,160 @@
...
+   *   The entity add or edit form.
+   * @param array $targets
+   *   The form fields to target for filtering.
+   *
+   * @return array
+   *   The options that are allowed for user selection.
+   */
+  public function userOptionsStorage(array &$form, array $targets) {
+
+    $cid = 'user-dropdown-access-' . $this->getUser()->id();
```
swirtSJW: Should include  namespace for what is creating it.
```suggestion
    $cid = 'userpermservice-dropdown-access-' . $this->getUser()->id();
```

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,160 @@
...
+    $allowed_options = [];
+
+    $cached_data = \Drupal::cache()->get($cid);
+    // If we have a cache, return it.
+    if (!empty($cached_data)) {
+      $allowed_options = $cached_data->data;
+    }
+    else {
+      // Return allowed form field options by filtering disallowed items.
+      $targets = ['field_office'];
```
swirtSJW: Major: You are passing in targets, but then overwriting it? Kind of short-circuits the ability to pass it in.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,160 @@
...
+                $access = $this->userAccess($option_key, 'node');
+                if ($access) {
+                  $allowed_options[] = $option_key;
+                }
+              }
+            }
+          }
+        }
+      }
+      $tags = [
```
swirtSJW: For tags, I think you want system tags https://www.drupal.org/docs/8/api/cache-api/cache-tags#syntax
like this.
`user:` . $this->getUser()->id()

I don't think there is one for form ids out of the box, and I don't think we need one here.

The only other one I think we want is one general one that would allow us to invalidate all like
`va_gov_backend_user_perms`

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
swirtSJW: Major: THere is a separate problem. The problem is that  you have this set up to accept different $targets, though you are overriding them.  If you alter to allow different targets being passed in, then this cache is not going to respect the different targets so you will get weird results because the targets you cache, may not match the targets being requested.  If the targets can change, you need to cache each target separately.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,160 @@
...
+   *   The user id.
+   * @param string $op
+   *   E.g., create, update, delete.
+   *
+   * @return bool
+   *   TRUE if the user has access, FALSE otherwise.
+   */
+  public function userAccess($entity_id, $entity_type, $user_id = NULL, $op = NULL) {
+
+    $account = $this->getUser();
```
swirtSJW:
```suggestion
    $account = $this->getUser($user_id);
```

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -84,10 +86,23 @@ function _va_gov_backend_add_vamc_regions_select(array &$form) {
   unset($form['title']['#size']);
 }
 
+/**
+ * Implements hook_entity_update().
+ */
+function va_gov_backend_user_update(EntityInterface $entity) {
+  // When the user object is updated, delete the allowed options cache.
+  Cache::invalidateTags(['allowed-dropdown-options-' . $entity->id()]);
```
swirtSJW: When you are saving the cache, if you use the suggestion I made to use one of the tags as `user:{nid}` then this whole bit is unnecessary as Drupal will already invalidate that tag on user update.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -84,10 +86,23 @@ function _va_gov_backend_add_vamc_regions_select(array &$form) {
...
+
 /**
  * Implements hook_form_alter().
  */
 function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+
+  $targets = ['field_office'];
+  _va_gov_backend_dropdown_field_access($form, $targets);
+
+  // _va_gov_backend_dropdown_field_access($form);
```
swirtSJW: Minor: Do we need this?

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -120,6 +135,37 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
...
+ *   The node form.
+ * @param array $targets
+ *   The node form fields to check.
+ */
+function _va_gov_backend_dropdown_field_access(array &$form, array $targets) {
+  $perms_service = \Drupal::service('va_gov_backend.user_perms');
+  // Grabs items user can access per section associations.
+  $allowed_options = $perms_service->userOptionsStorage($form, $targets);
+
+  $targets = ['field_office'];
```
swirtSJW: This was already passed in,  is it need here?

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: Oh good lord. Thanks for catching that, @swirtSJW !!

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: Good catch :+1:

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: nice

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: nope

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: nope

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,161 @@
...
+    // Return allowed form field options by filtering disallowed items.
+    foreach ($targets as $target) {
+
+      if (in_array($target, $form) && !empty($form[$target]['widget']['#options'])) {
+        $cid = 'userpermservice-dropdown-access-' . $target . '-' . $user_id;
+
+        $cached_data = \Drupal::cache()->get($cid);
+        // If we have a cache, return it.
+        if (!empty($cached_data)) {
+          $allowed_options = $cached_data->data;
```
swirtSJW: This is going to create a situation where only the last one found wins, because it will overwrite the previously found cache data.

You need more of like an array_merge followed by an array_unique.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,161 @@
...
+        }
+        foreach ($form[$target]['widget']['#options'] as $header_key => $option_header) {
```
swirtSJW: This is going to do all the work, regardless of the cache being found.   I think this all needs to be wrapped in an else so that it only gets processed if there is no cache found.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,161 @@
...
+          if (is_array($option_header) && !empty($option_header)) {
+            foreach ($option_header as $option_key => $option_item) {
+              $access = $this->userAccess($option_key, 'node', $user_id);
+
+              if ($access) {
+                $allowed_options[] = $option_key;
```
swirtSJW: I think you are going to want to keep these separate from $allowed_options for now since $allowed options is going to aggregate all allowed_options.   For now you want these to be kept separate.  
```suggestion
                $target_allowed_options[] = $option_key;
```
Which you will need to instantiate just before entering this foreach.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,161 @@
...
+            }
+          }
+        }
+        $tags = [
+          'va_gov_backend_user_perms',
+          'user:' . $user_id,
+        ];
+        // Cache for 24 hours.
+        $expire_time = time() + 24 * 60 * 60;
+        \Drupal::cache()->set($cid, $allowed_options, $expire_time, $tags);
```
swirtSJW:
```suggestion
        \Drupal::cache()->set($cid, $target_allowed_options, $expire_time, $tags);
```
Then right after this.  do an array_merge  into $allowed_options  and array_unique  on $allowed options.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,161 @@
...
+    $results = $query->execute()->fetchAll();
+
+    // Compare user sections against subject section to determine access.
+    return array_reduce(\Drupal::entityTypeManager()->getStorage('access_scheme')->loadMultiple(), function (AccessResult $carry, AccessSchemeInterface $scheme) use ($entity, $op, $account, $results) {
+      $status_class_name = get_class($scheme->getAccessScheme()->checkEntityAccess($scheme, $entity, $op, $account));
+      // Return true if we have our special snowflake Outreach listing node.
+      if ((count($results) > 0) && ($entity->id() === '736')) {
+        return TRUE;
+      }
+      if ($status_class_name === 'Drupal\Core\Access\AccessResultForbidden') {
```
indytechcook: Use `Drupal\Core\Access\AccessResultForbidden::class` instead of a string. It helps in static code analysis and PHPStorm class associations.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: Sorry - confused by what you mean / what this would look like.
Couldn't I just return $target_allowed_options if the cache is not set? It has the correct unique values already. Something like this:
```  public function userOptionsStorage(array &$form, array $targets, $user_id) {
    $allowed_options = [];
    // Return allowed form field options by filtering disallowed items.
    foreach ($targets as $target) {

      if (in_array($target, $form) && !empty($form[$target]['widget']['#options'])) {
        $cid = 'userpermservice-dropdown-access-' . $target . '-' . $user_id;

        $cached_data = \Drupal::cache()->get($cid);
        // If we have a cache, return it.
        if (!empty($cached_data)) {
          $allowed_options = $cached_data->data;
        }
        else {
          $target_allowed_options = [];
          foreach ($form[$target]['widget']['#options'] as $header_key => $option_header) {
            if (is_array($option_header) && !empty($option_header)) {
              foreach ($option_header as $option_key => $option_item) {
                $access = $this->userAccess($option_key, 'node', $user_id);

                if ($access) {
                  $target_allowed_options[] = $option_key;
                }
              }
            }
          }
          $tags = [
            'va_gov_backend_user_perms',
            'user:' . $user_id,
          ];
          // Cache for 24 hours.
          $expire_time = time() + 24 * 60 * 60;
          \Drupal::cache()->set($cid, $target_allowed_options, $expire_time, $tags);
          return $target_allowed_options;
        }
      }
    }

    return $allowed_options;

  }```

Comment (swirtSJW) [reply docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php]: Afraid not.  That would only return any that set by the last target,  it would totally obliterate any that were in previous targets.

File: docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php (hunk 36 above)
Comment (ethanteague) [reply docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php]: :+1: @indytechcook

File: docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php (hunk 35 above)
Comment (ethanteague) [reply docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php]: EDIT: Updating to say that per convo @swirtSJW & ET had, this won't work.

File: docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php (hunk 34 above)
Comment (ethanteague) [reply docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php]: :+1:

File: docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php (hunk 32 above)
Comment (ethanteague) [reply docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php]: :+1:

File: docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php [hunk 37]
```diff
@@ -0,0 +1,166 @@
...
+            }
+          }
+          $tags = [
+            'va_gov_backend_user_perms',
+            'user:' . $user_id,
+          ];
+          // Cache for 24 hours.
+          $expire_time = time() + 24 * 60 * 60;
+          \Drupal::cache()->set($cid, $target_allowed_options, $expire_time, $tags);
+          array_merge($target_allowed_options, $allowed_options);
```
swirtSJW: array_merge is method with a return so you have to catch the return.
```suggestion
          $allowed_options = array_merge($target_allowed_options, $allowed_options);
```

swirtSJW:
@ethanteague  There is one entry that is not showing parity.  Is "Events" the intentional snowflake?

swirtSJW:
@ethanteague  I found one problem.   Saving from workbench access is not invalidating the user### cache tag.  I guess only the user edit page does that automatically   So updating the user's entries in workbench access does not get reflected until a cache clear all is performed.   We don't want that.  
We will need to add a submit handler to that form save that invalidates the cache for that user https://www.drupal.org/docs/8/api/cache-api/cache-tags#s-invalidating

ethanteague:
> @ethanteague There is one entry that is not showing parity. Is "Events" the intentional snowflake?
> 
> image

Outreach events is the snowflake. I bet the reason `Events` is showing is because it is associated with one of those two facilities, which would make it an allowed option.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
ethanteague: :+1:

ethanteague:
> @ethanteague I found one problem. Saving from workbench access is not invalidating the user### cache tag. I guess only the user edit page does that automatically So updating the user's entries in workbench access does not get reflected until a cache clear all is performed. We don't want that.
> We will need to add a submit handler to that form save that invalidates the cache for that user
> submit handler: https://drupal.stackexchange.com/questions/223342/add-a-custom-submission-handler-to-a-form
> cache invalidation: https://www.drupal.org/docs/8/api/cache-api/cache-tags#s-invalidating

Thanks for spotting that, @swirtSJW - also appreciate you giving me the links to the howto's :+1: 
Just updated.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -120,6 +128,51 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
...
+/**
+ * Callback handler for workbench_access_assign_user form - resets perms.
+ *
+ * @param array $form
+ *   The workbench_access_assign_user form.
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
+ *   Provides an interface for an object containing the current state of a form.
+ */
+function va_gov_backend_workbench_assign_user_form_submit(array $form, FormStateInterface $form_state) {
+  $user_id = \Drupal::currentUser()->id();
```
swirtSJW: I don't think this is right.   We want it to be the user being saved in the form, not the current user.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -120,6 +128,53 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
   }
 }
 
+/**
+ * Callback handler for workbench_access_assign_user form - resets perms.
```
swirtSJW:
```suggestion
 * Callback handler for workbench_access_assign_user form - invalidate cache.
```

swirtSJW:
All QA testing passes.

docroot/modules/custom/va_gov_backend/src/Service/UserPermsService.php
```diff
@@ -0,0 +1,166 @@
+<?php
+
+namespace Drupal\va_gov_backend\Service;
+
+use Drupal\Core\Access\AccessResultForbidden;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\user\Entity\User;
+use Drupal\workbench_access\Entity\AccessSchemeInterface;
```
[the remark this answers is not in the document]
ethanteague: :+1: added @swirtSJW

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: :+1: @swirtSJW

PR #1056, opened 2020-02-25
VACMS-000: Removing table field drag and drop patch.

ethanteague:
## Description
Drag and drop patch applied to tablefield module isn't being rendered properly in frontend build. This pr removes this patch.
 - Expected: dragging and dropping rows in tablefield found on `node/767/edit` will render in new order here: `life-insurance/options-eligibility/sgli/`
 - Actual: new ordering is not respected, and additional column representing each row's weight is generated

## Testing done
Visual

## Screenshots
image

## QA steps
 - As an admin, go to `node/767/edit` and visually verify that you cannot drag table rows.
 - Save node, build front end, visually verify that weight column does not appear.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -248,8 +248,7 @@
                 "Bypass password requirement for password change if logged in with sso.": "https://www.drupal.org/files/issues/2018-05-01/simplesamlphp_auth-change-password-access-2968598-2.patch"
             },
             "drupal/tablefield": {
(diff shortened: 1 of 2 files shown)


howardbrande:
@ethanteague @swirtSJW I tested this using pr1085.ci.cms.va.gov. 
https://www.loom.com/share/64b8f66c90674a8fb8034661c21a85e2
- The drag and drop was removed but an extra column was added to the table. 
- I changed the number of columns from 5 to 4
- Looked correct on Preview
- Triggered a build but builds may be failing on CI

PR #1148, opened 2020-03-03
Add .env.global w/NODE_EXTRA_CA_CERTS=/etc/pki/tls/certs/ca-bundle.crt & PUPPETEER_SKIP_CHROMIUM_DOWNLOAD='TRUE'.

ElijahLynn:
Fixes #1049, but this is a mask and I don't like this long term. 

The error we are seeing with network flickers is below and this is the
most common one, it won't solve everything but will help stop the
bleeding.

We don't use the Chromium package in any of our tests. The FE team may
use it in theirs but we don't in CMS.

ERROR: Failed to download Chromium r594312! Set "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" env variable to skip download.
{ Error: self signed certificate in certificate chain
at TLSSocket.onConnectSecure (_tls_wrap.js:1051:34)
at TLSSocket.emit (events.js:189:13)
at TLSSocket._finishInit (_tls_wrap.js:633:8) code: 'SELF_SIGNED_CERT_IN_CHAIN' }
Script cd web && yarn install handling the va:web:install event returned with error code 1
Script @va:web:install was called via post-install-cmd

## Description

See _issueid_. 

## QA steps

As user _uid_ with _user_role_
1. Do this
1. Then that
1. Then validate Acceptance Criteria from issue
- [ ] This
- [ ] That
- [ ] The other thing

## Definition of Done
- [ ] Product release notes 
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)

PR #1162, opened 2020-03-04
Fix "self signed certificate in chain" error on deploys.

### LINKED ISSUE #1016: PR Environment Web build process flicker
AC
* PRs don't fail for random reasons 

**Describe the defect**
Verfiy platform task failed: http://devshop.cms.va.gov/node/12876

```
[4/4] Building fresh packages...
error /var/aegir/projects/demo/vamcupgrade/docroot/vendor/va-gov/web/node_modules/puppeteer: Command failed.
Exit code: 1
Command: node install.js
Arguments: 
Directory: /var/aegir/projects/demo/vamcupgrade/docroot/vendor/va-gov/web/node_modules/puppeteer
Output:
ERROR: Failed to download Chromium r594312! Set "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" env variable to skip download.
{ Error: self signed certificate in certificate chain
    at TLSSocket.onConnectSecure (_tls_wrap.js:1051:34)
    at TLSSocket.emit (events.js:189:13)
    at TLSSocket._finishInit (_tls_wrap.js:633:8) code: 'SELF_SIGNED_CERT_IN_CHAIN' }
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
Script cd web && yarn install handling the va:web:install event returned with error code 1
Script @va:web:install was called via post-install-cmd
```
It's just a flicker, the very next build was successful: http://devshop.cms.va.gov/node/12879
Note: We should ensure documentation covers this event and ensures those that are being trained of the possibility that this happens. Right now the CMS has no indicator of that happening

### LINKED ISSUE #1149: Provide training on Outreach and Events section to VEO users
Hi Howie,

I am connecting you with Tim Hudak from the Veterans Experience Office. He and his team will be using our Outreach and Events product, in partnership with Gary, Josh, and the OPIA team, to coordinate cross-enterprise Veteran outreach.Can you please set up Drupal access for Tim and team, and schedule training? Please Cc me and Jeff Barnes on everything.

As far as Drupal permissions are concerned, Tim and team should be part of the VACO governance pillar. They should be able to add outreach materials and events. They should not have access to other products powered by Drupal.

Tim, please reply all to this chain with names and VA email addresses of who needs access.Thanks,Andy

**This issue is a migration of VAGOV-6327 from Jira to help track 2020 work completed in ZenHub.**

ElijahLynn:
Supersedes #1148, which is borked in the CI DB table so I need a new PR#. 

This issue was masked by the fact that we were seeing network flickers with HTTPS handshakes and random failures for other files from different hosts as well. 

This issue was also difficult to reproduce as the first composer install is the only one that triggers the npm install script invocations on npm dependencies such as Puppeteer downloading Chromium. And deleting just web/node_modules wasn't enough to re-trigger the install scripts such as node_modules/puppeteer/install.js. 

Reproduce:
`rm -rf bin node_modules docroot/libraries docroot/core docroot/modules/contrib docroot/themes/contrib docroot/profiles/contrib/ docroot/vendor` (composer nuke - light) 
`composer install`

Solution:
We have a .env file that can be put down but that file isn't generated at the time composer is invoked the initial install (platform_verify) on CI. I tried loading a .env.global that is commited to VCS and that worked but not on the very first composer install. Somehow the install task doesn't get the ENV variables exported correctly to the subprocesses in the post-install-cmd section. I had to export them right into the shell process and this works for now. It also adds a bit more transparency on what is happening different from upstreams install. This will solve the issue in both CI and the CD ENVs (inside VAEC). 

 

Fixes #1016.
Fixes #1149.


composer.json
```diff
@@ -168,6 +168,9 @@
         "sort-packages": true,
         "platform": {
             "php": "7.1.3"
+        },
+        "preferred-install": {
+            "drupal/core": "source"
```
ElijahLynn: This is something @jonpugh found that will speed up deploys by 30-60 seconds. The TIC blocks Drupal zipballs and fails 3 times then finally downloads source, so this skips those 3 attempts and cuts right to source download. We can't leave comments in JSON so I left this in the comment and here.

composer.json
```diff
@@ -352,7 +355,7 @@
             "npm -v",
             "yarn -v",
             "cd web && yarn run install-repos",
-            "cd web && yarn install"
+            "export NODE_EXTRA_CA_CERTS=/etc/pki/tls/certs/ca-bundle.crt PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=TRUE; cd web && yarn install"
```
swirtSJW: If we are skipping the chromium download, does that mean that some of the vetsweb tests are not running because they  rely on it?

ElijahLynn: I am not actually sure what part of the vets-website tests use this but we don't invoke any of those tests right now. I suspect it is the accessibility test but I know that someone mentioned they use an external version of Chromium for one of them so we are not sure it is even needed. Some more context here https://dsva.slack.com/archives/CBU0KDSB1/p1582932468292300?thread_ts=1582931520.282600&cid=CBU0KDSB1 [INTERNAL].

But regardless, we don't invoke any of those tests, so saving the 30+ seconds here is worth it in my opinion.

ElijahLynn: This would be happening on vets-website builds too but those builds don't actually happen in the VAEC network.

PR #1177, opened 2020-03-06
VAGOV-1176 Fix "Cannot find profiles directory ./profiles".

ElijahLynn:
Fixes #1176.

PR #1174, opened 2020-03-05
VACMS-1120 Make table field not required.

swirtSJW:
## Description

See #1120  

## QA steps

As user _uid_ with _user_role_
1. Login with any editorial role
1. Then go to /node/748/edit
1. Edit one of the table paragraphs. 
1.  Delete the content out of one of the cells, make sure you leave no spaces in the cell.
1. Save the node
- [ ] validate that no red errors related to table being empty appear.

## Definition of Done
- [ ] Product release notes 
- [ ] Documentation has been updated, if applicable
- [ ] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)

PR #1182, opened 2020-03-06
VACMS-000 Break content model feature into two

swirtSJW:
Break content model fields into own feature.  

This is done to allow the spectool to output two diffrent features instead of concat erroring trying to do one big one.

## Testing done
automated tests pass.

## Screenshots
The behat tests in the spectool have been altered to have two separate spots to copy from on the content model.  One is the bundles, the other is fields.
image

## QA steps

1. Validate behat tests pass.

### DIFF
--- a/tests/behat/drupal-spec-tool/content_model_bundles.feature
+++ b/tests/behat/drupal-spec-tool/content_model_bundles.feature
@@ -1,0 +1,63 @@
+@api
+Feature: Content model bundles
+  In order to enter structured content into my site
+  As a content editor
+  I want to have content entity bundles that reflect my content model.
+
+  @dst @content_type @dstbundles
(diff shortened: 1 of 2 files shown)


swirtSJW:
This is not going to pass until https://github.com/department-of-veterans-affairs/va.gov-cms/pull/1174 gets merged in.

ElijahLynn:
Just merged in #1177 (via master) which will get rid of a bunch of failures in CI too.

ElijahLynn:
> This is not going to pass until #1174 gets merged in.

Looks like it passed, so that is a good start!

swirtSJW:
1174 has been merged so as soon as I resolve the conflict, we should be good to go

Issue #1158, opened 2020-03-03
CMS Export System: Tar content on updates


indytechcook:
When content is updated and after the tome sync occurs the content needs to be tared up.

Related to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/827

ElijahLynn:
Update on this: Working through an issue with EFS volume mounts being crazy slow, which will affect this task.

kevwalsh:
@ElijahLynn is that tracked in a separate ticket? or should this be moved back to in progress and have the AC updated?

PR #827, opened 2020-01-15
CMS Export System (CEX, formerly known as Tome Sync)

ElijahLynn:
Ongoing work to add and modify Tome Sync for the CMS Export System.

https://github.com/department-of-veterans-affairs/va.gov-team/issues/4876

PR #1197, opened 2020-03-10
VAGOV-1158 content export tar

indytechcook:
## Description

See #1158.  A new endpoint is added at `/export/content.tar` which provides a 302 to the actual file.

This also includes the code from https://github.com/department-of-veterans-affairs/va.gov-cms/pull/827.

## Testing done

* Create a tar of the `export-content` directory.
* Set the config variable `lando drush config:set va_gov_content_export.settings content_tar_uri public://export-content/content.tar`
* Visit `/export/content.tar`.  See the 302 to `/sites/default/files/content.tar`
* Remvoe the `content.tar` file.
* Visist `export/content.tar` and get a 403.

## Definition of Done
- [ ] Product release notes 
- [x] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/1197)


docroot/modules/custom/va_gov_content_export/src/Controller/ContentTar.php
```diff
@@ -0,0 +1,47 @@
...
+   *
+   * The main purpose of this page callback is to serve up the content.tar
+   * if it exists.  If the file does not exist then return a 404.
+   */
+  public function exportTar() {
+    // At this point we know we have a file because there is an access check
+    // in Drupal\va_gov_content_export\Access\ContentTarAccess which is
+    // called before we get to this point in the code.
+    $url = $this->contentTarLocation->getUrl();
+    return new RedirectResponse($url);
```
indytechcook: I had thought about streaming the file here also but we can't assume the file will be on the local file system.  I could be on an S3 bucket which would provide a significant performance issue.

indytechcook:
Canceling the PR as we are not going to use Drupal to serve the file

PR #1251, opened 2020-03-16
Fix issues with sync-files script on mac and zsh

indytechcook:
I have errors when trying to sync files locally, this fixes theme

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/1251)

### DIFF
--- a/scripts/sync-files.sh
+++ b/scripts/sync-files.sh
@@ -7,7 +7,8 @@ set -e
 cd $(git rev-parse --show-toplevel)/.dumps
 
 curl --remote-name --remote-header-name https://dsva-vagov-prod-cms-backup-sanitized.s3-us-gov-west-1.amazonaws.com/files/cms-prod-files-latest.tgz
-rm --force --recursive ../docroot/sites/default/files/*
+rm -rf ../docroot/sites/default/files/*
 echo "Extracting files to sites/default/files."
+mkdir -p ../docroot/sites/default/files
 tar --extract --gunzip --verbose --file cms-prod-files-latest.tgz --directory ../docroot/sites/default/files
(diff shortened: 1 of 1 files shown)


scripts/sync-files.sh
```diff
@@ -7,7 +7,8 @@ set -e
 cd $(git rev-parse --show-toplevel)/.dumps
 
 curl --remote-name --remote-header-name https://dsva-vagov-prod-cms-backup-sanitized.s3-us-gov-west-1.amazonaws.com/files/cms-prod-files-latest.tgz
-rm --force --recursive ../docroot/sites/default/files/*
+rm -rf ../docroot/sites/default/files/*
```
ElijahLynn: Ahh yes, BSD `rm` doesn't have the long options.

scripts/sync-files.sh
```diff
@@ -7,7 +7,8 @@ set -e
...
 echo "Extracting files to sites/default/files."
+mkdir -p ../docroot/sites/default/files
```
ElijahLynn:
```suggestion
mkdir --parents ../docroot/sites/default/files
```

indytechcook:
```
~/code/va/va.gov-cms/.dumps VAGOV-1237-staged-revision*
❯ mkdir --parents ../docroot/sites/default/files
mkdir: illegal option -- -
usage: mkdir [-pv] [-m mode] directory ...
```

scripts/sync-files.sh
oksana-c: @indytechcook , beat me to it 👍

PR #1252, opened 2020-03-16
Remove node revision which use the approved_by_reviewer state.

indytechcook:
## Description

Fix the node revisions which use the `approved_by_reviewer` moderation state.

## Testing done

* visited every node page/revision page and it did not error
** http://va-gov-cms.lndo.site/node/841/revisions
** http://va-gov-cms.lndo.site/node/992/revisions
** http://va-gov-cms.lndo.site/node/2587/revisions
** http://va-gov-cms.lndo.site/node/2628/revisions
** http://va-gov-cms.lndo.site/node/841
** http://va-gov-cms.lndo.site/node/992
** http://va-gov-cms.lndo.site/node/2587
** http://va-gov-cms.lndo.site/node/2628

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/1252)


docroot/modules/custom/va_gov_backend/va_gov_backend.install
```diff
@@ -139,3 +139,29 @@ function va_gov_backend_update_8002() {
...
+function va_gov_backend_update_8003() {
+  // These nid / revision ids are the ones in `approved_by_reviewer` status.
+  // 841 / 9646.
+  // 992 / 11407.
+  // 2587 / 15111.
+  $revisions_ids_to_draft = [9646, 11407, 15111];
+
+  foreach ($revisions_ids_to_draft as $vid) {
+    node_revision_delete($vid);
+  }
```
swirtSJW: Minor:  Generally when we have things happen in hook_update_n it happens out of sight.  If our drush was up-to-date , simply returning some message text that declares what was done should be good enough... but our drush is woefully outdated, so returned messages are never output :(   So it would be good to watchdog what happened, so that we have some record of the tree falling in the forest and no one around to hear it.

This is minor enough that I am not going to hold it back.

indytechcook: I added comments :)

indytechcook:
@ethanteague thanks!

indytechcook:
Log messages:

Recent log messages | VA gov CMS 2020-03-17 09-39-34

PR #1271, opened 2020-03-18
VACMS-1206 Add style to site_alerts.

swirtSJW:
## Description

See #1206 . 

## Testing done
viewed in both chrome and firefox.

## Screenshots
Medium Severity
image

High severity
image

low severity
image

## QA steps

1. Login as a user with admin privileges.
- [x] Validate that there is no site alert displayed at the bottom of the screen.  (assumes there are no active site alerts)
2. Visit /admin/config/system/site-alerts  and add a site alert with a severity of 'low'.  Give it a label  and a message.
- [x] Upon save, validate that there is "blue" tinted message at the bottom of the page that is fixed to the bottom of the page.
3. Edit the site alert you just created and change the severity to "Medium"  and save the site alert.
- [x] Validate that the message is now tinted yellow and fixed to the bottom of the screen.
4. Edit the site alert and set the severity to "High"  and save the site alert.
- [x] Validate that the message is now covering the active view port of the window and prevents the user from interacting with the elements of the page (with the exception of the admin toolbars).


kevwalsh:
*QA PASS*

*Design question*

* Why is the high alert yellow and not red? 

*Functionality*

- When a high alert is active, i can't edit admin/config/system/site-alerts without hacking the inspector. Can that path be excluded? I'm assuming that's a new patch
- Very minor: It would be great to get a the High/Medium/Low value as a column on the admin/config/system/site-alerts table. I'm assuming that's another patch

*Code questions*

- Why is the CSS in va_gov_backend and not the Seven subtheme, where it could be sassy?

swirtSJW:
> _Design question_
> 
> * Why is the high alert yellow and not red?

Jordana made that case, that red is scary and indicates they did something wrong, while in this case it is more of a message related to "nothing is wrong, but please don't do anything."

> _Functionality_
> 
> * When a high alert is active, i can't edit admin/config/system/site-alerts without hacking the inspector. Can that path be excluded? I'm assuming that's a new patch

When the site is in the middle of upgrade/deployment, admins shouldn't be doing anything either.  So that is intended.  All workers should stop what they are doing, including us.

> * Very minor: It would be great to get a the High/Medium/Low value as a column on the admin/config/system/site-alerts table. I'm assuming that's another patch  Yah that would be another patch.  This is only for admins and in our case, the alerts should not be managed by hand. They are managed by Jenkins.  While your suggestion is good, I am not sure it is one that will make a huge difference in our use case.   We could always patch later.  We already have three patches in play for this module.  The more we add, the more likely they are to start conflicting.
> 
> _Code questions_
> 
> * Why is the CSS in va_gov_backend and not the Seven subtheme, where it could be sassy?

I put it here because I know we are talking about switching themes, and I wanted it to work regardless of the active theme.

kevwalsh:
That all makes sense, although i think we'll want to get the CSS in the theme at some point.

PR #1279, opened 2020-03-20
VACMS-000 Update node_link_report to 1.12

swirtSJW:
## Description

Just a simple update to resolve a notice. that won't be visible until php is updated.


swirtSJW:
Closing this in preference of #1280

PR #1305, opened 2020-03-23
[WIP] VACMS-1239: hook_update to sync field_listing with field_office.

oksana-c:
## Description

**WIP**
**Latest revision date remains equal to previous revision date even though we set it explicitly. I'm now investigating this issue.** 

## QA steps

1. Do this
1. Then that
- [ ] This
- [ ] That
- [ ] The other thing

## Definition of Done
- [ ] Product release notes 


docroot/modules/custom/va_gov_backend/va_gov_backend.install
```diff
@@ -139,3 +139,36 @@ function va_gov_backend_update_8002() {
...
+      $node->setNewRevision(TRUE);
+      // New revision will inherit content moderation status from default rev.
+      $node->isDefaultRevision(TRUE);
+      $node->setRevisionLogMessage('VACMS Team: Synced field_listing with field_office for upcoming content model updates.');
+      $node->setRevisionCreationTime(Drupal::time()->getRequestTime());
+      $node->setChangedTime(Drupal::time()->getRequestTime());
+      // Setting revision as super admin.
+      $node->setRevisionUserId(1);
+      $node->save();
+    }
```
swirtSJW: Normally for a hook_update_N we want to return a message that should be output after a successful run of the function.  However, due to our use of an otdated version of Drush, these messages are not getting displayed in the terminal when they run.   Sadly our updates run silently, which is not good.  As a replacement/workaround, may you please have it log a message that indicates how many nodes were updated.

oksana-c: yes, can do that. Currently, if a node is not updated properly, there will be a log that comes from a function used by hook_entity_presave(), but can definitely log a summary.

docroot/modules/custom/va_gov_backend/va_gov_backend.install
```diff
@@ -139,3 +140,46 @@ function va_gov_backend_update_8002() {
...
+      $node->setRevisionLogMessage('VACMS Team: Synced field_listing with field_office for upcoming content model updates.');
+      $node->setRevisionCreationTime(Drupal::time()->getRequestTime());
+      $node->setChangedTime(Drupal::time()->getRequestTime());
+      // Setting revision as super admin.
+      $node->setRevisionUserId(1);
+      $saved = $node->save();
+      $updated = (is_int($saved) && $saved > 0) ? $updated + 1 : $updated;
+    }
+
+    $log_level = ($updated === count($nodes)) ? LogLevel::INFO : LogLevel::ERROR;
```
swirtSJW: Nice touch.

Issue #1332, opened 2020-03-25
Errors on config import locally after db import.


swirtSJW:
**Describe the defect**
Upon config import following importing the db locally with the ./scripts/sync-db.sh  An error occurs during config import

`Importing /app/.dumps/cms-prod-db-sanitized-latest.sql...

Import complete!
Cache rebuild complete.                                                                                                                                                                                                                              [ok]
No database updates required                                                                                                                                                                                                                         [success]
Drupal\Core\Entity\Exception\NoCorrespondingEntityClassException: The Drupal\user\Entity\User class does not correspond to an entity type. in Drupal\Core\Entity\EntityTypeRepository->getEntityTypeFromClass() (line 98 of                          [error]
/app/docroot/core/lib/Drupal/Core/Entity/EntityTypeRepository.php) #0 /app/docroot/core/lib/Drupal/Core/Entity/EntityBase.php(531): Drupal\Core\Entity\EntityTypeRepository->getEntityTypeFromClass('Drupal\\user\\Ent...')
#1 /app/docroot/vendor/drush/drush/lib/Drush/User/User8.php(26): Drupal\Core\Entity\EntityBase::load(0)
#2 /app/docroot/vendor/drush/drush/lib/Drush/Boot/DrupalBoot.php(558): Drush\User\User8->load_by_uid(0)
#3 /app/docroot/vendor/drush/drush/includes/bootstrap.inc(354): Drush\Boot\DrupalBoot->bootstrap_drupal_login()
#4 /app/docroot/vendor/drush/drush/includes/bootstrap.inc(473): drush_bootstrap(6, 6)
#5 /app/docroot/vendor/drush/drush/lib/Drush/Boot/BaseBoot.php(59): drush_bootstrap_to_phase(6)
#6 /app/docroot/vendor/drush/drush/includes/preflight.inc(67): Drush\Boot\BaseBoot->bootstrap_and_dispatch()
#7 /app/docroot/vendor/drush/drush/drush.php(12): drush_main()
#8 {main}.
[...]

PR #1333, opened 2020-03-25
VACMS-1332 Resolve error on db sync.

swirtSJW:
## Description

See #1332 

## QA steps
1. run `./scripts/sync-db.sh `
- [ ] Validate that the process runs fully without errors and results in a functional site.


swirtSJW:
I don't want to merge this yet.  I want to make sure this is an issue everyone is seeing.  I don't want this to mask an issue.

.lando.yml
```diff
@@ -9,6 +9,7 @@ events:
     # @TODO: Change to `composer yaml-tests va/deploy` once the "filter argument" feature is in place.
     - appserver: cd $LANDO_WEBROOT && /app/bin/drush cache-rebuild -y
     - appserver: cd $LANDO_WEBROOT && /app/bin/drush updatedb -y
+    - appserver: cd $LANDO_WEBROOT && /app/bin/drush cache-rebuild -y
```
ElijahLynn, line 12: @swirtSJW Can you add a comment here on why this is here so that we can maybe remove it in the future.

Issue #1328, opened 2020-03-25
Outreach events URL showing an event number instead of the event URL


howardbrande:
**Describe the defect**
Some of the events are showing a proper URL such as https://www.va.gov/outreach-and-events/events/midland-virginia-veterans-resource-summit/

however, others are showing with just a number such as https://www.va.gov/outreach-and-events/events/events-5/ 

is correct in the prod.cms.va.gov/node/2486/edit

**To Reproduce**
1. Go to https://www.va.gov/outreach-and-events/events/midland-virginia-veterans-resource-summit/ and then 
https://www.va.gov/outreach-and-events/events/events-5/ 
2. See error

**Expected behavior**
The event URL should have the event name from that automatically generated URL from the CMS

**Screenshots**

**Desktop (please complete the following information if relevant, or delete):**

**Additional context**

## Labels
- [/] Issue type (red) (defaults to "Defect")
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [/] CMS objective (orange) (not needed for bug tickets)
- [ ] CMS-supported product (black)

Issue #1191, opened 2020-03-09
Add pathauto to VAMC banner alerts and health services nodes


kevwalsh:
## User Story or Problem Statement

As a VAMC editor working on a VAMC nodes that don't have have FE stand-alone versions, i still want decent URL aliases, so that i know "where i am". 

## Acceptance Criteria
* VAMC banner alert pattern: `/[field_administration]/vamc-banner-alert/[date created]/[node title]`
* VAMC system health service `/[node:field_office]/health-services/[node:title]`
* VAMC facility health service `/[node:field_office]/[node:field_facility]/health-service/[node:title]`
* we have validated that no node types are missing pathauto. 

## Implementation steps
1. Update pathauto settings
2. Deploy to prod
3. Run pathauto updates on prod (consider an update hook for this too)

oksana-c:
Revised patterns:

-  VAMC banner alert pattern `[node:field_administration:entity:name]/vamc-banner-alert/[node:changed:html_date]/[node:title]`
-  VAMC system health service `[node:field_administration:entity:name]/system-health-services/[node:title]`
-  VAMC facility health service `facility-health-services/[node:title]`

PR #1336, opened 2020-03-26
VACMS-1191: New and refactored path patterns

oksana-c:
## Description

See #1191 , #1328. 

- Added path pattern for banners.
- Deleted duplicate pattern for events.
- Added pattern for VAMC system health services & facility health service
- Normalized path patterns naming to match content types.
- Added hook_update() to regenerate aliases for select CTs. Also included `event` CT here - see #1328 

## Testing done

Visual

## QA steps

1. Review path patterns at /admin/config/search/path/patterns
1. Check if aliases are generated according to defined patterns /admin/config/search/path/patterns by  filtering Content view by each of targeted content types (Event, VAMC system banner alert with situation updates, VAMC facility health service, VAMC system health service, Publication, Support Service) 
1. View a node from each of targeted content types and verify that new revision was not created when the node was saved via hook_update


docroot/modules/custom/va_gov_backend/va_gov_backend.install
```diff
@@ -226,3 +226,45 @@ function va_gov_backend_update_8004() {
...
+ */
+function va_gov_backend_update_8005() {
+  $updated = 0;
+  $failed = 0;
+  $failed_nids = [];
+
+  // Adding event ct here, to cover any aliases from VACMS-1328.
+  $bundles = [
+    'event',
+    'full_width_banner_alert',
```
swirtSJW: I don't think we want to trigger a save to existing banner alerts saves are capable of triggering a send to GovDelivery which can trigger a lot of repeat email to people.   Is there an essential reason to update banner alert aliases?

docroot/modules/custom/va_gov_backend/va_gov_backend.install
```diff
@@ -226,3 +226,45 @@ function va_gov_backend_update_8004() {
...
+
+    Drupal::logger('va_gov_backend')
+      ->log(LogLevel::INFO, 'Path Alias for %bundle: Successfully updated path alias for %updated out of %count entities. Failed to update %failed out of %count. %failed_nids', [
+        '%bundle' => $bundle,
+        '%count' => count($entities),
+        '%updated' => $updated,
+        '%failed' => $failed,
+        '%failed_nids' => 'Failed nids: ' . implode(', ', $failed_nids) . '.',
+      ]);
+  }
```
swirtSJW: Generally what kind of numbers are we expecting here? Less than 50, less than 100, less than 1000...?

oksana-c: not much - total 545 entities, that's why didn't batch them

snippets from log:
- Path Alias for support_service: Successfully updated path alias for 16 out of 16 entities.
- Path Alias for regional_health_care_service_des: Successfully updated path alias for 123 out of 123 entities.
- Path Alias for outreach_asset: Successfully updated path alias for 177 out of 177 entities.
- Path Alias for health_care_local_health_service: Successfully updated path alias for 157 out of 157 entities.
- Path Alias for event: Successfully updated path alias for 72 out of 72 entities.

docroot/modules/custom/va_gov_backend/va_gov_backend.install
oksana-c: when Kevin and I talked about aliases, we knew there's likely something that we may have missed. This is exactly the kind of feedback I needed. I'll reach out to connect on GovDelivery separately. 

There are only 2 banners on prod. It is not necessary to update aliases for them. I'll make that adjustment.

docroot/modules/custom/va_gov_backend/va_gov_backend.install
swirtSJW: Awesome thank you.

Issue #1089, opened 2020-02-27
Fix drupal views used by "create media" screens in D8.8


kevwalsh:
Drupal media views should be prefiltered, and filters for media type should not be visible. 

![ZoomAnnotationContentWnd.jpg](https://images.zenhubusercontent.com/5d6e57ad330f64311395f08b/61fdb704-9668-4658-98bc-4e24e9d89302)

kevwalsh:
@swirtSJW can you groom this?

PR #1337, opened 2020-03-26
VACMS-1089 Get medialibrary widget working properly.

swirtSJW:
## Description

See #1089 

THis PR essentially reverts the views to what media_library defines, then adds our customizations back to it.

## QA steps

1. As a content admin got to /node/add/documentation_page
1. Then scroll down to "Main Content" and click the "Add Content block" button.
1. Select the "EMBEDDED IMAGE" block type
1. Click "Add media" button.
  - [x] Validate that the media widget appears and only contains media of type images (no movies, no documents)
image

  - [x] Validate that the filters and disply toggle match 
image

5. Add another content block, by clicking "Add content block" 
6. Scroll down and select "LINK TO FILE OR VIDEO"
7.  Click "Add media" button.
   - [x] Validate that when the "Document" tab is selected, only documents show.
   - [x] Validate that when the "Image" tab is selected, only images show.
   - [x] Validate that when the "Video" tab is selected, only videos show show.
   - [x] Validate that name serach filter works by typing in "map" and then clicking "Apply filters"  You should only see maps.
   - [x] Validate that owner works by selecting "VA Butler health care while on the image tab and only a few results should be returned.
  - [x] Validate that selecting media and clicking "Insert selected" actually selects the media and inserts it into the node.

### DIFF
--- a/config/sync/views.view.media_library.yml
+++ b/config/sync/views.view.media_library.yml
@@ -4,21 +4,21 @@ status: true
 dependencies:
   config:
     - core.entity_view_mode.media.media_library
+    - image.style.media_library
     - taxonomy.vocabulary.administration
   enforced:
     module:
       - media_library
   module:
+    - image
     - media
     - media_library
     - taxonomy
     - user
-_core:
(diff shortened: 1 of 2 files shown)


kevwalsh:
PR #l337!

Issue #1375, opened 2020-04-01
Default Owner for VBA facility, NCA facility, Veterans Center nodes


oksana-c:
## Description

Set default owner for VBA facility, NCA facility, Veterans Center nodes:
- Owner for Veteran Centers should be “VHA facilities”  (TID 8, already exists)
- Owner for NCA facilities should be “NCA facilities” (needs creating)
- Owner for VBA facilities should be “VBA facilities” (needs creating)

ADDITIONAL:
- Add NCA and VBA owners to Facility Status view owner filter.

## Acceptance Criteria
- Owner for Vet Centers is “VHA facilities
- Owner for NCA facilities is “NCA facilities”
- Owner for VBA facilities is “VBA facilities”

## Labels
- [/] Issue type (red) (defaults to task)
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [ ] CMS objective (orange)
- [ ] CMS-supported product (black)

Issue #1369, opened 2020-04-01
Dashboard: Facility Filter, Search and Bulk Edit


oksana-c:
## Description

Dashboard
* As an editor who wants to edit operating status for multiple facilities, i can use a bulk edit screen to add values to both Operating Status and Operating status - more info for facilities, so that i can update my content faster. 
* As an editor i can filter by Operating Status (critical) -- never null and Operating status - more info (nice to have) (keyword? Is empty/ is not empty?) so that i can accomplish my tasks faster. 
* Similar to the “Update individual facility status” on /node/1010/edit
* URL: /admin/content/facility-status 

## Acceptance Criteria
- [ ]  Content Admin is able to filter facilities by type, operating status, operating status -more info
- [ ]  Content Admin is able to bulk edit facilities fields: operating hours, operating status -more info

## Labels
- [/] Issue type (red) (defaults to task)
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [ ] CMS objective (orange)
- [ ] CMS-supported product (black)

PR #1383, opened 2020-04-02
VACMS-1369: Facility governance dashboard.

oksana-c:
## Description

See #1369 . 

- Adds view page "Facility Status" at **/admin/content/facility-status**
- View is only available for users w/ roles: Content Admin, Administrator

**NOTE:** 
- due to the defect #1384 , "Modify field values" VBO action is not functional and may need to further configured once the defect is addressed. 
- Owner filter currently has only two options; once additional facility owner term are created on prod, I'll add them within #1375 .

## Testing done

Visual + CR

## Screenshots
Screen Shot 2020-04-02 at 3 37 54 PM

## QA steps

As user with Content Admin role (e.g. 1203 david.conlon@va.gov)
- [ ] visit /admin/content/facility-status
- [ ]  test filters
     - [ ] update some VAMC facilities to fill Operations status- more info to test full text search
     - [ ] search using full text filters
     - [ ] search using Operating status and Owner filters
- [ ]  Test permissions: log in as a user without Administrator or Content Admin permission. Eg. chanelle.blakely@va.gov . Confirm you get Access Denied response when visiting `/admin/content/facility-status`
 
## Definition of Done
- [ ] Product release notes 

### DIFF
--- a/config/sync/views.view.facility_governance.yml
+++ b/config/sync/views.view.facility_governance.yml
@@ -11,13 +11,17 @@ dependencies:
     - node.type.nca_facility
     - node.type.vba_facility
     - node.type.vet_center
+    - system.menu.admin
     - taxonomy.vocabulary.administration
     - user.role.administrator
     - user.role.content_admin
   content:
+    - 'taxonomy_term:administration:2c331a6d-b525-4f0c-8bea-4ecde41c7ef0'
     - 'taxonomy_term:administration:a8b1bb11-67a8-4489-a648-6d35c04be9cb'
     - 'taxonomy_term:administration:a9526e03-dda4-4dcc-b928-fcc5e56f4e5f'
+    - 'taxonomy_term:administration:edadc39e-dbaa-4fe1-b6a8-344357d6abfe'
   module:
+    - content_moderation
     - node
     - options
     - taxonomy
@@ -57,7 +61,7 @@ display:
       exposed_form:
         type: basic
         options:
-          submit_button: Search
+          submit_button: Filter
           reset_button: true
           reset_button_label: Reset
           exposed_sorts_label: 'Sort by'
@@ -67,7 +71,7 @@ display:
       pager:
         type: full
         options:
-          items_per_page: 10
+          items_per_page: 50
           offset: 0
           id: 0
           total_pages: null
@@ -102,6 +106,8 @@ display:
             field_operating_status_facility: field_operating_status_facility
             field_operating_status_more_info: field_operating_status_more_info
             field_administration: field_administration
+            moderation_state: moderation_state
+            changed: changed
             operations: operations
           info:
             views_bulk_operations_bulk_form:
@@ -137,12 +143,26 @@ display:
               separator: ''
               empty_column: false
               responsive: ''
+            moderation_state:
+              sortable: false
+              default_sort_order: asc
+              align: ''
+              separator: ''
+              empty_column: false
+              responsive: ''
+            changed:
+              sortable: true
+              default_sort_order: desc
+              align: ''
+              separator: ''
+              empty_column: false
+              responsive: ''
             operations:
               align: ''
               separator: ''
               empty_column: false
               responsive: ''
-          default: '-1'
+          default: changed
           empty_table: false
       row:
         type: 'entity:node'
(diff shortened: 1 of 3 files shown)


kevwalsh:
A few changes: 

- [ ] View should have some no results behavior, like "No facilities match those search criteria".
- [ ] Page it at 50 instead of the 10 default (that's typically our standard for these kinds of pages)
- [ ] Can you update the Owner filter to use the two new Owner fields create by Steve: NCA facilities and VBA facilities. 
- [ ] Add two new fields: Moderation state and Last updated. The latter should be sortable, default DESC. 
- [ ] Add a filter for Moderation state, since 7 (out of 2000+!) will be published.

kevwalsh:
Also, as user David.conlon@va.gov, i tried the VBO action to update field values, and it worked, i updated Facility status for two facilities. 

- [ ] One last thing: let's add a menu item under /admin/content for now, with a high weight so that it goes to the bottom

config/sync/views.view.facility_governance.yml
```diff
@@ -0,0 +1,850 @@
...
+        options:
+          disable_sql_rewrite: false
+          distinct: false
+          replica: false
+          query_comment: ''
+          query_tags: {  }
+      exposed_form:
+        type: basic
+        options:
+          submit_button: Search
```
swirtSJW, line 60: Double checking... should this be search, or filter, or apply.   We have a mixture in use around the site right now.  It would be good to standardize.

oksana-c: I'll poke around. Agreed, we do need to develop a standard.

oksana-c:
updated view
Screen Shot 2020-04-03 at 10 38 21 AM

oksana-c:
@kevwalsh I just added form UI improvements that I mentioned on call today. Please go over Modify field values form again once env builds and before this PR is merged. Also, @swirtSJW - see code updates. Not the most elegant way of altering the form, but it should do it in this iteration.

Screenshot of resulting form when two types of facilities are selected:

Screen Shot 2020-04-03 at 2 06 22 PM

kevwalsh:
Nice! QA Pass. when we decide to iterate on the form alter, we might want to look at some of the other VBO functionality on the site too.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -130,6 +130,63 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
...
+  if (!empty($entity_data)) {
+    foreach ($entity_data as $bundle => $data) {
+      if (is_array($data)) {
+        $bundle_fields[$bundle] = $data;
+      }
+    }
+  }
+
+  foreach ($bundle_fields as $bundle => $fields) {
+    if (is_array($fields)) {
```
swirtSJW, line 169: Minor: Is this check needed?  You were careful higher up to only populate it if was an array.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -130,6 +130,63 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
...
+            // Remove all facility content type fields that are not related
+            // to Operating status and Op info from vbo content type form.
+            unset($form['node'][$bundle][$field_name]);
+          }
+        }
+      }
+    }
+  }
+
+  // Remove multi-value field options, since none of the fields we're editing
```
swirtSJW, line 187: This one scares me a bit because we are making an assumption about future fields based on current fields.   I have a fear this will cause one of us much head scratching one day, but I don't have a better solution at this time.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -130,6 +130,63 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
     // @todo Consider disabling it instead of hiding it.
     $form['field_office']['#attributes']['class'] = 'hidden';
   }
+
+  if ($form_id === 'views_bulk_operations_configure_action') {
+    $build_info = $form_state->getBuildInfo();
+    // We need to target only the VBO "Modify field values" form from
+    // Facility Status tool.
+    if ($build_info['args'][0] === 'facility_governance' && $build_info['args'][1] === 'page_1') {
```
swirtSJW, line 138: Minor:  It would be nice if this logic also resided inside `_va_gov_backend_vbo_facility_status_form_ui(); so that the logic is more self contained and not adding to the clutter in this hook. (not your fault, there is a lot in there already)

swirtSJW:
This looks good.  Just waiting on tests to pass.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
oksana-c: this check is needed. The first check verifies whether bundle has any fields, second check is for field-specific render arrays, because we do not want to unset any non-array objects.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
oksana-c: I 👂 u

docroot/modules/custom/va_gov_backend/va_gov_backend.module
oksana-c: I absolve myself from any responsibility ....

joking

on the serious note, I did have this condition in the helper func originally, but consciously decided to move back to hook, so that I don't have to wrap the whole function in it or add a return.

swirtSJW: No return needed,  just pass by reference. :)

docroot/modules/custom/va_gov_backend/va_gov_backend.module
swirtSJW: Maybe I am misunderstanding.   When you are populating $bundle_fields  you are populating it with $data, only if it is an array.  When you are processing  $bundle_fields  you iterating it as $fields ... which is what you just put in as $data.  So $data = $fields   so if $data was an array, then $fields has to be an array

PR #1408, opened 2020-04-06
VACMS-1163: Updating new field with old field data in update hook.

ethanteague:
## Description
See #1163 

## Testing done
Visual

## Screenshots
image

image

## QA steps
 - [ ] As an admin, go to `/admin/structure/types/manage/page/form-display` and move disabled `Intro text` field to bottom of enabled list and save
- [ ] Go to `node/3014/edit` and visually verify that wysiwyg field appears at bottom of page, and is populated with content

### DIFF
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -400,14 +400,16 @@ function va_gov_backend_pathauto_alias_alter(&$alias, array &$context) {
 
 /**
  * Implements hook_entity_presave().
+ *
+ * @todo: Remove node page wysiwyg sync after VACMS-1163 is on prod.
  */
 function va_gov_backend_entity_presave(EntityInterface $entity) {
   $bundle_types = ['event', 'news_story', 'press_release', 'outreach_asset'];
 
   if ($entity->getEntityTypeId() === 'node' && in_array($entity->bundle(), $bundle_types)) {
-    _va_gov_backend_sync_office_listing_fields($entity);
     _va_gov_backend_feature_bump($entity);
   }
+
   if ($entity->getEntityTypeId() === 'node' && $entity->bundle() === 'page') {
     _va_gov_backend_sync_wysiwyg_fields($entity);
   }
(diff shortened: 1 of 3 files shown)


docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -404,10 +404,13 @@ function va_gov_backend_pathauto_alias_alter(&$alias, array &$context) {
 function va_gov_backend_entity_presave(EntityInterface $entity) {
   $bundle_types = ['event', 'news_story', 'press_release', 'outreach_asset'];
 
-  if ($entity->getEntityTypeId() == 'node' && in_array($entity->bundle(), $bundle_types)) {
+  if ($entity->getEntityTypeId() === 'node' && in_array($entity->bundle(), $bundle_types)) {
     _va_gov_backend_sync_office_listing_fields($entity);
     _va_gov_backend_feature_bump($entity);
   }
+  if ($entity->getEntityTypeId() === 'node' && $entity->bundle() === 'page') {
```
swirtSJW, line 411: We need @todos in here to remove these.   The others above should be gone already.    Please create another ticket to slate these for removal.

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -227,3 +228,48 @@ function va_gov_db_update_8008() {
...
+ */
+function va_gov_db_update_8009(&$sandbox) {
+  // Grab our nodes and set the count.
+  if (empty($sandbox['total'])) {
+    $nids = \Drupal::entityQuery('node')
+      ->condition('type', 'page')
+      ->execute();
+    $sandbox['total'] = count($nids);
+    $sandbox['current'] = 0;
+  }
```
swirtSJW, line 243: Nice use of sandbox!

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -227,3 +228,48 @@ function va_gov_db_update_8008() {
...
+
+  // Batch 25 at a time. Low number, but we have less than 300.
+  $nodes_per_batch = 25;
+
+  // Establish our place in the process.
+  $nids = \Drupal::entityQuery('node')
```
swirtSJW, line 249: You can actually do the full query of nids and save them in $sandbox to keep the state so that you don't have to keep getting them each time
`$sandbox['nids_process'] = [Full query here]`

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -227,3 +228,48 @@ function va_gov_db_update_8008() {
...
+  // Save the nodes to update fields, and count where we are.
+  foreach ($nids as $nid) {
+    $node = Node::load($nid);
+    $node->save();
+    $sandbox['current']++;
+  }
+
+  // Tell drupal we processed some nodes.
+  Drupal::logger('va_gov_db')
+    ->log(LogLevel::INFO, '%current nodes processed', [
```
swirtSJW, line 263: When you can be specific, be specific.
```suggestion
    ->log(LogLevel::INFO, 'Page nodes %current nodes saved to populate the html intro field', [
```

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -227,3 +228,48 @@ function va_gov_db_update_8008() {
...
+      '%current' => $sandbox['current'],
+    ]);
+
+  // Determine when to stop batching.
+  if ($sandbox['total'] === 0) {
```
swirtSJW, line 268: Major:  Sandbox `total` is set on the first run and is never updated after that.   It will never be 0.
$sandbox['#finished'] = ($sandbox['current'] / $sandbox['total']);  is fine all by itself as it will become equal to 1 when current becomes the same number as total.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: :+1: @swirtSJW https://github.com/department-of-veterans-affairs/va.gov-cms/issues/1421

docroot/modules/custom/va_gov_db/va_gov_db.install
ethanteague: :+1: @swirtSJW

docroot/modules/custom/va_gov_db/va_gov_db.install
ethanteague: Apologies, @swirtSJW - updating my pr sans this, b/c I don't understand what this should look like.

swirtSJW: It can live as is... but we should discuss so you can understand the idea.

PR #1433, opened 2020-04-09
VACMS-5673 cms analytics 

ethanteague:
## Description
See department-of-veterans-affairs/va.gov-team#5673

## QA
Was performed by @jonwehausen


ElijahLynn:
> Drupal\Core\Database\DatabaseExceptionWrapper: SQLSTATE[40001]:      [error]
> Serialization failure: 1213 Deadlock found when trying to get lock;
> try restarting transaction: INSERT INTO {cache_config} (cid, expire,
> created, tags, checksum, data, serialized) VALUE

Working on this in https://github.com/department-of-veterans-affairs/va.gov-cms/issues/1441

ElijahLynn:
Latest failure is because of https://github.com/department-of-veterans-affairs/va.gov-cms/issues/1441

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -301,15 +304,68 @@ function va_gov_backend_inline_entity_form_entity_form_alter(&$entity_form, $for
...
+  $sections = [];
+  $current_path = \Drupal::service('path.current')->getPath();
+  $request = \Drupal::request();
+  // Grab our page title.
+  if ($route = $request->attributes->get(RouteObjectInterface::ROUTE_OBJECT)) {
+    $page_title = \Drupal::service('title_resolver')->getTitle($request, $route);
+    $sections['pageTitle'] = $page_title;
+  }
+  $node = \Drupal::routeMatch()->getParameter('node');
+  if ($node instanceof NodeInterface) {
```
swirtSJW, line 341: Reviewer confusion (so might be nothing, might be major): Something seems odd here.   At this point $node can either be a node object or a node id depending on whether this was called from the node or a revision. 

https://www.drupal.org/project/drupal/issues/2730631

But then you check to see if it is a node instance, and if it is (meaning it is already the node object) you reload the node object.   Seems ackwards.  Seems like if it is not a node interface, then you should load it.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -301,15 +304,68 @@ function va_gov_backend_inline_entity_form_entity_form_alter(&$entity_form, $for
...
+    $loaded_node = \Drupal::entityTypeManager()->getStorage('node')->load($node->id());
+    $last_path_arg = end(explode('/', $current_path));
+    // If we are on node view, get the alias.
+    if (is_numeric($last_path_arg)) {
```
swirtSJW, line 345: Seems like a lot of assumptions here.  Why is the presence of a path arg the check to see if what you are on has an alias.  Maybe check the route instead?

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: You are right, @swirtSJW. We don't wanna track revision page stuff, so that makes this even simpler.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: Yeah, that makes sense :+1:

Issue #939, opened 2020-02-07
WYSIWYG for Benefits Detail page intro text


swirtSJW:
## User Story or Problem Statement

As a editor, I need a way to safely use html in the intro text so I can make the intro text full featured and valid html.

The current use has a plain text field with help text about allowed tags, however the plain text is field is not validated on the way in, it is only filtered at render time, however GraphQL is not a render, it is data, so nothing is preventing the field from containing all kinds of html tags as well as faulty html.
HTML "allowed" via current help text
```<em> <strong> <br> <p> <a>```

HTML tags allowed by field settings 
```<a> <b> <big> <code> <del> <em> <i> <ins> <pre> <q> <small> <span> <strong> <sub> <sup> <tt> <ol> <ul> <li> <p> <br> <img>```

## Acceptance Criteria

1. On content type page (benefit detail page) `field_intro_text`  is replaced with `field_intro_text_limited_html` 
2. All contents of the field_intro_text are moved to field_intro_text_limited_html  (roughy 233 nodes)
3. The field field_intro_text_limited_html uses a WYSIWIG that applies a limited html filter. 
4.  The allowed html tags should be `<em> <strong> <br> <p> <a>`
5.  Behat tests updated and passing
6.  Update the frontend query to  use the new field. (timing critical)
7. Collab w/ Rachel on helptext

## Implementation details

1. A hook_update_n should be added to the va_gov_db module to load all the nodes and copy the data from the old field to the new and save the data. 
2. This will have to go out in **FOUR** releases:
    1. CMS - Add the new field and make it hidden on node view and node edit.
    2. CMS - Update hook moves the content and hides the old field from node edit but not node view. Make new field visible on node edit and node view.
    3. VAWEB - update GraphQL query and template
    4. CMS remove field_intro_text from the `page` content type

## Labels
- [/] Issue type (red) (defaults to Enhancement)
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [ ] CMS Objective (orange)
- [ ] CMS-supported product (black)

swirtSJW:
@kevwalsh  Can you update the AC item to list what tags we do want to allow.  There is a disconnect between the help text and what is currently "allowed"

kevwalsh:
@swirtSJW updated! sorry i missed this on Friday.

PR #1434, opened 2020-04-09
VACMS-939: Updating wysiwyg text format to fix tag output.

ethanteague:
## Description
See #939 - oriignal update hook didn't set text format for new wysiwyg field, resulting in tags being encoded in output. This pr sets text format for new field, and prevents this from happening.

## Testing done
Visual

## Screenshots
image

## QA steps
 - [ ] Visit `education/about-gi-bill-benefits/montgomery-active-duty` and visually verify that intro text is not wrapped in visible <p> tags.

### DIFF
--- a/docroot/modules/custom/va_gov_db/va_gov_db.install
+++ b/docroot/modules/custom/va_gov_db/va_gov_db.install
@@ -270,3 +270,45 @@ function va_gov_db_update_8009(&$sandbox) {
   $sandbox['#finished'] = ($sandbox['current'] / $sandbox['total']);
 
 }
+
+/**
+ * Set field_intro_text_limited_html field to new wysiwyg type.
+ */
+function va_gov_db_update_8010(&$sandbox) {
+  // Grab our nodes and set the count.
+  if (empty($sandbox['total'])) {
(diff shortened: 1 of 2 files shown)


docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -270,3 +270,45 @@ function va_gov_db_update_8009(&$sandbox) {
   $sandbox['#finished'] = ($sandbox['current'] / $sandbox['total']);
 
 }
+
+/**
+ * Set field_intro_text_limited_html field to new wysiwyg type.
+ */
+function va_gov_db_update_8010(&$sandbox) {
```
swirtSJW, line 277: Minor: Since the code is exactly the same, it would have sufficed to just in increment the N number on the existing one.

ethanteague: I thought about that, and only added the new one to preserve the old update hook comment - glad to change if you think best, @swirtSJW

swirtSJW: Nah this is only a minor thing.  No need to change.

Issue #1436, opened 2020-04-09
Rename decision reviews sidebar nav 


kevwalsh:
## Acceptance Criteria
* left nav shows up on /decision-reviews/board-appeal/

PR #1437, opened 2020-04-09
VACMS-1436: Rename Decision Reviews menu.

kevwalsh:
## Description

See #1436

## Testing done
None yet. 

## QA steps

- [x] On PR environment created by this PR, publish all /decision-reviews benefits detail pages and then trigger a content build. 
- [x] decision-reviews/board-appeal/ should show a left nav

### DIFF
--- a/config/sync/system.menu.decision-reviews-benefits-h.yml
+++ b/config/sync/system.menu.decision-reviews-benefits-h.yml
@@ -9,6 +9,6 @@ third_party_settings:
     access_scheme:
       73: '73'
 id: decision-reviews-benefits-h
-label: 'Decision reviews benefits hub'
+label: 'Decision reviews'
 description: ''
 locked: false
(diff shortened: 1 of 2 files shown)


config/sync/system.menu.decision-reviews-benefits-h.yml
```diff
@@ -9,6 +9,6 @@ third_party_settings:
     access_scheme:
       73: '73'
 id: decision-reviews-benefits-h
-label: 'Decision reviews benefits hub'
+label: 'Decision reviews'
```
ncksllvn, line 12: The `benefits hub` aligns with the naming conventions of the other hub sidebars, but the `fieldAdministration` property on the Decision Reviews pages is only `Decision reviews`. So updating that `fieldAdministration` on all the pages would be more consistent with the rest of the pages but probably more work

kevwalsh: oh, are we talking about _right_ rail? fieldAdministration matters for right rail on the hub landing page (which it shouldn't, but it does for now), but it really should never matter for the _left_ nav on benefits detail landing pages. i think a few of us should get on the phone to understand what problem we are trying to solve here.

kevwalsh: PR environment for review: http://pr1437.web.ci.cms.va.gov/decision-reviews/higher-level-review/

swirtSJW:
image

Issue #1389, opened 2020-04-03
Push facility operating status data to Lighthouse


kevwalsh:
## User Story or Problem Statement

As a veteran, i want the latest facility status available via facility locator tool.

As an editor of a facility's operating status, i want my content to go immediately to the services that display facility status to veterans, so that i know they are get 

## Acceptance Criteria
- [ ] Lighthouse API receives the data within X seconds of saving new operating status info
- [ ] Success is logged in drupal log
- [ ] CMS-team slack is notified with certain problems -- TBD. 
- [ ] Possible next steps documented in a github issue 

## Assumptions
If Drupal has a facility that Lighthouse doesn't?

## Implementation steps
1. A queuer that will queue items on node_save if `field_operating_status_facility` or `field_operating_status_more_info` has been changed, which could be changed manually or by migration
1. Queue trigger to process the queue
1. Call the Lighthouse write/post API
1. Notification if the queue is stalled
1. Slack admin to work on getting a custom slack webhook notification

oksana-c:
push workflow - https://www.lucidchart.com/documents/edit/3b19f443-8d0a-4cd7-ac77-c1c6addda4fd/OdhJVqF-g3cm

PR #1438, opened 2020-04-10
[obsolete] VACMS-1389: content push api

oksana-c:
## Description

See #1389. 

## QA steps

1. Do this
1. Then that
- [ ] This
- [ ] That
- [ ] The other thing

## Definition of Done
- [ ] Product release notes 

### DIFF
--- a/docroot/modules/custom/content_push_api/content_push_api.links.menu.yml
+++ b/docroot/modules/custom/content_push_api/content_push_api.links.menu.yml
@@ -1,0 +1,20 @@
+# todo: Need to decide where these links should live exactly.
+# For now, placed them in Admin > Content menu right below, Facility Status dash link.
+content_push_api.admin_home:
+  title: Content Push API
+  description: 'Content Push API Administration.'
+  route_name: content_push_api.admin_home
+  parent: system.admin_content
+  weight: 101
+content_push_api.admin_queue:
+  title: Queue
+  description: 'Review and process Content Push Queue on demand.'
+  route_name: content_push_api.admin_queue
+  parent: content_push_api.admin_home
+  weight: 0
+content_push_api.admin_settings:
+  title: Settings
+  description: 'Configure Content Push API settings.'
+  route_name: content_push_api.admin_settings
+  parent: content_push_api.admin_home
+  weight: 1
--- a/docroot/modules/custom/content_push_api/content_push_api.module
+++ b/docroot/modules/custom/content_push_api/content_push_api.module
@@ -1,0 +1,65 @@
+<?php
+
+/**
+ * @file
+ * Contains content_push_api.module.
+ */
+
+use Drupal\Core\Entity\EntityInterface;
+
+/**
+ * Implements hook_entity_insert().
+ *
+ * Queues new items for sync.
+ */
+function content_push_api_entity_insert(EntityInterface $entity) {
+  _content_push_api_add_item_to_queue($entity);
+}
+
+/**
+ * Implements hook_entity_update().
+ *
+ * Queues updated items for sync.
+ */
+function content_push_api_entity_update(EntityInterface $entity) {
+  _content_push_api_add_item_to_queue($entity);
+}
+
+function _content_push_api_add_item_to_queue(EntityInterface $entity) {
+  if ($entity->getEntityTypeId() === 'node') {
+    // @todo: The content_moderation_state is an entity and it is also
+    // updated on entity save, so we need to figure out how to elegantly run
+    // queue item creation only for nodes.
+    // If there's a need to compare updated vs. original fields values, use the
+    // $entity->original in Payload Decorator.
+    $payload = Drupal::service('content_push_api.payload');
+    $queue_ops = Drupal::service('content_push_api.add_to_queue');
+    $endpoint = Drupal::service('content_push_api.endpoint');
+
+    $data['nid'] = $entity->id();
+
+    // Set a key based on which the endpoint will be
+    // defined during queue execution.
+    $data['endpoint_key'] = $endpoint->endpointKey($entity);
+
+    // Modifier that will be used in the endppoint url to target specific
+    // entity.
+    $data['modifier'] = $endpoint->endpointModifier($entity);
+
+    // Set payload. Default payload provided by this module is empty. You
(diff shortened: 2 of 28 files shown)


docroot/modules/custom/content_push_api/content_push_api.links.menu.yml
```diff
@@ -0,0 +1,20 @@
+# todo: Need to decide where these links should live exactly.
+# For now, placed them in Admin > Content menu right below, Facility Status dash link.
+content_push_api.admin_home:
+  title: Content Push API
+  description: 'Content Push API Administration.'
+  route_name: content_push_api.admin_home
+  parent: system.admin_content
```
swirtSJW, line 7: Maybe this should live under 'Web services' ?

This really just a push api.   In our case it is related to content, but in someone else's use case it could be connected to a user action, or an error handler.

docroot/modules/custom/content_push_api/content_push_api.module
```diff
@@ -0,0 +1,65 @@
...
+use Drupal\Core\Entity\EntityInterface;
+
+/**
+ * Implements hook_entity_insert().
+ *
+ * Queues new items for sync.
+ */
+function content_push_api_entity_insert(EntityInterface $entity) {
+  _content_push_api_add_item_to_queue($entity);
+}
```
swirtSJW, line 17: I think this is mixing concerns.   Where to call the service to add to add something to the queue should happen in the local custom module..  The concern of this module should just be to provide the queue service and the queue worker.

docroot/modules/custom/content_push_api/content_push_api.module
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Implements hook_entity_update().
+ *
+ * Queues updated items for sync.
+ */
+function content_push_api_entity_update(EntityInterface $entity) {
+  _content_push_api_add_item_to_queue($entity);
+}
+
+function _content_push_api_add_item_to_queue(EntityInterface $entity) {
```
swirtSJW, line 28: I don't think this handling should be in the contrib module.  It should reside in the custom.  Otherwise you end up making too many assumptions and be coming opinionated about what people should be doing with it.

docroot/modules/custom/content_push_api/content_push_api.module
```diff
@@ -0,0 +1,65 @@
...
+  if ($entity->getEntityTypeId() === 'node') {
+    // @todo: The content_moderation_state is an entity and it is also
+    // updated on entity save, so we need to figure out how to elegantly run
+    // queue item creation only for nodes.
+    // If there's a need to compare updated vs. original fields values, use the
+    // $entity->original in Payload Decorator.
+    $payload = Drupal::service('content_push_api.payload');
+    $queue_ops = Drupal::service('content_push_api.add_to_queue');
+    $endpoint = Drupal::service('content_push_api.endpoint');
```
swirtSJW, line 38: There are a lot of different services here.   I was thinking a single service that  would be invoked and have payload and enpoint parameters set  as methods within the add_to_queue service   something like this 
https://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/docroot/modules/custom/va_gov_govdelivery/src/Service/ProcessStatusBulletin.php#L105-L121

swirtSJW: I think abstracting it out to three different services makes it a nightmare to document how anyone else should use the services and get them working together.

docroot/modules/custom/content_push_api/src/Form/ContentPushQueueForm.php
```diff
@@ -0,0 +1,165 @@
...
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $queue = $this->queueFactory->get('content_push_queue');
+    $queue_items = $this->getItems('content_push_queue');
+
+    $rows = [];
+
+    foreach ($queue_items as $item) {
+      // @todo: display payload data in readable format.
```
swirtSJW, line 84: nice!

docroot/modules/custom/content_push_api/src/Form/ContentPushSettingsForm.php
```diff
@@ -0,0 +1,91 @@
...
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $config = $this->config('content_push_api.settings');
+
+    $form['endpoint_host'] = [
+      '#type' => 'textfield',
+      '#title' => $this->t('API Endpoint'),
+      '#description' => $this->t(
+        'Stores CMS Facility API endpoint host in database - THIS IS NOT RECOMMENDED.
```
swirtSJW, line 65: nice

docroot/modules/custom/content_push_api/src/Form/ContentPushSettingsForm.php
```diff
@@ -0,0 +1,91 @@
...
+        "application/json"'),
+      '#default_value' => $config->get('header_content_type'),
+    ];
+
+    $form['logging'] = [
+      '#type' => 'details',
+      '#title' => $this->t('Logging and Notifications'),
+    ];
+
+    $form['logging']['slack'] = [
```
swirtSJW, line 124: I think making the contrib module require slack is extreme.  Not everyone uses slack.   It fits our use case, but it would be better just to have  a way to let the custom module handle the external triggers... maybe creating an event and event subscriber.?

docroot/modules/custom/content_push_api/src/Service/Request.php
```diff
@@ -0,0 +1,112 @@
...
+    $sending_data = Json::encode($payload);
+
+    try {
+      $response = $this->httpClient->request('POST', $endpoint, [
+        'headers' => $headers,
+        'body' => $sending_data,
+      ]);
+      return $response->getStatusCode();
+    }
+    catch (ServerException $e) {
```
swirtSJW, line 98: Nice use of multiple catches!

ElijahLynn:
Latest failure is because of https://github.com/department-of-veterans-affairs/va.gov-cms/issues/1441

oksana-c:
started a separate PR. closing this.

Issue #1441, opened 2020-04-10
[cms-ci] PR builds failing with `SQLSTATE[40001]:      [error] Serialization failure: 1213 Deadlock found when trying to get lock; try restarting transaction: INSERT INTO {cache_config}`


ElijahLynn:
Multiple PRs (#1433, #1438, and #1304) are now failing with:
> SQLSTATE[40001]:      [error]
Serialization failure: 1213 Deadlock found when trying to get lock;
try restarting transaction: INSERT INTO {cache_config}

I attempted to resolve by setting:
`transaction-isolation = READ-COMMITTED` in /etc/my.cnf per https://www.drupal.org/project/drupal/issues/2833539#comment-12500749, which we have applied to the upstream ENVs RDS param group a while ago, and never applied to CI. But they are still happening now. 

> mysql> SELECT @@global.tx_isolation, @@session.tx_isolation;
> +-----------------------+------------------------+
> | @@global.tx_isolation | @@session.tx_isolation |
> +-----------------------+------------------------+
> | READ-COMMITTED        | READ-COMMITTED         |
> +-----------------------+------------------------+

ElijahLynn:
This all started happening yesterday I think, so I want to be cognizant that an upstream DB change may have exacerbated this issue, and to look at code changes in the past 24-48 hours. Would love to be able to reproduce locally.

ElijahLynn:
We are testing now the results of #1445 being deployed to PROD to see if this resolves the issue.

ElijahLynn:
- [ ] Note to link the Post-Mortem/RCA to this document.

PR #1445, opened 2020-04-10
Remove # from standbox drush update

indytechcook:
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/1445)

### DIFF
--- a/docroot/modules/custom/va_gov_db/va_gov_db.install
+++ b/docroot/modules/custom/va_gov_db/va_gov_db.install
@@ -267,7 +267,7 @@ function va_gov_db_update_8009(&$sandbox) {
     ]);
 
   // Determine when to stop batching.
-  $sandbox['#finished'] = ($sandbox['current'] / $sandbox['total']);
+  $sandbox['finished'] = ($sandbox['current'] / $sandbox['total']);
 
 }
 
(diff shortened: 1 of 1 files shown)


ElijahLynn:
Confirmed this is updating the schema now. 
> drush sql:query "select name,value from key_value where name = 'va_gov_db'"

Might be related to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/1441.

indytechcook:
Also see: https://api.drupal.org/api/examples/batch_example%21batch_example.module/8.x-1.x

ElijahLynn:
And here is with this PR:
image

And current master shows like this, repeating the update hook name 10 times:
image

indytechcook:
I'm merging this even with the deadlock issue so we can get this deployed since is it causing updb to break.

ElijahLynn:
Caught this error before it was destroyed, just putting it here for now:

> [4/4] Building fresh packages...
> 
> error An unexpected error occurred: "ENOENT: no such file or directory, lstat '/var/aegir/projects/cms/pr1445/docroot/vendor/va-gov/web/node_modules/global-agent/node_modules/core-js/es/array/virtual/join.js'".
> 
> info If you think this is a bug, please open a bug report with the information provided in "/var/aegir/projects/cms/pr1445/docroot/vendor/va-gov/web/yarn-error.log".
> info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
> 
> Script export NODE_EXTRA_CA_CERTS=/etc/pki/tls/certs/ca-bundle.crt PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=TRUE; cd web && yarn install handling the va:web:install event returned with error code 1
> 
> Script @va:web:install was called via post-install-cmd

http://devshop.cms.va.gov/node/19482

PR #1485, opened 2020-04-14
VACMS-1054 Add migration yaml tasks

swirtSJW:
## Description

See _issueid_. 

## QA steps

1. Do this
1. Then that
- [ ] This
- [ ] That
- [ ] The other thing

## Definition of Done
- [ ] Product release notes 


tasks-background.yml
```diff
@@ -0,0 +1,38 @@
+# These are tasks that are called by Jenkins similar to cron..
+
```
ElijahLynn, line 2: When we have the URL to the job we need to put it here.

PR #1517, opened 2020-04-16
VACMS-1521: claimItem in the database and memory queues does not use e…

oksana-c:
…xpire correctly

## Description

This is a follow up to queue item processing, queue item lease expire dates.

## QA steps

**Problem statement**: in certain circumstances (yet unknown/unreproducible), queue items lease expire timestamps are not cleared on queue item release. If expiration date is cleared, the queue worker will not touch that item ever again.

Expected behavior: if the queue item has expire value set to a timestamp that signifies past data, the queue worker should claim the item during the next processing task and either successfully process it or release back to queue w/ expire timestamp set to `0`.

[Core patch](https://www.drupal.org/project/drupal/issues/2893933#comment-13413895) applied in this PR fixes the problem.

[See Slack thread for context](https://dsva.slack.com/archives/CT4GZBM8F/p1587059574195200)

Also, phpunit tests were originally failing after core patch applied due to `govdelivery_bulletins` `claimItem(0)` which now should only be a positive integer. I already updated post_api to fix the same issue and posted a [patch](https://www.drupal.org/project/govdelivery_bulletins/issues/3128425) for `govdelivery_bulletins` which I applied here. @swirtSJW , if you'd like to create a new release w/ that patch and include version update in this PR - it'd be great.

**To reproduce and test the fix for this issue:**

1. `git checkout master`
1. open va_gov_post_api.module in IDE docroot/modules/custom/va_gov_post_api/va_gov_post_api.module and on **line 65**  change the condition `if (!empty($data['payload']) && $facility_id) {` to `if (!empty($data['payload'])) {`
1. run perms.feature tests `lando behat --tags=perms`, wait until tests are complete
    - [ ] visit https://va-gov-cms.lndo.site/admin/config/post-api/queue  and confirm that 3 out of 4 items have expiration dates that are in the past
    - [ ] Process the queue. Confirm that expire dates left unchanged - this means that _none_ of the items w/ expire dates were processed
1. git checkout this branch
1. `lando composer install`
   - [ ] visit https://va-gov-cms.lndo.site/admin/config/post-api/queue  and confirm that 3 out of 4 items have expiration dates that are in the past
    - [ ] Process the queue. Confirm that expire dates are now set to `0` - this means that _all_ of the items w/ expire dates were processed and released correctly

## Definition of Done
- [ ] Product release notes 

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -224,8 +224,8 @@
                 "Content lock causes redirect response preventing migration rollback.": "https://www.drupal.org/files/issues/2018-08-28/2951652-4.patch"
             },
             "drupal/core": {
-                "2815221 - Add quickedit to the latest-revision route": "https://www.drupal.org/files/issues/2019-03-05/2815221-116.patch",
-                "Allow menu items which link to unpublished nodes to be selected in the parent item selector": "https://www.drupal.org/files/issues/drupal-core-allow-menu-items-to-link-to-unpublished-2807629-7-8.3.x.patch"
+                "2807629 - Allow menu items which link to unpublished nodes to be selected in the parent item selector": "https://www.drupal.org/files/issues/drupal-core-allow-menu-items-to-link-to-unpublished-2807629-7-8.3.x.patch",
+                "2893933 - claimItem in the database and memory queues does not use expire correctly": "https://www.drupal.org/files/issues/2020-01-07/2893933-34.patch"
             },
             "drupal/entity_browser": {
                 "Entity display plugin Rendered Entity is missing cache tags/keys": "https://www.drupal.org/files/issues/2020-03-31/rendered-entity-cache-keys-tags-3123876-2.patch"
--- a/composer.json
+++ b/composer.json
@@ -237,6 +237,9 @@
                 "Allow to position the group in the advanced (sidebar) column": "https://www.drupal.org/files/issues/2018-07-26/2652642-59.patch",
                 "Incompatibility with inline entity form": "https://www.drupal.org/files/issues/2018-10-30/field_group-inline_entity_form_compatibility-2986704-3.patch"
             },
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -225,8 +225,8 @@
                 "Content lock causes redirect response preventing migration rollback.": "https://www.drupal.org/files/issues/2018-08-28/2951652-4.patch"
             },
             "drupal/core": {
-                "2815221 - Add quickedit to the latest-revision route": "https://www.drupal.org/files/issues/2019-03-05/2815221-116.patch",
```
indytechcook, line 228: Why was this removed?

oksana-c: it's a duplicate. The original reference is installed by `lightning_core/composer.json`, so we don't need to apply it again in here.  Patches combined from ./composer.json and lightning_core/composer.json are also recorded in our main composer.lock file  fhttps://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/composer.lock#L4052

oksana-c: our composer files needed some love for some time...

composer.json
```diff
@@ -225,8 +225,8 @@
...
-                "Allow menu items which link to unpublished nodes to be selected in the parent item selector": "https://www.drupal.org/files/issues/drupal-core-allow-menu-items-to-link-to-unpublished-2807629-7-8.3.x.patch"
+                "2807629 - Allow menu items which link to unpublished nodes to be selected in the parent item selector": "https://www.drupal.org/files/issues/drupal-core-allow-menu-items-to-link-to-unpublished-2807629-7-8.3.x.patch",
```
swirtSJW, line 228: Thank you for including the issue number  :)

oksana-c:
@swirtSJW you brought up a valid concern. Thanks,

P.S. connected to #1521

Issue #1546, opened 2020-04-21
Stories at a VA Pittsburgh are missing the /pittsburgh-health-care/ part of URL


kevwalsh:
**Describe the defect**
URL pattern for stories is missing the field_office (via field_story_listing)

**To Reproduce**
1. Create a story and assign to Pittsburgh
2. URL will be stories/[title]

**Expected behavior**
URL should be /pittsburgh-health-care/stories

PR #1547, opened 2020-04-21
VACMS-1546: Fix URL aliases for events news stories and press releases.

kevwalsh:
## Description

See #1546. 

## Testing done
Manual

## QA steps

As user ryan.stubblebine@va.gov with content_publisher role
1. Create a story, event, and news release
1. Assign all to Pittsburgh (owner and listing)
1. URLs should all start with /pittsburgh-health-care


oksana-c:
@kevwalsh @swirtSJW we need to set the same pattern for Publications.

PR #1560, opened 2020-04-21
Add mysql init commands for read commits

indytechcook:
## Description

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/1560)


ElijahLynn:
If we move to doing all of this in settings.php, I want to remove our custom RDS param group and any changes to the CI my.cnf as well. So let's create a follow up issue for that if/when merged.

PR #1603, opened 2020-04-23
VACMS-1528 Disable revision autoclean cron and set limits on unlimited.

swirtSJW:
The setting of cron was adding an item to the queue for every node saved.
This was causing the queue table to grow out of control to the point
where it got too big to process on cron.

## Description

See _issueid_. 

## QA steps

As user _uid_ with admin
1. Go to here /admin/config/content/revisions-autoclean
- [ ] Validate that the cron option is not enabled 
image

- [ ] Validate that all "Limit revisions for node type *" have a revision count number set that is NOT 0.

### DIFF
--- a/config/sync/node_revisions_autoclean.settings.yml
+++ b/config/sync/node_revisions_autoclean.settings.yml
@@ -1,11 +1,11 @@
-enable_on_cron: 1
+enable_on_cron: 0
 enable_on_node_update: 1
 _core:
   default_config_hash: UblN00sCWoyDV-yZe4SE80vKQEQmZHLWmxtaQmJLLaI
 node:
   documentation_page: 50
   event: 50
-  health_care_local_facility: 50
+  health_care_local_facility: 500
   health_care_region_detail_page: 50
   health_care_region_page: 50
   landing_page: 50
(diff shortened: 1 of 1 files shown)


config/sync/node_revisions_autoclean.settings.yml
```diff
@@ -18,18 +18,21 @@ node:
...
-  publication_listing: 0
-  press_releases_listing: 0
-  leadership_listing: 0
+  event_listing: 50
+  publication_listing: 50
+  press_releases_listing: 50
+  leadership_listing: 50
   vba_facility: 50
-  nca_facility: 0
+  nca_facility: 50
```
swirtSJW, line 30:
```suggestion
  nca_facility: 200
```

config/sync/node_revisions_autoclean.settings.yml
```diff
@@ -18,18 +18,21 @@ node:
...
   veterans_center: 0
-  vet_center: 0
+  vet_center: 50
```
swirtSJW, line 32:
```suggestion
  vet_center: 200
```

config/sync/node_revisions_autoclean.settings.yml
```diff
@@ -18,18 +18,21 @@ node:
...
   full_width_banner_alert: 50
-  event_listing: 0
-  publication_listing: 0
-  press_releases_listing: 0
-  leadership_listing: 0
+  event_listing: 50
+  publication_listing: 50
+  press_releases_listing: 50
+  leadership_listing: 50
   vba_facility: 50
```
swirtSJW, line 29:
```suggestion
  vba_facility: 200
```

Issue #1243, opened 2020-03-13
List a user's sections on their user/UID page


kevwalsh:
AC
* As a CMS user visiting my user page, i see a list of all the sections i belong to (Workbench access Sections taxonomy), with links to those section pages, so that i can access my content easily.
* As a content_admin or admin who bypassess workbench access, display the existing  taxonomy tree view that's currently visible on /sections page.
* As a user I want to see my sections in a flat list, without hierarchy

Implementation
* Drupal view? 
* Layout (getting this on the page) 
* Use/Modify UserPermsService to return taxonomy ID per user access

Issue #996, opened 2020-02-14
Implement a service that combines role and Section


kevwalsh:

kevwalsh:
Where's the PR for this? Not sure how to QA this.

Issue #927, opened 2020-02-07
Redirect users on login to their content (section taxonomy term)


kevwalsh:
### User Story or Problem Statement

Most CMS users only edit content in one section of the site. Eg

- a Public Affairs Officer at VAMC Pittsburgh only edits VAMC Pittsburgh content
- a Web Manager at edits VISN 4 content (and VAMC system content "below" it)
- a VHA editor edits VHA content (health care benefits hub, VISN 4 content, etc)
- a Content admin on the Public Websites team edits content in sections across the site. 

Ideally upon login, these users would be redirected to a dashboard of their content. 

Content admins and admins, who are typically not assigned to a section, can just be redirected to /admin/content, for now.

### Affected users and stakeholders

* All  users

## Acceptance Criteria
* ~~user X will be redirect to Section Y~~
* ~~user _foo_ will be redirected to Section _bar_~~
* ~~user A will be redirected to Section B~~
* ~~users who are members of more than one section are redirected to a list of sections  they belong to (on user profile page?)~~
*  Send all users to their user page where they will see a view block that shows the section links for any section they belong to.

## Runbook  
This is particularly relevant for feature launches, or for making changes to the content model that require front end changes as well.
1. Step_to_get_to_production_X  
1. Step_to_get_to_production_Y
1. Step_to_get_to_production_Z

## Possible tickets to create for this epic

- [ ] Spike task, with some of the following attributes:
* Identify the baseline against which we will measure improvements
* Ideation (sketching, cross-displinary meeting of minds around how the problem may be solved)
* Analysis of the drupal landscape
* Possible implementation(s)
* Proof of concept or prototype
* UX writing discovery

- [ ] User research
- [ ] Design
- [ ] UX writing
- [ ] Usability testing (eg with a prototype or other design stimula)
- [ ] Implementation
  - [ ] A drupal service that can be called to get  user's "membership"  (getMemberships()) and the memberships parents (getMembershipsParents())  This will also be needed for #930 
- [ ] Documentation (What needs changing or adding to the documentation?
- [ ] Product release notes (How will we describe this feature to CMS users?)
- [ ] Training (Will any users require additional training?)
- [ ] Metrics (How will we measure any improvements)

## Labels for  this issue
- [/] Issue type (red) (defaults to Epic)
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [ ] CMS objective (orange)
- [ ] CMS-supported product (black)

[...]

PR #1683, opened 2020-05-04
VACMS-1243: Render user sections on their profiles.

oksana-c:
## Description

See #1243. 

- adds custom block that renders user sections on user profile
   - block output varies depending on whether user has no sections assigned, some section assigned, root section assigned "All sections" or user has roles Administrator or Content Admin
   - users can now view own and other users' sections by visiting their profiles (given they have permission to view other users' profiles)

Note: decision was made to add custom block instead of View, due to the following factors:
- control of pages on which the block appears is much simpler via custom route check - e.g. render only on /user/* would render block on user view page but also on user edit pages and others. We'd need to add another contrib module in order to set correct path pattern in Block UI
- logic for differentiating users with zero, some and all sections would have to live in preprocess functions of a theme that will likely be changed soon. Custom block landed itself well to this requirement w/ much less code needed
 
**User has NO sections assigned** - ( opened UX writing issue for @rachel-kauff to iterate on placeholder lingo - #1684 )
Screen Shot 2020-05-04 at 1 54 21 PM

**User has some sections assigned**
Screen Shot 2020-05-04 at 1 57 44 PM

**User is assigned to a root section which provides access to all sections OR user has role Content Admin or Administrator**
Screen Shot 2020-05-04 at 1 55 52 PM

## QA steps

As user 1

- [x] visit https://va-gov-cms.lndo.site/admin/people, open profiles of several users in that list while comparing sections displayed in SECTION column with those displayed on user profile View page
- [x] login as `Shawnee.Blair@va.gov` and visit `/user` page - confirm you see Sections block w/ a. message `Contact #cms-support for VA section assignment.`
- [x] login as `Laura.Walsh1@va.gov` and visit `/user` page - confirm you see 4 sections
- [x] as an admin - add a new section for Laura Walsh , then view profile page as user `Laura.Walsh1@va.gov` and confirm you see updated list of sections immediately after page refresh
- [x] login as `howie.brande@va.gov`. Howie does not have roles "Content Admin" or "Administrator", but has access to root section. Confirm you see a tree of all sections when viewing `/user`  page as Howie
- [x] as an admin or user w/ role "Content Admin", view `/user` page and confirm you see a tree of all sections 

## Definition of Done
- [ ] Product release notes 
- [x] A link has been provided to the originating GitHub issue (or connected to it via ZenHub)

### DIFF
--- a/docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php
+++ b/docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php
@@ -82,7 +82,7 @@ class UserSections extends BlockBase implements ContainerFactoryPluginInterface 
     // Get currently viewed profile.
     $user = $this->routeMatch->getParameter('user');
 
-    // Other users need to see only sections assigned to their profiles.
+    // Get ids of sections assigned to user profile.
     $query = $this->database->select('section_association__user_id', 'sau');
     $query->join('section_association', 'sa', 'sau.entity_id = sa.id');
     $query->condition('sau.user_id_target_id', $user->id());
@@ -102,6 +102,8 @@ class UserSections extends BlockBase implements ContainerFactoryPluginInterface 
       return $view->buildRenderable();
     }
 
+    // Use has access only to some sections.
+    // Compose list.
     $entity_storage = $this->entityTypeManager->getStorage('taxonomy_term');
     $terms = $entity_storage->loadMultiple($results);
 
--- a/docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php
+++ b/docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php
@@ -1,0 +1,144 @@
+<?php
+
+namespace Drupal\va_gov_user\Plugin\Block;
+
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Database\Connection;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Url;
+use Drupal\views\Views;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a block of links to User Sections.
+ *
+ * @Block(
+ *   id = "va_gov_user_sections",
+ *   admin_label = @Translation("Sections")
+ * )
+ */
+class UserSections extends BlockBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * The route match.
+   *
+   * @var \Drupal\Core\Routing\RouteMatchInterface
+   */
+  protected $routeMatch;
+
+  /**
+   * Database connection.
+   *
+   * @var \Drupal\Core\Database\Database
+   */
+  private $database;
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
+    return ['user'];
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition, RouteMatchInterface $route_match, Connection $database, EntityTypeManagerInterface $entity_type_manager) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->routeMatch = $route_match;
+    $this->database = $database;
+    $this->entityTypeManager = $entity_type_manager;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('current_route_match'),
+      $container->get('database'),
+      $container->get('entity_type.manager')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    // Get currently viewed profile.
+    $user = $this->routeMatch->getParameter('user');
+
+    // Other users need to see only sections assigned to their profiles.
+    $query = $this->database->select('section_association__user_id', 'sau');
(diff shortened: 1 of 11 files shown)


docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    // Get currently viewed profile.
+    $user = $this->routeMatch->getParameter('user');
+
+    // Other users need to see only sections assigned to their profiles.
```
oksana-c, line 85:
```suggestion
    // Get ids of sections assigned to user profile.
```

docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php
```diff
@@ -0,0 +1,144 @@
...
+    if ($user->hasRole('content_admin') || $user->hasRole('administrator') || in_array('administration', $results)) {
+      $view = Views::getView('sections_tree');
+      $view->setDisplay('block_1');
+      $view->preExecute();
+      $view->execute();
+
+      return $view->buildRenderable();
+    }
+
+    $entity_storage = $this->entityTypeManager->getStorage('taxonomy_term');
```
oksana-c, line 105:
```suggestion
    // Use has access only to some sections.
    // Compose list.
    $entity_storage = $this->entityTypeManager->getStorage('taxonomy_term');
```

docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    // Get currently viewed profile.
+    $user = $this->routeMatch->getParameter('user');
+
+    // Get ids of sections assigned to user profile.
+    $query = $this->database->select('section_association__user_id', 'sau');
+    $query->join('section_association', 'sa', 'sau.entity_id = sa.id');
```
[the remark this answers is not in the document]
swirtSJW: I like idea of creating this module, va_gov_user.  It makes sense to consider moving the UserPermsService.php from va_gov_backend and into this module.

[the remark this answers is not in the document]
oksana-c: @swirtSJW 
agreed on moving the UserPermsService.php 👍 

re: `->getSections( $user)` -  was there a separate issue open for this?

docroot/modules/custom/va_gov_user/src/Service/UserPermsService.php
```diff
@@ -57,6 +74,45 @@ private function getUser($user_id = NULL) {
...
+    // Get ids of sections assigned to user profile.
+    $query = $this->database->select('section_association__user_id', 'sau');
+    $query->join('section_association', 'sa', 'sau.entity_id = sa.id');
+    $query->condition('sau.user_id_target_id', $user->id());
+    $query->fields('sa', ['section_id']);
+    $results = $query->execute()->fetchCol();
+
+    if (($key = array_search('administration', $results)) !== FALSE) {
+      unset($results[$key]);
+      $sections['administration'] = 'All sections';
```
oksana-c, line 101: not thrilled about having non-term section option included in term return array - `administration` = access to "All sections". If there are any ideas of how we can work around it more elegantly, let me know. Also added a `NOTE: ...`  to method return value. to document this anomaly.

docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php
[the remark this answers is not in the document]
swirtSJW: I don't believe it made it into an issue.  It was Just Ethan and I kicking around what the service would provide for our needs on 
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/927

I think this was the stub ticket that should have had the details in it, but I never added them before it got closed. 
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/996

docroot/modules/custom/va_gov_user/src/Service/UserPermsService.php
oksana-c: more context - we do need to return the `administration` option because otherwise we dont have a way of discerning whether of not the user has access to All sections.

docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php
```diff
@@ -0,0 +1,150 @@
...
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('current_route_match'),
+      $container->get('va_gov_user.user_perms')
+    );
+  }
```
swirtSJW, line 73: I  really need you to teach me about this style.   I get it... but I don't get it well enough.

docroot/modules/custom/va_gov_user/src/Service/UserPermsService.php
```diff
@@ -57,6 +65,45 @@ private function getUser($user_id = NULL) {
...
+   *   Array of user sections where keys are term ids and values are term names.
+   *   Special key 'administration' signifies user's assignment to ALL sections.
+   *   NOTE: consumers of this method will have to take care of array key
+   *   'administration' according to their use case.
+   */
+  public function getSections(AccountInterface $user) {
+    $sections = [];
+
+    // Get ids of sections assigned to user profile.
+    $query = $this->database->select('section_association__user_id', 'sau');
```
swirtSJW, line 84: Doesn't a call to `userOptionsStorage`  get all the sections the person has access to?  It should be all cached and ready to go.   I thought we could have bypassed this query.

oksana-c: hm, good question - I had to go back and take a good look at it. We cannot use it that method, because it returns only some of the sections that user has access to. 
Caching of block results is accounted for in block plugin. So it only needs to run once per each user/user path and cache is invalidated as soon as user is assigned a new section or unassigned from anything (Workbench access form save).

swirtSJW: Ok.  I guess that makes sense.    Thank you.

swirtSJW:
image


image



image

swirtSJW:
@rachel-kauff Release notes should not be a part of this as from the user's point of view, nothing has changed yet until we remove the redirect to /sections, which is not part of this ticket.

Issue #1705, opened 2020-05-07
Content editor can bypass editorial workflow permissions on bulk edit


kevwalsh:
**Describe the defect**
As a content editor (but not content_publisher), i can bulk edit content and set it to Published, from a section page.

**To Reproduce**
1. Log in as Laura.Walsh1@va.gov
2. Go to `/section/health-care-benefits-hub`
3. You can see lots of actions

**Expected behavior**
You shouldn't be able to see these actions unless you are content publisher

PR #1710, opened 2020-05-07
VACMS-1705: Enable and configure actions permissions.

kevwalsh:
## Description

See #1705. 

The original scope was small, but once i looked at the configuration for action permissions it made sense to apply a number of them to the existing use case. 

### Content editors shouldn't be able to bulk publish content, unless they have content_publisher role as well. 
1. Log in as Laura.Walsh1@va.gov
2. Go to http://pr1710.ci.cms.va.gov/section/health-care-benefits-hub
3. No bulk publishing. 

Only admins and users should be able to bulk update users. 

###  Content publishers should be able to bulk publish content in their section, but no other options. and no bulk publishing in other sections. 

1. Log in as ryan.stubblebine@va.gov
2. Go to http://pr1710.ci.cms.va.gov/section/veterans-health-administration/vamc-facilities/va-coatesville-health-care
3. No bulk publishing. 

FAIL. Content publishers can still bulk publish content that they shouldn't have access to. 🤔  

### Content admins should still be able to bulk publish content in any section, but no other options

1. Log in as randi.hecht@va.gov
2. Go to http://pr1710.ci.cms.va.gov/section/health-care-benefits-hub
3. "Publish latest revision" is an available action


kevwalsh:
Despite the fail on the second user story, i think this should not prevent merge. We could spin off a separate ticket for it.

swirtSJW:
QA 1: Pass
image

QA 2: Fail
image   Covered by a separate ticket so, ignoring this fail.

QA 3: Pass
image

Issue #998, opened 2020-02-14
Implement redirects based on section membership


kevwalsh:
AC
~~ Redirect single-section user to their Section page (taxonomy/term/tid)~~
~~Redirect multi section user or content_admin or administrator to user page (user/UID)~~
- [x] Send all users after login to their user page where they will see the sections they belong in.

Implementation

- [x] Uninstall redirect_after_login module
- [ ] ~~either specifiy the home page as /user/ or set a redirect  (investigate both and implement one.~~ This was not needed as the the default behaviour when no home pages is specified (like our current site is) is to rout users to their user page.

PR #1709, opened 2020-05-07
VACMS-998 Uninstall module redirect_after_login

swirtSJW:
## Description

See #998. 

## QA steps

1. As anon user try to visit the home page http://pr1709.ci.cms.va.gov/
   - [x] Validate that log in screen appears. 

2. Login as selina.cooper@va.gov 
   - [x] Validate that after login, you are on /user/1343 
   - [x] validate you only see the following sections
image
   - [x] Validate that when you try to go to the home page  http://pr1709.ci.cms.va.gov/ you end up redirected to /user/1343

3. Logout by going to http://pr1709.ci.cms.va.gov/user/logout
4. Login as yourself as an admin user.  
   - [x] Validate that you end up on your user page  and you see a complete hierarchy of sections.
   - [x] Validate that when you try to go to the home page  http://pr1709.ci.cms.va.gov/ you end up redirected to your user page.

## Definition of Done
- [x] QA'd
- [ ] Product release notes written

PR #1735, opened 2020-05-08
VACMS-998 Remove redirect_after_login from codebase.

swirtSJW:
## Description

See #998  

THis module has already been uninstalled and made its way to production.  This is simply the followup to remove it from the codebase.

## QA steps

There is no QA as there is nothing to test.


swirtSJW:
Followup to #1709

Issue #1476, opened 2020-04-14
Facility post queue should respect Lighthouse API rate limit


oksana-c:
## Description

https://www.drupal.org/project/post_api/issues/3127573

Slack conversation: https://dsva.slack.com/archives/CT4GZBM8F/p1586815780004700

## Acceptance Criteria

* Facility post queue respects Lighthouse API rate limit and does not fill the log with 429 request responses

## Labels
- [/] Issue type (red) (defaults to task)
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [ ] CMS objective (orange)
- [ ] CMS-supported product (black)

PR #1740, opened 2020-05-11
VACMS-1476: Post API rate limit config.

oksana-c:
## Description

See  #1476 

https://www.drupal.org/project/post_api/issues/3127573

## Screenshots

**New setting: API rate limit per minute**
_defaults to 0, in which case no rate limiting will be applied_
_screenshot depicts current facility api rate limit - 1000_
Screen Shot 2020-05-11 at 12 03 17 PM

## QA steps

As user 1
- [x] confirm Api rate limit is set to 1000 at /admin/config/post-api/config
- [x] truncate `queue` table for a more controlled experiment ` lando drush sql-query "truncate table queue;"`
- [x] queue up VAMC facility nodes (total 1296) at /admin/config/va-gov-post-api/facility-force-queue
- [x] ensure you dont have endpoint/ apikey set locally. This will assist in verifying the amount of items processed under a set rate limit
- [x] process queue manually at /admin/config/post-api/queue . Locally you should not have more than 1000 items processed due to api rate limiting but also due to max runtime of 60 secs. Item "expire" date is a reliable indicator that the this item was processed. Confirm this by browsing to page 100 (/admin/config/post-api/queue?page=100) - you should see no expire dates in the table after the item with id 1000. Go back a page or two and check which item was processed last. In my local experiments, around 990 items are processed within the limit. Your mileage may vary.

## Definition of Done
- [ ] Product release notes 


swirtSJW:
I reviewed the code and found one issue in post_api related to negative time that needs to be addressed.  Other that that, all looks great.

swirtSJW:
Tests are passing in devshp but not updating here. 
image

PR #1776, opened 2020-05-18
VACMS-import Wilkes-Barre Services.

sundarka1:
## Description

See _issueid_. 

## QA steps
Node data
image

- [x] 1 Lookup gets correct VHA service name and description.
- [x] 2 WYSIWIG content looks correct.  
- [x] 3 Text format correct
- [x] 4 draft time, author  correct.
- [x] 5 VAMC system correct.
- [x] 6 Owner set correctly

Revisions data
image
- [x] 1 revision date and author 
- [x] 2 Log message meaningful
- [x] 3 owner
- [x] 4 parent
- [x] 5 & 6 Recent change log matches.

Validate "Primary Care"
Primary Care appears in the service lookup in two places.  Validate that the child not the parrent is being set.

image

- [x] 1  Primary Care selected
- [x] 2 vid 37 selected.

## Definition of Done
- [ ] Product release notes 


swirtSJW:
Missing item: Author and revision information is needed
https://dsva.slack.com/archives/CDHBKAL9W/p1588339541369200?thread_ts=1588339541.369200

config/sync/migrate_plus.migration.regional_health_care_service.yml
```diff
@@ -20,17 +20,26 @@ source:
...
+      field_administration: Owner
+
 process:
   field_service_name_and_descripti: field_service_name_and_descripti
   field_body: field_body
-  field_region_page: field_region_page
-  field_administration: field_administration
+  field_region_page:
+    plugin: default_value
+    default_value: 2361
```
swirtSJW, line 39: Major: The value for this is in your CSV as "Health System" but it is not being used because the source is not defined..  We don't want to hard code this value in as te default  for 'VA Wilkes-Barre health care'  That makes this migration specific to Willkes-Barre.    It should get the system from the data so that the migration can import all of them.  Otherwise we would have to have a migration for each system and we lose the ablility to map them.

config/sync/migrate_plus.migration.regional_health_care_service.yml
```diff
@@ -20,17 +20,26 @@ source:
...
+  field_administration:
+    plugin: default_value
+    default_value: 186
```
swirtSJW, line 42: Major: This exists in the CSV as the column  "Owner", but then this line is overiding the data in all cases with  186 because the `source` of that field has not been defined.  This should   taxonomy lookup.

config/sync/migrate_plus.migration.regional_health_care_service.yml
```diff
@@ -20,17 +20,26 @@ source:
...
     - id
   column_names:
-    -
-      id : 'The unique identifier for the service'
-    -
-      field_service_name_and_descripti: 'Health Service'
-    -
-      field_body: 'Full Upload'
+    0:
+      id: Id
```
swirtSJW, line 24: minor:  The description lost some meaning here.

config/sync/migrate_plus.migration.regional_health_care_service.yml
```diff
@@ -20,17 +20,26 @@ source:
...
+    1:
+      field_service_name_and_descripti: Health Service
+    2:
+      field_body: Full Upload
+    3:
+      field_region_page: Health System
+    4:
+      field_administration: Owner
+
 process:
```
swirtSJW, line 34: Major:  You want to make sure you add the process plugin recipe to get the timestamps, revision messages, and authors showing up. https://www.drupal.org/docs/8/modules/migration-tools/migrate-process-plugin-recipes#s-revisions-process-recipe

config/sync/migrate_plus.migration.regional_health_care_service.yml
```diff
@@ -20,17 +20,26 @@ source:
...
   field_service_name_and_descripti: field_service_name_and_descripti
   field_body: field_body
```
swirtSJW, line 36: Major:  The body field since it is WYSIWYG field  has two elements.  ! for the value of the field, the other is for the format.
```
process:
  'body/format':
    plugin: default_value
    default_value: full_html
  'body/value': example_matched_field
```
This example comes from this documentation https://www.drupal.org/docs/8/api/migrate-api/migrate-process-plugins/migrate-process-overview

docroot/modules/custom/va_gov_migrate/data/regional-health-care-service.csv
```diff
@@ -1,3 +1,3 @@
-"Id","Health Service","Full Upload"
-"VAMC-1","Addiction and substance abuse treatment","We help Veterans on their recovery journey through outpatient and inpatient services."
-"VAMC-2","Audiology and Speech Pathology","Our Audiology and Speech Pathology clinic provides diagnostic and rehabilitative services, VA compensation and pension examinations, and screening for employee hearing conservation."
+Id,Health Service,Full Upload,Health System,Owner,Uid
```
swirtSJW, line 1: Major:  The user id should not be part of the data.     Hard coded drupal id's should not be in your migration source.

config/sync/migrate_plus.migration.regional_health_care_service.yml
```diff
@@ -21,19 +21,61 @@ source:
...
+    default_value: true
+  revision_log:
+    plugin: default_value
+    default_value: "Update of status by migrations."
+  revision_timestamp:
+    plugin: callback
+    callable: time
+  revision_uid:
+    plugin: default_value
+    default_value: uid
```
swirtSJW, line 64: Major: This should not be coming from your migration source.   It also should not be your user name.   The content changes should not be attributed to you(uid: 1268), they should be attributed to CMS Migrator (uid: 1317), just like in our other migrations, and the recipe I provided.

sundarka1: So we can use the uid:1317 as default value in configuration.

config/sync/migrate_plus.migration.regional_health_care_service.yml
```diff
@@ -21,19 +21,59 @@ source:
...
-  field_body: field_body
-  field_region_page: field_region_page
-  field_administration: field_administration
+  'body/format':
+    plugin: default_value
+    default_value: full_html
+  'body/value': field_body
+  field_region_page:
+    plugin: default_value
+    default_value: field_health_system
```
swirtSJW, line 41: Major: This is not really what the default value plugin is for.   It is to be used when there is no value in the source.
These should be entity lookups.

config/sync/migrate_plus.migration.regional_health_care_service.yml
```diff
@@ -21,19 +21,59 @@ source:
...
+  field_administration:
+    plugin: default_value
+    default_value: field_owner
```
swirtSJW, line 44: Major: This is not really what the default value plugin is for.   It is to be used when there is no value in the source.
These should be entity lookups.

docroot/modules/custom/va_gov_migrate/data/regional-health-care-service.csv
```diff
@@ -1,3 +1,3 @@
-"Id","Health Service","Full Upload"
-"VAMC-1","Addiction and substance abuse treatment","We help Veterans on their recovery journey through outpatient and inpatient services."
-"VAMC-2","Audiology and Speech Pathology","Our Audiology and Speech Pathology clinic provides diagnostic and rehabilitative services, VA compensation and pension examinations, and screening for employee hearing conservation."
+Id,Health Service,Full Upload,Health System,Owner
+VAMC-1,33,"We help Veterans on their recovery journey through outpatient and inpatient services. Our residential facility helps Veterans deal with substance abuse, homelessness, mental health, and unemployment through its 24-hour therapeutic and educational programs like:",2361,186
+VAMC-2,47,"Our Audiology and Speech Pathology clinic provides diagnostic and rehabilitative services, VA compensation and pension examinations, and screening for employee hearing conservation.,VA Wilkes-Barre health care""",2361,186
```
swirtSJW, line 3: Major:  There should not be node, or other entity ids in your migration source.  Health service, Health system and Owner should be entity lookups and not be node id's

config/sync/migrate_plus.migration.regional_health_care_service.yml
swirtSJW: The process plugin you want is entity_lookup https://www.drupal.org/docs/8/api/migrate-api/migrate-process-plugins/list-of-process-plugins-provided-by-migrate-plus

swirtSJW:
@sundarka1 
This is looking better.   It looks like you have a git conflict that may have been resolved incorrectly during your rebase. 
image


If this is working for you locally and is ready to be tested, please complete the section of the ticket related to what testing you have done and more importantly what manual testing steps should be done to confirm that this migration is working the way it is expected to work.

cc: @SgardnerPMP

config/sync/migrate_plus.migration.regional_health_care_service.yml
SgardnerPMP: Agreed.  Spoke to Sundar and his next commit should have 'VA Wilkes-Barre health care' removed from script and will rely on csv file to provide system data.  Please confirm that this has been complete in latest commit.  And we can resolve this Major.  Thank you.

config/sync/migrate_plus.migration.regional_health_care_service.yml
```diff
@@ -14,23 +14,67 @@ label: 'Node - VAMC System Health Service.'
...
+      field_health_system: 'Health System'
 process:
-  field_service_name_and_descripti: field_service_name_and_descripti
+  field_service_name_and_descripti:
+    plugin: entity_lookup
+    entity_type: taxonomy_term
+    value_key: name
+    bundle_key: vid
+    bundle: health_care_service_taxonomy
+    source: field_service_name_and_descripti
```
swirtSJW, line 40: Since `Primary care` appears in the taxonomy in two places, once as a parent (vid = 155) that should never be selected, and once as the child (vid = 37), we need to make sure that if the entity lookup returns 155, that we convert that to use 37.   The [static_map process plugin](https://api.drupal.org/api/drupal/core%21modules%21migrate%21src%21Plugin%21migrate%21process%21StaticMap.php/class/StaticMap) will do that like this:
```suggestion
  field_service_name_and_descripti:
   -
      plugin: entity_lookup
      entity_type: taxonomy_term
      value_key: name
      bundle_key: vid
      bundle: health_care_service_taxonomy
      source: field_service_name_and_descripti
    -
      plugin: static_map
        map:
          155: 37
        bypass: TRUE
```

swirtSJW: @sundarka1 I am not seeing this change added.  Was there a reason you chose not to add it?  

I called it out specifically in Slack.
image

cc: @SgardnerPMP

config/sync/migrate_plus.migration.regional_health_care_service.yml
```diff
@@ -14,23 +14,67 @@ label: 'Node - VAMC System Health Service.'
...
+    callable: time
+  new_revision:
+    plugin: default_value
+    default_value: true
+  revision_default:
+    plugin: default_value
+    default_value: true
+  revision_log:
+    plugin: default_value
+    default_value: 'Update of status by migrations.'
```
swirtSJW, line 68: @sundarka1  I have commented on this a few times now.   Please revise this message to be meaningful to what is actually happening in this migration.   This is not a facility status migration.
cc: @SgardnerPMP

config/sync/migrate_plus.migration.regional_health_care_service.yml
sundarka1: Services dropdown list has only one "Primary care". That value is selected correctly with entity lookup.
This selects correctly(37).

swirtSJW: @sundarka1 This is doing an entity lookup, which knows nothing about the dropdown list which does not show the parent.   The entitly lookup is just getting lucky that it is returning the right one.   The suggestion I made, needs to be there so we are not relying on luck for it to get the right one.
cc: @SgardnerPMP

swirtSJW: image

swirtSJW:
Closing temporarily to cause devshop to rebuild.

Issue #1657, opened 2020-05-01
Jenkins Deploys failing due to test timeouts


indytechcook:
We have several jenkins jobs failing due to performance tests failing.  This is occurring more consistently:

```
       Time: 55.14 seconds, Memory: 136.00MB
   
       There was 1 failure:
   
       1) tests\phpunit\LoginPerformance::testLoginPerformance with data set #0 (2)
   
       Operation took 2.021 seconds which is longer than the benchmark of 2 seconds.
   
       Failed asserting that '2.021' is less than 2.
   
       /var/www/cms/tests/phpunit/PerformanceLoginTest.php:47
```

ElijahLynn:
I dug into this locally with Blackfire and it _appeared_ to be the "Find content by section" block on the `/user/UID` path that was adding a half second to the response. The block is installed by va_gov_user.module that is loaded as the last request from `$this->drupalLogin($author);`. I profiled 5 runs with it enabled and without and noticed a 22.57% (.46 seconds) increase in request time, from 2.06 (disabled) > 2.52 (enabled) seconds. The [Blackfire profile diff can be viewed here, showing which functions increased time](https://blackfire.io/profiles/compare/aa7a5f32-9d74-4ae5-9fd8-fd1af596ff71...fa483559-4955-4e0a-ad75-d68fc13c3e0a/graph?settings%5Bdimension%5D=cpu&settings%5Bdisplay%5D=focused&settings%5BtabPane%5D=nodes&selected=&callname=Drupal%5CCore%5CRender%5CRenderer%3A%3AexecuteInRenderContext). 

![image.png](https://images.zenhubusercontent.com/5cdc72c50b7efe3f164150b2/6223b8d1-1165-41db-979d-7457744b377d)

![image.png](https://images.zenhubusercontent.com/5cdc72c50b7efe3f164150b2/9e25dfc7-5cbd-4524-b49b-4386f3d4b7cd)

But _then_ I realized I had my settings.local.php to `$settings['cache']['bins']['page'] = 'cache.backend.null';` which is not what DEV/STAGING have. So I unset that and ran again and now my benchmarks were showing a -32.21% (decrease) in response time for `/user/UID`, from 0.75 (disabled) > 0.51 (enabled) seconds. 

![image.png](https://images.zenhubusercontent.com/5cdc72c50b7efe3f164150b2/b7d191f2-7228-4015-99da-5ec2d77f3593)

[...]

PR #1808, opened 2020-05-21
Set DEV to match STAGING for CSS/JS/CACHE/GZIP.

ElijahLynn:
@swirtSJW, @oksana-c and I discussed this earlier today and noticed DEV did not match STAGING while investigating #1657. We discussed either having a DEV environment that was different, to enable more debugging on it, but if that was the case to not run tests on it that passed a commit status, because they would be different than PROD. We didn't make a decision to do that just yet, so in the mean time I think since tests do run on both DEV/STAGING and pass commit status then we should make DEV match STAGING/PROD in the meantime.


indytechcook:
> I think it might be prudent to not have the environments match exactly, but make it so that success on dev can not approve passing a deployment of prod.

This would be accomplished by having `yaml-test` run as `--dry-run`

ElijahLynn:
> > I think it might be prudent to not have the environments match exactly, but make it so that success on dev can not approve passing a deployment of prod.
> 
> This would be accomplished by having `yaml-test` run as `--dry-run`

I am down with having STAGING be the single source of truth for tests passing. Let's do it!

indytechcook:
Canceling this for https://github.com/department-of-veterans-affairs/devops/pull/6992

Issue #1651, opened 2020-04-30
Site Alerts should not cause a cache rebuild during a deploy


indytechcook:
## Summary

When a site alert is making an Ajax request and a cache rebuild is occurring at the same time, the site alert ajax request tries to rebuild the cache the same time it's being cleared.  This has the potential to use database locks and possible deadlocks, which will become more frequent as our scale increases. 

## Final Solution:

**Return static HTML for `/ajax/site_alert` requests.**

During the initializing of `settings.php` a deploy mode check was added.  The deploy mode check looks for the environmental variable `VA_GOV_IN_DEPLOY_MODE`.  If the variable exists then the logic checks what path is being requested.  If the path is `ajax/site_alert` a static HTML response is sent back to the ajax call which the `site_alert`module is making.  The static HTML is returned by throwing a `SuccessHTTPException` which will cause Drupal to stop the bootstrap/initialization process, not try to fill any caches and just return the HTML in the message.

## Possible solutions:

* Create a custom PHP file which will bootstrap drupal to the database, check a maintenance status (assuming the site is in maintenance mode) and returning instead of continuing the bootstrap process
* Override the kernal (or hook into)
* Have a "deploying" `.htaccess` file which will return 404 for the ajax request

swirtSJW:
One idea might be a temporary addition to .htaccess, to simply 404 any request for the ajax endpoint during the deploy.

indytechcook:
The first PR to add is ready: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/1860.

A follow up PR is needed on the devops repo to to rename the settings file from docroot/sites/default/settings/deploy-settings.skip_cache.php to docroot/sites/default/settings/settings.skip_cache.php during the deploy process.

indytechcook:
I'm holding off on the devsops changes until @ElijahLynn finishes #1719 and #1850

indytechcook:
This will be resolved with with https://github.com/department-of-veterans-affairs/va.gov-cms/issues/2851

PR #1826, opened 2020-05-22
Do not cache on site alert requests

indytechcook:
## Description

See https://github.com/department-of-veterans-affairs/va.gov-cms/issues/1651.

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/1826)


docroot/sites/default/settings.php
```diff
@@ -168,6 +168,40 @@
   include $app_root . '/' . $site_path . '/settings/settings.fast_404.php';
 }
 
+// Adjust cache settings if we are on a site alert
+if ($_SERVER['REQUEST_URI'] === 'ajax/site_alert') {
```
swirtSJW, line 172: Would adding a check for maintenance mode make this more performant?  So that caching the site_alert works normally, but then when the site is in maintenance mode, we don't allow the hit to cache.

swirtSJW: or maybe we just use the check to redirect to an empty html file?

indytechcook:
Replacing with: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/1860

PR #1844, opened 2020-05-26
[WIP] VACMS-9237 | 9738: Adding modules and config for service address hardening.

ethanteague:
## Description
See https://github.com/department-of-veterans-affairs/va.gov-team/issues/9237

## Testing done
Visual

## Screenshots
image

## QA steps

...


docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/ValidNumberValidator.php
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Validates the ValidNumber constraint.
+ */
+class ValidNumberValidator extends ConstraintValidator {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validate($items, Constraint $constraint) {
+    foreach ($items as $item) {
```
oksana-c, line 17:
```suggestion
    // Get parent entity id. In this case, the value returned is an ID of a `phone_numbers` paragraph.
    // You can then find paragraph parents.
    $parent_entity_id = $items->getEntity()->id();

    foreach ($items as $item) {
```

oksana-c: the difference here is that we were targeting single item instead of item collection to find a parent

PR #1860, opened 2020-05-27
Add and Configure Skip Cache module

indytechcook:
## Description

This avoids cache on the `/ajax/site_alert` call when the file is rename from `docroot/sites/default/settings/deploy-settings.skip_cache.php` to `docroot/sites/default/settings/settings.skip_cache.php` during the deploy process.

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/1860)

### DIFF
--- a/docroot/sites/default/settings.php
+++ b/docroot/sites/default/settings.php
@@ -183,5 +183,10 @@ if (file_exists($app_root . '/' . $site_path . '/settings/settings.local.php')) 
   include $app_root . '/' . $site_path . '/settings/settings.local.php';
 }
 
+// Should Skip Patches
+if (file_exists($app_root . '/' . $site_path . '/settings/settings.skip_cache_check.php')) {
+  include $app_root . '/' . $site_path . '/settings/settings.skip_cache_check.php';
+}
+
 $settings['tome_content_directory'] = 'public://cms-export-content';
 $settings['tome_files_directory'] = 'public://cms-export-files';
--- a/docroot/sites/default/settings/deploy-settings.skip_cache.php
+++ b/docroot/sites/default/settings/deploy-settings.skip_cache.php
@@ -1,0 +1,46 @@
+<?php
+
+use Symfony\Component\HttpFoundation\Request;
+
+$settings['skip_all_caches_enabled'] = TRUE;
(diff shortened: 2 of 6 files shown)


docroot/sites/default/settings.php
```diff
@@ -183,5 +183,10 @@
   include $app_root . '/' . $site_path . '/settings/settings.local.php';
 }
 
+// Should Skip Patches
+if (file_exists($app_root . '/' . $site_path . '/settings/settings.skip_cache_check.php')) {
```
swirtSJW, line 187: Instead of relying on Jenkins to move this file around, could we just make this switch contingent on being in maintenance mode?   That way it would behave the same on all environments. and not get stuck in some wierd state if the file did not get removed.

indytechcook:
The maintenance flag is stored in the state which is stored in the DB.  We
do not have access to the DB at the point.

docroot/sites/default/settings.php
indytechcook: The maintenance flag is stored in the state which is stored in the DB.  We do not have access to the DB at the point.

docroot/sites/default/settings/deploy-settings.skip_cache.php
```diff
@@ -0,0 +1,46 @@
...
+  'migrate',
+  'page',
+  'render',
+  'rest',
+  'static',
+  'tome_static',
+  'toolbar',
+];
+
+include $app_root . '/' . $site_path . '/settings/skip_all_caches.inc';
```
indytechcook, line 41: This line needs to be deleted

indytechcook, line 41:
```suggestion
```

indytechcook, line 41:
```suggestion
```

ElijahLynn:
Adding the fact that the new `skip_all_caches` module you created is upstreamed here > https://www.drupal.org/project/skip_all_caches.

swirtSJW:
@indytechcook I am going to close this as I know you are taking a differrent approach.   We can always re-open this if we need it.

Issue #1904, opened 2020-06-01
Event location help text is confusing


howardbrande:
**Describe the defect**
On the Create event node, physical address of a VA facility is not populating when selecting a VA facility from the 'Facility location' dropdown

**To Reproduce**
1. Log in to http://cmstraining.demo.ci.cms.va.gov as howard.brande@va.gov 
2. Click on 'Manage' > 'Content' > '+ Add content' > 'Event' 
3. Enter
- Name of event: Test
- Date and time: Start: today's date and time. End: same but later time
- For governance select: VA Pittsburgh health care 
- Event listing: Choose 'VA Pittsburgh health care: event'
- Scroll down to location
- Location type: Choose 'VA facility...'
- Facility location: Select 'Pittsburgh VA Medical Center - university Drive'
- Scroll down and select 'save and continue editing'
- You will receive an error message saying 'Please fill out this field'  (Street address) 
- Change 'Country' to 'None'
- Select 'save and continue editing'
- nose will then save as epxected

**Expected behavior**
Based on the help text, when selecting a VA facility for facility location, a physical address is not needed. Changing the country to 'None' is intended for online event

**Screenshots**
https://www.loom.com/share/79558a246a534707805eec41bd0758e8

**Desktop (please complete the following information if relevant, or delete):**

**Additional context**

## Labels
- [/] Issue type (red) (defaults to "Defect")
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [/] CMS objective (orange) (not needed for bug tickets)
- [ ] CMS-supported product (black)

PR #1906, opened 2020-06-02
VACMS-1904: Improve help text for event facility fields.

kevwalsh:
## Description

See #1904. 

## Testing done

Manual.

## QA

Go to /node/add/event. For the facility field, help text no longer instructs user to not include the address.


kevwalsh:
@swirtSJW Can you merge this for today? Howie's doing an events training tomorrow and it would be great to include it.

Issue #1872, opened 2020-05-29
Add navigable "child" sections to a user's profile page sections listing


kevwalsh:
## Problem: 

The sections listing on a user's profile page only shows a user's sections membership, but does not also show the child sections that they also have access to. 

![Edward_Boehme_va_gov___VA_gov_CMS.jpg](https://images.zenhubusercontent.com/5d6e57ad330f64311395f08b/5dbd5722-7b62-437a-a0d4-e00bcb2d663c)

## Example:

A [VAMC upgrade team member](https://prod.cms.va.gov/user/1300) shows VHA facilities access, but not all the child facilities. 

## Hypothesis

By adding child sections (possibly with a collapsible navigation) users will be able to find their section, and their content, faster.

Issue #1302, opened 2020-03-23
Get accessibility tests passing


swirtSJW:
### User Story or Problem Statement

As a developer I need our accessibility tests to pass so that when we see failures, we can take action to address them.

As a CMS user I need our CMS to meet accessibility requirements so that I can use the site without restrictions.

This is an epic, so that each failure can get its own issue.

### Affected users and stakeholders

* CMS_editors_of_X_product
* Developers of Y teams

### Assumptions
* composer va:test:accessibility adequately tests the accessibility of the CMS.

(How will these assumptions be validated?)

## Acceptance Criteria
* composer va:test:accessibility  passes.

## Possible tickets to create for this epic

- [ ] Spike task, with some of the following attributes:
* Identify the baseline against which we will measure improvements
* Ideation (sketching, cross-displinary meeting of minds around how the problem may be solved)
* Analysis of the drupal landscape
* Possible implementation(s)
* Proof of concept or prototype
* UX writing discovery

## Labels for  this issue
- [/] Issue type (red) (defaults to Epic)
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [ ] CMS objective (orange)
- [ ] CMS-supported product (black)

PR #1980, opened 2020-06-05
VACMS-1872: Render child sections.

oksana-c:
## Description

See #1872 . 

### Access to All Sections
Screen Shot 2020-06-05 at 7 49 03 AM

### Access to some sections

Screen Shot 2020-06-05 at 7 53 17 AM

## QA steps

1. Do this
1. Then that
- [ ] This
- [ ] That
- [ ] The other thing

## Definition of Done
- [ ] Product release notes 


swirtSJW:
508 Issues:  Needs to use aria to reflect its state.  Also needs to work on keyboard navigation similar to 
https://designsystem.digital.gov/components/accordion/
You could look at node_link_report to see how it is handled.
See this page to test keyboard  and aria handling
http://pr1980.ci.cms.va.gov/locations/franklin-r-sousley-campus

swirtSJW:
The other change I noticed is that I no longer have sections.
image

oksana-c:
@todo: add user path to accessibility testiny

oksana-c:
alrighty! switched this baby to use accordion pattern from VADS and added other style adjustments. It is now 100% compliant and we won't need another iteration just to cover 508.

Users with "bypass workbench access" permission will see all sections.

Screen Shot 2020-06-05 at 2 26 48 PM

also, added /user path to accessibility check. Let's get ready for some warnings 😄 :octocat:

oksana-c:
@kevwalsh ^^ fyi

oksana-c:
Another update in this PR that is worth noting - **\Drupal\va_gov_user\Service\UserPermsService::getSections** method was refactored to return an array of all sections that user has access to, keyed by TIDs. 

This update removes tight coupling to Workbench Sections key `administration`, which (if returned by getSections, signified that the user has access to _all sections_.  Removing the dependency on this key will allows us to rely on specific tids and prevent us from adding any proprietary logic (tech-debt) which function would rely on workbench-specific definitions.

---

/user path that I added to accessibility tests in a previous commit shows 1 contrast error. Detected contrast error is coming from the administrator toolbar and is unrelated to the "sections" block added in this PR. We have #1302 open to fix accessibility tests that are global to the application.

swirtSJW:
@oksana-c I like the accessibility improvements and the switch to full accordions.   
One minor concern is the pattern of putting a  link in the button. area.

It is not accounted for on either 
https://designsystem.digital.gov/components/accordion/
or 
https://design.va.gov/components/accordions

In this case it creates some additional 508 issues.
1.  On tabbing the focus seems to go from right to left as it jumps from the plus sign to the link texts. 
 This is counter to movement of the eye and reading direction.
2.  Since there is no text in the button itself, the screen reader just reads "Toggle",   and then reads the section link, but by that time the reader  is on the section link, so by the time they are informed what the toggle does, they are on the section link which takes them to the section.
I think you have to  put the heading in the button, without the link, then repeat the list item within the link inside the accordion content.
3.  The shading behavior on hover.   There are two hover zones that carry out actions.  The link text and the plus, but there is also a hover zone in the middle that responds as though it should do something, but does nothing.  Addressing issue number 2 would also resolve this issue.
image

If we want to get this out, as-is, and target this issue with a follow-up, I am okay with that as long as we commit to fixing it.    @kevwalsh Any thoughts on timing?

oksana-c:
after thorough deliberation, decision was made to use "menu" pattern for Sections. Screenshots for the most common scenarios of section assignment are attached below

Screen Shot 2020-06-08 at 12 04 36 PM

Screen Shot 2020-06-08 at 12 06 24 PM

Screen Shot 2020-06-08 at 12 06 37 PM

oksana-c:
@swirtSJW ,  Give this another whirl once devshop env builds.

@kevwalsh is onboard w/ the latest "menu" pattern and the display of child sections as currently implemented (i.e. not nested). Our goal is to reduce the number of clicks the user takes to get to the content of any section. By allowing all child sections to be visible without extra collapsible areas, we are moving closer to that goal.

Please test keyboard navigation between sections while paying special attention to the order in which parent section link and collapse button are tabbed. Press "return" while focused on the collapse button and tab through child sections.

P.S. I had to switch parent link render array to markup element instead o f a link. I am not fond of that solution, but there's no other way that I know of to achieve the order we need w/ render arrays. We aim for perfect, knowing it unattainable...

oksana-c:
closing the loop here:
> There is a bit of mismatch between hover target and link target. but it is a minor issue.

current implementation follows core menu pattern styles for link targets. We are releasing as-is with an understanding that further iterations may follow with a potential pattern change.

Issue #1600, opened 2020-04-23
Disable workflow participants module


kevwalsh:
## Description
Remove the Workflow Participants module and related custom code.

## Acceptance Criteria
- [ ] Workflow participants module is disabled.

## Implementation Steps
- [ ] Disable `workflow_participants` and `va_gov_notifications`. PR: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/1987
- [ ] Remove `workflow_participants` from composer and `va_gov_notifications` from code base

indytechcook:
PR for disabling the `workflow_participants` module and removing the related views.  The `va_gov_notificaitons` module was already disabled.

https://github.com/department-of-veterans-affairs/va.gov-cms/pull/1987

swirtSJW:
1987 has been merged.  It is ok to proceed with a PR to remove the module from composer.

indytechcook:
Send PR is in: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/2000

PR #1987, opened 2020-06-08
VACMS-1600 disable workflow participants

indytechcook:
## Description

See #1600.

## Testing done

* Edit a currently published node (http://va-gov-cms.lndo.site/node/3004/edit)
  * Verify the Participants tab is no longer visible.
* Change the state of the node to draft.
* Take the new draft through the workflow.  Errors should not occur.
* Look in watchdog for any new errors/notices.

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/1987)


indytechcook:
> Approved, just waiting on tests to pass.

This seems to be taking while......

PR #1988, opened 2020-06-08
VACMS-9167: Upgrade facility endpoint from v0 to v1.

ethanteague:
## Description
See https://github.com/department-of-veterans-affairs/va.gov-team/issues/9167
Upgrades facilities api endpoint from v0 to v1. This should just be a change in the endpoint path - the data structure that is changing isn't associated with any of the data we are consuming.

## Screenshots
N/A


ethanteague:
> @ethanteague can you confirm what migrations you ran to test this?

None, @swirtSJW - I couldn't connect to the api endpoint with the dev creds you provided (received 400's). I pushed this pr to see if I could get @oksana-c 's test to pass, and try to run migrations via the ui in devshop.

PR #1998, opened 2020-06-09
VACMS-1600 disable workflow participants

indytechcook:
## Description

See #1600.

## Testing done

* Edit a currently published node (http://va-gov-cms.lndo.site/node/3004/edit)
  * Verify the Participants tab is no longer visible.
* Change the state of the node to draft.
* Take the new draft through the workflow.  Errors should not occur.
* Look in watchdog for any new errors/notices.

This PR replaced #1987 due to issues with devshop.

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/1998)


tests/behat/drupal-spec-tool/views.feature
```diff
@@ -29,7 +29,6 @@ Feature: Views
 | Media library | media_library | Media | Enabled | Find and manage media. |
 | Moderated content | moderated_content | Content revisions | Enabled | Find and moderate content. |
 | Moderation history | moderation_history | Content revisions | Enabled |  |
-| My Workflow  | my_workflow | Content | Enabled | Content a user has access to that is ready for moderation  |
```
swirtSJW, line 32: @indytechcook These need to be removed from the actual SpecTool spreadsheet as that is the source of truth for these tests.   I just removed them as there is an issue currently logjammed with Beth's forms related commits.

PR #2000, opened 2020-06-09
VACMS-1600: Remove workflow participants from composer.

indytechcook:
## Description

See #1600. 

Follow up of #1998 to remove the workflow participants from composer.  DO NOT MERGE UNTIL AFTER DEPLOY


composer.lock
```diff
@@ -4809,62 +4818,6 @@
             ],
             "time": "2017-12-06T20:35:40+00:00"
         },
-        {
-            "name": "drupal/dynamic_entity_reference",
```
swirtSJW, line 4813: This removal strikes me as odd.

indytechcook: It does look odd, I just ran `lando composer remove drupal/workflow_participants`

Issue #2036, opened 2020-06-15
Javascript failing in IE


swirtSJW:
**Describe the defect**
There is a javascript error happening in IE that is likely breaking a lot of other javascript items in various places.  

Examples:
1. Collapsible sections on user page are not collapsible.
2. Date widget on event date field does not appear.
3 Expandable menus in main navigation does not expand.

On first glance it appears that the error is coming from the google tag manager related code.

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information if relevant, or delete):**
 - OS: Windows
 - Browser IE11.3630

<img width="597" alt="R03-V11-Server_Specialty_Desktop_1" src="https://user-images.githubusercontent.com/643678/84687814-4debf880-af0c-11ea-9581-a1aa118ebb0f.png">
<img width="607" alt="R03-V11-Server_Specialty_Desktop" src="https://user-images.githubusercontent.com/643678/84687818-4e848f00-af0c-11ea-8402-7173a4b3f375.png">
<img width="720" alt="R03-V11-Server_Specialty_Desktop_2" src="https://user-images.githubusercontent.com/643678/84687819-4f1d2580-af0c-11ea-9435-f3f3e489e004.png">

PR #2039, opened 2020-06-15
VACMS-2036: Change arrow functions to old school for ie11.

ethanteague:
## Description

See #2036 

## Testing done
Console

Using our crossbrowser testing tools account, with windows 10 and ie11 (which uses a real windows machine), I'm not able to replicate this as user ethan.teague, or user kevin.walsh on user page (no console errors, js works fine). I do see a console error on the node/add/event page related to ie11 not liking arrow functions, but js is working fine on said page (not being blocked by gtm script). I also see errors being thrown b/c drupal core uses forEach method that ie11 doesn't understand, so this could also potentially cause issues. I don't have cag access, so I'm doing this in the blind - swapping out arrow functions in gtm script with old school js functions.


swirtSJW:
Thanks for this Ethan.  Can you confirm that verified that Tag manager actions are still working?

ethanteague:
> Thanks for this Ethan. Can you confirm that verified that Tag manager actions are still working?

I didn't check the gtm dash from my local @swirtSJW , but the push event / payload was sending & properly formed. Once this is merged in, we should get with Jon W. about confirming that events are still registering properly (low risk - the push / and payload are still the same, so events should still be registering as before).

Issue #1645, opened 2020-04-29
Warning: Creating default object from empty value in Drupal\menu_link_content\Plugin\Menu\MenuLinkContent->updateLink()


oksana-c:
## Description

Warning: Creating default object from empty value in Drupal\menu_link_content\Plugin\Menu\MenuLinkContent->updateLink() (line 237 of /var/www/cms/docroot/core/modules/menu_link_content/src/Plugin/Menu/MenuLinkContent.php)

PHP warning is fixed by applying patch `#2` in https://www.drupal.org/project/drupal/issues/2775665 but the error is originating from docroot/modules/custom/disable_node_menu_item/disable_node_menu_item.module line 83 and we should investigate the reason.

<img width="1531" alt="Screen Shot 2020-04-21 at 3 39 45 PM" src="https://user-images.githubusercontent.com/16477724/79917416-b1afe480-83e7-11ea-9c9b-d13096678dbf.png">

## Acceptance Criteria
* the cause of the warning is identified and addressed. Warning not logged in dblog

## Labels
- [/] Issue type (red) (defaults to task)
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [ ] CMS objective (orange)
- [ ] CMS-supported product (black)

PR #2104, opened 2020-06-22
VACMS-1645: Fix disable_menu_link warning.

oksana-c:
## Description

See #1645 . 

This PR fixes PHP warning specified in #1645 
Original assumption was that this warning may have contributed to DB locks.
Unfortunately, this fix will not have any significant effects on reducing DB locks, since database writes have only been reduced on account of less messages written to DBlog.

## QA steps

As Admin user:
- [x] edit node 725 - `/node/725/edit` 
- [x] update menu link title in "MENU SETTINGS" fieldset
- [x] Save the node using "Save" button (not "Save and continue")
- [x] review menu links `/admin/structure/menu/manage/careers-employment-benefits` and confirm your title change took effect
- [x] review log `/admin/reports/dblog` and confirm that it does not contain warning like this `Warning: Creating default object from empty value in Drupal\menu_link_content\Plugin\Menu\MenuLinkContent->updateLink() (line 237 of /var/www/cms/docroot/core/modules/menu_link_content/src/Plugin/Menu/MenuLinkContent.php)...`
- [x] edit node 725 - `/node/725/edit` and uncheck "Enable in menu" checkbox in "MENU SETTINGS" fieldset. Save the node and verify that menu link is now disabled in `/admin/structure/menu/manage/careers-employment-benefits` and log `/admin/reports/dblog` does not contain warnings

### DIFF
--- a/docroot/modules/custom/disable_node_menu_item/disable_node_menu_item.module
+++ b/docroot/modules/custom/disable_node_menu_item/disable_node_menu_item.module
@@ -78,7 +78,6 @@ function disable_node_menu_item_submit(array $form, FormStateInterface $form_sta
     $menu_links = $menu_link_manager->loadLinksByRoute('entity.node.canonical', ['node' => $node->id()]);
     if (count($menu_links)) {
       foreach ($menu_links as $menu_link) {
-        $definition = $menu_link->getPluginDefinition();
         $definition['enabled'] = $form_state->getValue(['menu', 'link_enabled']);
         $menu_link_manager->updateDefinition($menu_link->getPluginId(), $definition);
       }


docroot/modules/custom/disable_node_menu_item/disable_node_menu_item.module
```diff
@@ -78,7 +78,6 @@ function disable_node_menu_item_submit(array $form, FormStateInterface $form_sta
     $menu_links = $menu_link_manager->loadLinksByRoute('entity.node.canonical', ['node' => $node->id()]);
     if (count($menu_links)) {
       foreach ($menu_links as $menu_link) {
-        $definition = $menu_link->getPluginDefinition();
```
oksana-c, line 81: we do not need to override all properties in plugin definition within `disable_node_menu_item` scope

by leaving only `enabled` property for override, we're eliminating warnings listed in #1645

swirtSJW: Nice discovery.

PR #2124, opened 2020-06-25
VACMS-9237 facility service hardening - limited approach 06/25

ethanteague:
## Description
This is an incremental step towards "hardening" the facility service content type. At present, there is one big wysiwyg blob where users enter facility appointment, contact, and location service data. 
Our end goal is to replace this blob with targeted fields. To that end, this pr begins this replacement process by adding three appointment field select fields to the facility service content type:
- Referral required?
- Walk-ins accepted?
- Online scheduling available?

And a new phone number paragraph type, attached to the `field_phone_numbers_paragraph` field. This new phone paragraph allows for users to choose the # type (e.g. : phone number, fax, sms, tel), and uses states management to hide an extension field if the phone number type is not selected. Furthermore, there is validation logic in place that checks to see if the number entered in the phone/fax/sms/tel field has a format that matches one of the four patterns expected when allowing for the formats this field takes:
- 123-456-7890
- 123
- 12345
- 123456

For now, there isn't enough data collection provided by this incremental step to allow for these new fields/paragraph fields to be used to replace the existing wysiwyg blob, so front end build templates and graphql have not been altered to expect anything other than the current wysiwyg field. Per facilities-cms teams meeting on 06/24 and further discussion with @mmiddaugh, we will not be using any of our new location / appt / phone number content model fields for front end build until we have a complete replacement for the data being collected in the current wysiwyg field (although these fields can be used for content entry for output in the future). Once we have a complete replacement, we will then add a toggle for users to decide if they would like to use the new content model fields for output, or the existing wysiwyg field - this selection will then be applied to the front end build.

Going forward, we will need to leverage @rachel-kauff 's content smithing skills for wrangling the original poc pr: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/1844 for each incremental addition to the wysiwyg replacement.

## Testing done
Visual, behat

## Screenshots
* The phone number drag and drop verbiage will need a separate task to mirror https://app.mural.co/t/vagov6717/m/vagov6717/1591734393833/db39cbc183722673ab91b46e5365508271b6eae8 (per Rachel, it's okay to punt on this)
image
image

## QA steps
- As an administrator, visit `/node/4758/edit` 
- [x] Visually verify that new appointments fieldset with three new select lists appears.
image

- [x] Visually verify that new phone number paragraph appears
- [x] Toggle off phone to any other selection, and verify that extension field disappears
- [-] ~~Validate the revision log field is in the workflow section on the bottom of the edit form.~~
  It is not there, but in looking at this content type beforehand, there is no workflow section on this content type, so it is beyond the scope of this ticket to get it added.  
- [x] With phone option selected, enter 941-9151811, and verify that when saving node, field error occurs
- [x] validate that the link in there error takes you to the field.

- [x] With phone option selected, enter 941-915-1811, and verify that when saving node, no field error occurs.
- [x] Validate that the view makes sense. 

image


swirtSJW:
@ethanteague I am wondering about this:
image

Is there are reason that the type is required, but the number is not.  Seems like having one without the other is problematic.

config/sync/core.entity_form_display.node.health_care_local_health_service.default.yml
```diff
@@ -53,18 +74,54 @@ content:
...
+    settings: {  }
+    third_party_settings: {  }
+    type: options_select
+    region: content
 hidden:
   created: true
   moderation_state: true
   path: true
   promote: true
-  revision_log: true
```
swirtSJW, line 67: I added a QA step to make sure the revision log is where it is supposed to be.  I've noticed on other content exports that the revision log field goes missing or magically moves to the upper right sidebar.

config/sync/core.entity_form_display.node.health_care_local_health_service.inline_entity_form.yml
```diff
@@ -55,11 +59,14 @@ content:
...
   created: true
+  field_online_scheduling_availabl: true
+  field_phone_numbers_paragraph: true
+  field_referral_required: true
   field_regional_health_service: true
+  field_walk_ins_accepted: true
   moderation_state: true
   path: true
   promote: true
-  revision_log: true
```
swirtSJW, line 62: also odd will have to check it passes QA step

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/ValidNumberValidator.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+class ValidNumberValidator extends ConstraintValidator {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validate($items, Constraint $constraint) {
+    foreach ($items as $item) {
+      // Check for phone number, then short codes, then tty. No match = proceed.
+      if (!preg_match("/^\d{3}-\d{3}-\d{4}$/", $item->value) && !preg_match("/^\d{3}$/", $item->value) && !preg_match("/^\d{5}$/", $item->value) && !preg_match("/^\d{6}$/", $item->value)) {
```
swirtSJW, line 19: It would be nice to have each of these pregmatches on its own line with a comment indicating what it is checking for. Then we could pass in some helpful response based on what it failed for  "Invalid tty form... should look like ___"

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -287,19 +292,44 @@ function va_gov_backend_field_widget_paragraphs_form_alter(&$element, FormStateI
       ];
     }
   }
+
```
swirtSJW, line 295: Need some nice helpful comments here to indicate what the intent is.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -287,19 +292,44 @@ function va_gov_backend_field_widget_paragraphs_form_alter(&$element, FormStateI
...
+  if (in_array($paragraph_entity_reference_field_name, ['field_phone_numbers_paragraph', 'field_appointment_phone_numbers'])) {
+    $widget_state = WidgetBase::getWidgetState($element['#field_parents'], $paragraph_entity_reference_field_name, $form_state);
+
+    $paragraph_instance = $widget_state['paragraphs'][$element['#delta']]['entity'];
+    $paragraph_type = $paragraph_instance->bundle();
+
+    if ($paragraph_type === 'phone_number') {
+      $selector = sprintf(':input[name="field_phone_numbers_paragraph[%d][subform][field_phone_number_type]"]', $element['#delta']);
```
swirtSJW, line 303: Comment describing what is being done and why..

docroot/modules/custom/va_gov_backend/va_gov_backend.module
swirtSJW: It would also be nice to have this all in its own nicely named function to keep hooks field_widget_paragraphs_form_alter() from becoming tangled up with all kinds of disconnected intents.  Then just call the nicely named function in the hook.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -287,19 +292,44 @@ function va_gov_backend_field_widget_paragraphs_form_alter(&$element, FormStateI
...
   if ($entity_form['#bundle'] === 'health_care_local_facility') {
     $entity_form['title']['widget'][0]['value']['#attributes']['disabled'] = 'disabled';
   }
 }
 
+/**
+ * Implements hook_entity_bundle_field_info_alter().
+ */
+function va_gov_backend_entity_bundle_field_info_alter(&$fields, EntityTypeInterface $entity_type, $bundle) {
+  if ($entity_type->id() === 'paragraph' && $bundle === 'phone_number' && isset($fields['field_phone_number'])) {
```
swirtSJW, line 328: Add a comment describing what and why.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/ValidNumber.php
```diff
@@ -0,0 +1,25 @@
...
+namespace Drupal\va_gov_backend\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Checks that the submitted value is a valid number.
+ *
+ * @Constraint(
+ *   id = "ValidNumber",
+ *   label = @Translation("Valid Number", context = "Validation"),
```
oksana-c, line 12:
```suggestion
 *   id = "ValidPhoneNumber",
 *   label = @Translation("Valid Phone Number", context = "Validation"),
```

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/ValidNumberValidator.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Drupal\va_gov_backend\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+/**
+ * Validates the ValidNumber constraint.
+ */
+class ValidNumberValidator extends ConstraintValidator {
```
oksana-c, line 11:
```suggestion
 * Validates the ValidPhoneNumber constraint.
 */
class ValidPhoneNumberValidator extends ConstraintValidator {
```

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -287,19 +292,44 @@ function va_gov_backend_field_widget_paragraphs_form_alter(&$element, FormStateI
...
+    $fields['field_phone_number']->addConstraint('ValidNumber');
```
oksana-c, line 329:
```suggestion
    $fields['field_phone_number']->addConstraint('ValidPhoneNumber');
```

config/sync/core.entity_form_display.node.health_care_local_health_service.default.yml
ethanteague: Thank you, sir. :+1:

config/sync/core.entity_form_display.node.health_care_local_health_service.inline_entity_form.yml
ethanteague: Thank you, sir. :+1:

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/ValidNumberValidator.php
ethanteague: Great idea, @swirtSJW :+1:

ethanteague:
> good morning @ethanteague
> please see added inline suggestions to rename `ValidNumber` constraint into `ValidPhoneNumber` to make it clear that it's not just any ol' number constr

> @ethanteague I am wondering about this:
> image
> 
> Is there are reason that the type is required, but the number is not. Seems like having one without the other is problematic.

Yeah.......I'll make that phone field required, @swirtSJW :+1:

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/ValidNumber.php
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/ValidNumberValidator.php
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: :+1:

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: :+1:

config/sync/field.storage.node.field_online_scheduling_availabl.yml
```diff
@@ -0,0 +1,27 @@
...
+    - node
+    - options
+id: node.field_online_scheduling_availabl
+field_name: field_online_scheduling_availabl
+entity_type: node
+type: list_string
+settings:
+  allowed_values:
+    -
+      value: 'yes'
```
swirtSJW, line 15: Sorry, I didn't spot this before. It would be nice if the value was a bool for logic and for when if this gets exposed to Lighthouse. The label could be Yes No but the values should be 1 or 0.

config/sync/field.storage.node.field_walk_ins_accepted.yml
```diff
@@ -0,0 +1,27 @@
...
+    - node
+    - options
+id: node.field_walk_ins_accepted
+field_name: field_walk_ins_accepted
+entity_type: node
+type: list_string
+settings:
+  allowed_values:
+    -
+      value: 'yes'
```
swirtSJW, line 15: binary values please

config/sync/field.storage.paragraph.field_phone_number_type.yml
```diff
@@ -0,0 +1,33 @@
...
+    - options
+    - paragraphs
+id: paragraph.field_phone_number_type
+field_name: field_phone_number_type
+entity_type: paragraph
+type: list_string
+settings:
+  allowed_values:
+    -
+      value: phone
```
swirtSJW, line 15: This value should be `tel` so that on the front end, nothing needs to be done to make tel:####    the same way it would make fax:### or sms:##3   The value could then be used directly as protocol and the Label would be what gets made visible to the display.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -287,19 +292,70 @@ function va_gov_backend_field_widget_paragraphs_form_alter(&$element, FormStateI
       ];
     }
   }
+
+  if (in_array($paragraph_entity_reference_field_name, ['field_phone_numbers_paragraph', 'field_appointment_phone_numbers'])) {
+    _va_gov_backend_phone_paragraph_states($element, $form_state, $context);
```
swirtSJW, line 297: Minor: good function names have a verb to bring clarity to what they **do**.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -287,19 +292,70 @@ function va_gov_backend_field_widget_paragraphs_form_alter(&$element, FormStateI
...
   }
 }
 
+/**
+ * Implements hook_entity_bundle_field_info_alter().
+ */
+function va_gov_backend_entity_bundle_field_info_alter(&$fields, EntityTypeInterface $entity_type, $bundle) {
+  // Add phone number validation to our phone number field.
+  // Preg matches check for length and format.
+  // If not TTY, Shortcode, or 10 digit dashed format = error thrown.
```
swirtSJW, line 353: Minor:  Too much detail here.   That is reaching beyond its concern and describing how the validation function works... which is defined in the validation function.  Documentation is good but if it is too far removed from what is actually doing the work, then it can easily become outdated and misleading.
```suggestion

```

config/sync/field.storage.node.field_walk_ins_accepted.yml
ethanteague: :+1:

config/sync/field.storage.node.field_online_scheduling_availabl.yml
ethanteague: :+1:

config/sync/field.storage.node.field_referral_required.yml
```diff
@@ -0,0 +1,27 @@
...
+    - node
+    - options
+id: node.field_referral_required
+field_name: field_referral_required
+entity_type: node
+type: list_string
+settings:
+  allowed_values:
+    -
+      value: 'yes'
```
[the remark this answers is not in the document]
ethanteague: :+1:

swirtSJW:
@ethanteague I found an error that has a 508 implication and an editorial issue.
When a phone field does not validate you get an error message with a link that is supposed to jump to the  field with the error.
The link points to 
`#edit-field-phone-numbers-paragraph-0-subform-field-phone-number`
Unfortunately nothing with that id exists.  Only these exist.
`edit-field-phone-numbers-paragraph-0-subform-field-phone-number-0-value`
`edit-field-phone-numbers-paragraph-0-subform-field-phone-number-0-value--description`
`edit-field-phone-numbers-paragraph-0-subform-field-phone-number-wrapper`
`edit-field-phone-numbers-paragraph-0-subform-field-phone-number-type-wrapper`

So the link does not work.

swirtSJW:
Not to goldplate this puppy, but can we do a bit of layout adjustment here?
image

It seems like a bad UX to have the extension output ahead of the number.   It would be nice to tighten up this display to it is a bit more compact and grouping so that when you have more than one, you can tell where one starts and one ends
image

ethanteague:
> Two issues that I see are noted in other comments:
> 
> 1. Jump link to error does not work.  (UX and 508 issue)
> 2. Preview output is jumbled, long and confusion when there are multiples.

@swirtSJW :+1:

config/sync/core.entity_form_display.node.landing_page.default.yml
```diff
@@ -118,7 +118,7 @@ third_party_settings:
       weight: 3
       format_type: fieldset
       format_settings:
-        description: 'Crisis alerts draw attention to information that will be useful to veterans experiencing an active crisis. You can <a href="/block/add/alert?destination=/admin/content/alerts">create a new crisis alert</a>, or reuse an existing one.'
+        description: 'Crisis alerts draw attention to information that will be useful to veterans experiencing an active crisis. You can <a href="http://vagovcms.lndo.site/block/add/alert?destination=/admin/content/alerts">create a new crisis alert</a>, or reuse an existing one.'
```
swirtSJW, line 121: This looks like an out of date db forcing itself back into yml

config/sync/core.entity_form_display.node.va_form.default.yml
```diff
@@ -3,9 +3,7 @@ langcode: en
 status: true
 dependencies:
   config:
-    - entity_browser.browser.alert_blocks
```
swirtSJW, line 6: Looks like this is unrelated changes due to out of sync code or db

Issue #2127, opened 2020-06-26
Generate schema for transformer team


kevwalsh:
## User Story or Problem Statement

As a writer of CMS export transformers, i need schema information. 

See threads in #cms-transformer channel, particularly [this one](https://dsva.slack.com/archives/C01512KN35G/p1592981545115700).

## Acceptance Criteria
- [ ] Transformer team has a way of getting dynamic schema information about all CMS entities

## Approach options
1. GraphQL. 

- Whenever Content export is tarred, it includes the GraphQL output about schema in a file in the export (eg put it in sites/default/files/export-content/)  
- This would allow other consumers (eg Lighthouse) to access.

2. JSON API

- not currently "supported" by our team, despite being native

## Related

https://github.com/department-of-veterans-affairs/va.gov-cms/issues/1851

swirtSJW:
I have got a working model using the schema provided by OpenAPI.  I deliverd the output to the FE transformer team for validation and am not working on making it so they can get it.   Then can work on getting the file saved.

PR #2173, opened 2020-07-02
VACMS-2127 generate entity schemas

swirtSJW:
## Description

See #2127 . 

AC:
- [x] A dynamic request for the content schema can be achieve with api_consumer role.
- [x] An exact copy of the content schema is available in the CMS export tar.

## QA steps
As an unauthenticated user
1. Visit `/openapi/jsonapi?_format=json`
   - [ ] Validate that a basic auth window appears
image
   - [ ] validate that u:api p:drupal8  allows you to get to the page (will not work in lando)
   - [ ] Validate page with a ton of json appears 
image

2. visit /cms-export/content
   - [ ] Validate that you are prompted with a download of 
image

3. Save the tar and then open it.  
image
    - [ ] 1 Validate that inside the tar is a directory path of `cms-export-content/meta/`
    - [ ] 2 Validate a file `schema.json`
    - [ ] 3. Validate the date and time represent when you first downloaded this file.
    - [ ] 4 Validate that the file is approx 4MB
4. Open the `schema.json` file. 
    - [ ] Validate that it contains a ton of json  (may make your editor unhappy).

### DIFF
--- a/docroot/modules/custom/va_gov_content_export/src/Archive/ArchiveDirectory.php
+++ b/docroot/modules/custom/va_gov_content_export/src/Archive/ArchiveDirectory.php
@@ -5,6 +5,7 @@ namespace Drupal\va_gov_content_export\Archive;
 use Alchemy\Zippy\Archive\ArchiveInterface;
 use Alchemy\Zippy\Zippy;
 use Drupal\Core\File\FileSystemInterface;
+use Drupal\va_gov_content_export\Event\ContentExportPreTarEvent;
 
 /**
  * Archive a file path for CMS Content Export.
@@ -62,6 +63,9 @@ class ArchiveDirectory {
 
     // Deleting the file before it's created improves performance.
     $this->fileSystem->delete($archiveArgs->getOutputPath());
+    // Dispatch the Pre Tar event so that subscribers can use it.
+    \Drupal::service('event_dispatcher')->dispatch(ContentExportPreTarEvent::CONTENT_EXPORT_PRE_TAR_EVENT, new ContentExportPreTarEvent("{$archiveArgs->getCurrentWorkingDirectory()}{$archiveArgs->getArchiveDirectory()}", $this->fileSystem));
+
     return $this->zippy->create($real_path, $files, TRUE);
   }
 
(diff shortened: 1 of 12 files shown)


docroot/modules/custom/va_gov_content_export/src/Archive/ArchiveDirectory.php
```diff
@@ -62,6 +63,9 @@ public function archive(ArchiveArgs $archiveArgs) : ArchiveInterface {
 
     // Deleting the file before it's created improves performance.
     $this->fileSystem->delete($archiveArgs->getOutputPath());
+    // Dispatch the Pre Tar event so that subscribers can use it.
+    \Drupal::service('event_dispatcher')->dispatch(ContentExportPreTarEvent::CONTENT_EXPORT_PRE_TAR_EVENT, new ContentExportPreTarEvent("{$archiveArgs->getCurrentWorkingDirectory()}{$archiveArgs->getArchiveDirectory()}", $this->fileSystem));
```
indytechcook, line 67: Should this be injected?  Or is this the drupal best practice?

indytechcook:
Just one question, otherwise LGTM

docroot/modules/custom/va_gov_content_export/src/Archive/ArchiveDirectory.php
swirtSJW: Not taboo... but not best practice  https://www.drupal.org/docs/drupal-apis/services-and-dependency-injection/services-and-dependency-injection-in-drupal-8#s-using-global-functions.
Good call I can inject it into the ArchiveDirectory service.

swirtSJW: @indytechcook I just pushed up a change to use inject.

docroot/modules/custom/va_gov_content_export/va_gov_content_export.services.yml
```diff
@@ -16,6 +16,10 @@ services:
     decorates: tome_sync.file_sync
   va_gov.content_export.archive_directory:
     class: Drupal\va_gov_content_export\Archive\ArchiveDirectory
-    arguments: ['@zippy', '@file_system', '@logger.channel.file']
```
swirtSJW, line 19: Removed the logger because it was not being used or even caught anywhere.

PR #2242, opened 2020-07-13
VACMS-000: Update gtm to use env-2 instead of env-5.

ethanteague:
## Description
Per @jonwehausen, gtm needs to be pointed to `env-2` (currently points to `env-5`). This pr replaces 2 instances of `env-5` with env-2

## Testing done
Codebase grep

## Screenshots
N/A

## QA steps
N/A


oksana-c:
linking original support request for reference https://github.com/department-of-veterans-affairs/va.gov-team/issues/11163

Issue #2243, opened 2020-07-13
PHPCS does not scan files with `.module` extension


oksana-c:
## Description 

phpcs does not scan `.module` files for coding standard errors.

Fixing this would entail fixing a bunch of errors coming from custom code that we have in codebase today.

## To reproduce

1. Open the following file in IDE - `docroot/modules/custom/disable_node_menu_item/disable_node_menu_item.module`
1. Remove space before the parenthesis in one of the `if` statements or make any other edit that would trigger phpcs warning/error
1. Run `lando composer va:test:cs` or `lando ssh` then `composer va:test:cs`

**Current behavior**
_phpcs is not identifying missing space as an error_

**Expected behavior**
_phpcs throws an error_

ElijahLynn:
This is working now: I followed the test case and it is failing as expected. 
![image](https://user-images.githubusercontent.com/1504756/94627560-c6d04f80-0272-11eb-8690-f55da133f513.png)

PR #2244, opened 2020-07-13
VACMS-2243: Add extensions that should be scanned by phpcs.

oksana-c:
## Description

See #2243. 

## Testing done

- [ ] Review phpcs test that ran for the first commit in this PR and confirm that `.module` files are now tested by code sniffer and are throwing errors + breaking the build.
- [ ] review phpcs tests for the second commit in this PR which addresses all previous phpcs errors and confirm all tests are now passing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -377,7 +377,7 @@
             "find docroot/modules/custom docroot/themes -name '*.install' -print0 | xargs -0 -n1 php -l > /dev/null"
         ],
         "va:test:cs": [
-            "phpcs --ignore=*.md,*.min.css,styles.css,wysiwyg.css,*/node_modules/*,*/simplesaml*/* --standard=./docroot/vendor/drupal/coder/coder_sniffer/Drupal/ruleset.xml docroot/modules/custom docroot/themes/custom --colors"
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -377,7 +377,7 @@
             "find docroot/modules/custom docroot/themes -name '*.install' -print0 | xargs -0 -n1 php -l > /dev/null"
         ],
         "va:test:cs": [
-            "phpcs --ignore=*.md,*.min.css,styles.css,wysiwyg.css,*/node_modules/*,*/simplesaml*/* --standard=./docroot/vendor/drupal/coder/coder_sniffer/Drupal/ruleset.xml docroot/modules/custom docroot/themes/custom --colors"
+            "phpcs --ignore=*.md,*.min.css,styles.css,wysiwyg.css,*/node_modules/*,*/simplesaml*/* --extensions=php,module,inc,install,profile,engine,theme,js,css --standard=./docroot/vendor/drupal/coder/coder_sniffer/Drupal/ruleset.xml docroot/modules/custom docroot/themes/custom --colors"
```
oksana-c, line 380: these extension should be tested by default, but for some reason aren't. If there is a better method of enabling tests for all these extensions, then I'd love to remove the `--extensions` param from this command.

NOTE: php files are tested fine without this modification, while `.module` files are not.

oksana-c:
NOTE: phpcs fixes that were missed in `.module` files are WIP.

swirtSJW:
Nice catch!

Issue #2064, opened 2020-06-17
"Save and continue editing" button data loss for new paragraphs


oksana-c:
**Describe the defect**

"Save and continue editing" button is not saving new paragraphs on "VAMC facility health services" content type and potentially others. We may want to disable "Save and continue editing" on all content types until we stabilize the behavior across the CMS.

**To Reproduce**
1. Create "VAMC health care facility" node
2. Add a paragraph
3. click on "Save and continue editing"
4. Paragraph content is not saved for the node

**Expected behavior**
Any new Paragraph content for "VAMC health care facility" node  is saved when "Save and continue editing" button is used to save the node.

**Research**

- can this be solved in the va_gov_build_trigger_form_alter_submit()  
- google analytics event data (how much use is this gettting?)

**Possible approaches**

1. Fix the button.

or 

2. Kill it and update users. 

**Screenshots**

**Desktop (please complete the following information if relevant, or delete):**

**Additional context**

rachel-kauff:
Can whoever works on this ping me when you decide whether to go with 1.) Fix the button 2.)Kill it and update users. so I can write the release note for this? Thanks!

ElijahLynn:
Reminder that the "save and continue" button, I think, is actually a repurposed native "Preview" button that has hook_form_preview_submit handlers or something like that. I don't know this, just re-stating what we talked about in a meeting recently.

ethanteague:
@kevwalsh @oksana-c - not able to reproduce this. I went to a new node form for vamc facility, and facility health service, and added paragraphs to both. When clicking the save and continue editing button on both instances, the paragraph was preserved. Confirmed that this wasn't just a browser form fill thing by going to admin content view and opening the newly created node.
![vamc_facility.png](https://images.zenhubusercontent.com/5e2b1d8b2f4d80671f879681/0c97aacf-aae9-4a6f-82f0-bcc89888464d)
![vamc_health_service.png](https://images.zenhubusercontent.com/5e2b1d8b2f4d80671f879681/386bc48f-b241-49cb-8a08-1502c7349d25)

oksana-c:
@ethanteague , interesting. I was on your POC branch for facility health service content model hardening when I identified this bug. I didn't think it was related to the work on your branch though, because you only added new fields/paragraph types there and no other logic that would influence Save and continue behavior. 

I'll see if I can reproduce this again tomorrow morning and let you know.

oksana-c:
[...]

PR #2245, opened 2020-07-13
VACMS-2064: Adding ief trigger submit to our preview submit.

ethanteague:
## Description
See #2064 
Turns out ief needs to have `#ief_submit_trigger` flag added to submit handler to attach parent entity, otherwise it's thrown away when parent form us saved.

## Testing done
Visual

## Screenshots
image

## QA steps
As and admin, go to `/node/72/edit`, and add an item to `Support Services` in right rail.
Click `Save and continue editing` and visually verify that item added is still available.

### DIFF
--- a/docroot/modules/custom/va_gov_build_trigger/va_gov_build_trigger.module
+++ b/docroot/modules/custom/va_gov_build_trigger/va_gov_build_trigger.module
@@ -61,7 +61,7 @@ function va_gov_build_trigger_form_alter(&$form, FormStateInterface $form_state,
  * @param \Drupal\Core\Form\FormStateInterface $form_state
  *   The form state array.
  */
-function va_gov_build_trigger_form_alter_submit(array &$form, FormStateInterface &$form_state) {
+function va_gov_build_trigger_form_alter_submit(array &$form, FormStateInterface $form_state) {
 
   // We want to associate our user with the revision.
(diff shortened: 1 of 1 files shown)


docroot/modules/custom/va_gov_build_trigger/va_gov_build_trigger.module
```diff
@@ -42,19 +42,26 @@ function va_gov_build_trigger_form_alter(&$form, FormStateInterface $form_state,
...
- * Implements hook_form_alter().
+ * Submit function for va_gov_build_trigger_form_alter.
+ *
+ * @param array $form
+ *   The form array.
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
+ *   The form state array.
  */
-function va_gov_build_trigger_form_alter_submit(&$form, FormStateInterface $form_state) {
+function va_gov_build_trigger_form_alter_submit(array &$form, FormStateInterface &$form_state) {
```
oksana-c, line 64:
```suggestion
function va_gov_build_trigger_form_alter_submit(array &$form, FormStateInterface $form_state) {
```

oksana-c: @ethanteague , was there a specific reason to pass `$form_state` by reference?

Issue #1744, opened 2020-05-12
Hide "Save and continue editing" button on content types that are not under moderation workflow


oksana-c:
## Problem statement

On content types that don't have moderation workflow (e.g. banner alerts), the "Save and continue editing" button has inconsistent behavior, so it should be removed.

"Save and continue editing:

- does not create a new revision (mixed results - still needs to confirmed)
- does not save revision log message
- does not produce a confirmation message on save similar to "VAMC system banner alert with situation updates XYZ has been updated."

## Acceptance criteria

- [ ] "Save and continue editing" button is removed for content types (1) that are not a part of editorial workflow (Only consists of `VAMC system banner alert with situation updates` content type)

Issue #1743, opened 2020-05-12
"Save and continue editing" button UX for content types under moderation workflow


oksana-c:
## User Story

As an editor, I want to clearly understand the purpose of "Save and continue editing" button.

## Problem statement

- "Save and continue editing" button always saves node in Draft state while disregarding
   - editorial workflow state selected by the editor in moderation widget
   - revision message entered by the editor (to be addressed in #1697) 
- "Save and continue editing" button should be renamed to "Save draft and continue editing"

## AC

- [ ] "Save and continue editing" button text is changed to "Save draft and continue editing"
- [ ] Moderation state is set to draft by default when "Save draft and continue editing" is used to save the revision

PR #2284, opened 2020-07-15
VACMS 1743 & 1744 save and continue

ethanteague:
## Description
See #1743 & #1744 

## Testing done
Visual

## Screenshots
image
image
image

## QA steps
- As an admin, go to `node/add/full_width_banner_alert` and visually verify that only `Save` button (in form actions section) is present.
- As an admin, go to `node/add/health_care_region_detail_page` and visually verify that save and continue button now reads `Save draft and continue editing`
- As an admin, go to `/node/5676/edit`
- Add a revision log message, set the state to published, and click the Save and continue editing button
- Go to `/node/5676/revisions` and visually verify that the message you entered appears next to the revision you just created and published state is indicated for the revision.


docroot/modules/custom/va_gov_build_trigger/va_gov_build_trigger.module
```diff
@@ -71,9 +74,8 @@ function va_gov_build_trigger_form_alter_submit(array &$form, FormStateInterface
   $node->setNewRevision(TRUE);
   $node->setRevisionCreationTime(\Drupal::time()->getRequestTime());
 
-  // Save it as unpublished and associate with user.
+  // Save it and associate with user.
   $node->setRevisionUserId($uid);
-  $node->set('moderation_state', 'draft');
```
oksana-c, line 76: @ethanteague , apologies for conflicting ACs in #1743 . I just updated those, but should have checked before you took it on.
Changing the button to 'Save draft and continue editing' assumes that the button will always create a new draft.
This behavior may be somewhat counterintuitive when we consider the moderation state dropdown right above it, although we're planning on removing the ability to change state from the edit form as a part of EWA work. 
For now, let's keep storing the Draft when 'Save draft and continue editing' button is clicked, while disregarding the state selected in mod dropdown.

docroot/modules/custom/va_gov_build_trigger/va_gov_build_trigger.module
```diff
@@ -38,18 +38,21 @@ function va_gov_build_trigger_node_insert(NodeInterface $node) {
  * Implements hook_form_alter().
  */
 function va_gov_build_trigger_form_alter(&$form, FormStateInterface $form_state, $form_id) {
-
+  // Add to node forms, but not to banners.
+  $banner_forms = ['node_full_width_banner_alert_edit_form', 'node_full_width_banner_alert_form'];
```
oksana-c, line 42: @ethanteague what do you think about using logic similar to what's described [here](https://drupal.stackexchange.com/questions/290210/getting-the-workflow-associated-to-a-node-type) for detecting content types not included in any workflows.

I can see us adding unmoderated CTs in the future, so would want to make sure that the logic we add now avoids manually managed edge cases.

docroot/modules/custom/va_gov_build_trigger/va_gov_build_trigger.module
```diff
@@ -38,18 +38,22 @@ function va_gov_build_trigger_node_insert(NodeInterface $node) {
...
       '#ief_submit_trigger' => TRUE,
       '#ief_submit_trigger_all' => TRUE,
       '#type' => 'submit',
       '#weight' => 10,
-      '#value' => 'Save and continue editing',
+      '#value' => 'Save draft and continue editing',
       '#submit' => ['::submitForm', 'va_gov_build_trigger_form_alter_submit'],
     ];
-
+    // Remove the default preview button from types not in workflow.
```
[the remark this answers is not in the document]
ethanteague: Updated, @swirtSJW :+1:

Issue #2399, opened 2020-07-23
Exclude documentation_page from content export. 


swirtSJW:
The content export currently contains files related to documentation_page content type which is not a publicly facing page.  It is meant for internal users.   It should be excluded from the content export.

PR #2400, opened 2020-07-23
VACMS-2399 Exclude documentation_page from content export.

swirtSJW:
## Description

See #2399. 

- [ ] Save a documentation page locally.  Validate that no tome file for that node is present in the export. (delete all the files first then it is easier to see that one was not created.)

- [ ] Run the full export via drush validate that no documentation_page nodes are present.

Post release steps:
- [ ] Run drush content bulk export on prod

### DIFF
--- a/docroot/modules/custom/va_gov_content_export/src/TomeExporter.php
+++ b/docroot/modules/custom/va_gov_content_export/src/TomeExporter.php
@@ -35,7 +35,7 @@ class TomeExporter extends Exporter {
     'content_moderation_state',
     'crop',
     'node.documentation_page',
-    'path_alias';
+    'path_alias',
     'site_alert',
     'user_history',
     'user_role',
--- a/docroot/modules/custom/va_gov_content_export/src/TomeExporter.php
+++ b/docroot/modules/custom/va_gov_content_export/src/TomeExporter.php
@@ -33,7 +33,10 @@ class TomeExporter extends Exporter {
    */
   protected static $excludedTypes = [
     'content_moderation_state',
+    'crop',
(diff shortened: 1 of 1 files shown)


docroot/modules/custom/va_gov_content_export/src/TomeExporter.php
```diff
@@ -33,9 +33,10 @@ class TomeExporter extends Exporter {
    */
   protected static $excludedTypes = [
     'content_moderation_state',
-    'user',
-    'user_role',
+    'node.documentation_page',
     'user_history',
+    'user_role',
+    'user',
```
indytechcook, line 39: While we are here, can you add `crop` and `site_alert` and maybe `path_alias`

indytechcook, line 39:
```suggestion
    'user',
    'site_alert',
    'crop',
```

swirtSJW: Good idea.

docroot/modules/custom/va_gov_content_export/src/TomeExporter.php
```diff
@@ -33,9 +33,13 @@ class TomeExporter extends Exporter {
    */
   protected static $excludedTypes = [
     'content_moderation_state',
-    'user',
-    'user_role',
+    'crop',
+    'node.documentation_page',
+    'path_alias';
```
indytechcook, line 38:
```suggestion
    'path_alias',
```

YIKES

PR #2418, opened 2020-07-25
VATEAM-11312 Add ListDataCompiler to va_gov_content_export.

swirtSJW:
## Description

See  https://github.com/department-of-veterans-affairs/va.gov-team/issues/11312. 

##Acceptance Criteria
- [x] On saving a list node, the reverse_field_list data is added to the content export for that node and include an array of list items by node ID : UUID pairs.
- [x] On saving a list item that belongs to one or more lists, the list content export will be updated to include current array of list items.
- [x] On full CMS content export, the reverse_field_list will be up-to-date on all list node exports.

## Testing done
All QA steps

## QA steps
As an admin:
1. Edit a list node `/pittsburgh-health-care/events`
    a.  Save the node
    - [ ] Inspect the file at `docroot/sites/default/files/cms-export-content/node.4d0b3e6f-37d6-46aa-9f71-0e5615511e91.json` and scroll down near the end of the content and validate that it has an element `reverse_field_list` that contains many NID : UUID pairs that represent list items that should appear on that list.

2. Add a new event 
   a. Go to /node/add/event and assign it to  event listing "VA Pittsburgh Health care events"  Fill out all required fields.
   b. Publish and save the node.  Make note of its node id.
   - [ ] inspect the file  `docroot/sites/default/files/cms-export-content/node.4d0b3e6f-37d6-46aa-9f71-0e5615511e91.json` and validate that the node id you just created is referenced in the element element `reverse_field_list`

3. Run the full content export  Steps -> https://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/READMES/cms-export.md#caveats
  - [ ] Inspect the file at `docroot/sites/default/files/cms-export-content/node.4d0b3e6f-37d6-46aa-9f71-0e5615511e91.json` and scroll down near the end of the content and validate that it has an element `reverse_field_list` that contains many NID : UUID pairs that represent list items that should appear on that list.

### DIFF
--- a/docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
+++ b/docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
@@ -2,10 +2,9 @@
 
 namespace Drupal\va_gov_content_export;
 
-use Drupal;
-use Drupal\Core\File\FileSystemInterface;
-use Drupal\Core\File\FileSystemInterface;
-use Symfony\Component\Serializer\Serializer;
+use Drupal\Core\Entity\ContentEntityInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\node\NodeInterface;
 
 /**
  * Class ListDataCompiler.
@@ -15,26 +14,25 @@ use Symfony\Component\Serializer\Serializer;
 class ListDataCompiler {
 
   /**
-   * A Drupal file system object.
+   * Drupal\Core\Entity\EntityTypeManagerInterface object.
    *
-   * @var \Drupal\Core\File\FileSystemInterface
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
    */
-  protected $fileSystem;
+  protected $entityTypeManager;
 
   /**
-   * A Symfony serializer object.
-   *
-   * @var \Symfony\Component\Serializer\Serializer
-   */
-  protected $serializer;
-
-  /**
-   * An array of excluded entity types.
+   * An array of entity types that should be treated as lists.
    *
    * @var string[]
    */
-  protected static $excludedTypes = [
-
+  protected static $listEntityTypes = [
+    'event_listing',
+    'health_services_listing',
+    'leadership_listing',
+    'locations_listing',
+    'press_releases_listing',
+    'publication_listing',
+    'story_listing',
   ];
 
   /**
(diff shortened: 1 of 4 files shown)


swirtSJW:
Lots of pseudo code at the moment.

docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
```diff
@@ -0,0 +1,219 @@
...
+
+  /**
+   * Update related lists.
+   *
+   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+   *   The entity which may reference lists.
+   * @param \Drupal\va_gov_content_export\TomeExporter $tomeExporter
+   *   The tome exporter.
+   *
+   * @todo This implementation creates one risk of failure in that, any list
```
swirtSJW, line 65: Created this issue to handle this later.  It is not a blocker for transformers
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/2464

docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
```diff
@@ -0,0 +1,219 @@
...
+   * item that has been moved from one list to another, will not know to remove
+   * it from the original list.
+   */
+  public function updateLists(ContentEntityInterface $entity, TomeExporter $tomeExporter) {
```
indytechcook, line 69: Perhaps inject `Drupal\tome_sync\ExporterInterface` instead?

docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
```diff
@@ -0,0 +1,219 @@
...
+  /**
+   * Check whether the entity is a list.
+   *
+   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+   *   The entity to check for being a list.
+   *
+   * @return bool
+   *   True is the entity is one known to be a list.
+   */
+  protected function isList(ContentEntityInterface $entity) : bool {
```
indytechcook, line 131: I think it would b a good idea to change this to a property on the entity object in the future and move the logic to check if an entity is a list to a dedicated class.  I can see this being useful to find this in the future.

docroot/modules/custom/va_gov_content_export/src/TomeExporter.php
```diff
@@ -72,12 +81,14 @@ public function __construct(
     EventDispatcherInterface $event_dispatcher,
     AccountSwitcherInterface $account_switcher,
     FileSyncInterface $file_sync,
-    AddBreadcrumbToEntity $addBreadcrumbToEntity
+    AddBreadcrumbToEntity $addBreadcrumbToEntity,
```
indytechcook, line 84: mixing camel and underscore case.

docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
swirtSJW: Can't do it.  Tried that first and it triggered an symphony circular reference error.

docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
swirtSJW: Good idea.

docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
indytechcook: really

indytechcook: Then just change `TomeExporter` to the `Drupal\tome_sync\ExporterInterface` i guess.

indytechcook: Ideally, we would split up the classes and put the logic to find if the field is a List to another service.

docroot/modules/custom/va_gov_content_export/src/TomeExporter.php
swirtSJW: Yes you did ;)

indytechcook: DAMN YOU PHPSTORM!!!

docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
swirtSJW: Shouldn't it be more specific?  It needs to be our extension of the class,not the more generic interface.   Isn't it safer to specify it as `Drupal\va_gov_content_export\TomeExporter`  ?

indytechcook: > Shouldn't it be more specific? It needs to be our extension of the class,not the more generic interface. Isn't it safer to specify it as Drupal\va_gov_content_export\TomeExporter

No, this method only wants to trigger an export on the different entities.  It shouldn't care what happens next.

The best fix is to split up the logic on this class to remove the circular reference.

swirtSJW: It has to care, because it is operating on the assumption that exporting it with our exporter.  If it is not using our exporter, it will not have the desired effect.

indytechcook: This is mixing concerns.  It's the Exporter's job to do/know this.

indytechcook: The desired effect is to update the export of related content.

swirtSJW: So not only is there a circular reference issue here.  There is a state issue.   Services can't / should not keep state.   But we have a need for the exporter to know whether it had been called by bulk or not.   The only way is to pass in the exporter not by dependency injection but at time of method call.

Issue #2513, opened 2020-07-30
Place Alert listing comes up when collapsing Q&A element and does not save however it does create a revision


howardbrande:
**Describe the defect**
When editing main content on a benefits detail page, when a Q&A section is added if you click on 'Collapse' the 'Place alert' listings come up. This only appears to happen on a additional Q&A section
**To Reproduce**
1. Go to https://prod.cms.va.gov/node/6785/edit
2. Scroll down to Main content to the Q&A section and click edit and then click collapse
3. Click Add content block and select Q&A
4. Enter dummy text for Section header, section intro, question and answer.
5. Click on 'Collapse' for both Section header and Question
 6. See 1st error Place alert listing come up (X out)
7. Go to save page as draft adding revision note
8. See 2nd error, the new Q&A does not show up on View or edit, it does however create anew revision

**Expected behavior**
The Place alert should not come up and changes to the page should be saved 

**Screenshots**
https://www.loom.com/share/1be49c523630450b82015eae54e2d679

**Desktop (please complete the following information if relevant, or delete):**

**Additional context**

## Labels
(You can delete this section once it's complete)
- [x] Issue type (red) (defaults to "Defect")
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [x] CMS objective (orange) (not needed for bug tickets)
- [ ] CMS-supported product (black)

PR #2523, opened 2020-07-30
VACMS-2513: Adding null check before firing method to prevent ajax failure

ethanteague:
## Description
See #2513 
This is being caused by an ajax error that is occurring b/c a method call is being made on an empty variable (`$parent`), here:
```
function _vagovadmin_get_node_parent(Paragraph $paragraph) {
  $parent_node = NULL;
  /** @var Drupal\paragraphs\Entity\Paragraph $parent */
  $parent = $paragraph->getParentEntity();
  if ($parent !== NULL && $parent->getEntityType()->id() === 'paragraph') {
    $parent_node = _vagovadmin_get_node_parent($parent);
  }
  elseif ($parent !== NULL && $parent->getEntityType()->id() === 'node') {
    $parent_node = $parent;
  }
  return $parent_node;
}
```
## Testing done
Steps defined in defect task

## QA steps
- As an admin, go to `node/6785/edit` 
- In main content, add a Q&A section.
- Add some text to the answer portion of the wysiwyg, and collapse it.
- Confirm collapse occurs, then expand, then recollapse.
- Verify that alert modal does not appear, and each interaction occurs as expected.


docroot/themes/custom/vagovadmin/vagovadmin.theme
```diff
@@ -58,10 +58,10 @@ function _vagovadmin_get_node_parent(Paragraph $paragraph) {
...
   $parent = $paragraph->getParentEntity();
-  if ($parent->getEntityType()->id() === 'paragraph') {
+  if ($parent !== NULL && $parent->getEntityType()->id() === 'paragraph') {
     $parent_node = _vagovadmin_get_node_parent($parent);
   }
-  elseif ($parent->getEntityType()->id() === 'node') {
+  elseif ($parent !== NULL && $parent->getEntityType()->id() === 'node') {
     $parent_node = $parent;
   }
   return $parent_node;
```
swirtSJW, line 67: If parent is null, this would be undefined.

swirtSJW: Disregard.  I missed seeing the declaration higher up.

Issue #2502, opened 2020-07-30
Iterate on Service Location editorial experience


kevwalsh:
# ACs: 

## Service Locations

**Form display**
- [x] Change form widget to Paragraphs experimental
- [x] Change "Paragraph" and "Paragraphs" string to "Service location" and "Service locations" 

**Address**
- [x] Clinic name, Building name/number, Wing, Floor or Room Number should be visible, even when “Use facility address” is checked. Order of fields TBD

**Phone number**
- [x] Help text: "Do not add an appointments-specific number, that’s lower on the screen."

**Hours**
- [x] Remove “The hours of operation is populated by Facility API” help text
- [x] Fix Table caption help text (remove the field altogether)
- [x] Hide Hours tables when “Use facility hours” is selected
- [x] Remove Add row and last two buttons

PR #2552, opened 2020-08-03
VACMS-2502 service mods et

ethanteague:
## Description
See #2502 

## Testing done
Visual, behat

## QA steps
As an admin, go to `/node/626/edit` and visually verify:
 - Form widget is now Paragraphs experimental
 - `Service location` is used for add button
 -  Clinic name, Building name/number, Wing, Floor or Room Number are visible
 -  Phone help text reads "Do not add an appointments-specific number, that’s lower on the screen."
 - Hours help text no longer appears
 - Hours caption field no longer appears
 - Hours table only appears when `Use facility hours` is selected
 - `Add row` and `Csv import` no longer appear


docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -436,41 +436,44 @@ function _va_gov_backend_modify_paragraph_interactions(array &$element, FormStat
...
-    // Each paragraph has an incrementing delta.
-    // Allows for controlling state / adding id for each unique widget delta.
-    // Otherwise, all elements will be affected.
-    $selector = sprintf(':input[name="field_phone_numbers_paragraph[%d][subform][field_phone_number_type]"]', $element['#delta']);
-    if ($paragraph_entity_reference_field_name === 'field_phone') {
-      $selector = sprintf(':input[name="field_service_location[%1$d][subform][field_phone][%1$d][subform][field_phone_number_type]"]', $element['#delta']);
+    // Instead of searching for delta and composing an element path piece by
+    // piece, we take the parent path of a field and process/transform it into
+    // a string that will point to a form element which affects the state of
+    // the field we're trying to target.
```
swirtSJW, line 463: Bonus points for good commenting

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -436,41 +436,44 @@ function _va_gov_backend_modify_paragraph_interactions(array &$element, FormStat
...
+    $element_path = $element['subform']['#parents'];
+    $parent_field = array_shift($element_path);
+    $element_path_string = implode('][', $element_path);
```
swirtSJW, line 466: If we are imploding on `][`  is it expected that this will have a `[` at the beginning and a `]` at the end?

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: Hat tip to @oksana-c !

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: Yeah, it always will - it's that thing that Drupal does when it outputs form element name attributes, @swirtSJW

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -436,41 +436,44 @@ function _va_gov_backend_modify_paragraph_interactions(array &$element, FormStat
...
+
+    if (!empty($element_path_string)) {
```
oksana-c, line 468:
```suggestion
    if (!empty($parent_field) && !empty($element_path_string)) {
```

oksana-c: I missed a check for a variable here.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
oksana-c: the actual closures are added here https://github.com/department-of-veterans-affairs/va.gov-cms/pull/2552/files#diff-5dc76f86e3dd373610cd6dd8d4189c3eR469

swirtSJW: I saw that they were added, but that's what cause me to question... aren't there already a set on there.  We are adding them around $element_path_string but it should already have them because of the implode.

oksana-c: > We are adding them around $element_path_string but it should already have them because of the implode.

brackets are only added in between array elements during an implode, not at the edges. Therefore we need to enclose `$element_path_string` when forming a selector

swirtSJW: Apologies... being an idiot.. was reading implode, but envisioning explode,   Please forgive me.

oksana-c: forgiven! remember that next time I blow a gasket 😄

Issue #2385, opened 2020-07-22
Rename listing page content types


kevwalsh:
## User Story or Problem Statement

[See this slack thread](https://dsva.slack.com/archives/CT4GZBM8F/p1595451491138200)

## Acceptance Criteria
* `Story listing` is renamed to `Stories list`
* `Event listing` is renamed to `Events list`
* `News releases listing is renamed to `News releases list`
* `Locations listing` is renamed to `Locations list`
* `Health services listing` is renamed to `Health services list`
* `Leadership listing` is renamed to `Leadership list`

## Implementation steps
- [ ] Update drupal labels and behat tests, and possibly other help text (worth searching the config for the string `listing`)

PR #2558, opened 2020-08-04
VACMS-2385: Update human names for listing types.

ethanteague:
## Description
See #2385

## Testing done
Visual / Behat

## QA steps
As an admin, confirm that each of the listing types defined in #2385 have had their human readable names changed on the content type page.


swirtSJW:
@ethanteague   Content type friendly names are used in the field behat featrue too so it will need to be updated as well.

ethanteague:
> @ethanteague Content type friendly names are used in the field behat featrue too so it will need to be updated as well.

D'oh! Updated, @swirtSJW

Issue #2347, opened 2020-07-20
CMS Export tar should be updated on every deploy


indytechcook:
During the deploy the full cms export should occur.

## AC

- [ ] During a code deploy the cms bulk export should run
- [ ] Drupal should serve the existing tar file
- [ ] No downtime for API requests coming in

**Implementation Notes**
- Write new Ansible role
- Use existing Jenkins deploy jobs
- Should  run right after updatedb & config:import so that the content model changes are present and updates have run
- Get tome sync working with drush 8. #2541

PR #2568, opened 2020-08-05
VACMS-2347 Do not tar files on cms export during deploy

indytechcook:
## Description

See #2347 

When the setting `va_site_deploy_mode` is set then the export endpoint should not re-tar the content.

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/2568)


docroot/modules/custom/va_gov_content_export/src/Controller/ContentExport.php
```diff
@@ -97,7 +109,9 @@ public function redirectToFile(string $export_type) : Response {
     $this->killSwitch->trigger();
     try {
       $archive_args = $this->getArchiveArgs($export_type);
-      $this->archiver->archive($archive_args);
+      if (!$this->siteStatus->inDeployMode()) {
```
swirtSJW, line 112: I'm not following this.  if it is in deploymode it will create an archive.   But if it is not in maintenance mode... should something else happen?

indytechcook: Deploy mode and maintenance mode are different. Deploy mode is specific to our setup.

Issue #2529, opened 2020-07-30
Extend reverse field lookup to other fields


swirtSJW:
Need the list of queries to try to match.

Discussion here https://dsva.slack.com/archives/C01512KN35G/p1596124795342500

PR #2573, opened 2020-08-05
VACMS-2529 Add reverse field data export for all reverse fields.

swirtSJW:
## Description

See #2529 . 

## Screenshots
Handles multiple references to the same target :+1: 
image

Handles multiple content types using same reference field
image

image

## QA steps

As user _uid_ with editorial privs 

1. Go to /node/370/edit   and save the node.   This node has both a  reference to a `field_list` and a `field_facility_location`  So it will update two different list type entities. and itself.
- [ ] validate that node.0b968788-cfb0-41aa-8206-2577105072fa.json  was saved (health_care_local_facility) 
- [ ] validate it  contains entries for `reverse_field_facility_location`
image

- [ ] validate that  node.4d0b3e6f-37d6-46aa-9f71-0e5615511e91.json   was saved (event_listing)
- [ ] validate it  contains entries for `reverse_field_listing`
image

 - [ ] validate that  node.7bb3b8f2-0641-4a63-a96c-93df7ac5874f.json was saved/ newly updated.  (event)
image

### DIFF
--- a/docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
+++ b/docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
@@ -131,7 +131,7 @@ class ListDataCompiler {
    * @return int
    *   The node id of a list to be updated.
    */
-  public function updateList(ContentEntityInterface $entity, $listNid, array $listFieldNames) {
+  public function updateList(ContentEntityInterface $entity, $listNid, array $listFieldNames) : int {
     // Check if $entity is the list, or if we need to get the list.
     if ($this->isList($entity) && ((int) $entity->id() === $listNid)) {
       // Add the lists to the list entity.
--- a/docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
+++ b/docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
@@ -23,16 +23,29 @@ class ListDataCompiler {
   /**
    * An array of entity types that should be treated as lists.
    *
-   * @var string[]
+   * Keyed by content type, values: array of fields pointing to content type.
+   *
+   * @var array
+   *
+   * @todo It would be nice to pull these dynamically rather than hard code.
+   * However that would need to be cached in order to make this faster.
    */
   protected static $listEntityTypes = [
-    'event_listing',
-    'health_services_listing',
-    'leadership_listing',
-    'locations_listing',
-    'press_releases_listing',
-    'publication_listing',
-    'story_listing',
+    'event_listing' => ['field_listing'],
+    'health_care_local_facility_servi' => ['field_location_services'],
+    'health_care_local_facility' => ['field_facility_location'],
+    'health_care_local_health_service' => ['field_local_health_care_service_'],
+    'health_care_region_page' => ['field_region_page'],
+    'leadership_listing' => ['field_office'],
+    'office' => ['field_office'],
+    'person_profile' => ['field_leadership'],
+    'press_releases_listing' => ['field_listing'],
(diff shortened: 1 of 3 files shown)


docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
```diff
@@ -23,26 +23,37 @@ class ListDataCompiler {
...
    *
-   * @var string[]
+   * Keyed by content type, values: array of fields pointing to content type.
+   *
+   * @var array
+   *
+   * @todo It would be nice to pull these dynamically rather than hard code.
+   * However that would need to be cached in order to make this faster.
    */
   protected static $listEntityTypes = [
```
indytechcook, line 33: Let's make a finder service we can inject, that way in the future, perhaps we can make this dynamic.  It would have methods like `getReferenceFieldsForEntity` and `getAllFields` or something like that.

docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
```diff
@@ -52,10 +63,25 @@ class ListDataCompiler {
    */
   public function __construct(EntityTypeManagerInterface $entityTypeManager) {
     $this->entityTypeManager = $entityTypeManager;
+    $this->extractListingFields();
+  }
+
+  /**
+   * Extract field names from $this->listEntityTypes into $this->listingFields.
+   */
+  protected function extractListingFields() {
```
indytechcook, line 72: This logic would be in the mentioned service instead.

docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
```diff
@@ -67,52 +93,54 @@ public function __construct(EntityTypeManagerInterface $entityTypeManager) {
...
-   * @param string $listFieldName
-   *   The field machine name to use as the reverse entity reference.
+   * @param array $listFieldNames
+   *   The field machine names to use as the reverse entity reference.
    *
    * @return int
    *   The node id of a list to be updated.
    */
-  public function updateList(ContentEntityInterface $entity, $listNid, $listFieldName) {
+  public function updateList(ContentEntityInterface $entity, $listNid, array $listFieldNames) {
```
indytechcook, line 134:
```suggestion
  public function updateList(ContentEntityInterface $entity, $listNid, array $listFieldNames) : int {
```

type hint the return

docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
```diff
@@ -153,16 +181,16 @@ protected function getListNids(ContentEntityInterface $entity) {
     $listNids = [];
     if ($this->isList($entity)) {
       // This is a list so just return the nid.
-      $listNids[$entity->id()] = 'field_listing';
+      $listNids[$entity->id()] = static::$listEntityTypes[$entity->bundle()];
     }
     else {
       // Look to see if it has any listing fields.
-      foreach (static::$listingFields as $listingField) {
+      foreach ($this->listingFields as $listingField) {
```
indytechcook, line 188: This would the new service as well.

docroot/modules/custom/va_gov_content_export/src/ListDataCompiler.php
swirtSJW: Yah that would be good.   The catch is detecting what are the right entity references... not all entity references qualify and I haven't come up with a way in my head to programatically look them up, but be smart enough to know whether it qualifies.  Trading a hand curated exclusion list is just as bad in my head as this hand curated inclusion list.

PR #2567, opened 2020-08-05
VACMS-2315 2316

ethanteague:
## Description
See #2315 & #2316 

## Testing done
Visual and behat

## QA steps
@oksana-c to verify that everything looks okay (please).

Issue #2533, opened 2020-07-31
Normalize and re-enable Behat test for @paragraph_fields tag


oksana-c:
## Description

Behat test that analyses a table of paragraph fields against Drupal fields queried from DB produces the following non-descriptive error:
```
Notice: Detected differences that cannot yet be displayed. See https://github.com/TravisCarden/behat-table-comparison/issues/1.
```

Visual inspection of "Paragraph fields" tab in Drupal Spec tool reveals no conflicts w/ the current DB state.

Paragraph fields test (tag `@paragraph_fields`) is temporarily disabled in order to unblock merges of critical content model changes. We cannot afford to keep this test disabled for too long since this opens a potential for technical debt accumulation.

## Acceptance Criteria
* @paragraph_fields Behat test and Spec tool tab are reconciled and @paragraph_fields test exclusion condition is removed from `tests.yml`. see https://github.com/department-of-veterans-affairs/va.gov-cms/pull/2511/commits/46ce22d1cd2eedea83b091d3def6fe5d23444716

PR #2579, opened 2020-08-06
VACMS-2533 fix paragraph test and reformat all tests

ethanteague:
## Description
See #2533 
This pr adds a patch to the TableEqualityAssertion class required by spec tool. At present, when duplicate entries are detected in a .feature file, an ambiguous / unhelpful error is thrown. This pr outputs a message indicating a duplicate has been detected, and the duplicate line.

This pr also reformats dst .feature files to match the formatting in the spec tool google sheet.

Depends on https://github.com/department-of-veterans-affairs/va.gov-cms/pull/2567

## Testing done
Behat

## Screenshots
image

## QA steps
Add a duplicate line to any dst feature file and run the test - visually verify that the duplicate line message and line appear in the error message.


composer.json
```diff
@@ -308,6 +308,9 @@
             },
             "drupal/workflow_participants": {
                 "Fix revision uncaught type error by extending class.": "https://www.drupal.org/files/issues/2019-08-16/workflow_participants-revision-error-3075426-1.patch"
+            },
+            "traviscarden/behat-table-comparison": {
+                "Throw meaningful message for duplicate entry.": "https://patch-diff.githubusercontent.com/raw/TravisCarden/behat-table-comparison/pull/7.patch"
```
swirtSJW, line 313: I think if we are specifically adding this patch, then it may be worth explicitly requiring the package.   As in 
`lando composer require traviscarden/behat-table-comparison`

composer.json
```diff
@@ -19,7 +19,7 @@
         "drupal-composer/drupal-scaffold": "^2.0.0",
         "drupal/address": "^1.4",
         "drupal/admin_toolbar": "^1.24",
-        "drupal/auto_entitylabel": "^2.1@beta",
+        "drupal/auto_entitylabel": "3.0-beta3",
```
swirtSJW, line 22:
```suggestion
        "drupal/auto_entitylabel": "^3.0-beta3",
```
We want this to be able to float upward from here.

swirtSJW: Lets leave as is. will fix on thenext  PR

Issue #2352, opened 2020-07-20
Behat test for "Save as draft and continue editing" action


oksana-c:
## Description

"Save as draft and continue editing" action is powered by custom code, which at time is prone to error. We need a test that covers "Save as draft and continue editing" action functionality to ensure there are no regressions in the future

## Acceptance Criteria

- [ ] test coverage for "Save as draft and continue editing" action is available and running in CI process

## Implementation

- [ ] Behat test that includes the following steps:
   - [ ] new content is created; "Save as draft and continue editing" button is clicked; confirm that user is on the edit form page; visit new content; confirm that entered content is on the page; confirm that content is in draft state
   - [ ] same steps as above, but for content edit action - e.g. performed on existing node

PR #2595, opened 2020-08-07
VACMS-2352: Adding save and continue behat testing.

ethanteague:
## Description
See #2352 

## Testing done
Behat

## Screenshots
N/A

## QA steps
N/A - ci

### DIFF
--- a/tests/behat/features/save_and_continue.feature
+++ b/tests/behat/features/save_and_continue.feature
@@ -19,34 +19,34 @@ Feature: Save and continue button works as expected.
     And I fill in "#edit-field-checklist-0-subform-field-checklist-sections-0-subform-field-checklist-items-0-value" with the text "Behat save and continue new test checklist item 1"
     And I fill in "#edit-field-administration" with the text "5"
     And I press "Save draft and continue editing"
+
     # Confirm our values
-    Then I visit the "" page for a node with the title "Behat save and continue new test"
-    Then I should see "Behat save and continue new test"
-    And I should see "test meta title"
-    And I should see "test meta description"
-    And I should see "test button label"
-    And I should see "Behat save and continue new test section header 1"
-    And I should see "Behat save and continue new test section header 2"
-    And I should see "Behat save and continue new test checklist item 1"
-    # Make sure additional revisions are saved
-    Then I visit the "edit" page for a node with the title "Behat save and continue new test"
+    Then I should not see "error has been found:"
+    And I should see "Behat save and continue new test"
+    And "#edit-field-meta-title-0-value" should have the "value" with "test meta title"
+    And "#edit-field-description-0-value" should have the "value" with "test meta description"
+    And "#edit-field-buttons-0-subform-field-button-label-0-value" should have the "value" with "test button label"
+    And "#edit-field-buttons-0-subform-field-button-link-0-uri" should have the "value" with "<nolink>"
+    And "#edit-field-checklist-0-subform-field-section-header-0-value" should have the "value" with "Behat save and continue new test section header 1"
+    And "#edit-field-checklist-0-subform-field-checklist-sections-0-subform-field-section-header-0-value" should have the "value" with "Behat save and continue new test section header 2"
+    And "#edit-field-checklist-0-subform-field-checklist-sections-0-subform-field-checklist-items-0-value" should have the "value" with "Behat save and continue new test checklist item 1"
+    And the "5" option from "#edit-field-administration" should be selected
+
+    # Make sure additional edits are saved
     And I fill in "Page title" with "Behat save and continue new test - edited"
-    And I fill in "#edit-field-meta-title-0-value" with the text "test meta title - edited"
     And I fill in "#edit-field-description-0-value" with the text "test meta description - edited"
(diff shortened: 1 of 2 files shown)


tests/behat/features/save_and_continue.feature
```diff
@@ -0,0 +1,52 @@
...
+    And I fill in "#edit-field-description-0-value" with the text "test meta description"
+    And I fill in "#edit-field-buttons-0-subform-field-button-label-0-value" with the text "test button label"
+    And I fill in "#edit-field-buttons-0-subform-field-button-link-0-uri" with the text "<nolink>"
+    And I fill in "#edit-field-checklist-0-subform-field-section-header-0-value" with the text "Behat save and continue new test section header 1"
+    And I fill in "#edit-field-checklist-0-subform-field-checklist-sections-0-subform-field-section-header-0-value" with the text "Behat save and continue new test section header 2"
+    And I fill in "#edit-field-checklist-0-subform-field-checklist-sections-0-subform-field-checklist-items-0-value" with the text "Behat save and continue new test checklist item 1"
+    And I fill in "#edit-field-administration" with the text "5"
+    And I press "Save draft and continue editing"
+    # Confirm our values
+    Then I visit the "" page for a node with the title "Behat save and continue new test"
```
swirtSJW, line 23: "" ?

swirtSJW: Is this step even necessary as after pressing the button you end back up there anyway.

tests/behat/features/save_and_continue.feature
```diff
@@ -0,0 +1,52 @@
...
+    Then I visit the "edit" page for a node with the title "Behat save and continue new test"
+    And I fill in "Page title" with "Behat save and continue new test - edited"
+    And I fill in "#edit-field-meta-title-0-value" with the text "test meta title - edited"
+    And I fill in "#edit-field-description-0-value" with the text "test meta description - edited"
+    And I fill in "#edit-field-buttons-0-subform-field-button-label-0-value" with the text "test button label - edited"
+    And I fill in "#edit-field-checklist-0-subform-field-section-header-0-value" with the text "Behat save and continue new test section header 1 - edited"
+    And I fill in "#edit-field-checklist-0-subform-field-checklist-sections-0-subform-field-section-header-0-value" with the text "Behat save and continue new test section header 2 - edited"
+    And I fill in "#edit-field-checklist-0-subform-field-checklist-sections-0-subform-field-checklist-items-0-value" with the text "Behat save and continue new test checklist item 1 - edited"
+    And I press "Save draft and continue editing"
+    Then I visit the "" page for a node with the title "Behat save and continue new test - edited"
```
swirtSJW, line 41: ?  necessary?

ethanteague: Yep. Otherwise you end up on the edit page, @swirtSJW - the above step sends the user to the view page.

tests/behat/features/save_and_continue.feature
ethanteague: Yep. Otherwise you end up on the edit page, @swirtSJW - the above step sends the user to the view page.

ethanteague:
> Just a question

Save and continue sends the user back to the node form. Setting empty quotes in this step sends the user to the view page (desired behavior for this step).

tests/behat/features/save_and_continue.feature
swirtSJW: Sorry, this seems like an anti-pattern for a test.  The purpose of the "Save and continue" button is that it saves the values and returns you to the edit form.  In this case we are forcing the test to do something unnatural and then trying to assert that it did what it is supposed to do.   In its current state, this test is only testing that "Save and continue" saves, but is completely blind to the "continue" portion.

swirtSJW: We ought to be able to assert the data was saved at the same time asserting the continue.   And it would probably be good to assert that no errors are visible after the submit.   By forcing the page to go to the view, and then testing... we have lost the ability to check for a submission error.

ethanteague: Great points, @swirtSJW . Here's my thought process: You can't get to the view page if you have form errors, because saving / node creation fails when the form doesn't submit, so the test would fail immediately - sending the test to view confirms save and continue success. Another reason why I want to send the test to the view page as opposed to remaining on the edit page is because @howardbrande reported that form paragraph data was being lost when clicking the save and continue button (the primary reason for this test). I'm concerned that checking these values on the form page after pressing the button isn't a rock solid confirmation that the data has been saved. Sending the test to the view page, and checking the values there is a true confirmation that data has not been lost.

swirtSJW: `You can't get to the view page if you have form errors, because saving / node creation fails when the form doesn't submit, so the test would fail immediately`
How certain are you of that?  The test is not testing for errors and is being overridden to jump to view page.  I think it would be a safer (less assumptions) to have it test things on the node form as save and continue is supposed to.   Then add a last check which will go to the view page and validate that a value that was set on the first save is there and a value that was altered on the second save is also there.   That would be enough to establish that both saves worked.

ethanteague: Makes sense, @swirtSJW :+1: - updated

Issue #2560, opened 2020-08-04
[tracking] Dependabot failing on timeout


ElijahLynn:
Upstream tracking here > https://github.com/dependabot/dependabot-core/issues/2416

acrollet:
There has been some movement on the upstream issue: https://github.com/dependabot/dependabot-core/issues/2416#issuecomment-759480528

acrollet:
Dependabot has been reliable since #3303, closing this.

PR #2614, opened 2020-08-08
Remove indirect dependencies to avoid dependabot 30 minute timeout

ElijahLynn:
## Description

https://github.com/department-of-veterans-affairs/va.gov-cms/issues/2560
https://github.com/dependabot/dependabot-core/issues/2416

## Testing done
This is the test. 

## QA steps

1. Do this
1. Then that
- [ ] This
- [ ] That
- [ ] The other thing

## Definition of Done

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,7 +9,6 @@ updates:
   open-pull-requests-limit: 10
   allow:
   - dependency-type: direct
-  - dependency-type: indirect
   ignore:
   - dependency-name: acquia/drupal-spec-tool
     versions:


indytechcook:
We Added all sub-dependencies because we had missed a security update on Drupal module which was a dependency of one four other Drupal modules.  We rely on composer to manage our dependencies in Drupal.

While us removing the sub-dependencies will make it work, it introduced a security risk we will have to come up with a process to monitor.

ElijahLynn:
I am thinking that once Dependable timeout is increased we can go back to the sub-dependencies and revert this change. Maybe it will only be a week or so. :crossed_fingers:

Issue #1873, opened 2020-05-29
Add "recent content i've edited" listing to user profile page. 


kevwalsh:
## Problem

People want to find content they're actively working on quickly.  

## Hypothesis

By placing actively worked on content on a user profile page, users will be able to edit the content they need immediately after logging in, with just one click. 

## Long-term context

We'll be developing these user profiles page as dashboards.

## Ideas

* "Temporary dashboard"
  * Borrow the "Recent activities" pattern from the node view page, showing Title of node, latest editor (even if it wasn't user), last revision log message, moderation state, date. 
  
## Acceptance criteria

- [ ] the block with recent content is rendered in the user profile page based on the available prototype - https://xd.adobe.com/view/9f94e624-9803-44f7-b799-5059e55796d1-cac7/screen/6b41da4e-852b-415a-b54a-df23184d4ff2?fullscreen
- [ ] users always see the latest content they've edited
- [ ] the view does not slow down user profile page load - login performance test is passing
- [ ] Coordinate Announcement with @rachel-kauff https://github.com/department-of-veterans-affairs/va.gov-cms/issues/4176 
   - Slack thread w/ announcement https://dsva.slack.com/archives/CT4GZBM8F/p1613583575276600?thread_ts=1613583550.276500&cid=CT4GZBM8F 

## Implementation

* Create a view, place on page using method from the Section listings
* Writing labels
* Internal feedback 
* Any small iteration
* @jordana to provide svg file for the lock icon
* lock icon is just an icon, not linked to anything
* view block has a distinct name e.g. `block_user_recent_content`

## Prototype 

https://xd.adobe.com/view/9f94e624-9803-44f7-b799-5059e55796d1-cac7/?fullscreen

ethanteague:
@swirtSJW and I were talking about this task, @oksana-c - caching isn't really feasible for this block in it's current format. We need to limit the amount of data that is being returned. At present, there is no time or content limit restriction on the table. So, if a user made a revision two years ago, and someone else made a new revision this am, that content would show up on the top of the user table. We should figure out some reasonable limits before working out caching.

oksana-c:
the need for "recently edited content" view was resurfaced by @davidconlon

msbtterswrth:
We could leverage the existing 'recent content' block from core and add it to the profile pages as seen below.  There are a few things we could do to improve usability, such as:

- adding a 'view more' button that links out to a paginated page
- adding another layer of filter for most recent revision only
[...]

PR #2633, opened 2020-08-11
[DO NOT MERGE] VACMS-1873: Add recent activity block for revisions to user page

ethanteague:
## Description
See #1873 

## Testing done
Visual / Behat

## Screenshots
image

## QA steps
- As an admin, visit some user pages for active editors: user/34, user/89, etc.
- Click on a few of the titles, and visit the revisions tab, ensuring that the user id for the user you are inspecting has a revision in the stack.


config/sync/block.block.views_block__recent_content_activity_block_2.yml
```diff
@@ -0,0 +1,31 @@
...
+status: true
+dependencies:
+  config:
+    - views.view.recent_content_activity
+  module:
+    - system
+    - views
+  theme:
+    - vagovadmin
+id: views_block__recent_content_activity_block_2
```
swirtSJW, line 12: moderate:  Is it possible to have this not be generic `_block_2`?   I'd prefer it be meaningful to what it is.

swirtSJW: image

ethanteague: Apologies for missing your screenshot after you were kind enough to take the time to show me how to resolve, @swirtSJW !
Updated :+1:

ethanteague:
> @Ethan this view is mostly working. A few notes:
> 
> 1. lock icon is not showing on devshop. I see the following errors in console
> 
> Screen Shot 2020-08-13 at 6 58 03 AM
> 
> this may be due to SOCKs proxy policy, Kev and I identified this issue before w. Bitter font. As much as I love the efficiency of font awesome, I think a safer route would be to add an svg icon to the theme. Tagging @jordanamello for the icon.
> 1. please switch to the date format as shown on prototype + let's make the moderation status in caps.
> 2. when viewing another user's profile, the help text under "Content" title should change to `user.email.va.gov has worked on the following content.` See [prototype](https://xd.adobe.com/view/9f94e624-9803-44f7-b799-5059e55796d1-cac7/screen/84782285-5767-44de-b312-62ede02e1c36?fullscreen)
> 3. Let's add some space  between the sections block and "Content" title
> 4. I browsed several different user profiles and it seems that the content displays as expected. When looking at Howie's profile http://pr2633.ci.cms.va.gov/user/1345 , the content list is empty. Please confirm that this is even possible

@oksana-c Howie has two profiles - the one you looked at (1345) isn't active. The one he uses is 86. I just checked, and it shows all of his activity.

jordanamello:
Hey @ethanteague , here are a few additional comments:

- +1 to @oksana-c 's comment about more space between the sections block and "Content" title.

- Let's add more space between the lock icon and the end of the title. I believe 10px is enough.

- I'd like to propose another order for the columns: TITLE, STATE, LAST CHANGED, LAST CHANGED BY, OWNER, CONTENT TYPE. By default, the table can still be ordered by LAST CHANGED. @oksana-c thoughts here?

Thanks!

ethanteague:
Updated, @oksana-c

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
```diff
@@ -0,0 +1,99 @@
...
+   * Return the rendered view.
+   *
+   * @return object
+   *   Recent user activity data.
+   */
+  public function getView() {
+    $result = FALSE;
+    $view = Views::getView('recent_content_activity');
+
+    if (is_object($view)) {
```
swirtSJW, line 42: Do we we want to look for an instance of view interface instead of any ol object?

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * Return the rendered view.
+   *
+   * @return object
+   *   Recent user activity data.
+   */
+  public function getUserHeader() {
+    $current_user = \Drupal::currentUser()->id();
```
swirtSJW, line 61: I think we want to inject this into the plugin constructor.
https://www.drupal.org/docs/8/api/services-and-dependency-injection/dependency-injection-in-plugin-block

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
```diff
@@ -0,0 +1,99 @@
...
+    $current_path = \Drupal::request()->getpathInfo();
```
swirtSJW, line 62: and here too.

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
```diff
@@ -0,0 +1,99 @@
...
+   */
+  protected function blockAccess(AccountInterface $account) {
+    return AccessResult::allowedIfHasPermission($account, 'access content');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function blockForm($form, FormStateInterface $form_state) {
+    $config = $this->getConfiguration();
```
swirtSJW, line 87: Why load this if we are not doing anything with it?

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
```diff
@@ -0,0 +1,99 @@
...
+   * {@inheritdoc}
+   */
+  protected function blockAccess(AccountInterface $account) {
+    return AccessResult::allowedIfHasPermission($account, 'access content');
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function blockForm($form, FormStateInterface $form_state) {
```
swirtSJW, line 86: Is this required by the interface?   We don't seem to be doing anything with it.

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
ethanteague: Yeah, better safe than sorry, @swirtSJW :+1:

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
ethanteague: Good q - I thought it was, but not sure. I'll remove and test, @swirtSJW

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
ethanteague: Man......I just realized I've had a pretty whacked understanding of dependency injection. I learned something this morning, @swirtSJW. Thank you!

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
ethanteague: :+1:

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
```diff
@@ -0,0 +1,121 @@
...
+    $result = FALSE;
+    $view = Views::getView('recent_content_activity');
+
+    if (is_object($view)) {
+      $view->setDisplay('block_user_recent_content');
+      $view->execute();
+      $rendered_view = $view->render();
+
+      // Render the view.
+      $result = \Drupal::service('renderer')->render($rendered_view);
```
oksana-c, line 86: ideally this service would be injected in the container

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
```diff
@@ -0,0 +1,121 @@
...
+   * Return the rendered view.
+   *
+   * @return object
+   *   Recent user activity data.
+   */
+  public function getUserHeader() {
+    $current_user = $this->account->id();
+    $current_path = $this->path->getPath();
+    $arg = explode('/', $current_path);
+    $string = t('You have worked on the following content.');
```
oksana-c, line 102:
```suggestion
    $string = $this->t('You have worked on the following content.');
```

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
```diff
@@ -0,0 +1,121 @@
...
+    if (!empty($arg[2]) && $arg[2] !== $current_user) {
+      $user = User::load($arg[2]);
+      $string = t(':user has worked on the following content.',
```
oksana-c, line 105:
```suggestion
      $string = $this->t(':user has worked on the following content.',
```

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
```diff
@@ -0,0 +1,121 @@
...
+      $container->get('current_user'),
+      $container->get('path.current')
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    $output = $this->getUserHeader() . '<br />' . $this->getView();
```
oksana-c, line 64:
```suggestion
    $output = '<p>' . $this->getUserHeader() . '</p>' . $this->getView();
```

docroot/modules/custom/va_gov_user/src/Plugin/Block/RecentUserContent.php
ethanteague: Doesn't seem to work when I check for the view interface, @swirtSJW

oksana-c: `getView()` returns either a viewExecutable or nothing at all https://api.drupal.org/api/drupal/core%21modules%21views%21src%21Views.php/function/Views%3A%3AgetView/8.2.x 

you can follow the model that `getView` follow, which is simply 

```
if ($view) {
    // render.
  }
```

ethanteague:
@oksana-c @swirtSJW per our performance testing discussion on 8/17, I ran some load time tests. Steve's initial hunch was correct - the `tests/phpunit/PerformanceLoginTest.php` test was looking for an instance of AccountInterface, and I was passing it a fully loaded User object, so the results returned were from an anonymous user login page. To fix this, I manually plugged in the username and password creds in `docroot/core/tests/Drupal/Tests/UiHelperTrait.php`, and confirmed that this did in fact load the logged in user page with the subject block (see attached phpunit html output for user ethan.teague). I ran tests for users who all (with the exception of danielle.thierry) have several pages of revisions, and the results all came in below our page load benchmark of three seconds. 

These results are all for no caching enabled on the subject block and view. I will update this pr, accordingly, but happy to change if you all don't like this approach.
danielle thierry
Jennifer Lee27@va gov
Margaret Gannon@va gov
rachel kauff
Randi Hecht@va gov
[results_ethan_teague.txt](https://github.com/department-of-veterans-affairs/va.gov-cms/files/5098517/results_ethan_teague.txt)

ethanteague:
> thanks for getting these benchmarks @ethanteague !
> initial thoughts: the page load time for less active users seems to not be an issue at all. But our numbers for Rachel's and Jen Lee's profiles do not leave us any room for growth. The 2.66 and 2.99 seconds load time will start breaking our 3 second performance test pretty soon and that can unexpectedly block deployments. Likely not the risk we want to take.
> 
> Also, just seen this in Slack https://dsva.slack.com/archives/CT4GZBM8F/p1598277301089600?thread_ts=1598277018.089500&cid=CT4GZBM8F
> 
> We can of course lift our page load goal to 5 seconds while we are developing a better caching/query strategy, although now I'm thinking that there's high likelihood that we will also resurface our good ol' DB locks issue if we were to release this view as-is.
> thoughts?

I only saw a difference in performance when a user had less than 25 results (the per page limit). So, > 25 is all the same - we won't take a performance hit as users who have more than 25 results grow their revisions. I think we are safe to proceed with this model, for now, knowing that we are ultimately going to redesign / refactor, @oksana-c

swirtSJW:
Has some conflicts

swirtSJW:
Given that we are already having cases where the login performance test is preventing builds from passing fairly often, I would like to hold on releasing this until @ElijahLynn has had a chance to blackfire our current login and see what is causing the current failures... then we can blackfire this one too and see what happens.

ElijahLynn:
> iven that we are already having cases where the login performance test is preventing builds from passing fairly often, I would like to hold on releasing this until @ElijahLynn has had a chance to blackfire our current login and see what is causing the current failures... then we can blackfire this one too and see what happens.

This may or may not be me but here is the story. https://github.com/department-of-veterans-affairs/va.gov-cms/issues/2788

config/sync/views.view.recent_content_activity.yml
```diff
@@ -0,0 +1,833 @@
...
+          table: node_field_data
+          field: title
+          relationship: none
+          group_type: group
+          admin_label: ''
+          label: 'Content title'
+          exclude: false
+          alter:
+            alter_text: true
+            text: '{{ title }} <i class="fas fa-lock display-icon-{{ is_locked }}" aria-label="This content was locked on {{ timestamp }}. You can break the lock by following the link provided on the content edit page."></i>'
```
oksana-c, line 256:
```suggestion
            text: '{{ title }} <i class="fas fa-lock display-icon-{{ is_locked }}" title="This content was locked on {{ timestamp }}. You can break the lock by following the link provided on the content edit page."></i>'
```

oksana-c: aria-label is useful for accessibility, but the `title` attr can cover both accessibility reqs and FE user experience. The original ask on one of our 16th mins was to provide a native hover tooltip text in addition to the icon.

PR #2641, opened 2020-08-12
VACMS-11163: Add production environment variable id switch.

ethanteague:
## Description
See https://github.com/department-of-veterans-affairs/va.gov-team/issues/11163
Per @jonwehausen , we need a switch to differentiate production from all other environments. So, if we are on prod, gtm is environment 2. All others are tracked as environment 5.

## Testing done
Visual inspection in console.

### DIFF
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -568,7 +568,9 @@ function _va_gov_backend_gtm_settings() {
  */
 function va_gov_backend_page_bottom(array &$page_bottom) {
   // Add GTM noscript iframe to body.
-  $gtm_noscript = Markup::create('<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-WQ3DLLB&gtm_auth=hHvHnY1eIx1IOSJMAnOBLA&gtm_preview=env-2&gtm_cookies_win=x" height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>');
+  $env_name = \Drupal::config('environment_indicator.indicator')->get('name');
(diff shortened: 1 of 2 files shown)


docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -568,7 +568,9 @@ function _va_gov_backend_gtm_settings() {
  */
 function va_gov_backend_page_bottom(array &$page_bottom) {
   // Add GTM noscript iframe to body.
-  $gtm_noscript = Markup::create('<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-WQ3DLLB&gtm_auth=hHvHnY1eIx1IOSJMAnOBLA&gtm_preview=env-2&gtm_cookies_win=x" height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>');
+  $env_name = \Drupal::config('environment_indicator.indicator')->get('name');
+  $env_num = $env_name === 'Production' ? '2' : '5';
```
swirtSJW, line 572: We don't have an environment named `Production` so I think this will always be 5.
Here are the available names
image

ethanteague: This pulls from `$config['environment_indicator.indicator']['name']` in the environment settings file, @swirtSJW - the local lando settings file is set to "Lando" (Tested on my local and it returns "Lando"). Production settings is set to: `$config['environment_indicator.indicator']['name'] = 'Production';`

swirtSJW: Ahh... blarrgh   Sorry for this.  We are overriding the name in settings.prod.php to something that is not even an option in config.  Not your fault.... just complaining about a disconnect.

Issue #2637, opened 2020-08-12
Some reverse fields are missing


swirtSJW:
From slack  https://dsva.slack.com/archives/C01512KN35G/p1597238927424000

 2. Some reverse fields are expected to exist in the JSON files, but do not. E.g.: node-office is an office node which is expecting the reverseFieldListingNode field

Investigation:

There are only two offices:
- [x] VA Central Office -> Node 126  -> node.c715d441-b6cb-4067-92eb-0650e97b7d62.json 

- [x] Outreach and events -> Node 735  -> c71e29a0-431f-44a5-ac5c-cee0c0ef0ceb

both have entries for `reverse_field_office`

swirtSJW:
The original request from the transformer team was  had mentioned that the transformers were expecting reverseFieldListingNode   but that is not really a field.   That seems like a class of processing.    reverse_field_office does exist in the export.  
The one example node you referenced lives here https://www.va.gov/outreach-and-events/ and does not seem to have any content on it that would result from anything other than the processing of
```
"reverse_field_office": [
        {
            "target_type": "node",
            "target_bundle": "event_listing",
            "target_id": "736",
            "target_uuid": "30295cab-89be-4173-a33c-8c0ca0a85d07"
        },
        {
            "target_type": "node",
            "target_bundle": "publication_listing",
            "target_id": "737",
            "target_uuid": "931072aa-94ed-4072-b3d4-d8735052a073"
        }
    ]
```
With a sample of 1, I am having a hard time determining if there is anything else I should be expecting on this content type.

This conversation is here in slack https://dsva.slack.com/archives/C01512KN35G/p1597353559443100

swirtSJW:
Based on conversation in the slack thread, I am closing this out.

Issue #2636, opened 2020-08-12
Some reverse fields are empty


swirtSJW:
From Slack https://dsva.slack.com/archives/C01512KN35G/p1597238927424000

1. Some reverse fields exist in the JSON files, but they're always empty arrays. therefore, we can't really test them. E.g.: node-health_services_listing and node-leadership_listing

as far as node-leadership_listing I used "entityId": "4414" in which returns with an empty array

- [x] node-leadership_listing  - No field_office ever points or is allowed to point at this content type.  It should be removed.
- [x] node-health_services_listing

swirtSJW:
I think the empty data for reverse_field_office in leadership_listing nodes is a misdirection.   I looked through the whole system and there are no nodes with field_office that point to or are even allowed to point to leadership_listing nodes. so then there would be  nothing that ever shows up in reverse_field_office . I am going to remove that reverse_field from the content type.  It should have never been there.

I have a feeling that the grpahQL query was originally looking for it, but then the content moddle changed and the graph query was never updated to remove it.

PR #2653, opened 2020-08-13
VACMS-2637 & 2636 modify reverse entity reference export.

swirtSJW:
## Description

See #2636 and #2637.


swirtSJW:
This resolves some of the missing data, but not all.   Need to get this out while I work on the remainders.

PR #2699, opened 2020-08-18
VACMS-2676 Update regional health care service data.

swirtSJW:
## Description

See _issueid_. 

## Data Issues that need resolving
- [x] inconsistent separators:  commas on some lines, semicolons on all the others.
- [x] Some newlines in mid-row.
- [x] Fix redundant `VA` in `VA Sheridan VA health care` entries
- [x] links are not properly formed:  Example: ` <a href=https://www.mentalhealth.va.gov/res-vatreatmentprograms.asp>www.mentalhealth.va.gov</a>`
- [x] Split  the migration to use the region (system) and the service as the unique ID for the migration.  This will split the data from Wilkes-Barre, but that is ok since we are not planning to ever "deblob" content on regional services, only facility services.
- [x]  re-using unique id's Example: `VAMC-1`  has already been assigned to VA Wilkes-Barre health care  
**UPDATE** This can be de-prioritized since we are no longer using that column as the unique id.
- [x] One entry using service name `Primary Care` rather than `Primary care`  - REMOVED since it appeared as duplicate.
- [x] 2 rows with id `Surgery-VA Eastern Colorado health care`  Can't have repeats of unique ID.
- [x] 3 rows with id `Rehabilitation and prosthetics-VA Eastern Colorado health care` Can't have repeats of unique ID
- [x] discrepancy with lookup for `VA Oklahoma City health care` in Drupal and `VA Oklahoma health care` in CSV. Which one is right?

## Screenshot
image


swirtSJW:
A quick scan of the data form the initial import/commit has some issues
- [ ] inconsistent separators:  commas on some lines, semicolons on all the others.
- [ ] Some newlines in mid-row.

image

swirtSJW:
Updated csv resolved the issue of changing separators.

SgardnerPMP:
Revised .csv file with new unique identifier, remediated all href links, all "VA" duplicates remediated at core playbook,  and confirmed that all line breaks have been fixed.   Solid .csv all the way down, submitted with "new" at the end of the file name.  Thank you.

swirtSJW:
Just ran the migration locally and getting a fatal error.   Looking into it 
image

swirtSJW:
There are 4 rows that are not migrating  I am looking to figure out why.
image

swirtSJW:
@SgardnerPMP I am seeing two issues with this spreadsheet.  
There seems to be a duplicate of rows at 158 and 159.  The only difference being that one is using 'Primary Care' and the other 'Primary care'
image

I am guessing that 158 is the wrong row and should be removed?  Can you confirm?

Also on those rows there is another issue related to encapsulation of the cells.  The cells are being encapsulated with single quotes, but then when it encounters the word `Women's` where the single quote is not escaped, so it is breaking the encapsulation.  When you export the csv from the spreadsheet and tell it to encapsulate, it should escape any internal quotes automatically.

swirtSJW:
@SgardnerPMP  I am seeing repeats of the Unique ID on  the following rows:

Surgery-VA Eastern Colorado health care
On line 171 and 172
https://github.com/department-of-veterans-affairs/va.gov-cms/pull/2699/files#diff-aece06f0e5f21af548ff9957348d7d3cR171-R172




Rehabilitation and prosthetics-VA Eastern Colorado health care
Is  used on 166, 167. amd 168.
https://github.com/department-of-veterans-affairs/va.gov-cms/pull/2699/files#diff-aece06f0e5f21af548ff9957348d7d3cR166-R168

If there are supposed to be multples, we need increment numbers after the uid so they will be different.   If they are just duplicates,  tell me to remove.

swirtSJW:
@SgardnerPMP  I have resolved the single quote encapsulation  but I need your guidance on what to do with the duplicate entries.  Should they just be removed or is there a problem with the data?

SgardnerPMP:
Ran select * Unique on ID field and then double checked why these duplicates went through the first time.  confirmed duplicates have been removed.  And now resubmitted for testing.

docroot/modules/custom/va_gov_migrate/config/install/migrate_plus.migration.facility_health_care_service.yml
```diff
@@ -0,0 +1,91 @@
+uuid: d6f0d43f-2e80-4224-9013-982136c095c8
```
swirtSJW, line 1: This file was already in config/sync but had never been started in the config/install which is our usual pattern, so just adding it to config/install for consistency.

swirtSJW:
Running the migration locally looks good.
image

swirtSJW:
Running on devshop
image

Then executing

image

Titles and draft states are good.
image


Entity lookups are good.
image

swirtSJW:
@SgardnerPMP I am seeing failures on system services in Oklahoma.

image


It seems to come down to a difference between how the VAMC System is entered.
image
The value to be looked up in the system is `VA Oklahoma City health care` but the CSV is doing a lookup for `VA Oklahoma health care`
image


What I am not clear on is, do we have it wrong in the system value in Drupal, or do we have it wrong in the CSV?

SgardnerPMP:
Hi Steve, I think we have it wrong in CMS.  To be accurate, we should update the "Owner" Value to represent "VA Oklahoma City health care" and I can then update the .csv to match.  Slack me when you get this and we can discuss next steps.  Thank you.

swirtSJW:
image
  Looks good.

Issue #2687, opened 2020-08-17
Child sections on /section pages should be sorted by term weight


kevwalsh:
**Describe the defect**
Child sections are out of order on some /section pages 

**To Reproduce**
Go to section/veterans-health-administration/visn-4 note Pittsburgh appears first.
Go to section/vamc-facilities and note VISN 4 appears first

**Expected behavior**

Sections should be listed in the same order as 

/admin/structure/taxonomy/manage/administration/overview 

**Screenshots**
![image](https://user-images.githubusercontent.com/643678/90424280-dcefcb00-e08b-11ea-9a1e-0855bddb104a.png)

PR #2713, opened 2020-08-19
VACMS-2687: Sort child terms by weight, then name.

kevwalsh:
## Description

PR #2688 wasn't sufficient for #2687.  In many cases, taxonomy term weight of child terms is all 0, and editors will expect alphabetical order. 

## QA

Go to /taxonomy/term/163... terms all have weight 0, and should be sorted alphabetically

### Before

VISN_2___VA_gov_CMS

### After

VISN_2___VA_gov_CMS


swirtSJW:
@dependabot merge

Issue #2756, opened 2020-08-24
Metatag image urls are oddly pathed


swirtSJW:
Metatag image paths are odd. 

![image](https://user-images.githubusercontent.com/5752113/91204472-28305c00-e6d2-11ea-9045-db23395e66d7.png)

Related conversation https://dsva.slack.com/archives/C01512KN35G/p1598292814004300

swirtSJW:
This is going out today 8/26

swirtSJW:
This was re-opened as this change broke the actual FE build when building prod due to jenkins not being able to access the actual prod url

swirtSJW:
This was on hold while the new deploy process was in resolved.   This can now proceed through testing all the way to prod.

PR #2761, opened 2020-08-25
VACMS-2756 Add file_public_base_url to lando dev,staging and prod env…

swirtSJW:
## Description

See #2756 . 

## Testing done
1. Saved a node with a defined image `/node/2299/edit`
2. inspected the output of that page in the content export 
`node.5f30741e-d04b-4d0e-858a-3ab94334e3ca.json`
3. examined metatag image paths to make sure they contained the proper domain and path.
image

4. Examined path to page and made sure it was unaffected.

image

5. Examined view of the page to make sure the rendering of the image still worked
`/wilkes-barre-health-care/locations/northampton-county-va-clinic`
image

## QA steps
Testing wont be possible in CI because those paths are dyanmic and unaffected by these changes.   Will need to be tested on dev or staging.

As an admin:

- [ ]  Save a node with a defined image `https://staging.cms.va.gov/node/2299/edit`
- [ ]  inspect the output of that page in the content export 
`docroot/sites/default/files/cms-export-content/node.5f30741e-d04b-4d0e-858a-3ab94334e3ca.json`
- [ ]  examine metatag image paths in the export file to make sure it contains the proper domain and path like this
`"image_src": "https:\/\/staging.cms.va.gov\/sites\/default\/files\/2020-04\/Northampton_480x330.jpg",`
- [ ] scroll down to the path element and make sure the path to the node is root relative like 
`"path": "\/wilkes-barre-health-care\/locations\/northampton-county-va-clinic"`
- [ ] Examine the view mode of the page `https://staging.cms.va.gov/wilkes-barre-health-care/locations/northampton-county-va-clinic` and validate the page still shows the image.

## Definition of Done

### DIFF
--- a/docroot/sites/default/settings/settings.dev.php
+++ b/docroot/sites/default/settings/settings.dev.php
@@ -22,6 +22,7 @@ $config['system.logging']['error_level'] = 'all';
 $config['environment_indicator.indicator']['bg_color'] = '#05F901';
 $config['environment_indicator.indicator']['fg_color'] = '#000000';
 $config['environment_indicator.indicator']['name'] = 'Development';
+$settings['file_public_base_url'] = 'https://dev.cms.va.gov/sites/default/files';
 
 
 $settings['trusted_host_patterns'] = [
--- a/docroot/sites/default/settings/settings.lando.php
+++ b/docroot/sites/default/settings/settings.lando.php
@@ -25,6 +25,8 @@ $config['environment_indicator.indicator']['name'] = 'Lando';
 // Link to this file locally since lando can not access prod where the real
 // file exists.  You will need to copy the file from the same path on prod.
 $config['migrate_plus.migration.va_node_form']['source']['urls'] = ['http://va-gov-cms.lndo.site/sites/default/files/migrate_source/va_forms_data.csv'];
+$settings['file_public_base_url'] = 'http://va-gov-cms.lndo.site/sites/default/files';
+
 
 $settings['trusted_host_patterns'] = [
     // For ALB/ELB Healthchecks.
--- a/docroot/sites/default/settings/settings.lando.php
+++ b/docroot/sites/default/settings/settings.lando.php
@@ -27,7 +27,6 @@ $config['environment_indicator.indicator']['name'] = 'Lando';
 $config['migrate_plus.migration.va_node_form']['source']['urls'] = ['http://va-gov-cms.lndo.site/sites/default/files/migrate_source/va_forms_data.csv'];
 $settings['file_public_base_url'] = 'http://va-gov-cms.lndo.site/sites/default/files';
 
-
 $settings['trusted_host_patterns'] = [
     // For ALB/ELB Healthchecks.
     'localhost',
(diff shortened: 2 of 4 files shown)


docroot/sites/default/settings/settings.dev.php
```diff
@@ -22,6 +22,7 @@
 $config['environment_indicator.indicator']['bg_color'] = '#05F901';
 $config['environment_indicator.indicator']['fg_color'] = '#000000';
 $config['environment_indicator.indicator']['name'] = 'Development';
+$settings['file_public_base_url'] = 'https://dev.cms.va.gov/sites/default/files';
```
ElijahLynn, line 25: I don't like that we need to specify the scheme here (http vs https) but I think that is the way it needs to be and is an issue I would need to take up on the Drupal core issue queue.

ElijahLynn:
I want to deploy this to DEV, before merge to test this out and see if it solves our issues. After today's merge window/deploy, let's push this branch to this repo (not fork), build and deploy to DEV and test there. 

I am still reading through https://www.drupal.org/project/drupal/issues/2528988 and https://www.drupal.org/node/2753591 to understand why $base_url, $base_host was removed etc. The rational I am reading so far is ultimately that D8 can autodetect the host, I assume from the HTTP_HOST header from an Apache/browser request and in CLI with Drush by passing in `drush -uri=example.com`. Which I wonder if we should just change the devops scripts to pass in that option actually.

docroot/sites/default/settings/settings.dev.php
swirtSJW: Is there a problem with specifying the scheme in our setup?

swirtSJW:
@ElijahLynn  My concern about handling on the request is that concevable someone logged into prod.cms.va.gov and saves a node, then the path in the content export file is going to be based on prod.cms.va.gov and then if someone else is on cms.va.gov and they hit save, then the image path in the export is going to be cms.va.gov.    That is going to mean transformers will have two cases to transform rather than just one.   More complexity for them, and I am not sure what it gains us.

ElijahLynn:
> @ElijahLynn My concern about handling on the request is that concevable someone logged into prod.cms.va.gov and saves a node, then the path in the content export file is going to be based on prod.cms.va.gov and then if someone else is on cms.va.gov and they hit save, then the image path in the export is going to be cms.va.gov. That is going to mean transformers will have two cases to transform rather than just one. More complexity for them, and I am not sure what it gains us.

Gotcha, we currently redirect all users that request cms.va.gov to prod.cms.va.gov. e.g. only one effective domain any user can use.

docroot/sites/default/settings/settings.dev.php
ElijahLynn: No actual issue.

ElijahLynn:
K, I tested this in DEV with `cd /var/www/cms; sudo -u apache bash -c "source /etc/sysconfig/httpd; /usr/local/bin/drush va-gov-cms-export-all-content --process-count=8 --entity-count=500 --delete-existing 2>&1"` then `cd docroot/sites/default/files; grep -r "image_src" .` and it works, see screenshot below:
image


Worth noting that I grepped before a full export again and saw a bunch of prod.cms.va.gov URLs already in there from PROD. 

image


> Testing wont be possible in CI because those paths are dyanmic and unaffected by these changes. Will need to be tested on dev or staging.

I tested on CI and the export is going to the sites/default/files folder which CI doesn't currently work with and instead uses a path of e.g. `sites/pr123.ci.cms.va.gov/files`. So we need a follow up story to get this working with CI, but worth noting that we may be switching to a new system soon and can rework that then. 

Also worth noting is that the export on CI is much faster, and it might be because we are using a c5.9xlarge, which is 36 cores and likely has faster EBS networking etc. A full export on CI was:
- 8 cores, 500 = ~55 seconds
- 16 cores, 250 = ~35 seconds
- 24 cores, 250 = ~30 seconds

image
image
image

Issue #2430, opened 2020-07-27
Remove seconds field from start and end date on events content type


rachel-kauff:
On the events content type, the start / end date fields have the user specify the time down to the second the event will start. Removing the seconds field will improve user experience by omitting unnecessary information.  While we're at it, let's update the AM/PM labels to the format specified by the VA content style guide. E.g. 10:30 a.m.

AC
- [ ] Start / end date fields use the following format: **00:00 AM/PM**, not 00:00:00 AM/PM 
- [ ] Labels in in the time field **use a.m. or p.m.** not AM or PM
- [ ] FE templates are not affected by the date format change. Spin up FE locally and verify dates display as expected

![Screen Shot 2020-07-27 at 9.33.38 AM.png](https://images.zenhubusercontent.com/5e399cf5446c5b4ca0e31e42/38f2027b-79fa-4313-9c6e-c46a93023cfb)

Related to help text improvements efforts for VAMC content types: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/2318

PR #2762, opened 2020-08-25
VACMS-2430: Remove seconds input from datetime fields.

ethanteague:
## Description
See #2430 

## Testing done
Front end build / visual

## Screenshots
image

## QA steps
- As an admin, go to `node/6920/edit` and visually verify that seconds input does not appear in date widget.
- Change time, ensure node saves without errors.


docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -38,6 +38,24 @@ function va_gov_backend_help($route_name, RouteMatchInterface $route_match) {
   }
 }
 
+/**
+ * Implements hook_element_info_alter().
+ */
+function va_gov_backend_element_info_alter(array &$types) {
+  $types['datetime']['#process'][] = 'va_gov_backend_datetime_set_format';
```
swirtSJW, line 45: I don't think we want to add this blindly.   I think we only want to set this if $types['datetime'] is set.

swirtSJW: Also `$type` is D7,  `$info` is D8   and `$types` is Dnothin.

swirtSJW:
Has some conflicts that need resolving.

ethanteague:
> Has some conflicts that need resolving.

Updated, @swirtSJW :+1:

Issue #1883, opened 2020-05-29
EWA: Entity meta data block


oksana-c:
## User Story

As a content editor, I need quick access to node's meta data information when I'm viewing node's latest or older revisions. 

## Acceptance criteria

- [ ] an entity meta data block is displayed in the right sidebar on the node view page per [prototype](https://xd.adobe.com/view/62bb878c-565e-4120-6449-3ea443b6064f-148a/?fullscreen)
<img width="440" alt="Screen Shot 2020-06-04 at 12 29 05 PM" src="https://user-images.githubusercontent.com/16477724/83801003-9cc0a500-a665-11ea-9496-85e8e33eb426.png">

## Implementation

- block's content is proprietary to the VACMS and will live in a custom module
- the following information should be included:
    - [ ] Owner: contextual to currently viewed revision
    - [ ] content type
    - [ ] Va.gov URL: taken from latest published revision or N/A [TBD in #1900]

PR #2781, opened 2020-08-26
VACMS-1883: Add EWA entity metadata block.

ethanteague:
## Description
See #1883 

## Testing done
Visual / Behat

## Screenshots
image

## QA steps
As any user, go to the view page for any node, and visually verify that EWA metadata block appears.


docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -0,0 +1,149 @@
...
+
+use Drupal\Core\Block\BlockBase;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Link;
+use Drupal\Core\Url;
+use Drupal\node\NodeInterface;
+use Drupal\taxonomy\Entity\Term;
+use Symfony\Component\DependencyInjection\ContainerInterface;
```
swirtSJW, line 13: minor: alpha sort would be nice.

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -0,0 +1,149 @@
...
+      $block_items['VA.gov URL'] = Link::fromTextAndUrl($url, Url::fromUri($url))->toString();
+    }
+    $output = '';
+    foreach ($block_items as $block_key => $block_item) {
+      $output .= '<div><span class="ewa-title"><strong>' . $block_key . ': </strong></span><span class="ewa-content">' . $block_item . '</span></div>';
+    }
+    return [
+      '#markup' => $output,
+      '#cache' => [
+        'tags' => ['node:' . $nid],
```
swirtSJW, line 80: Nice

docroot/modules/custom/va_gov_ewa/va_gov_ewa.info.yml
```diff
@@ -0,0 +1,5 @@
+name: 'VA.gov EWA'
+description: 'Provides blocks and other functionality for uEWA'
```
swirtSJW, line 2: Words would be good here so that in a couple years when we are all off the project people aren't saying "what the bleep is uEWa?!?"

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
swirtSJW: Wondering if we also have to cache this based on term: tid   to cover if someone edits a term name, we don't want old use floating around.

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function build() {
+    $block_items = [];
+    $block_items['Owner'] = $this->getSections();
+    if ($this->routeMatch->getParameter('node') instanceof NodeInterface) {
+      $nid = $this->routeMatch->getParameter('node')->id();
+      $url = 'https://va.gov' . Url::fromRoute('entity.node.canonical', ['node' => $nid])->toString();
```
swirtSJW, line 69: Hmmm this is going to create links to pages that actually do not exist on based on va.gov.    @kevwalsh  and I were just discussing about the need for a centralized list of bundles that do not exist as pages the front end.

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
ethanteague: Not sure how to do this, @swirtSJW. What would the syntax look like with multiple tags?

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
ethanteague: Okay. Sounds like we need to discuss this, @swirtSJW @kevwalsh

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
swirtSJW: I am pretty sure it would look something like
`'tags' => ['node:' . $nid, 'term:' . $tid1,  'term:' . $tid2],

ethanteague: Thanks, @swirtSJW - I'll start testing it that way :+1:

ethanteague: Yep. Changing tags is updated automagically in the block. Updated in code, @swirtSJW

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
swirtSJW: What I am thinking is a UI in va_gov_backend that provides a text field that we can add machine names for any that have no front end page.  Then the various places that need access to it can get that config.  

example form https://git.drupalcode.org/project/node_link_report/-/blob/8.x-1.x/src/Form/NodeLinkReportAdminForm.php#L96-107

then on save you clean up 
https://git.drupalcode.org/project/node_link_report/-/blob/8.x-1.x/src/Form/NodeLinkReportAdminForm.php#L172
with this
https://git.drupalcode.org/project/node_link_report/-/blob/8.x-1.x/src/Form/NodeLinkReportAdminForm.php#L190-199

The advantage of this over checkboxes for each content type would be that  you can make one config call and get all the values instead of needing to call all the fields and re-assemble them.

ethanteague: Seems legit to me. Let’s say a content type doesn’t have a frontend page - what are we gonna output for the link? Nothing? Or maybe say “current page?”

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -0,0 +1,172 @@
...
+    if (!empty($term_ids)) {
+      foreach ($term_ids as $term_id) {
+        $all_tags[] = 'taxonomy_term:' . $term_id;
+      }
+    }
+
+    return [
+      '#markup' => $output,
+      '#cache' => [
+        'tags' => $all_tags,
```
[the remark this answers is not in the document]
oksana-c: I'm adding a method/comment that allows to cover all necessary cache context based on currently viewed node above in the block annotation section ^^^


also, we can remove the `'#cache' => [` part completely now based on the context within annotation proposed above.

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -0,0 +1,172 @@
...
+use Drupal\taxonomy\Entity\Term;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Provides a block of Meta information for entities.
+ *
+ * @Block(
+ *   id = "va_gov_entity_meta",
+ *   admin_label = @Translation("Entity Meta Display")
+ * )
```
oksana-c, line 21:
```suggestion
*   context = {
*     "node" = @ContextDefinition("entity:node", label = @Translation("Node"))
*   }
 * )
```

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -0,0 +1,172 @@
...
+   * Returns max of 3 section hierarchy of breadcrumbs and section term ids.
+   *
+   * @return array
+   *   All section breadcrumbs in hierarchy and term ids.
+   */
+  public function getSections() {
+    if ($this->routeMatch->getParameter('node') instanceof NodeInterface) {
+      $sections = [];
+      $tids = [];
+      $node = $this->routeMatch->getParameter('node');
```
oksana-c, line 107:
```suggestion
      $node = $this->getContextValue('node');
```

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  /**
+   * Returns max of 3 section hierarchy of breadcrumbs and section term ids.
+   *
+   * @return array
+   *   All section breadcrumbs in hierarchy and term ids.
+   */
+  public function getSections() {
+    if ($this->routeMatch->getParameter('node') instanceof NodeInterface) {
```
oksana-c, line 104:
```suggestion
    if ($this->getContextValue('node') instanceof NodeInterface) {
```

oksana-c: ^ using value from block context system

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
oksana-c: if the content type doesn't have a FE page, we don't show VA.gov URL part at all 👍

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
oksana-c: ^ allows Drupal to take care of caching automatically for you. Since we only display this block on nodes.

oksana-c:
let's also remove the existing "Owner" block on nodes, since we show the owner value in the new Entity meta block

Screen Shot 2020-08-30 at 1 11 56 PM

docroot/modules/custom/va_gov_ewa/src/Plugin/Block/EntityMetaDisplay.php
oksana-c: if the content type doesn't have a FE page, we don't show VA.gov URL part at all 👍

oksana-c: additional criteria: 
VA.gov URL will be displayed only if the node has at least one "published" revision and the latest revision is not "archived"

swirtSJW: As a future enhancement it could display a link the page where the content appears... if it is service  or list content that would be displayed on a parent page..   But I don't think it needs to be able to do that now.

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -0,0 +1,197 @@
...
+    $sections = $this->getSections()['links'];
+    $block_items['Owner'] = $sections;
+    $nid = '';
+    if ($this->routeMatch->getParameter('node') instanceof NodeInterface) {
+      $node = $this->routeMatch->getParameter('node');
+      $block_items['Content Type'] = $node->type->entity->label();
+
+      // Make sure this bundle is okay to display on va.gov.
+      $exclude_types_string_raw = $this->configFactory->getEditable('workflow_assignments_admin.settings')->get('types_to_exclude');
+      $exclude_types_string = $exclude_types_string_raw;
```
swirtSJW, line 88: Minor:  Unused reassignment.

docroot/modules/custom/va_gov_workflow_assignments/src/Form/WorkflowAssignmentsAdminForm.php
```diff
@@ -0,0 +1,76 @@
+<?php
```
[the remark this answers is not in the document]
ethanteague: My bad, @swirtSJW - I missed the namespace suggestion. Will update.

docroot/modules/custom/va_gov_backend/src/Form/ExclusionTypesAdminForm.php
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+    $config = $this->config('exclusion_types_admin.settings');
+
+    $form['types_to_exclude'] = [
+      '#type' => 'textarea',
```
oksana-c, line 27: _This is a suggestion, your call whether to implement._

If we were to make this field a set of checkboxes `'#type' => 'checkboxes',` with options pulled from all available node types, then we would not need to process values entered into the text area in `convertSort()` and the selection in UI would be faster than finding an appropriate node type machine name.

to pull all available node types:
```
$types = \Drupal::entityTypeManager()
  ->getStorage('node_type')
  ->loadMultiple();
```
this example is for quick reference, doesn't use DI but should.

swirtSJW: If we can easily get an array of content types out of that on reading the config, that would work.  I was thinking we would need to query the config out of each checkbox in oder to build our array on the read side as opposed to jest exploading the one config, but maybe not.

ethanteague: Yooooooooooooo - @oksana-c !
image

ethanteague: Errrr......actually.....looks like we need to ->get() name / machine name on each of these. These properties are protected. So, code would look like this:
`    $types = $this->entityTypeManager
      ->getStorage('node_type')
      ->loadMultiple();
    foreach ($types as $type) {
      $options[$type->get('type')] = $this->t(':name', [':name' => $type->get('name')]);
    }`

ethanteague: Updated to use checkboxes.

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -0,0 +1,195 @@
...
+      $node = $this->routeMatch->getParameter('node');
+      $block_items['Content Type'] = $node->type->entity->label();
+
+      // Make sure this bundle is okay to display on va.gov.
+      $exclude_types = $this->configFactory->getEditable('exclusion_types_admin.settings')->get('types_to_exclude');
+      // If active revision is not published, it doesn't get included in build.
+      // Sufficient to check moderation state on node = active revision.
+      if (!in_array($node->bundle(), $exclude_types) && $node->get('moderation_state')->getString() === 'published') {
+        $nid = $node->id();
+        $url = 'https://va.gov' . Url::fromRoute('entity.node.canonical', ['node' => $nid])->toString();
```
indytechcook, line 92: Is this meant to be hard coded to va.gov?

ethanteague: > Is this meant to be hard coded to va.gov?

It is, @indytechcook . The idea for this block is to give editors the real page link on va.gov if it is a content type that is allowed to be published to va.gov, and there is a published revision that lives on va.gov.

indytechcook: just a note, the url for the website has changed in the past.  It went from vets.gov to va.gov.

ethanteague: Good point. That said, the build will fail and alert us if the url changes and this block has bad links.

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -0,0 +1,195 @@
...
+      else {
+        $output .= $this->t('<div><span class="va-gov-entity-meta__title"><strong>@block_key: </strong></span><span class="va-gov-entity-meta__content">@block_item</span></div>',
+        [
+          '@block_key' => $block_key,
+          '@block_item' => $block_item,
+        ]);
+      }
+    }
+    return [
+      '#markup' => $output,
```
indytechcook, line 115: I'm going to approve this but in the future, it's best to use a theme function or render array.  If someone wanted to alter this block they would have to regex or rebuild the entire html.

indytechcook: For example, this is just list of links: https://api.drupal.org/comment/57318#comment-57318

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -0,0 +1,195 @@
...
+   *   All section breadcrumbs in hierarchy and term ids.
+   */
+  public function getSections() {
+    if ($this->getContextValue('node') instanceof NodeInterface) {
+      $sections = [];
+      $tids = [];
+      $node = $this->getContextValue('node');
+
+      // Grab our current section.
+      $tid = $node->get('field_administration')->getString();
```
indytechcook, line 133: This is some good logic. We should eventually move this into it's own service. The `SectionDiscovery` service or something.

indytechcook (approved):

Issue #2788, opened 2020-08-27
Deploys failing due to performanceLoginTest failures (too long)


ElijahLynn:
We are currently seeing a lot of intermittent failures (flickers) on our performanceLoginTest on our upper level environments which do not have a local MySQL (e.g. RDS), which is currently set to pass if under 3 seconds (previously 2 seconds, see #1657). I agreed on scrum this morning to bump up to 4 seconds if we prioritize a story for next sprint to do some performance profiling here. This story needs to be prioritized for Sprint 14 and we need to get this back down under 3 seconds.

I am making a PR for this issue to increase the time from 3-4 seconds, as a temporary adjustment so we can still deploy in the short term. 

AC:
- [ ]  Regression is identified
  - [x] Post profiling results if you do any (e.g. Blackfire, Xhprof)
  - [ ] Post summary of what the regression was
  - [x] Open any follow up stories as needed

Follow-up story Goal: Get 10 login tests in a row to pass with a threshold of 3 seconds on RDS environments 
`source /etc/sysconfig/httpd; PATH=$PATH:/usr/local/bin composer yaml-tests phpunit --dry-run 2>&1` (there is a better way to run just the login test but that is a good start).

See https://github.com/department-of-veterans-affairs/va.gov-cms/issues/1657#issuecomment-631755887 for previous identical issue.

ElijahLynn:
This was auto-closed, re-opened.

oksana-c:
@ElijahLynn , please include profiling and any expected outcomes as an AC of this story. Thanks.

ElijahLynn:
Done :heavy_check_mark:. 

The main outcome I care about is "Get 10 login tests in a row to pass with a threshold of 3 seconds on RDS environments", which I think would involve profiling but doesn't necessarily need to which is why I left it out.

ndouglas:
Working on this, but have a couple Blackfire profiles I wanna post before I lose them:

- [HEAD: ~10s](https://blackfire.io/profiles/ec8b8fc0-02a2-4973-8bc4-84297b5c97ab/graph)
- [s/setMaxDepth(4)/setMaxDepth(1)/: ~3s](https://blackfire.io/profiles/74df9d00-08b3-49a0-aa34-a8779d6d108c/graph)

ElijahLynn:
[...]

PR #2789, opened 2020-08-27
Increase PerformanceLoginTest.php threshhold from 3 to 4, temporarily.

ElijahLynn:
## Description

https://github.com/department-of-veterans-affairs/va.gov-cms/issues/2788


ElijahLynn:
> @ElijahLynn , did you bring `file_public_base_url` settings back intentionally? I recall seeing an issue that was caused by explicitly setting this variable.

Ahh, no, good catch. I checked out the branch from the remote but forgot to do a `git fetch` so didn't have the revert. Fixing now.

Issue #2471, opened 2020-07-28
Unique title constraint for Q&A content type


oksana-c:
## User story

As a Content Editor, who adds and manages Q&A content, I would like to ensure I'm not creating duplicate Questions.

## Acceptance Criteria
- [ ]  `UniqueTitle` constraint is developed in `va_gov_backend` module
- [ ]  User-friendly error message is displayed for a scenario when the user enters a title/question that already exists - `Q&A with title ":title" already exists. <a href="/admin/content?title=&type=q_a&moderation_state=All&owner=All">Check existing Q&As before creating a new one.</a>` (@rachel-kauff to refine)
- [ ]  `UniqueTitle` constraint is applied to title field of the Q&A content type
- [ ]  the constraint is case-insensitive
- [ ] strip all punctuation before comparison

PR #2796, opened 2020-08-27
VACMS-2471: Adding q_a title constraint.

ethanteague:
## Description
See #2471 

## Testing done
Visual / form submit

## Screenshots
image

## QA steps

- As an admin, go to `node/add/q_a` and fill in required fields. In title field, enter `How do I get these benefits?` (already existing Q&A node title)
- Verify error appears with link to existing node.

### DIFF
--- a/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitle.php
+++ b/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitle.php
@@ -20,6 +20,6 @@ class UniqueTitle extends Constraint {
    *
    * @var \Drupal\va_gov_backend\Plugin\Validation\Constraint
    */
-  public $notUniqueTitle = 'Q&amp;A with title "%title" already exists. <a href="/node/:nid">Check the existing Q&A before creating a new one.</a>';
+  public $notUniqueTitle = 'Q&amp;A with title "%title" already exists. <a target="_blank" href="/node/:nid">Check the existing "%title" before creating a new one.</a>';
 
 }
--- a/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitle.php
+++ b/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitle.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Drupal\va_gov_backend\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Checks that the submitted value is a unique title.
+ *
+ * @Constraint(
+ *   id = "UniqueTitle",
+ *   label = @Translation("Unique Title", context = "Validation"),
+ *   type = "string"
+ * )
+ */
+class UniqueTitle extends Constraint {
+
+  /**
+   * The message that will be shown if the value is not a unique title.
+   *
+   * @var \Drupal\va_gov_backend\Plugin\Validation\Constraint
+   */
+  public $notUniqueTitle = 'Q&amp;A with title "%title" already exists. <a href="/node/:nid">Check the existing Q&A before creating a new one.</a>';
+
+}
--- a/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
+++ b/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
@@ -64,7 +64,7 @@ class UniqueTitleValidator extends ConstraintValidator {
     // Exclamation points, etc. are replaced with no space.
     // Underscores and double spaces are replaced with single space.
     $strip_punctuation_symbols = str_replace($remove_no_space, '', $string);
-    $strip_final = str_replace('  ', ' ', $strip_punctuation_symbols);
+    $strip_final = trim(str_replace('  ', ' ', $strip_punctuation_symbols));
     return $strip_final;
   }
 
--- a/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
+++ b/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
@@ -15,9 +15,10 @@ class UniqueTitleValidator extends ConstraintValidator {
    */
   public function validate($item, Constraint $constraint) {
 
+    // We will use this for comparison against results.
+    $original_comparator = $this->format($item->value);
     // Grab first word to use in search query.
-    $lower_case_value = strtolower($item->value);
-    $first_word = strtok($lower_case_value, ' ');
+    $first_word_comparator = strtok($original_comparator, ' ');
 
     // Per internal dev meeting, dependency injection doesn't
     // work well with constraints.
@@ -26,7 +27,7 @@ class UniqueTitleValidator extends ConstraintValidator {
     $node_storage = \Drupal::entityTypeManager()->getStorage('node');
     $query = $node_storage->getQuery();
     $query->condition('type', 'q_a');
-    $query->condition('title', $first_word . '%', 'LIKE');
+    $query->condition('title', $first_word_comparator . '%', 'LIKE');
     // Exclude the current entity.
     if (!empty($id = $this->context->getRoot()->getEntity()->id())) {
       $query->condition('nid', $id, '!=');
@@ -37,13 +38,46 @@ class UniqueTitleValidator extends ConstraintValidator {
     if (!empty($nodes)) {
       // See if we have any that match our field value.
       foreach ($nodes as $node) {
-        // Loose comparison just to be safe.
-        if ($lower_case_value == strtolower($node->get('title')->value)) {
-          $this->context->addViolation($constraint->notUniqueTitle, ['%title' => $item->value, ':nid' => $node->id()]);
+        // Compare our input title against db first word match.
+        if ($this->compare($original_comparator, $node->get('title')->value)) {
+          // Identical, so throw violation.
+          $this->context->addViolation($constraint->notUniqueTitle, ['%title' => $node->get('title')->value, ':nid' => $node->id()]);
         }
       }
     }
+  }
 
+  /**
+   * Returns formatted string for comparison.
+   *
+   * @return string
+   *   Returns a formatted string.
+   */
+  public function format($string) {
+    // Strip punctuation.
+    $strip_punctuation = trim(preg_replace('/[^a-z0-9]+/i', ' ', $string));
+    return strtolower($strip_punctuation);
+  }
+
+  /**
+   * Compare original with db result.
+   *
+   * @param string $a
+   *   The input.
+   * @param string $b
+   *   The db result.
+   *
+   * @return bool
+   *   Returns TRUE is strings identical.
+   */
+  public function compare($a, $b) {
+    $b_formatted = $this->format($b);
+    // Loose comparison just to be safe.
+    if ($a == $b_formatted) {
+      return TRUE;
+    }
+
+    return FALSE;
   }
 
 }
--- a/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
+++ b/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
@@ -62,9 +62,9 @@ class UniqueTitleValidator extends ConstraintValidator {
 
     // Some targeted replacements.
     // Exclamation points, etc. are replaced with no space.
-    // Underscores and double spaces are replaced with single space.
-    $strip_punctuation_symbols = str_replace($remove_no_space, '', $string);
-    $strip_final = trim(str_replace('  ', ' ', $strip_punctuation_symbols));
+    // Double spaces are replaced with single space.
+    $string_without_symbols = str_replace($remove_no_space, '', $string);
+    $strip_final = trim(str_replace('  ', ' ', $string_without_symbols));
     return $strip_final;
   }
 
--- a/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
+++ b/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
@@ -59,13 +59,12 @@ class UniqueTitleValidator extends ConstraintValidator {
     // We want to keep quotes and some other things,
     // So limit what we strip.
     $remove_no_space = ['?', '!', '.', '-'];
-    $remove_with_space = ['-', '  '];
 
     // Some targeted replacements.
     // Exclamation points, etc. are replaced with no space.
     // Underscores and double spaces are replaced with single space.
     $strip_punctuation_symbols = str_replace($remove_no_space, '', $string);
-    $strip_final = str_replace($remove_with_space, ' ', $strip_punctuation_symbols);
+    $strip_final = str_replace('  ', ' ', $strip_punctuation_symbols);
     return $strip_final;
   }
 
--- a/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
+++ b/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
@@ -65,7 +65,6 @@ class UniqueTitleValidator extends ConstraintValidator {
     // Underscores and double spaces are replaced with single space.
     $strip_punctuation_symbols = str_replace($remove_no_space, '', $string);
     $strip_final = trim(str_replace('  ', ' ', $strip_punctuation_symbols));
-    $strip_final = trim($strip_final);
     return $strip_final;
   }
 
--- a/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
+++ b/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
@@ -65,6 +65,7 @@ class UniqueTitleValidator extends ConstraintValidator {
     // Underscores and double spaces are replaced with single space.
     $strip_punctuation_symbols = str_replace($remove_no_space, '', $string);
     $strip_final = trim(str_replace('  ', ' ', $strip_punctuation_symbols));
+    $strip_final = trim($strip_final);
     return $strip_final;
   }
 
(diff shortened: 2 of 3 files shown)


docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitle.php
```diff
@@ -0,0 +1,25 @@
...
+ * )
+ */
+class UniqueTitle extends Constraint {
+
+  /**
+   * The message that will be shown if the value is not a unique title.
+   *
+   * @var \Drupal\va_gov_backend\Plugin\Validation\Constraint
+   */
+  public $notUniqueTitle = 'Q&A with title "%title" already exists. <a href="/node/:nid">Check the existing Q&A before creating a new one.</a>';
```
swirtSJW, line 23: This is html, should it be `Q&amp;A`?

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,37 @@
...
+   * {@inheritdoc}
+   */
+  public function validate($item, Constraint $constraint) {
+    // Per internal meeting dev meeting, dependency injection doesn't
+    // work well with constraints.
+    // Drupal contrib constraint modules call services directly.
+    // Okay to follow this pattern.
+    $query = \Drupal::entityQuery('node');
+    $query->condition('type', 'q_a');
+    $query->condition('title', $item->value);
```
swirtSJW, line 23: Should this be a 'like' condition so that it is case insensitive?
Otherwise we could end up with two Q and A's
"Are vets covered for COVID-19?"
"Are vets covered for covid-19?

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,37 @@
...
+    // Exclude the current entity.
+    if (!empty($id = $this->context->getRoot()->getEntity()->id())) {
+      $query->condition('nid', $id, '!=');
```
swirtSJW, line 26: Nice touch.

ethanteague: @swirtSJW I Only discovered realized this when I couldn't save an existing node on edit :laughing:

ethanteague:
Updated @swirtSJW @oksana-c :+1:

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,49 @@
...
+class UniqueTitleValidator extends ConstraintValidator {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validate($item, Constraint $constraint) {
+
+    // Grab first word to use in search query.
+    $lower_case_value = strtolower($item->value);
+    $first_word = strtok($lower_case_value, ' ');
```
swirtSJW, line 20: Discussion:  I like the approach of the search but we are going to have a predominance of "How", "What" , "When"   as start words.  If we going to arbitrarily pick a word and search for it, it might be more performant to choose the last word rather that the first.   The last  word will have much more variation than the first word on these QAs.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+    $nids = $query->execute();
+    $nodes = $node_storage->loadMultiple($nids);
+
+    if (!empty($nodes)) {
+      // See if we have any that match our field value.
+      foreach ($nodes as $node) {
+        // Loose comparison just to be safe.
+        if ($lower_case_value == strtolower($node->get('title')->value)) {
+          $this->context->addViolation($constraint->notUniqueTitle, ['%title' => $item->value, ':nid' => $node->id()]);
```
swirtSJW, line 42: Minor:  I think this help text is going to spit back what they already entered, which will already be visible in the field.   This should spit back the title `$node->get('title')->value` of the existing node that matches so it can be seen why it is a match.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,49 @@
...
+      $query->condition('nid', $id, '!=');
+    }
+    $nids = $query->execute();
+    $nodes = $node_storage->loadMultiple($nids);
+
+    if (!empty($nodes)) {
+      // See if we have any that match our field value.
+      foreach ($nodes as $node) {
+        // Loose comparison just to be safe.
+        if ($lower_case_value == strtolower($node->get('title')->value)) {
```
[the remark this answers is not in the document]
ethanteague: Yeah, I thought about that - I was thinking since we are doing a LIKE comparison it doesn’t make as much sense to strip punctuation since we aren’t as concerned about an exact match, and the only punctuation we might have to worry about in a title is a question mark or a period, neither of which will make a difference in a LIKE query. Whaddy’all think?

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
ethanteague: Here is where we could get into some trouble with punctuation. I like this idea, but can see a LIKE search becoming more complicated if we are keying off the last word - this is where we are more likely to have punctuation to worry about.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
ethanteague: Good point. I’ll change.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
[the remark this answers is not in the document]
ethanteague: Updated

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
oksana-c: Good point, @swirtSJW. Also, if we're showing what already exists, let's link that node title within the error message + `target="_blank"`

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -37,13 +38,46 @@ public function validate($item, Constraint $constraint) {
     if (!empty($nodes)) {
       // See if we have any that match our field value.
       foreach ($nodes as $node) {
-        // Loose comparison just to be safe.
-        if ($lower_case_value == strtolower($node->get('title')->value)) {
-          $this->context->addViolation($constraint->notUniqueTitle, ['%title' => $item->value, ':nid' => $node->id()]);
+        // Compare our input title against db first word match.
+        if ($this->compare($original_comparator, $node->get('title')->value)) {
+          // Identical, so throw violation.
+          $this->context->addViolation($constraint->notUniqueTitle, ['%title' => $node->get('title')->value, ':nid' => $node->id()]);
```
oksana-c, line 44: let's link the node title within the error message + target="_blank"

ethanteague: @oksana-c we have been discouraged in the past from opening links in a new window. I think it’s an accessibility thing - easy to make this change, just wanna make sure it’s ok.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
swirtSJW: Your search would not change.  Only the word you are searching for changes.
Typing in 
"How do I get paid?" would search for a LIKE match on 'paid' 
It would return things like:
"How to be paid for leave?"
"Can I get paid leave?"
"How do I get paid for benefits?
"This is how to get paid.
"How to get paid?"

Which is still going to be a smaller subset than searching and returning all the "Hows".

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,83 @@
...
+   *
+   * @param string $a
+   *   The input.
+   * @param string $b
+   *   The db result.
+   *
+   * @return bool
+   *   Returns TRUE is strings identical.
+   */
+  public function compare($a, $b) {
```
swirtSJW, line 73: Minor: "compare" is generic and does not speak to what it does.  A name like "isSame" lets you know more about what it is looking for and returning.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  /**
+   * Returns formatted string for comparison.
+   *
+   * @return string
+   *   Returns a formatted string.
+   */
+  public function format($string) {
```
swirtSJW, line 56: Minor:  This name I find confusing because it seems to me that it is removing formatting not adding it.  stripPunctuation seems to be more accurate.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,83 @@
...
+    // Strip punctuation.
+    $strip_punctuation = trim(preg_replace('/[^a-z0-9]+/i', ' ', $string));
```
swirtSJW, line 58: ?:   This seems like we need a multiple space remover. 
"How do I, get paid?"
would become "How do I  get paid"  (note double spaced after 'I') which would then not equate to being the same as "How do I get paid".

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,83 @@
...
+   * @param string $b
+   *   The db result.
+   *
+   * @return bool
+   *   Returns TRUE is strings identical.
+   */
+  public function compare($a, $b) {
+    $b_formatted = $this->format($b);
+    // Loose comparison just to be safe.
+    if ($a == $b_formatted) {
```
swirtSJW, line 76: Minor: https://www.php.net/manual/en/function.strcasecmp.php

If you use this, you can get rid of the call to strtolower in your format()

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
swirtSJW: I also think we want to allow apostrophes, as the difference between an apostrphe or not can be a completely different word.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
oksana-c: In the context of page view (not the form) we'd avoid target blank. In the context of a form though, we should use target blank for any links, just like we're doing with the Page title field Help text.

oksana-c: the scenario here would be:
The user is entering a question title which throws an error and shows a similar question that already exists. Instead of losing form data, the user is able to click on the link and open it in a different tab to verify that the existing question found indeed resembles what they started to enter in the form.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
ethanteague: Okay. Must be a bug somewhere else in the logic I had that was causing my last word LIKE search to miss. Will dig in more.

ethanteague: Ah - needs an addition of a wildcard on the _end_ to account for the punctuation mark.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
ethanteague: Arghhh - thought I had a good @swirtSJW sounding, verby kind of method when I plugged in `compare` :laughing:

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
ethanteague: I missed again on this one - was legit thinking, "Alright, this is a @swirtSJW sounding method name" :laughing:

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,93 @@
...
+   *
+   * @return string
+   *   Returns a formatted string.
+   */
+  public function stripPunctuation($string) {
+    // Strip targeted punctuation items.
+    // We want to keep quotes and some other things,
+    // So limit what we strip.
+    $remove_no_space = ['?', '!', '.', '-'];
+    $remove_with_space = ['-', '  '];
```
swirtSJW, line 62: super minor:  In the remove with no space, you are removing hyphens and replacing with nothing.   Then in the remove with space, you are replacing hyphens with spaces.... but there should be no hyphens left.

swirtSJW (dismissed):

swirtSJW:
This looks great Ethan.   Nice work.   Now I just need to figure out why Devshop lost the environment for 2796 completely.   
image

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
ethanteague: I don't know why I did that - gonna take out.

ethanteague: Removed, @swirtSJW

ethanteague:
> This looks great Ethan. Nice work. Now I just need to figure out why Devshop lost the environment for 2796 completely.
> image

Yeah, I have a couple/few instances that appear to be on the fritz. I tried to rebuild one earlier, but no luck, @swirtSJW

swirtSJW:
I did find an issue:
image

The way I bypassed the system was having a trailing space at the end of the title.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,92 @@
...
+    // We want to keep quotes and some other things,
+    // So limit what we strip.
+    $remove_no_space = ['?', '!', '.', '-'];
+
+    // Some targeted replacements.
+    // Exclamation points, etc. are replaced with no space.
+    // Underscores and double spaces are replaced with single space.
+    $strip_punctuation_symbols = str_replace($remove_no_space, '', $string);
+    $strip_final = str_replace('  ', ' ', $strip_punctuation_symbols);
+    return $strip_final;
```
swirtSJW, line 68:
```suggestion
    $strip_final = trim($strip_final);
    return $strip_final;
```

ethanteague:
> I did find an issue:
> image
> 
> The way I bypassed the system was having a trailing space at the end of the title.

I thought I had a trim in there, but I must have removed. Just added back in to address this issue (I had this happen when testing locally, a few days ago).
Thanks, @swirtSJW - updated.

swirtSJW:
I just added a commit to trim.   Will retest after it rebuilds.

swirtSJW:
woops... may have collided with each other.  LOL

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,93 @@
...
+    // We want to keep quotes and some other things,
+    // So limit what we strip.
+    $remove_no_space = ['?', '!', '.', '-'];
+
+    // Some targeted replacements.
+    // Exclamation points, etc. are replaced with no space.
+    // Underscores and double spaces are replaced with single space.
+    $strip_punctuation_symbols = str_replace($remove_no_space, '', $string);
+    $strip_final = trim(str_replace('  ', ' ', $strip_punctuation_symbols));
+    $strip_final = trim($strip_final);
```
swirtSJW, line 68:
```suggestion
```

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/UniqueTitleValidator.php
```diff
@@ -0,0 +1,92 @@
...
+    // Strip targeted punctuation items.
+    // We want to keep quotes and some other things,
+    // So limit what we strip.
+    $remove_no_space = ['?', '!', '.', '-'];
+
+    // Some targeted replacements.
+    // Exclamation points, etc. are replaced with no space.
+    // Underscores and double spaces are replaced with single space.
+    $strip_punctuation_symbols = str_replace($remove_no_space, '', $string);
+    $strip_final = trim(str_replace('  ', ' ', $strip_punctuation_symbols));
```
swirtSJW, line 67:
```suggestion
    // Double spaces are replaced with single space.
    $string_without_symbols = str_replace($remove_no_space, '', $string);
    $strip_final = trim(str_replace('  ', ' ', $string_without_symbols));
```

Issue #2388, opened 2020-07-23
Remove unused modules


swirtSJW:
## Description
Currently we have several modules that are not enabled and should just be removed from the codebase

- [ ] Acquia Connector
- [ ] Acquia Search
- [ ] Acquia Telemetry
- [ ] autosave_form
- [ ] lightning_contact_form  (contact form)
- [ ] Contact
- [ ] Contact Storage
- [ ] panels
- [ ] Moderation Dashboard

ethanteague:
Contact is a drupal core module, @swirtSJW - thinking you may have added that to the above list inadvertently.

ethanteague:
Increasing estimate on this one - a lot of the above are tied into lightning, and trying to remove is causing a daisy-chain of dependency errors.

PR #2808, opened 2020-08-29
VACMS-2388: Remove lightning and dependencies from composer and lock.

ethanteague:
In addition, add "composer/installers" to prevent autoloader errors occuring from loss of
autoload logic provided by lightning. See https://www.drupal.org/forum/support/installing-drupal/2019-02-14/php-fatal-error-cannot-redeclare-config_get_config

## Description
See #2388 

## Testing done
Removed all composer installed dirs: `vendor`, `core`, `libraries`, `modules/contrib`, ran `lando composer install` then `lando rebuild -y` then `./scripts/sync-db.sh` and confirmed successful rebuild. Then accessed local instance of site, visiting multiple nodes, user pages, and modules admin page. All looked good.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,7 @@
         "bower-asset/blazy": "^1.8",
         "bower-asset/cropper": "^4.1",
         "bower-asset/dropzone": "^5.5",
+        "composer/installers": "^1.2",
         "cweagans/composer-patches": "^1.6.4",
         "devshop/behat-drupal-extension": "dev-rewrite-last-page-output-links",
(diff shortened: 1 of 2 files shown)


ethanteague:
@swirtSJW updating the lock file is necessary after stripping out all the lightning / acquia stuff. The diff is huge, but I don't see any way around this - it's a one time thing to get us off the lightning crutch.

swirtSJW (dismissed):

indytechcook:
@ethanteague can you please resolve the conflicts! I'd love to get his merged in today.

indytechcook (approved):

Issue #2763, opened 2020-08-25
[R&S] Media list - videos component should allow to create new Videos


oksana-c:
## Description

Media list - videos component currently allows selection of existing  Video entities.

This component should also provide UI to create and attach new Video entity without leaving the form.

## Notes
- This was formerly blocked until Drupal 9 was released. We're not at Drupal 10, so this may now be possible.

## Acceptance Criteria
- [ ] Media list - Videos component provides UI for creating and attaching new Video entities without leaving the form in addition to selecting existing entities.

ethanteague:
@oksana-c this isn't going to be easy. https://www.drupal.org/node/3024762

ethanteague:
@oksana-c @kevwalsh what's the approach I should be taking on this task? Still need to figure out how to deal with this issue: https://www.drupal.org/node/3024762

ethanteague:
@oksana-c tried upgrading to drupal 8.9, but the issue remains - there isn't a way to add new media to the media browser.

oksana-c:
Drupal 9 will offer the "create inline" functionality. On hold until then.

swirtSJW:
Since this is labelled LC - Learning Center  I am moving this to Public Websites to establish if this is something that is still desired.

wesrowe:
@swirtSJW, how did it get on your radar this week? "Learning center" hasn't come up for us yet

wesrowe:
demoted to new/unref and "icebox" label because Outreach Materials Library, fka Learning Center, has been deprecated as of Fall 2022.

wesrowe:
@rlhecht, @DanielleThierryUSDSVA – Is the ability to add a video within the R&S Media List component something that would be high value for your team? If so, please share a little context as an input to our prioritization process.

FranECross:
@RLHecht, @DanielleThierryUSDSVA Please take a look at Wes's question above and let us know if this is still needed. I'll then check with Dave on prioritization if so. Thanks a bunch! 🎉 cc @jilladams 

PS: It looks like it was formerly blocked until Drupal 9 was released, and we're not at Drupal 10.

FranECross:
Closing old 2020 ticket; will revisit if it comes up again.

PR #2840, opened 2020-09-02
[WIP] VACMS-2763: create new video entities from Media list - Videos component

ethanteague:
## Description
See #2763 

## Testing done
Visual / Behat

## Screenshots
image
image

## QA steps
- As an admin go to `/node/add/media_list_videos` and visually verify that you see a video upload option. (Will require @oksana-c 's feedback)
- Add a youtube url and confirm save is successful


ethanteague:
@oksana-c will need to iterate on the upload experience. E.g., do we wanna add thumbnails to the fieldset, or do we wanna get even fancier and add it to the Media library field above, etc. Can be a big thing, or a small thing.

swirtSJW:
This seems to not agree with the AC from 2717

> Video media entities only allow YouTube as video provider. Vimeo and YouTube Playlist are removed from allowed options.

image

swirtSJW (changes requested):
Code looks fine.  Seems to be missing some of the AC on 2717

ethanteague:
> Code looks fine. Seems to be missing some of the AC on 2717

Thanks for catching that, @swirtSJW - updated.

swirtSJW:
image
 The language seems confusing to me even knowing the difference.
Suggesting we run this by UX team.
perhaps "Add media" button becomes "Choose from existing media"

swirtSJW (commented):
Code is all good.  But I think it should wait based on the language in the UX being confusing.

ethanteague:
@oksana-c I've created a separate pr to address the quick config stuff in 2717 (https://github.com/department-of-veterans-affairs/va.gov-cms/pull/2869) so that we can devote this pr to the heavy / deep-dive stuff in 2763.

Issue #2866, opened 2020-09-08
Devshpp install and deploy failures


swirtSJW:
![image](https://user-images.githubusercontent.com/5752113/92505899-a4e33000-f1d2-11ea-82a1-5693da60679f.png)

ElijahLynn:
I have these resolved now in #2874. Was a result of our Lightning removal. We have some more work to do still but I merged in enough to get CI working again.

PR #2874, opened 2020-09-08
VACMS-2866: Fix Yarn and consumer entity type missing in composer install

ElijahLynn:
## Description

See #2866.

This adds back the composer-plugin `oomphinc/composer-installers-extender` from the lightning-api that was removed in #2388.

STAGING didn't catch this because it actually does install DEV dependencies, it is only DevShop that has an initial step that does a `composer install --no-interaction --no-progress --no-dev --ansi` in the "install task" and then after that we have in tests.yml another one of `omposer install --dev --no-interaction --no-progress --ansi` which does get it, but we are failing before we get this. It was getting installed before as a dev dependency in lightning and normally CI would catch this but I think the PR that got this in had maybe got force pushed to once and it already had yarn installed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -126,6 +126,7 @@
         "michelf/php-markdown": "^1.8",
         "mouf/nodejs-installer": "^1.0",
         "npm-asset/yarn": "1.12.3",
+        "oomphinc/composer-installers-extender": "1.1.2",
         "phpunit/phpunit": "^7",
         "provision-ops/yaml-tests": "~1.9",
         "querypath/querypath": "^3.0",
--- a/composer.json
+++ b/composer.json
@@ -33,6 +33,7 @@
         "drupal/config_ignore": "^2.1",
         "drupal/config_override_warn": "^1.0",
         "drupal/config_split": "^1.4",
+        "drupal/consumers": "1.9",
         "drupal/content_lock": "^2.0@alpha",
         "drupal/core": "^8.8.8",
         "drupal/crop": "^2.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "e2338991981ce216160315c2a89ee394",
+    "content-hash": "adc1214f4b22cb1cc411237635f10ec3",
     "packages": [
         {
             "name": "alchemy/zippy",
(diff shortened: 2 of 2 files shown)


swirtSJW (commented):
Seems like a lot of changes here are unrelated to yarn.  Is that intentional?  Nevermind.  I see the note about it.

ethanteague (dismissed):

ethanteague:
> Seems like a lot of changes here are unrelated to yarn. Is that intentional?

Yeah, I think so, @swirtSJW - it's all that drush stuff that shouldn't have made it into the repo.

swirtSJW (approved):

ElijahLynn:
> > Seems like a lot of changes here are unrelated to yarn. Is that intentional?
> 
> Yeah, I think so, @swirtSJW - it's all that drush stuff that shouldn't have made it into the repo.

Yeah, I cheated with this PR since it was bugging me.

swirtSJW:
@ElijahLynn   Looks like this did not resolv the issue
image

ElijahLynn:
> > > Seems like a lot of changes here are unrelated to yarn. Is that intentional?
> > 
> > 
> > Yeah, I think so, @swirtSJW - it's all that drush stuff that shouldn't have made it into the repo.
> 
> Yeah, I cheated with this PR since it was bugging me, it seemed related enough but yeah, really should be a separate PR.

I moved this to https://github.com/department-of-veterans-affairs/va.gov-cms/pull/2876.

ElijahLynn:
> @ElijahLynn Looks like this did not resolv the issue
> image
> 
> Yarn!
> image
> 
> consumer
> image

Yeah, it wasn't that at all and instead was `oomphinc/composer-installers-extender` that went missing. `oomphinc/composer-installers-extender` is the composer-plugin that allows for specifying different installer types and folders etc. I initially installed it and kept getting an error but turns out that it needs to be the specific version I pinned in composer.json. Should pass now.

ElijahLynn:
Okay, so we have some more work to do. Some modules got removed but not uninstalled. Specifically consumers that is causing downstream errors with "consumer" entity type not found error. And if we look in the DB at the `key_value` table we see modules from https://git.drupalcode.org/project/lightning_api/-/blob/8.x-4.4/composer.json#L6-14 still there with schema's listed out. I have put back the "consumer" module for now so we can get functional PRs but we need more eyes on this and to follow the dependencies and probably will have to composer install these modules all back so that we can uninstall them properly and have their hooks clean up the DB. 

Here is a sample query:
> `select collection, name from key_value where name like '%consumer%';`

Issue #2851, opened 2020-09-04
Enable Maintenance mode


indytechcook:
When a deploy is occurring, a static html page should be show to users who are coming into the site during the deploy.  This page will allow the database to not be touched while the deploy is occurring. 

AC
* When a flag is set in the CMS, all pages should return a static HTML page.
* When a updb, config import is running then a static HTML page should be displayed
* Drush jobs should not run when deploy mode is enabled.
* Design and UI should reflect the site alert version, with these differences:
  * Page refresh every 30 seconds
  * No transparency

indytechcook:
This could follow a similar process as https://github.com/department-of-veterans-affairs/va.gov-cms/issues/1651

ElijahLynn:
Deploy Runbook (manual steps):

1. Go into existing instance (pre-deploy)
1. `sudo -u apache -s; cd /var/www/cms`
1. `git fetch`
1. `git checkout a8f64d577767d389aada2fd8886e785ffbbfe12f -- composer.json docroot/modules/custom/va_gov_backend/src docroot/sites/default/`
1. `source /etc/sysconfig/httpd; PATH=$PATH:/usr/local/bin composer dumpautoload`
1. Run Jenkins deploy job

PR #2878, opened 2020-09-09
Add static HTML deploy mode

indytechcook:
## Description

See  https://github.com/department-of-veterans-affairs/va.gov-cms/issues/2851.

## Testing done

Did a screenshare with @swirtSJW and @ElijahLynn.

## Definition of Done

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/2878)


docroot/modules/custom/va_gov_backend/src/Deploy/Plugin/CMSExport.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+namespace Drupal\va_gov_backend\Deploy\Plugin;
```
swirtSJW, line 3: Just noting this is not being used right now but will be in another PR.

docroot/sites/default/settings.php
```diff
@@ -169,6 +169,11 @@
   include $app_root . '/' . $site_path . '/settings/settings.fast_404.php';
 }
 
+// Deploy Settings
```
swirtSJW, line 172: I'd like to see some explanation here like...
```suggestion
// Jenkins moves this file into place during deploy, so if it is present we are in deploy mode.
```

docroot/sites/default/settings.php
```diff
@@ -184,5 +189,12 @@
   include $app_root . '/' . $site_path . '/settings/settings.local.php';
 }
 
+// The VA_GOV_IN_DEPLOY_MODE is set in settings.deploy.active.php.
```
swirtSJW, line 192: Could all this reside higher so that at lives with lines 173-176 so it reads nicer and is clearer?

docroot/sites/default/settings/settings.deploy.inactive.php
```diff
@@ -0,0 +1,3 @@
+<?php
+
```
swirtSJW, line 2:
```suggestion
// Ansible moves this file around to make deploy mode active or inactive.
```

swirtSJW (dismissed):
Just some minor code comment suggestions.   I really like this architecture Neil.  Well done.

docroot/sites/default/settings.php
indytechcook: I wanted it after the `settings.local.php` so I could test without actually renaming the file.

indytechcook:
Matching Devops PR: https://github.com/department-of-veterans-affairs/devops/pull/7433

swirtSJW (approved):

PR #2883, opened 2020-09-10
Fix issue with drush running during deploy

indytechcook:
Fixes as issue with drush running during deploy mode

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/2883)


indytechcook:
Close in favor of https://github.com/department-of-veterans-affairs/va.gov-cms/pull/2887

PR #2907, opened 2020-09-11
VACMS-000 Remove conflicting language on maintenance page.

swirtSJW:
## Description

The current language gives conflicting instructions.   This removes the conflict.
image

(scale of text did not change... that is just due to the screenshot sizes)
It will look like 
image

Cpde was also added to make the browser auto-refresh every 20 seconds to make this behave in a manner similar to the actual site alert screen.

### DIFF
--- a/docroot/sites/default/maintenance.html
+++ b/docroot/sites/default/maintenance.html
@@ -41,10 +41,11 @@
         font-size: 18px;
     }
 </style>
+<meta http-equiv="refresh" content="20" />
 </head>
 <body>
 <div class="site-alert">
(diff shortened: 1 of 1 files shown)


indytechcook:
This actually isn't accurate as the page will not autorefresh.

docroot/sites/default/maintenance.html
```diff
@@ -41,10 +41,11 @@
         font-size: 18px;
     }
 </style>
+<meta http-equiv="refresh" content="20" />
```
indytechcook, line 44: This works?

indytechcook (approved):
I tested the change locally by enabled site maintenance mode, getting the page.  Then I disabled maintenance mode and did not manually refresh.  The page refreshed automatically.

Issue #2880, opened 2020-09-10
Maintenance mode should allow CMS Export to download


indytechcook:
This is a follow up of https://github.com/department-of-veterans-affairs/va.gov-cms/issues/2851 to allow the CMS export tar to download if the site is in maintenance mode.

**AC**
* When maintenance mode is active and a cms export tar exists the response should be a redirect to the file.
* When maintenance mode is not active, a new file should be generated
* When maintenance mode is active and the cms export tar file does not exist, the resposee should be a 503.

indytechcook:
Need to also move the tar creation to the pre deploy task

indytechcook:
The idea is two fold.  First we have the cms export endpoint service a static version of the tar.  2. We export to the non-default directory and rsync files not overwriting newer files.  This does mean during a mass export new content will not be pushed out until the export is complete

PR #2964, opened 2020-09-22
When maintenace mode is enabled allow CMS content export tar to be available

indytechcook:
## Description

See #2880. 

## Testing done
* Enabled maintenance mode but added `putenv('VA_GOV_IN_DEPLOY_MODE=TRUE');` in `settings.local.php`
* Visited Home page and correct maintenance page was shown.
* Visited `/health` and correct `200` status was returned.
* Visited `/cms-export/content` and a `302` redirect was shown if the file exists.  Otherwise, the maintenance page with a `503` status is shown.

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/2964)


docroot/modules/custom/va_gov_backend/src/Deploy/Plugin/CMSExport.php
```diff
@@ -24,8 +37,85 @@ public function match(Request $request): bool {
...
    * {@inheritDoc}
    */
-  public function run(Request $request, string $app_root, string $site_path): ?bool {
-    $i = 1;
+  public function run(Request $request, string $app_root, string $site_path) {
+    $this->registerStreamWrapper();
+    try {
+      if ($this->fileExists()) {
+        $url = $this->getUrl();
+
```
swirtSJW, line 45: This looks good, it would be good to have a comment related to the "why"    
SOmething like.. "We are in deploy mode, but the file exists from a previous request, so return the existing (cached?) export because that is better than returning nothing."

swirtSJW (approved):

Issue #2825, opened 2020-09-01
Improve facility health service form for data call


kevwalsh:
## User Story or Problem Statement

As a PAO, i want good guidance on facility health service content

## Acceptance Criteria
- [ ] /node/add/health_care_local_health_service is ready for user guides / training and a data cull

## Implementation steps
<optional>

Issue #2504, opened 2020-07-30
Improve inline entity form on VAMC system health service node


kevwalsh:
Current experience when editing a VAMC system health service node (eg https://staging.cms.va.gov/node/6443/edit)

![Edit_VAMC_system_health_service_Social_work_at_VA_Wilkes-Barre_health_care___VA_gov_CMS.jpg](https://images.zenhubusercontent.com/5d6e57ad330f64311395f08b/e67e7b3d-5b97-411a-be16-8f617a39f238)

Clicking edit leads you to this screen
![Edit_VAMC_system_health_service_Social_work_at_VA_Wilkes-Barre_health_care___VA_gov_CMS.jpg](https://images.zenhubusercontent.com/5d6e57ad330f64311395f08b/0951f42f-f3a1-4bba-be95-ee87e67e2f11)

This experience will no longer make sense with new fields. We should either improve it or remove it altogether. 

If we remove it, it could be replaced with a content listing on the proofing screen (eg https://staging.cms.va.gov/node/6443) and/or on the Facility proofing screen (https://staging.cms.va.gov/node/2293).

![VA_Wilkes-Barre_health_care___Wilkes-Barre_VA_Medical_Center.jpg](https://images.zenhubusercontent.com/5d6e57ad330f64311395f08b/a6374869-d2d2-4843-825a-45a33701983c)

## Hypothesis

We should remove this altogether and possibly replace it with help text pointing to alternative paths.

PR #2965, opened 2020-09-22
VACMS-2825: Health service form improvements (2504 too).

kevwalsh:
## Description

See #2825 (mainly) but also #2504. 

## Testing done

Manual


kevwalsh:
this has been replaced by #2998

Issue #2975, opened 2020-09-23
Improve url aliases and node titles for health services


kevwalsh:
## User Story or Problem Statement

The URL alias and nodes for facility health services should indicate as much as possible about the content, including good breadcrumbs for users. 

## Acceptance Criteria
 * VAMC facility health services have their paths set to `[node:field_facility_location:entity:url:path]/[node:field_regional_health_service:entity:field_service_name_and_descripti]`
* VAMC system health services have their path set to `[node:field_region_page:entity:url:path]/health-services/[node:title]`
* An update hook in va_gov_db module  resaves all existing nodes of these types to use the new URL alias. this will have the added benefit of resaving some of the existing node titles, which are missing the their entity auto label setting. 
  * ideally, this would leave a revision log message, and be performed by "CMS bot" user

swirtSJW:
This will be resolved by https://github.com/department-of-veterans-affairs/va.gov-cms/issues/3287

PR #2994, opened 2020-09-24
VACMS-2975: Improve url aliases pattern for health services

acrollet:
## Description

See #2975.

## Testing done

Added phpunit test and ran locally.

## QA steps

Verify that Facility/System Health Service node path patterns are correct.
- [ ] Go to Configuration -> Search and Metadata -> URL aliases -> Patterns and verify that the patterns for VAMC facility & system health services match the required patterns in #2975 
- [ ] Validate that test items were removed.  

## Definition of Done


tests/phpunit/AliasesTest.php
```diff
@@ -0,0 +1,90 @@
...
+    // Create a VAMC facility local health care service node.
+    $local_service_node = $this->createNode([
+      'title' => 'Test Service at Test VA Medical Center',
+      'type' => 'health_care_local_health_service',
+      'field_facility_location' => $facility_node->id(),
+      'field_regional_health_service' => $service_node->id(),
+      'uid' => $author->id(),
+    ]);
+    $local_service_node->setPublished()->save();
+
```
swirtSJW, line 84: This looks great @acrollet.  Thank you for being thoughtful enough to add the test.   Just one request, can you delete these created nodes at the end of the test?   Right now all the front end dev and staging builds point to prod for their data, but we may go back to pointing them to staging, and if we did that, having test content remain published.  Behat tests remove what they create, but phpUnit tests don't.

swirtSJW (changes requested):
Nice work @acrollet.   Just one request. see in comments.

tests/phpunit/AliasesTest.php
acrollet: @swirtSJW thanks for the review - when I was running the test locally the nodes were deleted, but I'll go ahead and explicitly delete them :-)

acrollet: @swirtSJW as I look more, the documentation for ExistingSiteBase and the traits it's using is explicit that users, terms and nodes should be deleted at the end of the method. The one fly in the ointment is that if the test run fails mid-stream, any created entities would not be cleaned up. (This would also be true for manual deletion) I don't usually like relying on DB state for tests, but would it actually be better to test against existing content?

swirtSJW: Interesting find.  If they were actually deleted, I am not worried about them persisting if the tests fail.   That is not as critical.   Does make me wonder why some of our others are not being deleted.   I'll have to look into it.

acrollet: @swirtSJW good deal, just let me know if any changes are still needed?

swirtSJW:
Pausing on this as I am not sure this is the pattern we want
`va-eastern-colorado-health-care/system-health-services/recreation-and-creative-arts-therapy-at-va`

https://dsva.slack.com/archives/CT4GZBM8F/p1601053222076700

swirtSJW:
@acrollet when you have a chance can you updated these two QA steps to  be more specific?  What path should I go to, what should I expect to see?
```
- [ ] Facility Health Service node paths

 - [ ]  System Health Service node paths
```

acrollet:
@swirtSJW I've updated the QA steps.

kevwalsh:
> Pausing on this as I am not sure this is the pattern we want
> `va-eastern-colorado-health-care/system-health-services/recreation-and-creative-arts-therapy-at-va`
> 
> https://dsva.slack.com/archives/CT4GZBM8F/p1601053222076700

This looks like the old pattern that we are trying to move away from. perhaps the update hook was not successful? 

OLD pattern: '[node:field_administration:entity:name]/system-health-services/[node:title]'

NEW pattern: '[node:field_region_page:entity:url:path]/health-services/[node:title]'

`eastern-colorado-health-care/health-services/recreation-and-creative-arts-therapy-at-va-eastern-colorado-health-care` (if truncated a bit). 

There is a bit of repetition in there, but i think that's OK for now. These are not front end URLs, and there is some  minor concern about duplication without the duplicative `-at-va-eastern-colorado-health-care` at the end.

acrollet:
This is the URL I have for that node locally: ```/eastern-colorado-health-care/health-services/recreation-and-creative-arts-therapy-at-va-eastern```

@swirtSJW It appears that perhaps the update did not run successfully on your environment? It would be a bit chatty but I could perhaps add a log entry per node giving the alias before and after the node save?

swirtSJW:
> This is the URL I have for that node locally: `/eastern-colorado-health-care/health-services/recreation-and-creative-arts-therapy-at-va-eastern`
> 
> @swirtSJW It appears that perhaps the update did not run successfully on your environment? It would be a bit chatty but I could perhaps add a log entry per node giving the alias before and after the node save?

It did run them... that's where the screen capture in the QA steps came from.   It is possible we got a weird stay where devshop ran another build after I checked the hook_update ran., so that when I checked titles, they had not been updated yet.   I will check again on this recent build.

swirtSJW:
@acrollet It looks like either the pattern is not sticking... or the existing nodes are being saved but the path is not getting updated.
image
and I confirmed that node 7580 was updated in the hook_update
http://pr2994.ci.cms.va.gov/admin/reports/dblog/event/76

acrollet:
@swirtSJW ok, I'll pull an updated DB and try to repro locally.

acrollet:
I've updated this PR to remove the DB update so that another PR can operate with the updated config. (As discussed [on Slack](https://dsva.slack.com/archives/CT4GZBM8F/p1601053222076700))

swirtSJW (approved):

Issue #2910, opened 2020-09-14
Set text format to "Rich text limited" for Step paragraph field_wysiwyg


oksana-c:
## Acceptance Criteria
- [ ] default text format in `field_wysiwyg` within `step` paragraph is set to "Rich text limited" format and user cannot switch to any other text formats

PR #3000, opened 2020-09-25
VACMS-2910: enforce "Rich Text Limited" wysiwyg text format for step paragraph type

acrollet:
## Description

See #2910 

## Testing done

Created & ran behat test.

## QA steps

As a user with permission to create a "step by step" node
1. Create a "step by step" node
1. Confirm that text format selection is not allowed for the "Step" wysiwyg. (In the 'Add Step' section)
- [x] Text format selection is not allowed for steps
- [x] Available wysiwyg buttons are as expected for "Rich Text Limited" format

## Definition of Done


tests/behat/features/content_editing.feature
```diff
@@ -0,0 +1,14 @@
...
+  I need to have certain functionality available
+
+@content_editing
+  Scenario: Log in and confirm that "Step by step" node edit has the correct field settings
+    Given I am logged in as a user with the "administrator" role
+    And I am at "node/add/step_by_step"
+    Then I should see "Create Step-by-Step"
+    And I should see "Add Step"
+    # Confirm that text format selection is not allowed for the "Step" paragraph type.
+    And I should not see "Rich Text" in the "#edit-field-steps-0-subform-field-step-0-subform" element
```
swirtSJW, line 14: Great test.   Can we also assert that we should see  'Rich Text Limited'?

There is a bit of disconnect between ` I need to have certain functionality available` and a test that looks for what should not be there. ;)

acrollet: @swirtSJW hmm, excellent point, however Drupal doesn't show us the format in use when only one is available. We could perhaps try to confirm that a few of the appropriate WYSIWYG buttons are available and the text format select element is not present?

swirtSJW: oh right... with no choices it presents none.   Ok  Good enough reason.  ;)  Should we also validate plain text is not an option and therefore indirectly and loosely assert that only rich limited is present.

acrollet: @swirtSJW updated test pushed. I made a call that trying to look for the wysiwyg buttons with behat is likely to be quite time-consuming - lmk if you think it's worth investing the effort.

swirtSJW (approved):

Issue #3004, opened 2020-09-25
Make arguments work with lando behat/phpunit testing commands


acrollet:
## User Story or Problem Statement

As a developer, I need to be able to run individual tests with lando so that I can quickly validate new/updated tests.

## Acceptance Criteria
* lando behat tag arguments work as documented in [testing.md](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/READMES/testing.md)
- [ ] The ```lando behat --tags=perms``` command only runs tests tagged with "perms"

swirtSJW:
Interesting.  I don't know when this stopped working.  Not sure if it was with a lando update or behat update.

PR #3005, opened 2020-09-25
VACMS-3004: pass arguments to lando behat command

acrollet:
## Description

See #3004 

## QA steps
Validate Acceptance Criteria from issue:

- [ ]  The ```lando behat --tags=perms``` command only runs tests tagged with "perms"


swirtSJW (approved):
Thanks for fixing this.  I didn't know it was broken.

Tested locally on linux and this works.

PR #3012, opened 2020-09-28
VACMS-2975 (2): Re-save health services nodes to update titles/DB aliases

acrollet:
## Description

NB: This should not be merged until #3087  has been merged and is live.

See #2975.

## Testing done

Tested DB post update hook locally using new prod DB dump.

## QA steps

Verify that Facility/System Health Service node paths are correct after the database update has been completed.

- Go to the content listing, filter by Content Type: VAMC Facility Health Service, visit a few representative pages:
  - [ ] Validate that paths match the pattern "{Parent Facility node URL}/{Regional health service name & description}"
- Go to the content listing, filter by Content Type: VAMC System Health Service, visit a few representative pages:
  - [ ] Validate that paths match the pattern "{Parent Region node URL}/health-services/{node title}"
- [ ] validate that the hook_update_n completed 
image
- [ ] validate that  ~1400 nodes were updated via the hook_update_n
- [ ] Validate that updated nodes have proper revision message 
image

## Definition of Done


swirtSJW:
Holding on this one until after todays merge of part 1 lands on prod.

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -325,3 +325,69 @@ function va_gov_db_update_8018(&$sandbox) {
 
   return $messages;
 }
+
+/**
+ * Re-save all VAMC system & facility health service nodes.
+ */
+function va_gov_db_update_8019(&$sandbox) {
```
indytechcook, line 332: Can we switch to `hook_post_update_NAME` https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_post_update_NAME/8.8.x

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -325,3 +325,69 @@ function va_gov_db_update_8018(&$sandbox) {
...
+  }
+
+  // Run through a batch of 25.
+  $i = 0;
+  $nids = '';
+  foreach ($sandbox['nids_process'] as $revision => $nid) {
+    if ($i == 25) {
+      break;
+    }
+    $node = Node::load($nid);
```
indytechcook, line 357: The performance will be much better the nodes are bulk loaded

indytechcook: Using `array_splice` or something to the the nids to load load.

acrollet:
@indytechcook ready for re-review.

indytechcook (dismissed):

indytechcook:
Looks Great!!

swirtSJW:
This is still on hold as yesterday's merge of part 1  did not go out due not passing tests on staging. (not part 1's fault)

swirtSJW (approved):

swirtSJW:
This is cleared for merge as soon as tests pass.

swirtSJW:
@acrollet Looks like we may have an issue with this.  
image

swirtSJW (changes requested):
Titles seem to be failing.

swirtSJW:
Editing and saving one of the nodes with an empty title causes the title to be set correctly, so it is related to the atomated save.

acrollet:
pulling a new db and attempting to repro...

acrollet:
@swirtSJW After investigating, I at least have a workaround - changing the automatic label pattern from '[node:field_regional_health_service] - [node:field_facility_location]' to '[node:field_regional_health_service:entity:title] - [node:field_facility_location:entity:title]' lets the tokens be correctly replaced. We have a couple of options for dealing with this:

1. Update the auto-label patterns to explicitly specify the information we're pulling from the entity reference fields as above. This has the advantage of being clear about what data from the referenced entity is desired.  Unfortunately, I think this would take yet another deploy since the DB updates run before config import.
2. Attempt to repro this behavior in a d8 sandbox and raise an issue with a patch in the token module queue. (This is somewhat risky time-wise as it's not unknown whether this behavior can easily be reproduced with a base case or is due to interactions with contrib modules on the VA site) I did try adding a specific display mode for tokens, but that did not resolve the issue with pre-save.

swirtSJW:
@acrollet  We should do it right and update the pattern to make it more specific.  We deploy every day, so not a big deal to lose another day.

acrollet:
@swirtSJW sounds good, should we create a follow-on to look for and address this pattern elsewhere?

acrollet:
This will probably need to be re-rolled after #3087 goes live, assigning myself.

swirtSJW:
@kevwalsh The paths look right now, but the titles seem cumbersome.
Can you confirm that the titles are as they should be?
image

Is a facility service title supposed to have have a title that redundantly includes the system name?
image

swirtSJW:
It is like our concatenation of titles is cascading
image

kevwalsh:
We don't need the system name at the facility level. The green box in the last screenshot looks good (could be a "-" or an "at", i don't think it matters.)

eg 
Mental health care at Abie Abraham VA Clinic
or 
Mental health care - Abie Abraham VA Clinic

swirtSJW:
> We don't need the system name at the facility level. The green box in the last screenshot looks good (could be a "-" or an "at", i don't think it matters.)
> 
> eg
> Mental health care at Abie Abraham VA Clinic
> or
> Mental health care - Abie Abraham VA Clinic

The problem is, the system name is now part of the service name.  The service name does not exist by itself anymore.

kevwalsh:
We can still get to it, via a token, going through the VAMC system health service entity reference, then to the taxonomy term.

acrollet:
Closing in favor of https://github.com/department-of-veterans-affairs/va.gov-cms/pull/3154

swirtSJW (approved):
Looked at the results on devshop and the titles and paths look good.

Issue #2903, opened 2020-09-11
Move CMS Export deploy task to after post-deploy tasks (outside deploy mode) so deploys are faster


ElijahLynn:
From a slack conversation

>The idea is two fold.  First we have the cms export endpoint service a static version of the tar.  2. We export to the non-default directory and rsync files not overwriting newer files.  This does mean during a mass export new content will not be pushed out until the export is complete

_Originally posted by @indytechcook in https://github.com/department-of-veterans-affairs/va.gov-cms/issues/2880#issuecomment-691082140_

In addition: 
Currently the CMS Export that happens during deploy to accommodate for any schema and config changes happens inside the deploy window which increases the length of the deploy by 5-7 minutes right now. @indytechcook and I discussed moving this outside of the deploy window and that is fine as long as we specifically ensure that content that is updated during the CMS Export while the initial bulk export is in progress eventually makes it to the Export.

indytechcook:
CMS PR created to allow the export directory to be created: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/3014

indytechcook:
PR for devops change: https://github.com/department-of-veterans-affairs/devops/pull/7671

ElijahLynn:
Moving back to "in progress" as we discussed moving the tasks out into their own playbook in a Jenkins pipeline stage so it runs after the instance is back in service.

ElijahLynn:
https://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/READMES/devops/deploy-process.md

PR #3014, opened 2020-09-28
VACMS-2903 Allow export dir to be changed in the drush command

indytechcook:
## Description

See #2903. 

## Testing done
* Ran all cms export on with out the `export-dir` options: `lando drush va-gov-cms-export-all-content --process-count=8 --entity-count=500`
* Ran all cms export overriding the export directory: `lando drush va-gov-cms-export-all-content --process-count=8 --entity-count=500 --export-dir="public://cms-export-2"`
* Checked to see to see the new directory was created and contains contents
image

* Ran a diff on the two directories and they are the same.

image

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/3014)

### DIFF
--- a/docroot/sites/default/settings.php
+++ b/docroot/sites/default/settings.php
@@ -215,5 +215,8 @@ else {
 }
 $settings['file_public_base_url'] = "{$webhost}/sites/default/files";
 
+// Extend the limit of the tome cache table beyond 5000 rows.
+// See https://www.drupal.org/node/3162499.
+$settings['database_cache_max_rows']['bins']['tome_static'] = -1;
 $settings['tome_content_directory'] = 'public://cms-export-content';
 $settings['tome_files_directory'] = 'public://cms-export-files';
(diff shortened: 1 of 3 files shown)


docroot/sites/default/settings.php
```diff
@@ -215,5 +215,8 @@
 }
 $settings['file_public_base_url'] = "{$webhost}/sites/default/files";
 
+// Extend the limit of the tome cache table beyond 5000 rows.
+// See https://www.drupal.org/node/3162499.
+$settings['database_cache_max_rows']['bins']['tome_static'] = -1;
```
ElijahLynn, line 220: Per our call and your observation, we aren't using tome_static so we should removed this. 
```suggestion
```

ElijahLynn (approved):

swirtSJW (approved):

PR #3087, opened 2020-10-01
VACMS-2975: update system/facility health service node auto label settings

acrollet:
## Description

See #3012 - this PR updates the auto label settings to explicitly specify using the referenced entity labels.

## Testing done

Tested saving both CTs in browser and with new phpunit assertions.

## QA steps

- [x] Edit & save a facility health service node and confirm that the autogenerated title is correct
- [x] Edit & save a system health service node and confirm that the autogenerated title is correct

## Definition of Done


swirtSJW (approved):
Nice work adding the tests.

PR #3110, opened 2020-10-06
VACMS-12970: Add endpoint that outputs list of system menus for frontend consumption at build time.

ethanteague:
## Description
See https://github.com/department-of-veterans-affairs/va.gov-team/issues/12970

## Testing done
Visual

## QA
As an admin, go to `/graphql/explorer` and enter the following query: ```{
  sideNavMenus
}```
This is what you should see:
image


docroot/modules/custom/va_gov_graphql/src/Plugin/GraphQL/Fields/SideNavMenus.php
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function resolveValues($value, array $args, ResolveContext $context, ResolveInfo $info) {
+    $menus_array = menu_ui_get_menus(FALSE);
+    $systems_menus = [];
+    // Returns an array of drupal menu machine names.
+    foreach ($menus_array as $key => $value) {
+      if (strpos($key, 'va') !== FALSE || strpos($key, 'pittsburgh-health-care') !== FALSE) {
```
swirtSJW, line 31: I think we need to clarify with a comment why pittsburgh gets its own hardcoded exception because that is not clear from the code alone.

docroot/modules/custom/va_gov_graphql/src/Plugin/GraphQL/Fields/SideNavMenus.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Drupal\va_gov_graphql\Plugin\GraphQL\Fields;
+
+use Drupal\graphql\GraphQL\Execution\ResolveContext;
+use Drupal\graphql\Plugin\GraphQL\Fields\FieldPluginBase;
+use GraphQL\Type\Definition\ResolveInfo;
+
+/**
+ * Returns a list of the machine names of all the system menus on the site.
```
swirtSJW, line 10: Is this accurate?  I thought it is only returning health system menus.

Also your call of `menu_ui_get_menus(FALSE)` with the FALSE flag means it is excluding system menus and only doing custom menus.

swirtSJW (changes requested):
Nice work Ethan.  Just a couple of  documentation level things needed.

swirtSJW:
ALso needs a testing step about how to query this with the explorer and validate the results.

docroot/modules/custom/va_gov_graphql/src/Plugin/GraphQL/Fields/SideNavMenus.php
ethanteague: Yeah, it is only health system menus, @swirtSJW - I'll correct this :+1: 
Wrt setting that function to false, that is intentional. Health system menus will always be custom.

docroot/modules/custom/va_gov_graphql/src/Plugin/GraphQL/Fields/SideNavMenus.php
ethanteague: Good point - Pitt was added before we had an established `va-` prefix pattern, so it's an outlier. I'll note this, @swirtSJW :+1:

swirtSJW (approved):
Nicely done @ethanteague

PR #3120, opened 2020-10-06
VACMS-2811 add timezones to date fields

ethanteague:
## Description
See #2811 

## Testing done
Visual / Behat

## Screenshots
image
image
image

## QA steps
Visually verify that new smart time field (Date and time) appears, and is disabled on node and view forms for events, news releases, and situation update paragraphs.


kevwalsh (changes requested):
Great stuff.  Per zoom meeting, let's take care of these little details. 

Create_Event___VA_gov_CMS

ethanteague:
> Great stuff. Per zoom meeting, let's take care of these little details.
> 
> Create_Event___VA_gov_CMS

Updated, @kevwalsh

kevwalsh (dismissed):
i un-hid the new field and tested on http://pr3120.ci.cms.va.gov/node/add/event

That looks great. 

BUT: 

News release and Situation update should not have a daterange. they should be just single date fields. 

we could considering merging this as is, to unblock the FE as soon as possible for the event content type. But i think we'll need a different solution for news release and situation update.

ethanteague:
> i un-hid the new field and tested on http://pr3120.ci.cms.va.gov/node/add/event
> 
> That looks great.
> 
> BUT:
> 
> News release and Situation update should not have a daterange. they should be just single date fields.
> 
> we could considering merging this as is, to unblock the FE as soon as possible for the event content type. But i think we'll need a different solution for news release and situation update.

Just updated pr to only show the end date stuff on events, @kevwalsh . Here's what it looks like on a press release:
image

swirtSJW:
Needs conflict resolution

ethanteague:
> Needs conflict resolution

Updated, @swirtSJW :+1:

swirtSJW:
behat fail :(

ethanteague:
> behat fail :(

So, the failures look to be related to two terms that aren't on checklist - when the test tries to tick "Burials and memorials" or "Decision reviews and appeals," said terms aren't there. My pr isn't touching any term or checklist stuff, so I'm guessing this change occurred on prod, @swirtSJW @kevwalsh 
image
image

acrollet:
@ethanteague shoot, that must have been a recent change. We should probably refactor the test to just check the first three of those checkboxes. CC @kaise-lafrai I bet this is why the tests work locally for you.

swirtSJW (approved):

Issue #3131, opened 2020-10-07
Entity meta display should not appear on CMS help page


kevwalsh:
![CMS_help_center___Veterans_Affairs](https://user-images.githubusercontent.com/643678/95389958-99af1d00-08c2-11eb-94e6-bb9ff12f34cf.jpg)

ACs

- [ ] "Entity Meta Display" block is not displayed on the CMS Help Page content type nodes (currently displayed on all)

PR #3135, opened 2020-10-07
VACMS-3131: Hide EWA block on help pages.

kevwalsh:
## Description

See #3131 . 

## Testing done

Manual

## QA

1. As anon or any role, go to http://pr3135.ci.cms.va.gov/help, you shouldn't see the Entity meta blcok
1. go to any subpage, same should be true. 

CMS_help_center___Veterans_Affairs


swirtSJW:
This is only outdated based on  adocs commit, so I using the outdated test as good enough.

PR #2747, opened 2020-08-22
VACMS-2525: Menu toolbar improvements.

kevwalsh:
## Description

See #2525. 

## Testing done

Manual. 

# #2525 

### Before 
Disable_Node_Menu_Item_Settings___VA_gov_CMS

### After
kevin_walsh___VA_gov_CMS

## QA steps

Review as any user on PR environment AND locally

- [ ] Review string that shows the VACMS version for accuracy
- [ ] Review new Sections icon and all colors for accessibility
- [ ] To test prod environment indicator settings, overwrite lines 22-24 of settings.lando.php on your local with lines 22-24 of settings.prod.php and clear cache. 

## Notes

* This uninstalls moderation_sidebar. It offers nothing right now, but gets in the way of floating the Environment indicator to the right. We could probably remove it altogether. 
* on any screens that contain contextual links (like any node) there's an "edit" toolbar tab that appears on the top right. But Contextual links are only visible to admins right now, so i think this is fine. 

## Potential improvements

* Remove the Parentheses around VACMS v0.0.167, especially on prod. 

# #2660 CTA blocks

VAMC_guides_by_content_type___Veterans_Affairs

## QA steps

As admin, add the following classes to the new menu item attributes field for the top level menu items of the CMS Help Center block: 
* `icon cms-help-center`
* `icon helpdesk`
* `icon bug`
* `icon training`

Go to any page with the url `/help/*`, such as  /help/support/release-notes, and see these in the right rail. They should match the icons that appear in the in the CMS Help Center toolbar menu.

PR #3139, opened 2020-10-08
VACMS-2525 toolbar improvement C

swirtSJW:
## Description

This is a PR duplicate of #2747  just for the sake of devshop

## QA steps

1. Do this
1. Then that
- [ ] This
- [ ] That
- [ ] The other thing

## Definition of Done


swirtSJW:
Checks out on devshop 
image

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -830,20 +830,76 @@ function va_gov_backend_mail_alter(&$message) {
...
+  $environment = \Drupal::config('environment_indicator.indicator')->get('name');
   $branch = mb_strimwidth(exec('git rev-parse --abbrev-ref HEAD'), 0, 20, '...') ?? 'git unknown';
-  $tag_hash = trim(exec('git describe --tags')) ?? '--';
-  if ($branch === 'HEAD') {
-    // In BRD environment or checked out a tag/commit directly, show no branch.
-    $indicator = $tag_hash;
+  $commit_hash = trim(exec('git rev-parse --short HEAD'));
+  $git_tag = exec("git tag --points-at $commit_hash");
+  // Remove the 'cms/' from the tag that looks like 'cms/v0.0.190'.
+  $git_tag = str_replace('cms/v', '', $git_tag);
```
indytechcook, line 838: I'm a bit concerned this will not scale with a few hundred users at a time.  Can we create a follow up to set a drupal config during deploy?

swirtSJW: What does the number of users have to do with this?

indytechcook: Also, I have a feeling we won't have git available.  I've seen some security scans flag git being available as a red flag so I expect git to be removed in the future.

indytechcook: > What does the number of users have to do with this?

My bad on this as I didn't notice it was in `hook_rebuield`.

That being said, we shouldn't expect `git` to always being available.

Issue #1118, opened 2020-02-28
Content moderation views relationship error


swirtSJW:
**Describe the defect**
CONTENT MODERATION STATE VIEWS RELATIONSHIP
This installation contains one or more views which is using a relationship to the Content Moderation State entity. This relationship is deprecated and will be removed before 9.0.0. See this change record for information on removing this relationship or alternative solutions. 
https://www.drupal.org/node/3061099
Views that contain this relationship are: Moderation history, My Workflow, Right sidebar latest revision

**To Reproduce**
1. Go to /admin/reports/status
2.  See error

**Expected behavior**
No errors

**Screenshots**

![image](https://user-images.githubusercontent.com/5752113/75590243-e1c3b280-5a4a-11ea-9afb-f0d8621d23c0.png)

**Desktop (please complete the following information if relevant, or delete):**

**Additional context**

## Labels
- [/] Issue type (red) (defaults to "Defect")
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [/] CMS objective (orange) (not needed for bug tickets)
- [ ] CMS-supported product (black)

PR #3162, opened 2020-10-09
VACMS-1118 remove deprecated moderation relationship

acrollet:
## Description

NB: this PR depends on #3161 for a new behat condition and for clarity should not be reviewed until it has been merged.

See #1118 

## Testing done

Verified that updated views still show moderation states, verified that drupal status report issue is cleared, added behat tests for moderation views.

## QA steps

- [ ] Moderation sidebar view output is correct
- [ ] node/XXX/moderation-history output is correct


tests/behat/features/save_and_continue.feature
```diff
@@ -7,8 +7,14 @@ Feature: Save and continue button works as expected.
 @save_and_continue
   Scenario: Log in, edit, and save nodes with save and continue button and confirm revision saves changes.
     When I am logged in as a user with the "administrator" role
-    And I am at "node/add/checklist"
+
+    # Create beneficiaries term.
+    And I am at "admin/structure/taxonomy/manage/audience_beneficiaries/add"
+    And I fill in "Name" with "Awesome Veterans"
```
swirtSJW, line 13: Just in case devshop craps out in the middle of behat, it sometimes fails clean up fake content, which can confuse people in demo environments.  It would be nice if any fake content includes Behat in the title so we know where it came from.

acrollet: done! I also updated the similar test added in another branch.

swirtSJW (approved):

PR #3164, opened 2020-10-11
VACMS-000: Sync config with prod db.

ethanteague:
Db config code drift occurs organically over time. This pr remedies by bringing in differences
between prod db and config. This commit adds innocuous changes like quotes around strings, etc.
Syncing db with code is necessary to prevent developers from having to filter through code noise
when exporting config for commit.

Attn: @swirtSJW @kevwalsh @kaise-lafrai ^^^^^^^^^


swirtSJW (commented):
I just wanted to check on these three losses.  Do they seeim right?
image

ethanteague:
> I just wanted to check on these three losses. Do they seeim right?
> image

Yep. Holdovers from lightning that aren’t being used.

swirtSJW (approved):

PR #3179, opened 2020-10-14
VACMS-1299: fix missing labels on tablefield form

acrollet:
## Description

See #1299

## Testing done

Ran accessibility tests.

## Screenshots

Before:

Screen Shot 2020-10-14 at 3 37 31 PM

After:

Screen Shot 2020-10-14 at 3 38 16 PM

## QA steps

- [x] Verify that accessibility tests pass for /node/add/health_care_local_facility

PR #3181, opened 2020-10-14
VACMS-1301: fix aria allowed attributes

acrollet:
## Description

See #1301

this PR includes/depends on #3179

## Testing done

Ran accessibility tests

## Screenshots

Before:

Screen Shot 2020-10-14 at 4 48 59 PM

After:

Screen Shot 2020-10-14 at 4 49 31 PM

## QA steps

- [ ] Verify that individual address fields are required on /node/add/press_release
- [ ] Verify that accessibility tests pass


oksana-c (changes requested):
the Location field is no longer required if the country is not selected, so if the user omits the country field and submits the form - the location data will be missing.
Screen Shot 2020-10-16 at 8 15 56 AM

in the short-term, we can set "United States" as a default option for location country on this content type, so that other required fields are rendered on form load and play into the validation process.

acrollet:
@oksana-c thanks for surfacing that, I must have only looked at existing nodes when I was checking that.

acrollet:
@oksana-c I looked into contributing a patch for the root issue, but I'm not at all sure that consensus can be built for not allowing the required attribute on fieldset/details elements. In the meantime I've set the default country to 'US' and added a hook implementation to require the country field on press release nodes so that it cannot be unselected.

Screen Shot 2020-10-16 at 11 51 40 AM

acrollet:
@oksana-c ready for re-review when you have a chance :-)

swirtSJW (dismissed):
This all works for me.  Just waiting on hearing from @oksana-c if the solution of pres-setting the country works for her use case.

oksana-c (commented):
Screen Shot 2020-10-19 at 6 37 36 PM

expected behavior: the country is preselected to US and cannot be set to none and both address fields that appear after country selection are shown and marked as required.

Im not sure if devshop shows me something odd, but I dont see the two location fields along w/ the country. I may be missing something though

acrollet:
thanks for catching that @oksana-c! I hadn't seen it before but now seeing it on my local, browser cache perhaps.

acrollet:
Looks like I had missed exporting the default value setting for the location field - added that and a test to confirm that the fields are shown on the form.

acrollet:
I just confirmed that all fields are shown on the PR environment:

Screen Shot 2020-10-20 at 10 24 59 AM

acrollet:
@oksana-c all checks passed, this is ready for re-re-re-review ;-)

oksana-c (approved):
✅

Issue #3196, opened 2020-10-16
Q&A Answer Rich Text field refactoring


oksana-c:
## Description

Q&A Answer should be limited to 1 instance of Rich text field with a char limit of 1000 character including spaces.

## Acceptance Criteria

- [x] create new paragraph "Rich text - char limit 1000" `rich_text_character_limit` 
- [x] add existing `field_wysiwyg` to the paragraph | set to required | set cardinality to 1 | set form display config on the new paragraph to Text w/ counter w/ char limit of 1000 not including HTML markup
- [x] update `field_answer` on Q&A content type to allow reference to "Rich text - char limit 1000" only
- [x] write an update hook to move all existing values of `field_answer` Rich text paragraph into a new Rich text w/ counter paragraph
- [x] verify that all data is transferred correctly
- [x] update spec tool w/ new content model

oksana-c:
FE follow-up work for template refactoring https://github.com/department-of-veterans-affairs/va.gov-team/issues/15041

PR #3198, opened 2020-10-19
VACMS-3196: Q&A field_answer refactoring

oksana-c:
## Description

Part 1 of #3196 

**Goals**:

* allow only one instance of Rich text field to be attached as an answer on a node of Q&A content type
* allow maximum of 1000 characters per answer - current Rich text field allows unlimited chars
* editors should use Rich text limited text format for the answer - currently "Rich text" paragraph defaults to "Rich text" format with extended toolbar.

To accomplish these goals, a new paragraph with specific form display configuration needed to be created, which also requires a migration of existing paragraph value for field_wysiwyg to a new paragraph. See below

**Implementation:**

* created paragraph type `Rich text - char limit 1000` and configured form display mode to render rich text field w/ default (and only) text format - Rich Text Limited with a char counter - max 1000
* updated Q&A content type `field_answer` to allow only references to paragraph type `Rich text - char limit 1000` instead of `Rich text`
* switched `field_answer` cardinality from unlimited to single.

this change still keeps the data from the old paragraph that is attached to `field_answer` in tact, but non-editable. Edit by the user will throw an error on node validation.
**Part 2 of this effort is to run hook_update - ready in PR #3199**

We will need to deploy this config PR during regular deployment and then immediately follow up w/ an update hook PR. @swirtSJW , let's discuss and plan the sequence today at scrum 16th min.

## QA steps

- [ ] review several nodes of type Q&A and confirm that field "Answer" has an attached paragraph of type "Rich text - char limit 1000", the answer is populated w/ text and text format is set to "Rich Text Limited"
- [ ] confirm only one instance of Rich text paragraph can be added as an answer.
- [ ] review form /node/8225/edit and confirm that the field contains text that is 34 chars longer than set limit. This is expected, PW team will have to go in and edit all Q&As to adhere to max char standard.

## Definition of Done
- [ ] Content migration script has been run on production and any other applicable environments:
  ```
  # Run from Drupal's docroot:
  drush scr ../scripts/content/VACMS-3196-qa_rich_text_field_content_migration-2020-10.php
  ```

### DIFF
--- a/scripts/content/VACMS-3196-qa_rich_text_field_content_migration-2020-10.php
+++ b/scripts/content/VACMS-3196-qa_rich_text_field_content_migration-2020-10.php
@@ -1,0 +1,88 @@
+<?php
+/**
+ * @file VACMS-3196-qa_rich_text_field_content_migration-2020-10.php
+ *
+ * One-time migration for Q&A content type Answer field paragraphs.
+ */
+
+use Drupal\paragraphs\Entity\Paragraph;
+use Psr\Log\LogLevel;
+
+
+// Ensure that the new configuration is present.
+$field_answer_target = \Drupal::config('field.field.node.q_a.field_answer')->get('settings.handler_settings.target_bundles');
+if (!isset($field_answer_target['rich_text_char_limit_1000'])) {
+  print("The required paragraph configuration is not present, exiting!\n");
+  exit(1);
+}
+
+$updated = 0;
+$failed = 0;
+$failed_nids = [];
+
+$qas = \Drupal::entityTypeManager()->getStorage('node')->loadByProperties(['type' => 'q_a']);
+
+foreach ($qas as $qa) {
+  // Load the old paragraph instance if present.
+  $paragraph_old = $qa->get('field_answer')->first()->get('entity')->getTarget()->getValue();
+
+  // Proceed if the paragraph is present and not of the new type.
+  if ($paragraph_old instanceof Paragraph && $paragraph_old->bundle() == 'wysiwyg') {
+    $text_value = $paragraph_old->get('field_wysiwyg')->getValue()[0]['value'];
+
+    if (isset($text_value)) {
+      // Create a new paragraph instance with the previous paragraph's
+      // text and connect it to the Q&A node.
+      $paragraph_new = Paragraph::create([
+        'type' => 'rich_text_char_limit_1000',
+        'field_wysiwyg' => [
+          'value' => $text_value,
+          'format' => 'rich_text_limited',
+        ],
+      ]);
+      $paragraph_new->isNew();
+      $paragraph_new->save();
+
+      $new_answer = [
+        'target_id' => $paragraph_new->id(),
+        'target_revision_id' => $paragraph_new->getRevisionId(),
+      ];
+      $qa->set('field_answer', $new_answer);
+      $qa->setNewRevision(TRUE);
+      $qa->setRevisionCreationTime(\Drupal::time()->getRequestTime());
+      $qa->setRevisionUserId(1);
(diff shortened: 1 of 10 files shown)


swirtSJW (dismissed):

swirtSJW:
This looks good.  Waiting to merge based on need for discussion on deploy order.

acrollet:
@swirtSJW this PR is ready for review for potential merge/deployment today, thanks! :-)

scripts/content/VACMS-3196-qa_rich_text_field_content_migration-2020-10.php
```diff
@@ -0,0 +1,88 @@
...
+
+print(
+  t('Answer values for Q&A nodes were successfully migrated for @updated out of @count entities. @failed @failed_nids', [
+      '@count' => count($qas),
+      '@updated' => $updated,
+      '@failed' => $failed ? 'Failed to update ' . $failed . ' out of %count.' : NULL,
+      '@failed_nids' => count($failed_nids) ? 'Failed nids: ' . implode(', ', $failed_nids) . '.' : NULL,
+    ]
+  )
+);
```
indytechcook, line 88: Add a trailing line, this will file phpcs

indytechcook (approved):

scripts/content/VACMS-3196-qa_rich_text_field_content_migration-2020-10.php
acrollet: Ran phpcs and fixed a couple of other formatting issues too.

indytechcook (dismissed):
This code looks good.  I'll test on stage after it's merged.

scripts/content/VACMS-3196-qa_rich_text_field_content_migration-2020-10.php
acrollet: @indytechcook should we have a follow-on to add the scripts directory to the precommit hook?

scripts/content/VACMS-3196-qa_rich_text_field_content_migration-2020-10.php
```diff
@@ -0,0 +1,88 @@
...
+      $paragraph_new->save();
+
+      $new_answer = [
+        'target_id' => $paragraph_new->id(),
+        'target_revision_id' => $paragraph_new->getRevisionId(),
+      ];
+      $qa->set('field_answer', $new_answer);
+      $qa->setNewRevision(TRUE);
+      $qa->setRevisionCreationTime(\Drupal::time()->getRequestTime());
+      $qa->setRevisionUserId(1);
```
swirtSJW, line 54: We should be migrating as `1317` so it is logged as "CMS Migrator" rather than  user `1` = 'axcsd452ksey'

indytechcook, line 54:
```suggestion
      $qa->setRevisionUserId(1317);
```

indytechcook:
```suggestion
      $qa->setRevisionUserId(1317);
```

swirtSJW:
Log message looks good
image

swirtSJW:
Before and after seems to match
image

swirtSJW:
Test saved worked
image

swirtSJW:
The only deviation I find is this one on preview 
image

swirtSJW:
Logged updated count seems to match.
image

swirtSJW:
Messages were all truncated because of ampersand not being converted to html entity.   but no real issues with that other than cryptic message in log.  Not worth failing QA for this.
image

swirtSJW (approved):

PR #3199, opened 2020-10-19
VACMS-3196: Q&A field_answer refactoring - Part 2

oksana-c:
## Description

See #3198 . 

this is part 2 of #3198 and contains hook_update to migrate existing paragraph values into new paragraph

This PR needs to be rebased as soon as #3198 is merged.
without rebasing, this PR will not pass testing instruction since new config must exist for the hook to succeed.

## QA steps

- [ ] review several nodes of type Q&A and confirm that field "Answer" has an attached paragraph of type "Rich text - char limit 1000", the answer is populated w/ text and text format is set to "Rich Text Limited"
- [ ] confirm only one instance of Rich text paragraph can be added as an answer.
- [ ] review form /node/8225/edit and confirm that the field contains text that is 34 chars longer than set limit. This is expected, PW team will have to go in and edit all Q&As to adhere to max char standard.

### DIFF
--- a/config/sync/core.entity_form_display.paragraph.rich_text_char_limit_1000.default.yml
+++ b/config/sync/core.entity_form_display.paragraph.rich_text_char_limit_1000.default.yml
@@ -1,0 +1,34 @@
+uuid: 88a98cb8-98e1-4fbf-9816-94cca42258f9
+langcode: en
+status: true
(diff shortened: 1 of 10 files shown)


swirtSJW:
Closing this as #3198 has  it all.

Issue #2912, opened 2020-09-14
Force editors to save first as draft or in review, then publish on the proofing pages.


kevwalsh:
## User Story or Problem Statement

Users shouldn't be able to save as published in node edit forms.

## Acceptance Criteria
- [x] On a node edit form, "published" is not an available moderation state.
- [x] All content type forms, for which moderation is enabled, have moderation state widget at the top of the proofing page
- [x] release note to CMS users in Slack, email and CMS release log

## Implementation steps
- [ ] form alter for the moderation widget considering the context in which it is rendered - view or edit node page

Issue #2570, opened 2020-08-05
Database error: SQLSTATE[HY000] [14] unable to open database file


ElijahLynn:
This appears to be happen on Behat runs on BRD deploys, you can also see this in Drupal logs (dblog). Note, this doesn't actually break anything and is just the SQLite DB. 

Reproduce with `cd /var/www/cms; sudo -u apache bash -c "source /etc/sysconfig/httpd; PATH=$PATH:/usr/local/bin composer yaml-tests behat --dry-run 2>&1"` on DEV. 

I checked file permissions and directory permissions and `apache` has correct read on both. 

```
 ┃ And I press "Save"                                                                                                                                                                               
 ┃ # Drupal\DrupalExtension\Context\MinkContext::pressButton()                                                                                                                                      
 ┃                                                                                                                                                                                                  
 ┃ Then I should see "Benefits detail page VA blind and low vision rehabilitation services - EDITED has                  ┃ been updated."                                                           
                                                                                                                                                                                                    
 ┃ # Drupal\DrupalExtension\Context\MinkContext::assertPageContainsText()                                                                                                                           
 ┃                                                                                                                                                                                                  
 ┃ Aug 05 15:53:56 simplesamlphp ERROR [CL210e7a0c] Error loading session: Database error: SQLSTATE[HY000] [14] unable   ┃ to open database file                                                    
                                                                                                                                                                                                    
 ┃                                                                                                                                                                                                  
 ┃ Uncaught PHP Exception Exception: "Database error: SQLSTATE[HY000] [14] unable to open database file" at              ┃ /var/www/cms/docroot/vendor/simplesamlphp/simplesamlphp/lib/SimpleSAML/St
ore/SQL.php line 63                                                                                                                                                                                 
```

swirtSJW:
I'm not sure it is behat.  I am seeing these errors on prod.

ElijahLynn:
> I'm not sure it is behat. I am seeing these errors on prod.

Good observation and thanks for that.

ElijahLynn:
K, so this was a bit more involved than I thought. This is reproducible locally. 

Turns out this is stemming from an incorrect path, `sqlite:/../samlsessiondb.sq3`, being set for this file when Saml is being called from a GraphQL Sub Request when called from the Breadcrumb builder when being called from our addition to tome_sync_entity_update(). 

[...]

PR #3207, opened 2020-10-19
VACMS 2912: Force editors to save first as draft or in review, then publish on the proofing pages.

ndouglas:
## Description

See #2912.

I'm fairly new to this process, so any omissions/failures/errors/etc are unintentional.

## Testing done
Created and ran a Behat test.

## Screenshots

Edit Form:
Screen Shot 2020-10-19 at 3 00 23 PM

Proofing (Such As Is):
Screen Shot 2020-10-19 at 3 00 48 PM

## QA steps
As user with permission to create a node of a type with moderation enabled, e.g. `press_release`.
1. Open a form to create a new node of that type.
2. Confirm that "Published" is not an option for the Moderation State widget.
3. Save the node as a draft or for review.
4. Confirm that "Published" is now an option for the Moderation State widget.

## Definition of Done
- [ ] Release notes are forthcoming.

### DIFF
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -296,6 +296,12 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
     }
 
   }
+
+  // Add after_build callback for Press Release node forms.
+  if ($form_id === 'node_press_release_form' || $form_id === 'node_press_release_edit_form') {
+    $form['field_address']['widget']['#after_build'][] = 'va_gov_backend_press_release_address_after_build';
+  }
+
 }
 
 /**
@@ -303,14 +309,9 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
  */
 function va_gov_backend_field_widget_moderation_state_default_form_alter(&$element, FormStateInterface $form_state, $context) {
   // Remove "Published" from available moderation states.
-  $baseFormId = $form_state->getBuildInfo()['base_form_id'];
-  if ($baseFormId === 'node_form') {
+  $base_form_id = $form_state->getBuildInfo()['base_form_id'];
+  if ($base_form_id === 'node_form') {
     unset($element['state']['#options']['published']);
-  }
-
-  // Add after_build callback for Press Release node forms.
-  if ($form_id === 'node_press_release_form' || $form_id === 'node_press_release_edit_form') {
-    $form['field_address']['widget']['#after_build'][] = 'va_gov_backend_press_release_address_after_build';
   }
 }
 
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -308,7 +308,7 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
  * Implements hook_field_widget_WIDGET_TYPE_form_alter() for moderation_state widgets.
  */
 function va_gov_backend_field_widget_moderation_state_default_form_alter(&$element, FormStateInterface $form_state, $context) {
-  // Remove "Published" from available moderation states.
+  // Permit publication only from the proofing page, not the edit form.
   $base_form_id = $form_state->getBuildInfo()['base_form_id'];
   if ($base_form_id === 'node_form') {
     unset($element['state']['#options']['published']);
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -294,6 +294,18 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
     if (!in_array($node_type, $types_in_workflow)) {
       unset($form['actions']['preview']);
     }
+
+  }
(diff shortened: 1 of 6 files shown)


tests/behat/features/content_editing.feature
```diff
@@ -103,3 +103,18 @@ Feature: CMS Users may effectively create & edit content
...
+
+    # Confirm that the workflow state selector is present for the "press release" content type.
+    And the "#edit-moderation-state-0-state" element should exist
+
+    # Verify that the workflow state selector contains the "Draft" and "In review" options.
+    And I should see "Draft" in the "#edit-moderation-state-0-state" element
+    And I should see "In review" in the "#edit-moderation-state-0-state" element
+
+    # Verify that the workflow state selector does not contain the "published" option.
+    And I should not see "Published" in the "#edit-moderation-state-0-state" element
```
acrollet, line 120: @ndouglas nice work, thanks for adding the test! It would make sense to me for this test to run in a less privileged role, probably content admin. (I think I need to up my own game in that regard on a few existing tests...) It would also be nice to verify that we still see the "Published" option on the proofing page if there's not an existing test.

ndouglas: Sounds good.  I'll do both of those tomorrow.

ndouglas:
The tests in question pass locally but the command fails with an error akin to that in #2570, which @acrollet suggested was causing this issue.  I suspect that this might be the reason this test is failing in CI, since I'm not seeing an explicit error message associated with the test in the [Details](https://github.com/department-of-veterans-affairs/va.gov-cms/commit/be68f5f2d0fcdf8ca0d91c75de2ef85c3745a3eb#commitcomment-43424888).  I'm not sure of the best course of action here or if I'm diagnosing the issue correctly.

ndouglas:
This change thoroughly breaks [this test](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/298818aae49ebe5a9060e7ceced37bae1356b77c/tests/behat/features/perms.feature#L66).  I'll restructure that slightly to incorporate this change.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -294,12 +294,25 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
...
     $form['field_address']['widget']['#after_build'][] = 'va_gov_backend_press_release_address_after_build';
   }
+
+}
+
+/**
+ * Implements hook_field_widget_WIDGET_TYPE_form_alter() for moderation_state widgets.
+ */
+function va_gov_backend_field_widget_moderation_state_default_form_alter(&$element, FormStateInterface $form_state, $context) {
+  // Remove "Published" from available moderation states.
```
acrollet, line 311: trying to learn from @indytechcook here, should this be a "why" comment rather than a "what"?

tests/behat/Drupal/FeatureContext.php
```diff
@@ -148,17 +148,17 @@ public function theElementShouldNotExist($element) {
   /**
    * Check that an html element exists.
    *
-   * @param string $element
+   * @param string $selector
    *   The css selector.
    *
    * @Then the :element element should exist
```
acrollet, line 154: nice semantic improvement here! Not sure if there's great wording available but should we also change to something like the following?

"@Then the :element-selector element should exist"

(I don't know off-hand if behat will choke on the hyphen or not)

acrollet (changes requested):
Nice work! Just a couple of very minor comments :-)

tests/behat/Drupal/FeatureContext.php
ndouglas: I can find out pretty easily 😎

ndouglas: Yeah, it chokes, but `:selector` should work.

acrollet (approved):
EGc8sVIXoAIvcf-

tests/behat/features/content_editing.feature
```diff
@@ -128,6 +128,11 @@ Feature: CMS Users may effectively create & edit content
     And I fill in "Name" with "Test Office - BeHaT 404"
     And I press "Save"
     Then I should see "VA.gov URL" in the "#block-entitymetadisplay" element
+
+    # (Re-)Publish the node.
```
ndouglas, line 132: Hey @acrollet, I modified this test slightly to conform to the new behavior (not being able to publish directly from the node edit form).  Test passes and I think it makes sense, but I wanted to specifically call your attention to it in case I'm not thinking clearly this morning.  Which is a definite possibility.

tests/behat/features/perms.feature
```diff
@@ -89,6 +89,12 @@ Feature: Permissions
     And I am viewing an <type> with the title <title>
     And the "#edit-new-state" element should exist
     And I should see "published" in the "#edit-new-state" element
+
+    And I select "Published" from "Change to"
```
ndouglas, line 93: I think you mentioned at some point that this test needed followthrough, i.e. I should confirm that the node can indeed be published from the proofing page.  That's the purpose of this addition.

tests/behat/features/content_editing.feature
acrollet: I'm not sure I'm thinking any more clearly but it looks god!

acrollet (approved):

Issue #1421, opened 2020-04-07
Remove field_intro_text from the page content type and remove update hook data saving logic


ethanteague:
## Description
Remove va_gov_backend_entity_presave logic relating to page / wysiwyg after #1163 is on prod.
The logic calls ` _va_gov_backend_sync_wysiwyg_fields` and it should be removed.

## Acceptance Criteria
- [ ]  Logic that fires function in entity presave is removed, and function that is fired is removed. ` _va_gov_backend_sync_wysiwyg_fields`
- [ ] `field_intro_text` is gone from page content type
- [ ] CMS announcement for editors of `page` content type to tell them that intro text is more limited now. 

## Labels
- [/] Issue type (red) (defaults to task)
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [ ] CMS objective (orange)
- [ ] CMS-supported product (black)

ethanteague:
Front end piece (graphQL/template/transformer): https://github.com/department-of-veterans-affairs/vets-website/pull/15902

PR #3213, opened 2020-10-19
VACMS-1421: remove wysiwyg field sync code

acrollet:
## Description

See #1421

## Testing done

Tested saving 'page' type nodes and ensured intro content is preserved and updates are visible.

## QA steps

As a user with the content admin role
- [ ] Ensure that changes to the intro field on Benefits Detail Page content are made as expected

## Definition of Done

### DIFF
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -1042,18 +1042,12 @@ function va_gov_backend_pathauto_alias_alter(&$alias, array &$context) {
 
 /**
  * Implements hook_entity_presave().
- *
- * @todo: Remove node page wysiwyg sync after VACMS-1163 is on prod.
  */
 function va_gov_backend_entity_presave(EntityInterface $entity) {
   $bundle_types = ['event', 'news_story', 'press_release', 'outreach_asset'];
 
   if ($entity->getEntityTypeId() === 'node' && in_array($entity->bundle(), $bundle_types)) {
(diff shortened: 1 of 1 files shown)


acrollet:
@ethanteague should we have a follow-on to remove ```field_intro_text_limited_html``` from the page content type? It seems that it's no longer in use, but I'm not clear on the history :-)

swirtSJW (commented):
I am pretty sure we are still stuck on this one.   I don't think we ever made the front end switch from the old field to the new.   @ethanteague  can you confirm?

Last I recall, your front end PR was rolled back because we needed to wait on Table of Contents being supported... Which I think it is now, but I think the front end is still using the old field not the new.

swirtSJW:
> field_intro_text_limited_html

is the one that should be used, but we never made the switch to it.

ethanteague:
> I am pretty sure we are still stuck on this one. I don't think we ever made the front end switch from the old field to the new. @ethanteague can you confirm?
> 
> Last I recall, your front end PR was rolled back because we needed to wait on Table of Contents being supported... Which I think it is now, but I think the front end is still using the old field not the new.

That's right, @swirtSJW | @acrollet . Front end is still using the old field.

swirtSJW:
@acrollet I am not sure why this has a status of closed.  The ticket needs to performed, just not right now.  I am going to conver this to draft to make sure it does not get merged accidentally.

acrollet:
@swirtSJW weird, I don't think I closed it - anyway 👍 on leaving it in draft

swirtSJW:
This has been removed by an earlier merge.

Issue #3234, opened 2020-10-20
VA Forms importer does not correctly handle whitespace characters in the link_name field


acrollet:
**Describe the defect**
If whitespace characters are present in the link_name field of the CSV, a blank URL will be entered into the VA Form URL field instead of the generated URL. This causes a WSOD on the associated form node view page.

**Expected behavior**
The generated URL should be used if the field contains only whitespace.

PR #3236, opened 2020-10-20
VACMS-3234: trim link_name url field on import

acrollet:
## Description

See #3234 

## Testing done

Tested import with link_name field:
- empty
- only containing whitespace
- containing a url
- containing a url with leading/trailing whitespace

## QA steps

1. Run the VA forms import
- [ ] Confirm that the "VA Form 21P-4706b" node has a value for the "Link to form" field

### DIFF
--- a/config/sync/migrate_plus.migration.va_node_form.yml
+++ b/config/sync/migrate_plus.migration.va_node_form.yml
@@ -21,7 +21,7 @@ source:
     - 'https://prod.cms.va.gov/sites/default/files/migrate_source/va_forms_data.csv'
   delimiter: ','
   enclosure: '"'
-  escape: ''
+  escape: '\'
   header_offset: 1
   multi_value_delimiter: ','
   fields:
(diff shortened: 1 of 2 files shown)


config/sync/migrate_plus.migration.va_node_form.yml
```diff
@@ -21,7 +21,7 @@ source:
     - 'https://prod.cms.va.gov/sites/default/files/migrate_source/va_forms_data.csv'
   delimiter: ','
   enclosure: '"'
-  escape: ''
+  escape: '\'
```
acrollet, line 24: My local installation complained when this was empty, \ is the default for this field.

swirtSJW:
I just added the config changes to the va_gov_migrate/config/install.   We make changes there then run `lando migrate-sync` to make it so that the install and config/sync are in sync.   The main reason for wanting the config install, is that it has comments that get wiped out by config export.

https://dsva.slack.com/archives/CT4GZBM8F/p1603230114054700?thread_ts=1603220631.039200&cid=CT4GZBM8F

acrollet (commented):
The synced changes look good to me.

swirtSJW:
Tested locally with `lando mim va_node_form --update"  and verified that existing nodes and the troublesome form of http://va-gov-cms.lndo.site/find-forms/about-form-21p-4706b  resolved the missing url. 

image

swirtSJW (approved):

PR #3243, opened 2020-10-21
VACMS-000 Bump vetswebsite.

swirtSJW:
## Description

See _issueid_.


indytechcook (approved):
Lots of random stuff in here but it's all just metadata for packages.

Issue #2861, opened 2020-09-04
EWA: Entity meta block follow-up


oksana-c:
##  Acceptance Criteria

- [ ] "Entity Meta Display" block is displayed on node revision pages, e.g. /node/2/revisions/23101/view
- [ ] "Entity Meta Display" block displays "Va.gov URL" data based on the following rules
   - [ ] Va.gov URL is taken from latest published revision. Otherwise not displayed at all
       - [ ] logic for VA.gov URL display - if  ~env = prod and~ node has a published revision and node doesn't have an archived revision (archived rev ID is higher than latest published rev ID) -> display the URL
      - [ ] ~if latest published revision ID is greater than currently viewed revision ID AND URL alias of currently viewed revision is equal to latest published revision URL alias value, then Va.gov URL should be displayed using latest published revision alias. Otherwise, VA.gov URL should not be displayed, because it's no longer accurate.~ 
      - [ ] use the alias we have (since it's not revisionable)
      - [ ] if there's any published revision, check if URL exists on prod [200 OK], if it doesn't - gray it out and add "pending" label to it. Colors used to style inactive link are passing contrast test 

## Implementation

* `docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php`

kevwalsh:
AC moved to a separate issue, #3100,  in the icebox

- [ ]  "Entity Meta Display" block displays "Owner" data based on currently viewed revision.

kevwalsh:
Splitting this out because it's higher priority (actually, becoming a defect) in #3131

- [ ] "Entity Meta Display" block is not displayed on the CMS Help Page content type nodes (currently displayed on all)

acrollet:
@oksana-c regarding this AC:

>  if latest published revision ID is greater than currently viewed revision ID AND URL alias of currently viewed revision is equal to latest published revision URL alias value, then Va.gov URL should be displayed using latest published revision alias. Otherwise, VA.gov URL should not be displayed, because it's no longer accurate.

In my testing as of now, only one path alias is stored for a given node and all its revisions, whether using automatically generated or manually created alias. We could work around this for automatically aliased nodes by manually generating the alias for the revision in question and comparing against the live alias, but this wouldn't catch the case when the alias pattern has been updated. I'd suggest that we not display the va.gov URL when viewing any revision older than the currently published revision?

acrollet:
UX/UI in issue mini-spike:

[...]

PR #3255, opened 2020-10-21
VACMS-2861: entity meta display block updates

acrollet:
## Description

See #2861 

## Testing done

Added behat test, manually tested.

## Screenshots

Screen Shot 2020-10-21 at 3 21 45 PM

## QA steps

As a user with the content admin role:
1. Create a node without publishing it
1. [x] verify that the va.gov URL is not shown in the block
1. Publish the node
1. [x] verify that the va.gov URL is shown in the block, non-clickable with (Pending) appended
1. Archive the node
1. [x] verify that the va.gov URL is not shown in the block
1. Visit a node that has been previously published on production
1. [x] verify that the va.gov URL shows in the block and is clickable
1. Visit a revision page for a node
1. [x] verify that the block is shown

## Definition of Done


docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -75,28 +85,111 @@ public static function create(ContainerInterface $container, array $configuratio
...
     $block_items = [];
     $sections = $this->getSections()['links'];
     $block_items['Owner'] = $sections;
-    $nid = '';
+    $node = NULL;
+    $node_revision = NULL;
+    $display_url = TRUE;
+
+    // Drupal sometimes hands us a nid and sometimes an upcasted node object.
+    // This code should be future-proof, cf:
```
indytechcook, line 97:
```suggestion
    @TODO remove type checks when the patch is committed which looks to be 9.2.
    // This code should be future-proof, cf:
```

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -75,28 +85,111 @@ public static function create(ContainerInterface $container, array $configuratio
...
+    // https://www.drupal.org/project/drupal/issues/2730631
     if ($this->routeMatch->getParameter('node') instanceof NodeInterface) {
       $node = $this->routeMatch->getParameter('node');
+    }
+    elseif (is_numeric($this->routeMatch->getParameter('node'))) {
+      $node = \Drupal::entityTypeManager()->getStorage('node')->load($this->routeMatch->getParameter('node'));
+    }
+    if ($this->routeMatch->getParameter('node_revision') instanceof NodeInterface) {
```
indytechcook, line 105:
```suggestion

    if ($this->routeMatch->getParameter('node_revision') instanceof NodeInterface) {
```

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -75,28 +85,111 @@ public static function create(ContainerInterface $container, array $configuratio
...
+
+        // Do not show the URL if there is an archived revision that is newer
+        // than the most recent published revision.
+        if ($latest_published_revision_id && $latest_archived_revision_id && ($latest_archived_revision_id > $latest_published_revision_id)) {
+          $display_url = FALSE;
         }
       }
 
+      // The va.gov URL should be taken from the most recent published revision.
+      if ($display_url) {
```
indytechcook, line 160: Split this section out into it's own service name `VAGovStatus` or something.

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -75,28 +85,111 @@ public static function create(ContainerInterface $container, array $configuratio
    * {@inheritdoc}
    */
   public function build() {
+    $block = [];
     $block_items = [];
     $sections = $this->getSections()['links'];
     $block_items['Owner'] = $sections;
-    $nid = '';
+    $node = NULL;
```
indytechcook, line 92: Af first I had thought about spliting the logic to get a `node` and `node_revision` from a route out into it's own service but after more thought and the temporary nature of the code, let's put it into it's own method so it's easier to replace later.

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -75,28 +85,111 @@ public static function create(ContainerInterface $container, array $configuratio
...
+          $display_url = FALSE;
+        }
+      }
+
+      // If there is an archived revision with an ID higher than the latest
+      // published revision ID, do not attempt to show the URL.
+      //
+      // First, get the latest published revision ID.
+      if ($display_url) {
+        $query = \Drupal::database()->select('content_moderation_state_field_revision', 'cmr')
```
indytechcook, line 128: Inject the database object instead.

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -75,28 +85,111 @@ public static function create(ContainerInterface $container, array $configuratio
...
+      $node = \Drupal::entityTypeManager()->getStorage('node')->load($this->routeMatch->getParameter('node'));
+    }
+    if ($this->routeMatch->getParameter('node_revision') instanceof NodeInterface) {
+      $node_revision = $this->routeMatch->getParameter('node_revision');
+    }
+    elseif (is_numeric($this->routeMatch->getParameter('node_revision'))) {
+      $node_revision = \Drupal::entityTypeManager()->getStorage('node')->loadRevision($this->routeMatch->getParameter('node_revision'));
+    }
+
+    if ($node) {
```
indytechcook, line 112: This is another section which could be broken into it's own service.

tests/behat/features/content_editing.feature
```diff
@@ -97,3 +97,37 @@ Feature: CMS Users may effectively create & edit content
     Then I visit the "edit" page for a node with the title "Test Office - BeHaT"
     And the "menu[link_enabled]" checkbox should be checked
 
+@content_editing
```
indytechcook, line 100: YEAH FOR TESTS!!!  You are making the world a better place!

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -75,28 +85,111 @@ public static function create(ContainerInterface $container, array $configuratio
...
+        $exclude_types = $this->configFactory->getEditable('exclusion_types_admin.settings')->get('types_to_exclude');
+        if (in_array($node->bundle(), $exclude_types)) {
+          $display_url = FALSE;
+        }
+      }
+
+      // If there is an archived revision with an ID higher than the latest
+      // published revision ID, do not attempt to show the URL.
+      //
+      // First, get the latest published revision ID.
```
indytechcook, line 126: Comments should talk about why, let the code say what it's doing.

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -75,28 +85,111 @@ public static function create(ContainerInterface $container, array $configuratio
...
+          ->range(0, 1);
+        $result = $query->execute();
+        $latest_archived_revision_id = $result->fetchField();
+
+        // Do not show the URL if no revision has been published.
+        if (!$latest_published_revision_id) {
+          $display_url = FALSE;
+        }
+
+        // Do not show the URL if there is an archived revision that is newer
```
indytechcook, line 152: Here is another "what" comment.  If you are adding comments like this to code, ask yourself how can I make the code speak for itself.

indytechcook (changes requested):
Overall the general logic is good.  The code needs to be organized into smaller services.

For example, I see the following services:

* `WorkflowAssignmentConfigService`: In charge of getting different configurations.
* `NodeWorkflowQueryManager`: In charge of finding information about the node/revision in relation to a workflow/content moderation.
* `VAGovService`: In charge of communicating with the published content site.

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -75,28 +85,111 @@ public static function create(ContainerInterface $container, array $configuratio
...
+        // than the most recent published revision.
+        if ($latest_published_revision_id && $latest_archived_revision_id && ($latest_archived_revision_id > $latest_published_revision_id)) {
+          $display_url = FALSE;
         }
       }
 
+      // The va.gov URL should be taken from the most recent published revision.
+      if ($display_url) {
+        // Generate the va.gov URL.
+        $url = 'https://va.gov' . $node->toUrl()->toString();
```
indytechcook, line 162: Add some try/catch logic :)

```
The website encountered an unexpected error. Please try again later.

GuzzleHttp\Exception\ConnectException: cURL error 7: Failed to connect to va.gov port 443: Connection refused (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) in GuzzleHttp\Handler\CurlFactory::createRejection() (line 200 of vendor/guzzlehttp/guzzle/src/Handler/CurlFactory.php).
```

acrollet:
@indytechcook I still have work to do on writing services for the revision moderation statuses and va.gov URL status checking, but if you have a moment to glance over the changes so far I'd appreciate it. I'm just wanting to make sure the direction is correct and get feedback on the structure and naming for the ExclusionTypes service. Thanks!

docroot/modules/custom/va_gov_backend/src/ExclusionTypes.php
```diff
@@ -0,0 +1,70 @@
...
+   *   The configuration factory.
+   */
+  public function __construct(ConfigFactoryInterface $configFactory) {
+    $this->configFactory = $configFactory;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public static function create(ContainerInterface $container) {
```
indytechcook, line 33: You don't need this with a service.  The DI container will do it for you based upon what's definedi n the `servcies.yml` file.

indytechcook:
@acrollet this looks great so far.  The separation into different methods makes the code much easier read.  The service looks good.  I like that the service is specific to the exclude types rather than all config like I initially suggested.  The `typeIsExcluded` method is great.

I really like how this is progressing.  These two lines only make the code so much more readable

```
$node_revision = $this->getNodeRevision();
```

```
if ($this->vaGovUrlShouldBeDisplayed($node)) {
```

docroot/modules/custom/va_gov_backend/src/ExclusionTypesInterface.php
```diff
@@ -0,0 +1,34 @@
...
+   */
+  public function getExcludedTypes() : array;
+
+  /**
+   * Get Json array of excluded types.
+   *
+   * @return string
+   *   Json array.
+   */
+  public function getJson() : string;
```
indytechcook, line 24: What's this going todo? I didn't see it used yet.

acrollet: @indytechcook this is a stub per @swirtSJW, cf. https://dsva.slack.com/archives/CT4GZBM8F/p1603742611304500?thread_ts=1603734185.284100&cid=CT4GZBM8F

acrollet: @indytechcook thanks for the feedback!

indytechcook: Very cool!  Thanks for the link

acrollet:
@indytechcook this is ready for another round of feedback :-) Also if you have a little time tomorrow (also @ndouglas if you're interested) I'd love to discuss how best to mock the responses for the VaGovUrl service in behat tests - I haven't come up with an approach I love.

acrollet:
ok @indytechcook @ndouglas @ElijahLynn you can see the approach I came up for mocking the va.gov responses during behat testing here: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/3255/commits/3f9866ae436fdfb2e76611031760e24624d5284c

I don't completely love the approach, I would have preferred to keep the testing code within behat, but could not find an approach that would allow overriding a drupal service without creating a module. (Adding a services file using the container_yamls approach in settings.php would not work) Symfony has a neat way of doing [conditional arguments for services](https://symfony.com/doc/current/service_container/expression_language.html), but it's [not supported in Drupal](https://www.drupal.org/project/drupal/issues/2810587).

Feedback welcome :-)

docroot/modules/custom/va_gov_backend/modules/va_gov_backend_http_client/va_gov_backend_http_client.info.yml
```diff
@@ -0,0 +1,7 @@
+name: VA.gov Backend Http Client
+type: module
+description: Mocked HTTP Client for va_gov_backend VaGovUrl service
+core_version_requirement: ^8 || ^9
+package: 'Custom'
+dependencies:
+  - va_gov_backend
```
indytechcook, line 7: Add another line, this will fail phpcs.

docroot/modules/custom/va_gov_backend/modules/va_gov_backend_http_client/va_gov_backend_http_client.services.yml
```diff
@@ -0,0 +1,4 @@
+services:
+  va_gov_backend_http_client.http_client:
+    decorates: 'http_client'
+    class: Drupal\va_gov_backend_http_client\VaGovBackendHttpClient
```
indytechcook, line 4: Add another line, this will fail phpcs.

docroot/modules/custom/va_gov_backend/src/Form/ExclusionTypesAdminForm.php
```diff
@@ -51,16 +51,18 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
     }
+
     $form['types_to_exclude'] = [
       '#type' => 'checkboxes',
       '#options' => $options,
       '#title' => $this->t('Content types that are not individual pages on va.gov.'),
       '#description' => $this->t('Names of content types, that are not separate pages on va.gov'),
       '#weight' => '10',
-      '#default_value' => $config->get('types_to_exclude'),
+      '#default_value' => $config->get('types_to_exclude') ? $config->get('types_to_exclude') : [],
```
indytechcook, line 65:
```suggestion
      '#default_value' => $config->get('types_to_exclude') ?? [],
```

docroot/modules/custom/va_gov_backend/src/Service/VaGovUrl.php
```diff
@@ -0,0 +1,85 @@
...
+   *   The http client.
+   */
+  public function __construct(ClientInterface $httpClient) {
+    $this->httpClient = $httpClient;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function getVaGovUrlForEnvironment(String $environment) : string {
```
indytechcook, line 41: Allow this to be overridden by a `$settings` since the URL might be different for `demo` or `PR` environments.

docroot/modules/custom/va_gov_backend/src/Service/VaGovUrl.php
```diff
@@ -0,0 +1,85 @@
...
+        // Keep the timeout low so that we don't block page loads for too long.
+        $response = $this->httpClient->head($va_gov_url, ['connect_timeout' => 2, 'http_errors' => FALSE]);
+        return $response->getStatusCode();
+      }
+      catch (RequestException $e) {
+        if ($e->hasResponse()) {
+          return $e->getResponse->getStatusCode();
+        }
+
+        return 500;
```
indytechcook, line 75: Can you use `Symfony\Component\HttpFoundation\Response::HTTP_INTERNAL_SERVER_ERROR` instead?

docroot/modules/custom/va_gov_workflow_assignments/src/Service/EditorialWorkflow.php
```diff
@@ -0,0 +1,58 @@
...
+   */
+  public function getLatestArchivedRevisionId(NodeInterface $node) : int {
+    $query = $this->database->select('content_moderation_state_field_revision', 'cmr')
+      ->condition('content_entity_id', $node->id(), '=')
+      ->condition('moderation_state', 'archived', '=')
+      ->fields('cmr', ['content_entity_revision_id'])
+      ->orderBy('content_entity_revision_id', 'DESC')
+      ->range(0, 1);
+    $result = $query->execute()->fetchField();
+    return $result ? $result : 0;
```
indytechcook, line 41:
```suggestion
    return $result ?? 0;
```

docroot/modules/custom/va_gov_workflow_assignments/src/Service/EditorialWorkflow.php
```diff
@@ -0,0 +1,58 @@
...
+
+namespace Drupal\va_gov_workflow_assignments\Service;
+
+use Drupal\Core\Database\Connection;
+use Drupal\node\NodeInterface;
+
+/**
+ * Class EditorialWorkflow.
+ */
+class EditorialWorkflow implements EditorialWorkflowInterface {
```
indytechcook, line 11: Let's name it `EditorialWorkflowContentRepository` or something.  `EditorialWorkflow` is a bit generic.

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -187,11 +276,35 @@ private function getLink(Term $term, bool $first = FALSE) {
...
     ]), $term_url)->toString();
+  }
+
+  /**
+   * Determine whether the va.gov URL should be displayed.
+   *
+   * @return bool
+   *   Boolean value.
+   */
+  private function vaGovUrlShouldBeDisplayed($node) {
```
indytechcook, line 292:
```suggestion
  private function vaGovUrlShouldBeDisplayed($node) : bool {
```

docroot/modules/custom/va_gov_workflow_assignments/src/Plugin/Block/EntityMetaDisplay.php
```diff
@@ -67,36 +104,46 @@ public static function create(ContainerInterface $container, array $configuratio
...
-          $block_items['VA.gov URL'] = Link::fromTextAndUrl($url, Url::fromUri($url))->toString();
-        }
+
+    $block_items['Owner'] = $this->getSections($node, $node_revision)['links'];
+    $block_items['Content Type'] = $node->type->entity->label();
+
+    if ($this->vaGovUrlShouldBeDisplayed($node)) {
+      $va_gov_url = $this->vaGovUrl->getVaGovUrlForEntity($node);
+
+      if ($this->vaGovUrl->getVaGovUrlStatusForEntity($node) === 200) {
```
indytechcook, line 132: Let's have the servie determine success here instead of checking the status ourself

indytechcook (commented):
This is great!  I like how you mocked the test. It's a really good pattern for others to follow.

I made just a few comments.

docroot/modules/custom/va_gov_backend/src/Form/ExclusionTypesAdminForm.php
acrollet: TIL!

docroot/modules/custom/va_gov_workflow_assignments/src/Service/EditorialWorkflow.php
acrollet: Sure, I was thinking that it could eventually be a more general service that could eventually be asked questions about the va.gov editorial workflow, but I'm good with keeping its purpose narrow.

docroot/modules/custom/va_gov_backend/src/Service/VaGovUrl.php
acrollet: @indytechcook can you expand on this a little? Are you thinking of something like this in a Drupal settings.php?

```
$settings['va_gov_web_environments'] = [
    'prod' => 'https://prod.demo.va.gov',
    'staging' => 'https://staging.demo.va.gov',
    'dev' => 'https://dev.demo.va.gov',
];
```

docroot/modules/custom/va_gov_backend/src/Service/VaGovUrl.php
acrollet: I believe this is obviated by the new approach of returning success/failure, let me know if i missed it.

ndouglas:
I think that's a fine solution, given the limitations.  I thought I'd done something along these lines at some point with creating a and injecting a service dynamically using a closure, but I can't find any relevant code in my repos or SO or anywhere else, so I must be mistaken.

docroot/modules/custom/va_gov_backend/src/Service/VaGovUrl.php
indytechcook: No, just allow the url to be defined in `settings.php` directly.

docroot/modules/custom/va_gov_backend/src/Service/VaGovUrl.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+    catch (Exception $e) {
+      return '';
+    }
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function vaGovUrlForEntityIsLive(EntityInterface $entity, String $environment = 'prod') : bool {
```
indytechcook, line 61: This is nice.

docroot/modules/custom/va_gov_backend/src/Service/VaGovUrl.php
```diff
@@ -0,0 +1,87 @@
...
+    $va_gov_url = $this->getVaGovUrlForEntity($entity, $environment);
+
+    if (!empty($va_gov_url)) {
+      try {
+        // Keep the timeout low so that we don't block page loads for too long.
+        $response = $this->httpClient->head($va_gov_url, ['connect_timeout' => 2, 'http_errors' => FALSE]);
+        if ($response->getStatusCode() == 200) {
```
indytechcook, line 68: So I took a quick look at the `Symfony\Component\HttpFoundation\Response` object and there is a `isSuccessful` method!  

```
    /**
     * Is response successful?
     *
     * @return bool
     *
     * @final since version 3.2
     */
    public function isSuccessful()
    {
        return $this->statusCode >= 200 && $this->statusCode < 300;
    }
```

docroot/modules/custom/va_gov_backend/modules/va_gov_backend_http_client/src/VaGovBackendHttpClient.php
```diff
@@ -0,0 +1,26 @@
...
+ * Mock HTTP client for behat tests.
+ */
+class VaGovBackendHttpClient extends Client {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function head($uri, array $options = []) {
+    if (preg_match('/-200$/', $uri)) {
+      return new Response(200);
```
indytechcook, line 18: Ideally, you would use `Response::create()` but it doesn't really matter especially in a test.

indytechcook:
Sorry to be so nit-picky here.   I see this as a template for other developers to use in their code.

indytechcook:
> I think that's a fine solution, given the limitations. I thought I'd done something along these lines at some point with creating a and injecting a service dynamically using a closure, but I can't find any relevant code in my repos or SO or anywhere else, so I must be mistaken.

It's easier in PHPUnit because you can override the DI container plus I think you can use annotation in the actually class to say which lass to use to mock it but I might be wrong there.

docroot/modules/custom/va_gov_workflow_assignments/src/Service/EditorialWorkflow.php
indytechcook: I understand that thought.  Those sorts of classes turn into giant blobs of unmanageable code :). I've done it too many times to count.

acrollet:
> Sorry to be so nit-picky here. I see this as a template for other developers to use in their code.

not at all! Thanks for taking the time.

acrollet:
@ElijahLynn @indytechcook @olivereri one thought I had is that we want to be really, really sure that the va_gov_backend_http_client is never enabled on production. Do we have anything running in the deploy process that enforces that the enabled modules match config?

ndouglas:
I've used [config_readonly](https://www.drupal.org/project/config_readonly) for a year or two.  Might be an option generally.

indytechcook:
We run `config:import` on deploy.  I like the `config_readonly` solution.  

I do think disabled `devel` and `ui` type modules during a deploy is a good idea.  Just incase someone enabled and forget to disable.

docroot/modules/custom/va_gov_backend/src/Service/VaGovUrl.php
acrollet: We're actually dealing with [GuzzleHttp\Psr7\Response](https://github.com/guzzle/psr7/blob/master/src/Response.php) rather than the Symfony class. However we can get similar behavior by using its default behavior (follow redirects and throw exceptions for 4xx/5xx responses) by returning TRUE if there is no exception.

acrollet:
We may want a test that inspects the config for modules in a deny list and fails if it finds any, as it's a relatively easy developer mistake to update config and include the currently enabled modules.

ndouglas:
We could blacklist it in `config_split`.   I don't mind the idea of parsing the core.extensions.yml file and checking a blacklist though either.  Might prevent us from inadvertently enabling all sorts of potentially harmful or known-crappy modules.

indytechcook:
@ndouglas can you create a follow-up issue?

docroot/modules/custom/va_gov_backend/modules/va_gov_backend_http_client/src/VaGovBackendHttpClient.php
acrollet: GuzzleHttp\Psr7\Response doesn't actually have a create method, but I'm glad you mentioned it bc I found an issue with the 404 response.

acrollet:
> **EDIT**: Oh man, if I could travel back in time and do this on my ex-employer's old D7 repo 😩

😂

docroot/modules/custom/va_gov_backend/src/Service/VaGovUrl.php
acrollet: @indytechcook done, let me know if this is what you had in mind?

acrollet:
@indytechcook tests passed, ready for final(?) review.

tests/behat/Drupal/ContextBase.php
```diff
@@ -23,6 +23,22 @@ public function __construct() {
     $this->entityTypeManager = \Drupal::entityTypeManager();
   }
 
+  /**
+   * @BeforeFeature @mock_va_gov_urls
+   */
+  static function enableVaGovBackendHttpClient()
+  {
+    \Drupal::service('module_installer')->install(['va_gov_backend_http_client']);
```
ElijahLynn, line 31: This seems like it should work, but the install and uninstall seems like it may add 30-60 seconds onto the test run. Probably not an issue once we parallelize the tests though.

ElijahLynn:
> Do we have anything running in the deploy process that enforces that the enabled modules match config?

Yes, config split should work and can keep this module from ever being enabled on prod. I know Neil said to make a separate issue but I think it should be included in this PR if it is going to go in. 

https://github.com/department-of-veterans-affairs/va.gov-cms/tree/master/config

ndouglas:
> Yes, config split should work and can keep this module from ever being enabled on prod. I know Neil said to make a separate issue but I think it should be included in this PR if it is going to go in.

TBF Neil might've just been trying to get me to stop talking.

docroot/modules/custom/va_gov_backend/src/Service/VaGovUrl.php
```diff
@@ -0,0 +1,100 @@
...
+use Exception;
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Exception\RequestException;
+
+/**
+ * Class VaGovUrl.
+ */
+class VaGovUrl implements VaGovUrlInterface {
+
+  const WEB_ENVIRONMENTS = [
```
indytechcook, line 16: We do not need `WEB_ENVIRONMENTS` here since we will be setting the URL in `settings.php`

docroot/modules/custom/va_gov_backend/src/Service/VaGovUrl.php
```diff
@@ -0,0 +1,100 @@
...
+    $this->httpClient = $httpClient;
+    $this->settings = $settings;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function getVaGovUrlForEnvironment(String $environment) : string {
+    // Allow prod URL to be overridden by settings.php.
+    if ($environment === 'prod' && $this->settings->get('va_gov_prod_url', '')) {
```
indytechcook, line 54: Only look at `va_gov_frontend_url` in `Settings` for this value.  Default to `https://prod.demo.va.gov`

indytechcook (commented):
Looking Great!  One _last_ change and we are good to go!

acrollet:
NB: this PR now introduces a minor visual change, which is that the carats between owner sections are no longer linked. I view this as an improvement, because:
* There can be ambiguity about which section link the carats are part of
* It's consistent with standard Drupal breadcrumbs below the tabs

Just calling it out in case it's an important requirement.

tests/phpunit/VaGovUrlServiceTest.php
```diff
@@ -52,20 +52,20 @@ public function setUp() {
     parent::setUp();
 
     $this->container = new ContainerBuilder();
-    $this->config = [];
+    $this->config = ['hash_salt' => 'SCVSPZNSKKK5XCRJ1WLE'];
```
indytechcook, line 55: What's the story to why this was added?

acrollet: The test errored out locally without it. Not sure why it didn't before tbh.

indytechcook:
This all great!!

The last thing is to add the following to the settings.php files for the correct environment.

```
const WEB_ENVIRONMENTS = [
    'prod' => 'https://www.va.gov',
    'staging' => 'https://staging.va.gov',
    'dev' => 'https://dev.va.gov',
  ];
```

acrollet:
@indytechcook I must have misunderstood something in our previous conversation, I reverted the [BuildFrontend service](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/5ee5496eb7ea028a152a3d21d6376b7f498aa7e5/docroot/modules/custom/va_gov_build_trigger/src/Service/BuildFrontend.php#L36) to its previous state. I already added a 16th minute item for today to discuss the testing blocker, should we discuss this then as well?

indytechcook:
@acrollet Tests are working great!  It's more for the block display on dev/stage/prod.  So adding `va_gov_frontend_url` to each will have the URL in the block on each environment display correctly.  This isn't a blocker though and only required if the front end URL is different between environments.

indytechcook (dismissed):

ElijahLynn:
> > Do we have anything running in the deploy process that enforces that the enabled modules match config?
> 
> Yes, config split should work and can keep this module from ever being enabled on prod. I know Neil said to make a separate issue but I think it should be included in this PR if it is going to go in.
> 
> https://github.com/department-of-veterans-affairs/va.gov-cms/tree/master/config

Discussed in Slack (internal) https://dsva.slack.com/archives/CT4GZBM8F/p1604359354131000?thread_ts=1604336649.101600&cid=CT4GZBM8F and the only way this would ever get enabled is if someone manually enables on PROD, or puts a config dependency on it so I am fine with not doing that now. I just thought it was a pretty low lift to config split it if it was going to go in.

ndouglas:
maxresdefault

Issue #3323, opened 2020-10-27
Upgrade simplesaml php 


swirtSJW:
**Describe the defect**
There are several critical issues related to our outdated version of simplesamlphp  

1. https://github.com/department-of-veterans-affairs/va.gov-cms/network/alert/simplesamlphp-1.17.2/composer.lock/symfony%2Fhttp-foundation/open
2. https://github.com/department-of-veterans-affairs/va.gov-cms/network/alert/simplesamlphp-1.17.2/composer.lock/symfony%2Fdependency-injection/open
3. https://github.com/department-of-veterans-affairs/va.gov-cms/network/alert/simplesamlphp-1.17.2/composer.lock/robrichards%2Fxmlseclibs/open

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information if relevant, or delete):**

**Additional context**
Add any other context about the problem here. Reach out to the Product Managers to determine if it should be escalated as critical (prevents users from accomplishing their work with no known workaround and needs to be addressed within 2 business days).

**Tasks**
- [x] Update dependency
- [x] Update Drupal module
- [x] Add/remove/update any patches
- [x] Copy vendor folder over simplesamlphp-1.17 and re-apply any changes (diff the old version and folder) 
- [x] Update any post-install scripts in composer.json
- [ ] QA on DEV

ElijahLynn:
We are currently on 1.17.2 and 1.17.8 has some of the updates needed but not all, so we need to jump up a minor release to 1.18.8. This requires our drupal/simplesamlphp_auth to be updated as well as there was a dependency update just for this library update here https://www.drupal.org/node/3097283. 

I am making a PR for this now and testing on DEV. 

Related links for me:
https://github.com/simplesamlphp/simplesamlphp/releases
https://github.com/simplesamlphp/simplesamlphp/blob/simplesamlphp-1.19/docs/simplesamlphp-changelog.md
https://simplesamlphp.org/docs/stable/simplesamlphp-upgrade-notes-1.18
https://www.drupal.org/project/simplesamlphp_auth/releases/8.x-3.2

ElijahLynn:
Here is the PR for this change, https://github.com/department-of-veterans-affairs/va.gov-cms/pull/3336. However, I cannot test on DEV right now as even without this PR, I am getting a 403 Forbidden error so I will need to get that working first. 

[...]

PR #3336, opened 2020-10-29
VACMS-3323: Upgrade simplesamlphp library and Drupal module

ElijahLynn:
See #3323

This PR deletes over 4,000 files that were duplicated from vendor and were copied up outside of docroot. Now it just symlinks the docroot/vendor/simplesamlphp/{simplesamlphp,saml2} directories and copies the config/metadata folders so as not to have to duplicate them. 

This means it is not reviewable here and you need to check it out locally to review.

## Testing done
- [x] Deploy to DEV
- [x] Tested working login on DEV with my PIV card
- [x] Test that password change still works


ElijahLynn:
Oh, there is a folder in root for 1.17.2 that I need to figure out what has changed as well and put 1.18.8 there instead.

swirtSJW (approved):
Changes all look good.   If it tests out on dev or staging, it will be good to merge.

ElijahLynn:
Yay for tests...
Failure on features/user.feature:4 with:

> And I should see "Login with PIV or other Smartcard." in the "#edit-simplesamlphp-auth-login-link" element # Drupal\DrupalExtension\Context\MinkContext::assertElementContainsText()
> 
>       Element matching css "#edit-simplesamlphp-auth-login-link" not found. (Behat\Mink\Exception\ElementNotFoundException)

ElijahLynn:
> Oh, there is a folder in root for 1.17.2 that I need to figure out what has changed as well and put 1.18.8 there instead.

This is pushed now, and I included our custom modifications to the /config and /metadata folders as well.

ElijahLynn:
> Yay for tests...
> Failure on features/user.feature:4 with:
> 
> > And I should see "Login with PIV or other Smartcard." in the "#edit-simplesamlphp-auth-login-link" element # Drupal\DrupalExtension\Context\MinkContext::assertElementContainsText()
> > ```
> >   Element matching css "#edit-simplesamlphp-auth-login-link" not found. (Behat\Mink\Exception\ElementNotFoundException)
> > ```

There was a new config option that was disabled by default and I had to enable it and re-export it. This is pushed now too and this test passes locally.

config/stg/simplesamlphp_auth.settings.yml
```diff
@@ -32,3 +32,4 @@ role:
   population: ''
   eval_every_time: false
 autoenablesaml: true
+login_link_show: true
```
indytechcook, line 35: New settings?

ElijahLynn: Yeah, and there is an update hook to set the default value but it doesn't check to see if it was set before, at least in our case.

ElijahLynn:
I tested this on DEV and it is working, logging in. 

However, I don't like this pattern of committing simplesamlphp. I think if the folder needs to be outside of docroot, then it should be a post-install-cmd to copy it, and then lay down the config in our devops scripts, ideally. 

This PR is the same as what we have now though, so I don't want to let perfect be the enemy of good.

swirtSJW (approved):
Approved

ElijahLynn:
> I tested this on DEV and was able to log in.
> 
> However, I don't like this pattern of committing simplesamlphp. I think if the folder needs to be outside of docroot, then it should be a post-install-cmd to copy it, and then lay down the config in our devops scripts, ideally.
> 
> This PR is the same as what we have now though, so I don't want to let perfect be the enemy of good.

K, I finally got this working on DEV. I also took care of this issue and now we are not committing the entire library but just symlinking it and copying the two config/metadata files needed for it to work. This does require https://github.com/department-of-veterans-affairs/devops/pull/7968 to be merged simultaneously.

indytechcook (approved):

composer.json
```diff
@@ -296,8 +296,7 @@
                 "PHP 7.2 notice - \"Undefined offset: 1\" in JsonSchemaEncoder": "https://www.drupal.org/files/issues/2019-07-08/3063027-5-undefined-offset.patch"
             },
             "drupal/simplesamlphp_auth": {
-                "Don't attempt username sync when user was already matched by username": "https://www.drupal.org/files/issues/2019-08-14/2748731-23.patch",
-                "Bypass password requirement for password change if logged in with sso.": "https://www.drupal.org/files/issues/2018-05-01/simplesamlphp_auth-change-password-access-2968598-2.patch"
```
swirtSJW, line 300: These patches are no longer needed?

swirtSJW (approved):
Nice work getting all those files out of the repo with some symlink magic.

composer.json
ElijahLynn: "Bypass password requirement for password change if logged in with sso." is in this new release, but "Don't attempt username sync when user was already matched by username" is not yet committed and is still in this PR, one line below.

PR #3360, opened 2020-11-02
VACMS-000 Add exclusion to lando to prevent perm changes to settings.

swirtSJW:
## Description

As a follow up to 
https://github.com/department-of-veterans-affairs/va.gov-cms/commit/0052396271150022e42b52fe32231c7e9292594b

### CMS user-facing release note

- [ ] Yes, but it hasn't yet been written (this code is not ready to merge!).


ndouglas (approved):
I'm intimately familiar with this now (🤦) so I feel qualified to approve.

Issue #2934, opened 2020-09-16
Table field data contains broken HTML if editor forgets to close a tag


oksana-c:
## Problem statement

Table field does not offer a WYSIWYG for table cells, which forces editors to manually enter basic HTML tags if text needs to be decorated in any way. Most used tags are `<strong>`, `<em>`, `<ul>`, `<a href="">`

Table fields are quite small, especially when more than 3 table columns are added. It is easy for editors to forget to close an HTML tag. There is no feedback to users that some tags may be broken upon table save.

Content export exports table field values in a form of an array that has not been processed by any Drupal text format plugins which may catch and fix broken tags. All broken HTML persists on export. Sample export below:

```
"field_table": [
        {
            "value": {
                "0": [
                    "<strong>test",
                    ""
                ],
                "1": [
                    "<em>test",
                    ""
                ],
                "2": [
                    "<ul><li>list item<\/li>",
                    ""
                ],
                "3": [
                    "",
                    ""
                ],
                "4": [
                    "",
                    ""
                ],
                "caption": ""
            },
            "format": "rich_text",
            "caption": ""
        }
    ]

```

NOTE: this issue was identified by reviewing current function of tablefield  module and there are no known issues with table HTML on production related to broken markup returned by Drupal

Ideas

- [ ] triage FE process for building tables and confirm that FE does not process retrieved data to remove broken tags
- [ ] explore whether the CMS team should do additional preprocessing before generating table JSON  and open follow up story as needed
- [ ] use wysiwygs for individual cells (this approach will handle all of the html tag closing stuff via pre configured tidy library)

## Acceptance Criteria

- [ ] Table cells have individual wysiwygs.
**OR**
- [ ] Constraint violation occurs on save when malformed html is entered in cell

swirtSJW:
Looking into this deeper, we should really be making sure this is run through drupal  as processed html.   Similar to the way we handle WYSIWIG content.   Then the governance of the allowed html elements is still respected.    Otherwise it is just using raw input directly which is not safe and has no governance.

The other issue is that libxml testing does not catch `<strong>test` because non-closed tags are considered valid.

swirtSJW:
Using it processed as drupal does also ensures that our View tab preview is more in line with front end preview/render.

PR #3369, opened 2020-11-03
VACMS-2934: Add html formatting constraint checks to table cells.

ethanteague:
## Description
See #2934 
@kevwalsh @oksana-c have a look at the screenshot / form when qa'ing. 

**BAD NEWS**: You'll notice that every cell is highlighted when a violation is detected - I haven't been able to figure out a way (at least not with constraints) to target individual cells.

**GOOD NEWS**: Html entries are all being caught, and this approach will solve the challenges that editors are having with missed / mismatched tags breaking pages.

Just lemme know if this approach will suffice.

## Testing done
Visual 

## Screenshots
image

## QA steps
- As an administrator, go to `/node/471/edit`
- In Main content section, add a table paragraph
- Enter content with broken html tags in multiple cells
- Save form, and confirm error notifications occur and specify each improperly formatted html instance

### DIFF
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -926,6 +926,10 @@ function va_gov_backend_entity_bundle_field_info_alter(&$fields, EntityTypeInter
   if ($entity_type->id() === 'paragraph' && $bundle === 'phone_number' && isset($fields['field_phone_number'])) {
     $fields['field_phone_number']->addConstraint('ValidPhoneNumber');
   }
+  // Add tablefield cell html validation.
+  if ($entity_type->id() === 'paragraph' && $bundle === 'table' && isset($fields['field_table'])) {
+    $fields['field_table']->addConstraint('ValidCellHtml');
+  }
   // Add title duplication prevention validation to q_a nodes.
   if ($entity_type->id() === 'node' && $bundle === 'q_a' && isset($fields['title'])) {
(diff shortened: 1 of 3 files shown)


docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -184,10 +183,6 @@ function _va_gov_backend_add_vamc_regions_select(array &$form) {
  * Implements hook_form_alter().
  */
 function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form_id) {
-
-  // Disable path alias modification after lc node is published.
-  _va_gov_disable_alias_widget($form, $form_state);
-
```
swirtSJW, line 190: Looks like you are clobbering some previous work here.

swirtSJW (changes requested):
Bad rebase?   Clobbered a bunch of prior work.

ethanteague:
> Bad rebase? Clobbered a bunch of prior work.

Yeah - lemme sort this out. Gonna put this in draft state for now.

swirtSJW:
I'm trying a bit to refine this... make it less scary

swirtSJW:
@ethanteague  I just pushed up some changes to make the feedback more informative about which cells are bad... but this still does not seem like the right approach.   I commented as such on the issue.   We should re-group on this.

swirtSJW:
@ethanteague I am going to close this in favor of the processing approach rather than the validation approach.   It was a valiant effort on both our parts.. but this just aint the right dragon to slay,

Issue #2729, opened 2020-08-20
Anchor links behavior hides page headings


rachel-kauff:
When I add anchor links to user guides, it doesn’t account for the amount of space the admin toolbar takes up on the page, so when you click on the link it covers the heading that it was supposed to jump to. That's a confusing experience because it disorients users on the page.

This is an issue for users because I've structured user guides with anchor links to help users navigate lengthy or content-heavy pages, and integrated anchor links into the strategy for getting users to the right information. Howie plans to use the anchored links to user guides to provide help desk answers.

![Screen Shot 2020-08-20 at 8.27.29 AM.png](https://images.zenhubusercontent.com/5e399cf5446c5b4ca0e31e42/4cb5ebdf-87fd-4799-a7cb-1e6ed6c1bf0c)![Screen Shot 2020-08-20 at 8.27.35 AM.png](https://images.zenhubusercontent.com/5e399cf5446c5b4ca0e31e42/d19ae333-95a6-4ec0-b7fb-9cc36ac8b4f7) 

## Initial research

There's a rabbit hole of drupal issues related to something similar, starting with this 2009 thread. https://www.drupal.org/project/drupal/issues/546126

ElijahLynn:
Slack discussion here https://dsva.slack.com/archives/CT4GZBM8F/p1597938169005900 [internal].

kevwalsh:
There's a rabbit hole of drupal issues related to something similar, starting with this 2009 thread. https://www.drupal.org/project/drupal/issues/546126

swirtSJW:
Yeesh yah this is a bit of a mess. because it is largely handled by the browser which has not idea the top of the window is blocked from view.  Maybe a js behaviour that if a location has a # in it, it backscrolls the window 50px or whatever the thickness of those two menus are.

acrollet:
@rachel-kauff this will get deployed tomorrow :-)

PR #3383, opened 2020-11-04
VACMS-2729: Take toolbar height into account when going to anchor links

acrollet:
## Description

See #2729 

## Testing done

Manual testing on "How to edit detail pages — for VAMCs" page

## QA steps

As a user with the "authenticated" role:
1. Go to /help/product-guides/vamc/by-content-type/detail-pages
1. When clicking anchor links (e.g. "How to find existing Detail Page content to edit")
- [ ] The heading should not be hidden when the tooltray is horizontal
- [ ] The heading should not be hidden when the tooltray is vertical

As an anonymous user:
1. Go to /help/product-guides/vamc/by-content-type/detail-pages
1. When clicking anchor links (e.g. "How to find existing Detail Page content to edit")
- [ ] The heading should not be hidden

### CMS user-facing release note

- [ ] Yes, but it hasn't yet been written (this code is not ready to merge!).

### DIFF
--- a/docroot/themes/custom/vagovadmin/assets/js/anchor_links.js
+++ b/docroot/themes/custom/vagovadmin/assets/js/anchor_links.js
@@ -1,0 +1,43 @@
+/**
+ * @file
+ * Anchor link behaviors.
+ */
+
+(function ($, Drupal) {
+
+  'use strict';
+
+  /**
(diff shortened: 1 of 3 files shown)


ndouglas (approved):
LGTM and seems to have the desired effect both auth and anonymous.  Scaled up and down font size in browser and still worked.

PR #3415, opened 2020-11-07
VACMS-2934: Add html cleanup to table field.

ethanteague:
## Description
See #2934 

## Testing done
Visual / behat

## QA steps
- As an administrator, go to /node/471/edit
- In Main content section, add a table paragraph
- Enter content with broken html tags in multiple cells
- Save form, and confirm table displays properly on node save
- Return to edit form, and confirm broken html has been cured in table cells


docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1204,6 +1204,38 @@ function va_gov_backend_entity_presave(EntityInterface $entity) {
...
+  }
+}
+
+/**
+ * Clean up html tags in table fields.
+ *
+ * @param \Drupal\Core\Entity\EntityInterface $entity
+ *   The table paragraph.
+ */
+function _va_gov_backend_tidy_html(EntityInterface $entity) {
```
swirtSJW, line 1219: Naming here is a bit vague.  Could benefit from being named about what it actually does.  
```suggestion
function _va_gov_backend_apply_filter_to_tablefield_cells(EntityInterface $entity) {
```

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1204,6 +1204,38 @@ function va_gov_backend_entity_presave(EntityInterface $entity) {
...
     _va_gov_backend_disable_autopath_alias($entity);
   }
+
+  if ($entity->getEntityTypeId() === 'paragraph' && $entity->bundle() === 'table') {
+    _va_gov_backend_tidy_html($entity);
+  }
+}
+
+/**
+ * Clean up html tags in table fields.
```
swirtSJW, line 1214:
```suggestion
 * Apply html filter to tablefield on save so it is properly formatted on the way in.
 *
 * This is needed because the field data of table fields does not have a 'processed' 
 * property in graphQL or CMS Export. 
```

You'll need to add this one by hand since formatting of comments by the inline github suggestion always ends messed up.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1204,6 +1204,38 @@ function va_gov_backend_entity_presave(EntityInterface $entity) {
...
+  $format = $entity->get('field_table')->format;
+  $sanitized_items = [];
+  if (!empty($raw_table)) {
+    foreach ($raw_table as $row_key => $row_item) {
+      if ($row_key === 'caption') {
+        // Tablefield caption does not support html.
+        $sanitized_items['caption'] = strip_tags($row_item);
+      }
+      else {
+        foreach ($row_item as $cell_key => $cell_item) {
```
swirtSJW, line 1230: I think we need to be defensive here and check that is an array before dumping it into a foreach.   I can envision some goofy situation where an editor removes a row or something and we get unexpected results here which would create a whitescreen during node save which would be bad.

swirtSJW (changes requested):
I think you are on the right track here.  Just a few suggestions.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1194,6 +1194,38 @@ function va_gov_backend_entity_presave(EntityInterface $entity) {
   if ($entity->getEntityTypeId() === 'node' && in_array($entity->bundle(), _va_gov_backend_get_locked_alias_bundles())) {
     _va_gov_backend_disable_autopath_alias($entity);
   }
+
+  if ($entity->getEntityTypeId() === 'paragraph' && $entity->bundle() === 'table') {
```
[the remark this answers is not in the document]
ethanteague: That's a great point, @swirtSJW - didn't think of that :+1:

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1204,6 +1204,44 @@ function va_gov_backend_entity_presave(EntityInterface $entity) {
...
+ * Apply html filter to tablefield on save to properly format on the way in.
+ *
+ * This is needed because the field data of table fields does not have a
+ * 'processed' property in graphQL or CMS Export.
+ *
+ * @param \Drupal\Core\Entity\EntityInterface $entity
+ *   The table paragraph.
+ */
+function _va_gov_backend_apply_filter_to_tablefield_cells(EntityInterface $entity) {
+  $raw_table = $entity->get('field_table')->value;
```
swirtSJW, line 1226: This is still making the assumption it is in field_table... which bypasses other occurrences even if they were found in the earlier loop.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1204,6 +1204,44 @@ function va_gov_backend_entity_presave(EntityInterface $entity) {
   if ($entity->getEntityTypeId() === 'node' && in_array($entity->bundle(), _va_gov_backend_get_locked_alias_bundles())) {
     _va_gov_backend_disable_autopath_alias($entity);
   }
+
+  $field_definitions = $entity->getFieldDefinitions();
+  foreach ($field_definitions as $field) {
+    if ($field->getType() === 'tablefield') {
```
swirtSJW, line 1210: This logic is finding the presence of tablefield, but then not getting the field name, so it does not know what to alter.  
I suggest not doing the detection here.  Just call the function `_va_gov_backend_apply_filter_to_tablefield_cells` for all entities and have this detection in the function.  As it loops through detecting the `tablefield` it should be collecting the field names so then when it goes to clean those, it knows which fields to process.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
swirtSJW: Still need to be defensive here.

swirtSJW (changes requested):

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: oops - will fix :+1:

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1204,6 +1204,49 @@ function va_gov_backend_entity_presave(EntityInterface $entity) {
...
+ * This is needed because the field data of table fields does not have a
+ * 'processed' property in graphQL or CMS Export.
+ *
+ * @param \Drupal\Core\Entity\EntityInterface $entity
+ *   The table paragraph.
+ */
+function _va_gov_backend_apply_filter_to_tablefield_cells(EntityInterface $entity) {
+  $sanitized_items = [];
+
+  $field_definitions = Drupal::service('entity_field.manager')->getFieldDefinitions($entity->getEntityTypeId(), $entity->bundle());
```
swirtSJW, line 1223: Why was it necessary to load the service and call this for the bundle when you had direct access to the field definitions on the entity via `$entity->getFieldDefinitions()`?

That was working before. We just need to detect if the method is available first since not all entities support it.

ethanteague: Marking this as resolved, per slack convo summary: 
**Ethan Teague** Today at 9:38 AM
Howdy, @Steve Wirt - here's the problem I was running into / reason for calling the service to get field info.

image

**2 replies**

**Steve Wirt**  5 minutes ago
OK...   That is because our php is too out of date.  On our old php  method_exists returns a TRUE even if the method is private.  On current php  it returns FALSE if the method is private.
var_dump(method_exists(Bar::class, 'privateMethodTest'));
// PHP 7.4: bool(false)
// PHP 7.3: bool(true)
 (edited) 

**Steve Wirt**  5 minutes ago
Good solution then.

ethanteague:
Closing in favor of https://github.com/department-of-veterans-affairs/va.gov-cms/pull/3511

Issue #1117, opened 2020-02-28
Entity Field definitions


swirtSJW:
**Describe the defect**
On the module status page there is a large warning related to 4 different mismatched entity field definitions.   The scope of the impact of these mismatches is not known, but they can't be good.

Errors are related to:
```
Environment Switcher
The Environment Switcher entity type needs to be installed.
```

```
Migration
The Migration entity type needs to be installed.
Migration Group
The Migration Group entity type needs to be installed.
```
```
Content
The node.field_health_services_local_info field needs to be uninstalled.
The node.field_benefits field needs to be updated.
The node.field_facility_locator_api_id field needs to be updated.
```

**To Reproduce**
1. Go to '/admin/reports/status'
2. see error

**Expected behavior**
Not seeing that error 

**Screenshots**
![image](https://user-images.githubusercontent.com/5752113/75588475-bf2f9a80-5a46-11ea-8554-aaaf84def6b0.png)

**Additional context**

## Labels
- [/] Issue type (red) (defaults to "Defect")
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [/] CMS objective (orange) (not needed for bug tickets)
- [ ] CMS-supported product (black)

swirtSJW:
`drush entity-updates`  runs, but has no effect

```
17:10:09 » lando drush entity-updates
The following updates are pending:

environment_indicator entity type : 
  The Environment Switcher entity type needs to be installed.
migration_group entity type : 
  The Migration Group entity type needs to be installed.
migration entity type : 
  The Migration entity type needs to be installed.
node entity type : 
  The node.field_health_services_local_info field needs to be uninstalled.
  The node.field_benefits field needs to be updated.
  The node.field_facility_locator_api_id field needs to be updated.
Do you wish to run all pending updates? (y/n): y
Cache rebuild complete.                                                                                                                                                                                                                              [ok]
Finished performing updates.                                                                                                                                                    
```

kevwalsh:
@swirtSJW is estimating this at 8 because of uncertainty.

PR #3148, opened 2020-10-08
VACMS-2797: Add new tests to check for PHP notices, warnings & errors.

ElijahLynn:
## Description
This adds two new tests to check for PHP WARNINGs and ERRORs. I couldn't check for both so just made two new tests. 
#2797 

The first test run of this should fail, as we still have some existing code introducing an error in test runs. 

## Testing done
Locally tested on Lando. 

## QA steps
1. Create a PHP warning or error in code so it shows in the logs (normally a full test run might do this)
1. Run `lando yaml-tests php-`
1. Respective test should fail.

PR #3434, opened 2020-11-09
VACMS-1117: update mismatched entities and fields

acrollet:
## Description

See #1117 

## Testing done

- Pulled new prod DB, ran database update, verified that status errors are cleared
- Added test to confirm that no Drupal status errors are present

## QA steps

As an admin user:
1. Go to the status reports and confirm that errors are shown
1. Run database updates
- [x] No errors are shown on the Drupal status report

  - [ ] Merge and ping @rachel-kauff so she's ready to publish after deployment
  - [ ] Don't merge yet -- ping @rachel-kauff to prompt her to write and queue content

### DIFF
--- a/docroot/modules/custom/va_gov_db/va_gov_db.install
+++ b/docroot/modules/custom/va_gov_db/va_gov_db.install
@@ -99,7 +99,9 @@ function _va_gov_db_install_missing_entity_type($missing_entity_type) {
   $type_manager = \Drupal::entityTypeManager();
   $entity_type = $type_manager->getDefinition($missing_entity_type);
   \Drupal::entityDefinitionUpdateManager()->installEntityType($entity_type);
-  return "Installed missing entity type {$missing_entity_type}.";
+  \Drupal::logger('va_gov_db')->log(LogLevel::INFO, 'Installed missing entity type %missing_entity_type.', [
+    '%missing_entity_type' => $missing_entity_type,
+  ]);
 }
 
 /**
@@ -111,7 +113,9 @@ function _va_gov_db_install_missing_entity_type($missing_entity_type) {
 function _va_gov_db_update_node_field_storage_config($field_name) {
   if ($field_storage = FieldStorageConfig::loadByName('node', $field_name)) {
     $field_storage->save();
-    return "Updated field storage configuration for {$field_name}.";
+    \Drupal::logger('va_gov_db')->log(LogLevel::INFO, 'Updated field storage configuration for %field_name.', [
+      '%field_name' => $field_name,
+    ]);
   }
 }
 
@@ -374,12 +378,12 @@ function va_gov_db_update_8020(&$sandbox) {
     'migration_group',
   ];
   foreach ($missing_entity_types as $missing_entity_type) {
-    $messages[] = _va_gov_db_install_missing_entity_type($missing_entity_type);
+    _va_gov_db_install_missing_entity_type($missing_entity_type);
   }
 
   $out_of_date_fields = ['field_benefits', 'field_facility_locator_api_id'];
   foreach ($out_of_date_fields as $out_of_date_field) {
-    $messages[] = _va_gov_db_update_node_field_storage_config($out_of_date_field);
+    _va_gov_db_update_node_field_storage_config($out_of_date_field);
   }
 
   $entity_manager = \Drupal::entityManager();
@@ -389,8 +393,8 @@ function va_gov_db_update_8020(&$sandbox) {
     $definition_update_manager->uninstallFieldStorageDefinition(
       $previous_storage_definitions['field_health_services_local_info']
     );
-    $messages[] = "Deleted field storage configuration for field_health_services_local_info.";
+    \Drupal::logger('va_gov_db')->log(LogLevel::INFO, 'Deleted field storage configuration for field_health_services_local_info.');
   }
 
-  return implode(PHP_EOL, $messages);
+  return "Cleared mis-matched entity/field errors.";
 }
--- a/docroot/modules/custom/va_gov_db/va_gov_db.install
+++ b/docroot/modules/custom/va_gov_db/va_gov_db.install
@@ -369,7 +369,7 @@ function va_gov_db_update_8019(&$sandbox) {
 /**
  * Clear Drupal status report mis-matched entity/field errors.
  */
-function va_gov_db_update_8020(&$sandbox) {
+function va_gov_db_update_8020() {
   $messages = [];
 
   $missing_entity_types = [
--- a/docroot/modules/custom/va_gov_db/va_gov_db.install
+++ b/docroot/modules/custom/va_gov_db/va_gov_db.install
@@ -90,36 +90,6 @@ function _va_gov_db_change_text_field_max_length($field_name, $new_length) {
 }
 
 /**
- * Install a config entity type.
- *
- * @param string $missing_entity_type
- *   Config entity machine name.
- */
-function _va_gov_db_install_missing_entity_type($missing_entity_type) {
-  $type_manager = \Drupal::entityTypeManager();
-  $entity_type = $type_manager->getDefinition($missing_entity_type);
-  \Drupal::entityDefinitionUpdateManager()->installEntityType($entity_type);
-  \Drupal::logger('va_gov_db')->log(LogLevel::INFO, 'Installed missing entity type %missing_entity_type.', [
-    '%missing_entity_type' => $missing_entity_type,
-  ]);
-}
-
-/**
- * Re-save field storage configuration to clear error.
- *
- * @param string $field_name
- *   Node field machine name.
- */
-function _va_gov_db_update_node_field_storage_config($field_name) {
-  if ($field_storage = FieldStorageConfig::loadByName('node', $field_name)) {
-    $field_storage->save();
-    \Drupal::logger('va_gov_db')->log(LogLevel::INFO, 'Updated field storage configuration for %field_name.', [
-      '%field_name' => $field_name,
-    ]);
-  }
-}
-
-/**
  * Modify field_facility_locator_api_id field to be longer.
  */
 function va_gov_db_update_8001(&$sandbox) {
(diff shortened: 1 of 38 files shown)


acrollet:
@indytechcook @ElijahLynn I'd like your devops 👀s on this new test please :-)

acrollet:
FWIW we can ignore certain Drupal errors if we absolutely have to:

```
Adrian.Rollett@Agile6-A45 va.gov-cms % lando drush help core-requirements
Provides information about things that may be wrong in your Drupal installation, if any.

Examples:
 drush core-requirements                   Show all status lines from the Status Report admin page.                                
 drush core-requirements --severity=2      Show only the red lines from the Status Report admin page.                              
 drush core-requirements --pipe            Print out a short report in JSON format, where severity 2=error, 1=warning, and 0/-1=OK

Options:
 --field-labels                            Add field labels before first line of data. Default is on; use --no-field-labels to disable.                         
 --fields=<title, severity, description>   Fields to output. All available fields are: title, severity, sid, description, value (or 'summary'), reason, weight. 
 --format=<json>                           Select output format. Available: table, csv, html, json, list, var_export, yaml. Default is table.                   
 --full                                    Show the full output, with all fields included.                                                                      
 --ignore                                  Comma-separated list of requirements to remove from output. Run with --pipe to see key values to use.                
 --pipe                                    Equivalent to --format=json.                                                                                         
 --severity=<3>                            Only show status report messages with a severity greater than or equal to the specified value.

Topics:
 docs-output-formats                       Output formatting options selection and use.

Aliases: status-report, rq, core:requirements
```

ndouglas:
I'm not seeing any entity mismatch errors in the first place.  Did I miss something? 🤔

acrollet:
@ndouglas hmm, you should with a synced DB, see https://prod.cms.va.gov/admin/reports/status

ndouglas:
okay, I'm sure this is a local issue, carry on

ndouglas:
nvm, scrolled up and realized that I didn't see issues because your PR is so effective.  You were 2-3 steps ahead of me.  Confirmed working, just gonna review the rest.

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -88,6 +89,33 @@ function _va_gov_db_change_text_field_max_length($field_name, $new_length) {
...
+ *
+ * @param string $missing_entity_type
+ *   Config entity machine name.
+ */
+function _va_gov_db_install_missing_entity_type($missing_entity_type) {
+  $type_manager = \Drupal::entityTypeManager();
+  $entity_type = $type_manager->getDefinition($missing_entity_type);
+  \Drupal::entityDefinitionUpdateManager()->installEntityType($entity_type);
+  return "Installed missing entity type {$missing_entity_type}.";
+}
```
ndouglas: This might be just me, but I dislike returning strings from functions unless they're specifically string-formatting functions.  It has a little smell to me, like the function is fulfilling its operation and then formatting a string as a side-effect, but it's returning the formatted string instead of any easily manageable information about whether the operation succeeded or how it failed.  It also limits the reusability of the function, I think.

I'd just remove the string formatting and use a `try...catch` in the update hook, and format the string within the `try` block.

I'm not entirely sure this needs to be a separate function; the string-formatting is trivial (and I think should be split out anyway), and the code left is a call on each of two services, which could (infinitesimally more efficiently) be retrieved outside of the function and even outside of the loop.

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -88,6 +89,33 @@ function _va_gov_db_change_text_field_max_length($field_name, $new_length) {
...
+ *   Node field machine name.
+ */
+function _va_gov_db_update_node_field_storage_config($field_name) {
+  if ($field_storage = FieldStorageConfig::loadByName('node', $field_name)) {
+    $field_storage->save();
+    return "Updated field storage configuration for {$field_name}.";
+  }
+}
+
+
```
ndouglas: This returns either a formatted string or the **void**, and IDK, that leads to some existential angst personally.

I think it suffers from the same issue as above, though, where it's performing a purpose, and if it succeeds, it returns a string, if it fails, it returns diddly-squat, and if it fails it either dies screaming into the void or silently, without any useful feedback.

So I'm not sure it needs to be a separate function, but if it does, I'm not sure it should have the string-formatting commingled.

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -334,3 +362,36 @@ function va_gov_db_update_8019(&$sandbox) {
...
+    'migration',
+    'migration_group',
+  ];
+  foreach ($missing_entity_types as $missing_entity_type) {
+    $messages[] = _va_gov_db_install_missing_entity_type($missing_entity_type);
+  }
+
+  $out_of_date_fields = ['field_benefits', 'field_facility_locator_api_id'];
+  foreach ($out_of_date_fields as $out_of_date_field) {
+    $messages[] = _va_gov_db_update_node_field_storage_config($out_of_date_field);
```
ndouglas: This makes me think that a typo here would, like my `skip_permission_hardening` nightmare, just drift silently into the night without me ever noticing that I'd made a mistake.

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -334,3 +362,36 @@ function va_gov_db_update_8019(&$sandbox) {
...
+ * Clear Drupal status report mis-matched entity/field errors.
+ */
+function va_gov_db_update_8020(&$sandbox) {
+  $messages = [];
+
+  $missing_entity_types = [
+    'environment_indicator',
+    'migration',
+    'migration_group',
+  ];
```
ndouglas, line 379: Is there a way of retrieving these programmatically?  I think I don't understand the events which led to this problem on prod in the first place, so maybe I'm suggesting something asinine or that requires solving the halting problem or something first, IDK.

I think this seems like a great place to reuse code, but maybe it's not factored the best for that.  IDK if I understand the problem well enough in the first place to give meaningful feedback, though.

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -334,3 +362,36 @@ function va_gov_db_update_8019(&$sandbox) {
...
+  }
+
+  $out_of_date_fields = ['field_benefits', 'field_facility_locator_api_id'];
+  foreach ($out_of_date_fields as $out_of_date_field) {
+    $messages[] = _va_gov_db_update_node_field_storage_config($out_of_date_field);
+  }
+
+  $entity_manager = \Drupal::entityManager();
+  $previous_storage_definitions = $entity_manager->getLastInstalledFieldStorageDefinitions('node');
+  if (!empty($previous_storage_definitions['field_health_services_local_info'])) {
```
ndouglas, line 391: Ditto here with possibly retrieving this information programmatically, which might be blessed or cursed.

docroot/modules/custom/va_gov_db/va_gov_db.install
ndouglas: [dramatic reënactment of PHP flying off into the night with my data](https://www.youtube.com/watch?v=4aWhn0Hc8ps)

docroot/modules/custom/va_gov_db/va_gov_db.install
acrollet: @ndouglas yup, should be easy enough with [EntityDefinitionUpdateManager::getChangeList](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityDefinitionUpdateManager.php/function/EntityDefinitionUpdateManager%3A%3AgetChangeList/8.8.x)

I'm not completely sure without in-depth (read: time-consuming) research, but my guess is that these errors arose due to a timing mis-match between when [automatic entity update support was removed](https://www.drupal.org/node/3034742) and when these modules were installed. So, I wouldn't necessarily expect this issue to arise again.

docroot/modules/custom/va_gov_db/va_gov_db.install
acrollet: @ndouglas I believe these errors were caused by removing a field config without also adding a database update to remove the field data. This [new test](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/3434/commits/f62cca6441992dc23eb15563104c1789b0287340) should prevent us from introducing a commit which does the same in the future.

acrollet: @ndouglas in summary, I don't know that it's worth spending time to write and test code to retrieve the identifiers for either situation programmatically but I'm definitely open to it. :-)

acrollet: @ndouglas updated as mentioned on slack.

swirtSJW: I am not sure if we can attribute this to field deletion though.   This field `field_facility_locator_api_id` is still in use.  We are in an odd state and I don't know why,

swirtSJW (dismissed):

docroot/modules/custom/va_gov_db/va_gov_db.install
acrollet: @swirtSJW oops, I was imprecise! field_health_services_local_info is the field that was removed some time ago. The other two fields were in an 'out of date' state. I tried re-exporting their config, but that didn't result in any changes. I attribute it to some change(s) made in the codebase between when the field was first created and now. I don't have a better answer without some in-depth poking...

acrollet:
@ElijahLynn pinging for your review on the new test and help getting the devshop environment resurrected, thanks! :-)

ElijahLynn:
Just unstuck this job. Should build soon.

tests.yml
```diff
@@ -57,3 +57,7 @@ va/tests/behat:
   command: |
     cd tests/behat
     behat --colors
+
+va/tests/status-errors:
```
ElijahLynn, line 61: We should sync on naming here as I introduced two more here https://github.com/department-of-veterans-affairs/va.gov-cms/pull/3148/files: 

- va/tests/php-warning
- va/tests/php-error

Maybe:

```suggestion
va/tests/status-error:
```

tests.yml
```diff
@@ -57,3 +57,7 @@ va/tests/behat:
...
+  description: Check for Drupal status errors
+  command: drush $DRUSH_ALIAS core-requirements --severity=2 | grep . && exit 1 || echo "No Drupal status requirement errors were found."
```
ElijahLynn, line 63:
```suggestion
  command: drush $DRUSH_ALIAS core-requirements --severity=2 | grep . && exit 1 || echo "Success: No Drupal status requirement errors were found."
```

ElijahLynn:
Legit test failure:
> [Symfony\Component\Yaml\Exception\ParseException]                                                                                                                                                                          
  A colon cannot be used in an unquoted mapping value at line 63 (near "command: drush $DRUSH_ALIAS core-requirements --severity=2 | grep . && exit 1 || echo "Success: No Drupal status requirement errors were found."").

tests.yml
```diff
@@ -57,3 +57,7 @@ va/tests/behat:
   command: |
     cd tests/behat
     behat --colors
+
+va/tests/status-error:
+  description: Check for Drupal status errors
+  command: drush $DRUSH_ALIAS core-requirements --severity=2 | grep . && exit 1 || echo "Success: No Drupal status requirement errors were found."
```
ElijahLynn, line 63:
```suggestion
  command: drush $DRUSH_ALIAS core-requirements --severity=2 | grep . && exit 1 || echo 'Success: No Drupal status requirement errors were found.'
```

docroot/modules/custom/va_gov_db/va_gov_db.install
swirtSJW: hook_update_n is supposed to return strings. https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_update_N/8.2.x
If we were not running such outdated drush, we would see the return in the log.  It also shows if you ever use update.php (which we don't here) as a success response in Drupal messages.

swirtSJW: correction, in the terminal, not the log.

Issue #2457, opened 2020-07-28
Update table caption help text


oksana-c:
## Description

Table caption states the following `This brief caption will be associated with the table and will help screen reader better describe the content within.` , which suggests that this caption will not be visible on the rendered page.

## Acceptance Criteria
* Table caption help text is updated to indicate that the caption will be visible once page is rendered. Suggested help text: `This brief caption will be associated with the table, rendered as a table heading in bold and will help screen reader better describe the content within.`

## Implementation

Table field widget may require some hard overrides, e.g. via form_alter.

oksana-c:
@rachel-kauff , please review suggested help text before this ticket moves to engineering.

rachel-kauff:
@oksana-c You got it. A few questions: where can I see the UI where this help text will be implemented, for context? Also, is this something we're trying to accomplish this sprint (or sooner?)

oksana-c:
@rachel-kauff Sprint 13/14 . UI can be examined when adding a Table paragraph to Main Content on the Detail page.

rachel-kauff:
@oksana-c Perfect, thanks!

oksana-c:
@rachel-kauff , did you review table caption text before this went out?

PR #3435, opened 2020-11-09
VACMS-2457: Updates TableField caption help text.

ndouglas:
## Description

See #2457. 

## Testing done
Checked the field settings form for a couple of fields.

## Screenshots
Screen Shot 2020-11-09 at 2 30 28 PM

## QA steps
As an administrator:
- [ ] The settings form for [a TableField field on a given content type](http://va-gov-cms.lndo.site/admin/structure/types/manage/health_care_local_facility/fields/node.health_care_local_facility.field_facility_hours) should read `This brief caption will be associated with the table, rendered as a table heading in bold and will help screen reader better describe the content within.`

### DIFF
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -315,7 +315,7 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
 function va_gov_backend_form_field_config_edit_form_alter(&$form, FormStateInterface $form_state, $form_id) {
   $widget_type = $form['default_value']['widget'][0]['#type'] ?? NULL;
   if ($widget_type === 'tablefield') {
-    $form['default_value']['widget'][0]['caption']['#description'] = 'This brief caption will be associated with the table, rendered as a table heading in bold and will help screen reader better describe the content within.';
+    $form['default_value']['widget'][0]['caption']['#description'] = t('This brief caption will be associated with the table, rendered as a table heading in bold and will help screen reader better describe the content within.');
   }
 }
 
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -310,6 +310,16 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
 }
 
 /**
+ * Implements hook_form_FORM_ID_alter() for field_config_edit_form.
+ */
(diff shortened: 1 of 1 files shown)


docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -309,6 +309,16 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
 
 }
 
+/**
+ * Implements hook_form_FORM_ID_alter() for field_config_edit_form.
+ */
+function va_gov_backend_form_field_config_edit_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  $widget_type = $form['default_value']['widget'][0]['#type'] ?? NULL;
+  if ($widget_type === 'tablefield') {
+    $form['default_value']['widget'][0]['caption']['#description'] = 'This brief caption will be associated with the table, rendered as a table heading in bold and will help screen reader better describe the content within.';
```
acrollet, line 318: I believe we should use the translation API here, multi-lingual is coming at some point...

ndouglas: mmm, good point

ndouglas: that's something I've been lousy at tbh

swirtSJW (approved):

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -309,6 +309,16 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
 
 }
 
+/**
+ * Implements hook_form_FORM_ID_alter() for field_config_edit_form.
+ */
+function va_gov_backend_form_field_config_edit_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  $widget_type = $form['default_value']['widget'][0]['#type'] ?? NULL;
+  if ($widget_type === 'tablefield') {
+    $form['default_value']['widget'][0]['caption']['#description'] = t('This brief caption will be associated with the table, rendered as a table heading in bold and will help screen reader better describe the content within.');
```
swirtSJW, line 318: Non-issue: This probably came in from UX but we should avoid describing the Front end result in too much detail as we are making promises that are not up to the CMS to keep ;)

ndouglas: Agreed.  I think this is probably a common (anti-)pattern with decoupled Drupal.

acrollet (approved):

Issue #3332, opened 2020-10-28
Q&A content proofing experience


oksana-c:
## Description

We have a prototype that describes the [Edit experience for Q&As](https://xd.adobe.com/view/5cd15a8f-426d-4356-b81e-82920dd688be-0982/?fullscreen) in context of a standalone page.
Now we need to consider content experience when standalone indicator is enabled/disabled.
Current proofing experience displays all fields regardless of whether Q&A standalone page is enabled, which wrongfully suggests that when the page is live, all standalone page fields will also be displayed.

## Acceptance Criteria

- [x] if field_standalone_page != true, then the fields contained in the "Q&A page components" fieldset are not displayed on the Q&A node view page 
- [x] if field_standalone_page = true, then the fields contained in the "Q&A page components" fieldset are displayed on the Q&A node view page in the same order as they are presented in the Q&A edit form + using the same grouping under "Q&A page components" fieldset
- [x] hidden fields do not affect node links report. e.g. hard unset, without manipulating page styles

PR #3447, opened 2020-11-10
VACMS-3332: Add logic to toggle standalone qa entity view display.

ethanteague:
## Description
See #3332 

## Testing done
Visual / behat

## Screenshots
**Enabled:**
image

**Disabled:**
image

## QA steps
- As an administrator, go to `node/8241/edit` and toggle off standalone display, and save the node.
- Visually verify that latest revision display looks like the disabled screenshot.
- Go back to edit page, toggle on standalone display, and save the node.
- Visually verify that the latest revision display looks like the enabled screenshot.


docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -330,6 +330,42 @@ function va_gov_backend_field_widget_moderation_state_default_form_alter(&$eleme
   }
 }
 
+/**
+ * Implements hook_entity_view_alter().
+ */
+function va_gov_backend_entity_view_alter(array &$build, EntityInterface $entity) {
+  if ($entity->getEntityTypeId() === 'node' && $entity->bundle() === 'q_a') {
```
swirtSJW, line 337: minor: This logic could just as easily live inside that function so that the concerns are not handled in two places.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -330,6 +330,42 @@ function va_gov_backend_field_widget_moderation_state_default_form_alter(&$eleme
...
+/**
+ * Changes qa entity view display when standalone page option is selected.
+ *
+ * @param array $build
+ *   The entity display array.
+ * @param Drupal\Core\Entity\EntityInterface $entity
+ *   Instance of EntityInterface.
+ */
+function _va_gov_backend_qa_toggle_standalone_display(array &$build, EntityInterface $entity) {
+  $exclusions = [
```
swirtSJW, line 351: minor:  Defining this could be done within the  if statement below.   That way it is not defined, and taking up memory if is never going to be used.

moderate: `$exclusions` is very generic and does not speak to its intent.  It would need a comment to give some clarity as to the reason and intention.  Something like `$standalone_only_fields` would make it so the reason is understood right from the code.

swirtSJW (dismissed):

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: Great points :+1:

docroot/modules/custom/va_gov_backend/va_gov_backend.module
swirtSJW: That wasn't quite what I meant.  LOL  I meant move the if conditional into the function that was called.  Not move the logic from the function into the if conditional.   This works, just without the clarity that having it all in a function provides.  This is fine.

ethanteague: Ha! My bad, brother :+1:

swirtSJW (approved):

swirtSJW:
Testing looks good.

Non standalone
image

Yields 
image



checked Stand alone.
image

Yields
image

ethanteague:
Thanks for taking the time to test this, @swirtSJW

Issue #3253, opened 2020-10-21
Investigate why Grafana/Prometheus is showing downtime just after deploys


ElijahLynn:
Current downtime is measured in a once a minute script that returns a failure if prod.cms.va.gov does not return a 200 within 5 seconds. 

Possible issues discussed:
- Could be returning a 503 during/after deploy (if this is true, create a new issue for this) (doesn't seem to be the case, or if it is, the script is still counting it as a success)
- CMS Export is running just after deploy, causing responses to be longer than 5 seconds.

AC:
- [ ] No downtime shown/tracked when site is in deploy mode during a deployment in Grafana.

Tasks:
- [x] Set the User-Agent on request header
- [x] Set the deploy mode to check for a request header of User-Agent: curl.* (User-Agent: curl/7.61.1)

![image](https://user-images.githubusercontent.com/1504756/96786935-cde80a80-13a5-11eb-8b4e-8f37c5ba62a8.png) (the small gaps don't actually say 0, but say 1 (success) and is by design in the chart, this indicating no downtime, the big gap says 0 though which is during the export) This image is for a 15 minute deploy window. 

http://grafana.vfs.va.gov/d/dxf8a-6Zz/cms-dashboard?orgId=1&from=1603308563841&to=1603309484082&var-datasource=Prometheus%20(Utility)&var-environment=prod&var-app_name=cms&var-region=us-gov-west-1

ElijahLynn:
Got some good data on this right now:

I just did a manual "live deploy" job which normally runs after deploy, but failed earlier, so I replayed it around 12:46pm PT. Deploy finished at 12:39pm PT, for reference in the screenshots. I monitored server load through grafana and refreshed the login page (same as uptime check) and throughout all the requests never had a single one take longer than 1 second (and this is through a proxy tunnel too) and the majority where under 50ms. So the hypothesis that the CMS Export process could have been slowing things down is not validated. 

Which leads me to believe that it is something to do with the site being in deploy mode or the site alert functionality, which is still up when the export job (CMS Live) is run. I'll test that now on CMS-TEST. 



![image](https://user-images.githubusercontent.com/1504756/98596683-a1306000-228c-11eb-9c08-ce5ce1d683bc.png)
![image](https://user-images.githubusercontent.com/1504756/98596689-a2fa2380-228c-11eb-94d0-60b13ea35eb9.png)

ElijahLynn:
[...]

PR #3464, opened 2020-11-12
VACMS-3253 downtime status check

indytechcook:
## Description

See #3253. This update the health check deploy plugin to return 200 for prometheus.  See new tests for tests :)

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/3464)

### DIFF
--- a/docroot/modules/custom/va_gov_backend/src/Deploy/Plugin/HealthCheck.php
+++ b/docroot/modules/custom/va_gov_backend/src/Deploy/Plugin/HealthCheck.php
@@ -10,12 +10,19 @@ use Symfony\Component\HttpFoundation\Request;
  */
 class HealthCheck implements DeployPluginInterface {
 
+  public const USER_AGENT = 'curl-prometheus-check';
+
   /**
    * {@inheritDoc}
    */
   public function match(Request $request): bool {
     $current_path = $request->getPathInfo();
     if ($current_path === '/health') {
+      return TRUE;
+    }
+
+    $user_agent = $request->headers->get('User-Agent');
+    if ($user_agent === static::USER_AGENT) {
       return TRUE;
     }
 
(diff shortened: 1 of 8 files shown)


docroot/modules/custom/va_gov_backend/src/Deploy/Plugin/HealthCheck.php
```diff
@@ -10,6 +10,8 @@
  */
 class HealthCheck implements DeployPluginInterface {
 
+  public const USER_AGENT = 'curl-prometheus-check';
```
acrollet, line 13: Just seeing this constant name at first glance, I think I would expect it to refer to the user agent being used by the class to make a request. Perhaps something like VAGOV_DASHBOARD_USER_AGENT would be more self-documenting?

tests/phpunit/Deploy/DeployServiceTest.php
```diff
@@ -0,0 +1,70 @@
...
+    $settings = Settings::getAll();
+    $this->assertArrayEquals(
+      $orig_settings,
+      $settings,
+      'Deploy Serviced initiated Settings correctly.'
+    );
+  }
+
+  /**
+   * Test do make sure DeployPluginInterfaces are returned.
```
acrollet, line 54:
```suggestion
   * Test to make sure DeployPluginInterfaces are returned.
```

docroot/modules/custom/va_gov_backend/src/Deploy/Plugin/HealthCheck.php
```diff
@@ -19,6 +21,11 @@ public function match(Request $request): bool {
       return TRUE;
     }
 
+    $user_agent = $request->headers->get('User-Agent');
+    if ($user_agent === static::USER_AGENT) {
+      return TRUE;
```
acrollet, line 26: It feels like a comment explaining why we do this may be useful to future devs.

acrollet (commented):
just small stuff :-)

acrollet (approved):

ndouglas (approved):
LGTM.

ElijahLynn (approved):

ElijahLynn:
> There was 1 error:
> 
> 1) test\phpunit\Deploy\DeployServiceTest::testRun
> Error: Class 'test\phpunit\Deploy\DeployServiceMock' not found

indytechcook:
I'm still looking into why devshop isn't autoloading the mock class when lando does.

Issue #3449, opened 2020-11-10
Paragraphs should be forced to clone on node clone


indytechcook:
**Describe the defect**
When a node is cloned and the checkbox to clone a paragraph entity field is not checked then a paragraph is referenced from multiple nodes.  This reference breaks the concept of a paragraph having a single parent.

AC
* When clone a node with a paragraph, a new paragraph entity is always created.

indytechcook:
Related D.O issue: https://www.drupal.org/project/entity_clone/issues/3060223

PR #3474, opened 2020-11-13
VACMS-3449: Forces paragraphs to clone and hides their checkboxes on the clone form.

ndouglas:
## Description

See #3449. 

## Testing done
Attempted to clone a node with paragraphs and noted that the form no longer contains checkboxes for the paragraph children of the original node.  Clone completed successfully and verified in Views that this created new paragraphs whose parents were the newly cloned node.

## Screenshots
Screen Shot 2020-11-13 at 8 17 47 AM

## QA steps
As any user with the permission to clone a node:
1. Clone [a node with paragraphs](http://va-gov-cms.lndo.site/entity_clone/node/2768)
1. Verify that:
  - [ ] the clone form does not give the option to avoid cloning the paragraphs
  - [ ] the clone operation succeeds and the new node has paragraphs that are distinct from the paragraphs of the original node

## Definition of Done
It's just a config update.

### DIFF
--- a/config/sync/entity_clone.settings.yml
+++ b/config/sync/entity_clone.settings.yml
@@ -75,15 +75,11 @@ form_settings:
     hidden: true
     default_value: false
     disable: false
-  workflow_participants:
-    hidden: true
-    default_value: false
-    disable: false
   menu_link_content:
     hidden: true
     default_value: false
     disable: false
   paragraph:
     default_value: true
-    disable: false
-    hidden: false
+    disable: true
+    hidden: true


ndouglas:
@indytechcook Turns out this is already a thing, see http://va-gov-cms.lndo.site/admin/config/system/entity-clone .  Hey, at least it's done, right?  I just went ahead and hid the paragraph elements of the form too since it's wonky and unnecessary if we're always cloning them, but that can be undone obv.

indytechcook:
this is great!

indytechcook:
Config looks good

Entity clone settings | VA gov CMS 2020-11-13 10-14-45
Clone Content | VA gov CMS 2020-11-13 10-14-31

indytechcook (approved):

indytechcook:
Adding a note that the UX change of removing the options was reviewed with @kevwalsh

Issue #3470, opened 2020-11-12
Wrong revision is published via bulk "Publish latest revision" action


oksana-c:
**Describe the defect**

**To Reproduce**
1. Go to /admin/content/bulk?title=Psychology+Internships+and+Fellowships&type=health_care_region_detail_page&moderation_state=All&taxonomy_entity_index_tid_depth=All  
1. Use "Publish latest revision" action on "Psychology Internships and Fellowships" detail page
2. Note Oct 29 revision is published instead of Nov 5 - see "Recent changes" sidebar on the node view page.

<img width="1686" alt="Screen Shot 2020-11-12 at 11 45 01 AM" src="https://user-images.githubusercontent.com/16477724/98982836-36299800-24dd-11eb-92b3-6979c19a5938.png">
<img width="1362" alt="Screen Shot 2020-11-12 at 3 03 40 PM" src="https://user-images.githubusercontent.com/16477724/99001668-7ba78e80-24f8-11eb-8e6d-412fd3ad3e9f.png">

**Expected behavior**
Bulk  "Publish latest revision" action publishes latest revision.

## Acceptance criteria
- [ ] Bulk action "Publish latest revision" publishes latest revision
- [ ] validate that revisions are shown in correct order in "Recent changes" sidebar on nodes. If not -> open a follow up story.

## Implementation Details

The assumption is that reworking the View per notes below may address the defect of publishing the latest revision.

* views that need to show latest revision instead of latest published revision
     * /admin/content/*
     *  /admin/structure/views/view/taxonomy_term/edit/page_1
     *  /admin/content/va-forms

If the bulk action can be fixed without reworking view -> split Views rework into a new story.

acrollet:
This code lives in Drupal\va_gov_bulk\AdminModeration::publish()

oksana-c:
@indytechcook 🎩 off to your persistence to tackle the action issue first. I didn't realize it was a custom one. @acrollet 👏

acrollet:
@daddison48 this ended up being more like a 5 in terms of effort in case it's important for reporting.

PR #3478, opened 2020-11-13
VACMS-3470: fix publish latest revision action

acrollet:
## Description

See #3470.

## Testing done

Manual testing, added unit test for new service.

## QA steps

As an admin user:
1. Open the Bulk Edit Content page (/admin/content/bulk?title=Psychology+Internships+and+Fellowships&type=health_care_region_detail_page&moderation_state=All&taxonomy_entity_index_tid_depth=All)
1. Open the "Psychology Internships and Fellowships" node in a new tab and save a draft with a new title
1. On the Bulk Edit Content page, select the "Psychology Internships and Fellowships" and apply the "Publish Latest Revision" action
1. Validate that:
- [x] The published revision has the new title


indytechcook:
It looks good code wise.  I nice refactor.  I'm waiting until test pass to do the review in devshop.

ndouglas (dismissed):
I tested locally @indytechcook and it worked as expected.  Code LGTM too.

acrollet:
@oksana-c pinging in case you'd like to validate this.

docroot/modules/custom/va_gov_bulk/src/Service/ModerationActions.php
```diff
@@ -0,0 +1,103 @@
...
+
+    if ($node instanceof RevisionLogInterface) {
+      $node->setRevisionCreationTime($this->datetimeTime->getRequestTime());
+      $node->setRevisionLogMessage('Bulk operation create archived revision');
+      $node->setRevisionUserId($this->currentUser->id());
+    }
+    $node->save();
+
+    $node = $this->entityTypeManager->getStorage('node')->load($node->id());
+    $node->set('moderation_state', 'draft');
```
oksana-c, line 62: is there a reason why we would want to create a `draft` after creating `archived` version?

docroot/modules/custom/va_gov_bulk/src/Service/ModerationActions.php
```diff
@@ -0,0 +1,103 @@
...
+   *   The node.
+   */
+  public function unpublishCurrentRevision(NodeInterface $node) : NodeInterface {
+    $node->set('moderation_state', 'archived');
+
+    if ($node instanceof RevisionLogInterface) {
+      $node->setRevisionCreationTime($this->datetimeTime->getRequestTime());
+      $node->setRevisionLogMessage('Bulk operation create archived revision');
+      $node->setRevisionUserId($this->currentUser->id());
+    }
```
oksana-c, line 58: revision info is not applied on save. 

`$node instanceof RevisionLogInterface` - did you mean `NodeInterface`? do we need `if` statement here?

oksana-c (changes requested):
attaching annotated screenshots. We may need a 16th min on this one today

### testing VBO on existing content


let's rename these two custom actions while we're at it

rename actions

----

### Observations while creating new content then running it thru Unpublish bulk action

Revisions in content region seems to align, although no revision messages are added and expect that author and timestamp are also taken from the last revision before the VBO was ran. 
In the right sidebar, the view shows an extra Draft revision, which is expected, because current logic in this PR saves Archived and then Draft revision. I can guess the rationale why Draft was added as a last revision, although I think we should not add Draft after Archived for Unpublish action, since users need to get accustomed to the concept of "Archived" and make a conscious decision on whether or not content gets resurrected as Draft after been Archived.
Screen Shot 2020-11-16 at 7 02 02 AM

strangely, Draft is displayed as last status in the bulk content view
Screen Shot 2020-11-16 at 7 02 42 AM

docroot/modules/custom/va_gov_bulk/src/Service/ModerationActions.php
acrollet: good question, I assumed that this was an original requirement.

docroot/modules/custom/va_gov_bulk/src/Service/ModerationActions.php
acrollet: Good point, I think this action was originally meant to be available for all entities so I believe that was the reasoning behind `RevisionLogInterface`. However there were other dependencies on the object being a node, so I made the logic explicit and I think we can remove the `if` statement.

acrollet:
@oksana-c let's definitely 16th minute this and get on the same page since we're talking about changing the spec. :-)

acrollet:
I'm also seeing that the content creation time is shown in the 'Recent changes' view, which means that when a revision is published via bulk action, the date shown is when the draft was created rather than when it was published. Let's discuss whether to change that behavior as well.

acrollet:
Capturing action items discussed in 16th minute:

- [ ] change "Unpublish current revision" to "Archive selected content (Unpublish) "
- [ ] Archive action should not also create a new draft revision
- [ ] change "Publish Latest Revision" to "Publish latest revision"
- [ ] update "Recent changes" view to display revision update timestamp instead of revision created timestamp

oksana-c (approved):
both actions work well. I'm opening a follow up issue to use Revision Created time instead of Changed for sorting the right sidebar "Recent changes" view, since it's definitely out of scope here.
✅ good to merge from me

acrollet:
@indytechcook @ndouglas if I could get a re-pass on the CR from one of you that would be awesome, thanks :-)

ndouglas:
@acrollet What's the timetable?  I should be able to get to this today if it's not _too_ rushed, but it might be a couple hours.

acrollet:
I think we're past today's window to test and deploy, so any time today should be fine. cc @oksana-c

ndouglas (approved):
I confirmed that this worked as expected and reflected the update to the action name, etc.

acrollet:
many thanks @ndouglas !

PR #3476, opened 2020-11-13
VACMS-3182: Paragraph Remove Button

kaise-lafrai:
## Description

See _issueid_. https://app.zenhub.com/workspaces/vagov-cms-team-5c0e7b864b5806bc2bfc2087/issues/department-of-veterans-affairs/va.gov-cms/3182

## QA steps

- Login to preview environment
- Navigate to the following pages and confirm that the **Remove** button is no longer visible within the **Include Alert** section and that you can successfully **Save**.
  - [ ] `/node/add/checklist`
  - [ ] `/node/add/faq_multiple_q_a`
  - [ ] `/node/add/support_resources_detail_page`
  - [ ] `/node/add/media_list_images`
  - [ ] `/node/add/media_list_videos`
  - [ ] `/node/add/q_a`
  - [ ] `/node/add/step_by_step`

- Navigate to existing content types to confirm that the *Remove** button is no longer visible within the **Include Alert** section.
  - [ ] `checklist`
  - [ ] `faq_multiple_q_a`
  - [ ] `support_resources_detail_page`
  - [ ] `media_list_images`
  - [ ] `media_list_videos`
  - [ ] `q_a`
  - [ ] `step_by_step`

## Definition of Done
- [ ] Documentation has been updated, if applicable.

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change? 
- [ ] Yes, and it's written in issue ____ and queued for publication. 
  - [ ] Merge and ping @ rachel-kauff so she's ready to publish after deployment
- [ ] Yes, but it hasn't yet been written 
  - [ ] Don't merge yet -- ping @ rachel-kauff to prompt her to write and queue content
- [ ] No announcement is needed for this code change. 
  - [ ] Merge & carry on unburdened by announcements

PR #3520, opened 2020-11-19
VACMS-3182 paragraph remove button

swirtSJW:
## Description

THis is an exact copy of https://github.com/department-of-veterans-affairs/va.gov-cms/pull/3476 because devship fell off the rails completely and wil  not build the environement for 3476

See all testing steps on 3476


swirtSJW:
Closing this since its twin was merged.

PR #3515, opened 2020-11-18
VACMS-3507: implement alternative publishing workflow

acrollet:
## Description

See #3507 

## Testing done

Updated and ran behavioral tests, spot-checked content types.

## Screenshots

Screen Shot 2020-11-18 at 12 55 27 PM

## QA steps

As a content publisher:

1. For at least two different content types,
1. Validate Acceptance Criteria from issue:
- [ ] The Editorial workflow block on the edit page shows 'Save as' instead of 'Change to' before the moderation state selection widget
- [ ] When creating a new piece of content, the default state is 'Draft'
- [ ] When editing a published piece of content, the default state is 'Draft'

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [x] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

### Does this PR need review from a Product Owner

- [ ] `Needs PO review`

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change? 
- [x] Yes, and it's written in issue #3529  and queued for publication. 
  - [ ] Merge and ping @ rachel-kauff so she's ready to publish after deployment
- [ ] Yes, but it hasn't yet been written 
  - [ ] Don't merge yet -- ping @ rachel-kauff to prompt her to write and queue content
- [ ] No announcement is needed for this code change. 
  - [ ] Merge & carry on unburdened by announcements

PR #3525, opened 2020-11-19
VACMS-000: Sync db with config (remove drift / export noise).

ethanteague:
Prod and config have gotten out of sync, causing a lot friction when exporting code locally for commit.


acrollet:
FYI I also exported prod config in https://github.com/department-of-veterans-affairs/va.gov-cms/pull/3515

ethanteague:
Closing in favor of https://github.com/department-of-veterans-affairs/va.gov-cms/pull/3515

Issue #3468, opened 2020-11-12
Collect User ID via GA/GTM


oksana-c:
## User Story

As a researcher, I would 

## Description

Benefits of User ID tracking:

* **More accurate unique user count**
* **Data can be segmented by logins**. Logged in users (editors) behave differently than anonymous or guest users (CMS homepage, training materials). User ID tracking allows for comparison of the behavior of logged-in to logged-out users, providing more reliable metrics for both groups.
* **Additional reports in GA**. User ID tracking creates a new view that collects only unique user data - logged in only. Some granular reports in this view aren’t available in regular GA views, and even those that are aren’t based off accurate user metrics. Cross-device behavior data will become available.

---

The CMS team intends to track User ID as one of the User Scope dimensions for the CMS application. This will require enabling User ID tracking in GA and sending User ID in GTM dataLayer via pageView event.

Drupal assigns a unique User ID to each user which then, if included in GA reports, can be traced back to a specific person in the CMS. The CMS wants to prevent the exposure of any identifiable user data, so we agreed on the encryption method for tracking User ID which will prevent traceability back to a specific person. We plan on sending a hash of Drupal user id instead of sending the id directly. This way we are still able to group sessions by unique users, but also preventing traceable information from ending up in GA reports.

## Acceptance Criteria
- [x] `user_id` dataLayer value is populated with hashed Drupal UID on pageView event 
   * use Drupal's [hashBase64()](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Crypt.php/function/Crypt%3A%3AhashBase64/8.2.x ) to encrypt UID before passing to dataLayer
- [x] GTM container variable for User ID is added in draft container version [@oksana-c ]
- [x] User ID view exists for the CMS Prod GA property [@oksana-c]
- [x] dataLayer value is verified in Preview and Debug GTM mode [@oksana-c]
- [x] a request to publish latest version of the CMS Prod GTM container is submitted to Insights & Analytics team [@oksana-c]
- [x] User ID data is reflected in GA reports on production

PR #3554, opened 2020-11-23
VACMS-3468: Collect User ID via GA/GTM

ndouglas:
## Description

See #3468. 

## Testing done
Added function to Behat context to concisely express the test conditions and added Behat tests for test-created users of a couple of roles.  Manually tested with my account and an anonymous account.

I'm not crazy about how the tests for the newly created users don't compare against a known sane value.

For instance, I'd prefer the test to look something like this:
```gherkin
  @user
  Scenario Outline: Google Tag Manager should be provided with an appropriate hashed UID for the current user.
    Given I am logged in as a user with the <role> role
    And I am on "/user"
    And google tag manager data layer value for "userId" should be <hashed>
    Examples:
    | uid | hashed                                          |
    | 1   | 'a4ayc_80_OGda4BO_1o_V0etpOqiLx1JwB5S3beHW0s'   |
    | 2   | '1HNeOiZeFu7gP1lxi5tdAwGcB9i2xR-Q2jpmbuwTqzU'   |
    | 3   | 'TgdAhWK-24tgzgXB3s_jrRa3IjCWfeAfZAt-Rym0n84'   |
```

But this doesn't seem to be trivially accomplishable and adding that sort of mechanism is probably outside the bounds of this issue.  I'm a Behat newbie though, so I'll cheerfully accept suggestions/modifications/demands.

## Screenshots
Screen Shot 2020-11-23 at 10 00 29 AM
Screen Shot 2020-11-23 at 11 31 31 AM

## QA steps

As any user:
1. Find your Drupal user ID.
1. Then run the following command: `lando drush eval "echo \Drupal\Component\Utility\Crypt::hashBase64('<insert your user ID here>')"` to derive a user ID as hashed by Drupal's [`hashBase64()`](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Crypt.php/function/Crypt%3A%3AhashBase64/8.2.x) function.
1. Then load any page and enter the following in the Developer Console: `dataLayer[1].userId`
    - [ ] Validate that this value matches the hashed user ID returned by the console command in step 2.

### Definition of Done

- [ ] Acceptance Criteria in related issue are met (involves updating GTM and GA)

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `Needs PO review`


acrollet:
@ndouglas the one thing I can think of is to add a step condition something like:

`And the google tag manager userID value for the logged in user should match the base64 hash of their uid`

I believe that with the behat Drupal context it shouldn't be too difficult to grab the user ID, but it's not something I've done in the past either. Happy to co-work on this too.

acrollet:
@ndouglas oh jeez, that's pretty much what you already did. I'm going to finish my coffee before commenting again :p

ndouglas:
Well, I appreciate the validation 😁

oksana-c (commented):
quick 2 cents:
The goal of User ID is to be able to identify unique identifiable users (logged in).
I see on screenshots that we assign anonymous users an ID. We should only assign/send an ID for logged in users who have numeric UID that is not = 0

ndouglas:
Yeah, good point.  Fixed.

oksana-c (approved):
📈 ✅

Issue #3378, opened 2020-11-04
"Save and select" button fails on media image upload screen, when user unselects "reusable".


kevwalsh:
**Describe the defect**

When adding a new image to the media library in a Story, and making it reusable, the Save and select button fails. 

**To Reproduce**
1. Go to /node/add/news_story
2. Add media 
3. Add a file from your local machine and fill out required fields
4. Uncheck "reusable"
5. Click "save and select"
6. Note that you are returned to the media library browser, but nothing is selected. This is because non-reusable images do not appear on this screen. 

**Expected behavior**

There should be no option that fails to add an image. 

Possible solutions

1. Remove reusability from the content model. It's not in use yet. We can always bring it back later. 
2. Remove "Save and select" button when "Reusable" is checked. 
3. Remove "Save and select" button in contexts where media field is single value. There's no need to be returned to a multi-select form in those contexts.  This has the advantage of a less confusing UI (only one possible action). 
 
Some background: 
* Media was installed with D8.6, and Media updates in 8.8 changed the UI, including the browser and button behavior. 

**Screenshots**

See this video 

https://www.loom.com/share/645115f02bb24fc89094dfc371aa0a20

![Create_Story___VA_gov_CMS.jpg](https://images.zenhubusercontent.com/5d6e57ad330f64311395f08b/a40940d5-033c-40bc-9f40-7db9073e1054)

kevwalsh:
@oksana-c @ndouglas 

Decision reached on Nov 25: "Save and select" should only be visible on media browser screens when Reusable is checked. This allows for editors who want multiselect checkboxes on the ensuing screen to access it when their content is reusable.

ndouglas:
I dug into `#states` trying to figure out what was going wrong, and frankly it defeated me.

The issue is that the dialog loads the form inline and then uses JS to clone the buttons and move them into the dialog, then hides the cloned buttons.  This seems a bit janky to me, but it also breaks `#states`.  A patch was written a year or two ago that copies the `data-drupal-states` attribute from the copied button to the clone, which in theory should allow the clone to observe the state changes, but this stopped working (assuming it ever worked reliably).  The state changes continue to affect the original button and not the clone.  I've played with a few approaches, from forwarding messages to simply deleting the original button, and none of them have worked reliably (and almost surely cause issues elsewhere).  This is a bit of a complicated form operation within a dialog, and I think both the dialog and states components are... ehh.

[...]

PR #3614, opened 2020-12-02
VACMS-3378: Selecting and unselecting "Reusable" shows and hides the "Save and select" button...

ndouglas:
…in the media library dialog.

## Description

See #3378.

This is ugly.

As documented in #3378, I attempted to solve this using the official solution of the Forms API `#states` system, but that clashes horribly with the Core `dialog` system and both seem at least a little broken individually.  I attempted yesterday to fix the intersection between the two, and realized after some time that I'm not the man to do it.  At least, not yet.

The solution I ended up going with took two parts:
- added a custom class to the `field_media_in_library` checkbox widget to make it easier for jQuery to identify cleanly.  I could've used regex tests but I felt this might be cleaner and more transparent.
- attached a custom library to the media upload form which scans for appearances of that specific checkbox widget and listens to their events; when the checkbox is unchecked, it hides the dialog button, and when the checkbox is checked, it shows the dialog button.  Additionally, it will check the existing state of the checkbox and set the visibility of the button accordingly, which is significant because reloads of the form (for instance, in response to errors) can present the checkbox and button in mismatched states.

I bet there's a better way to do this, and I'm completely open to suggestions/corrections/pitchfork-shaking demands.  If there is, I couldn't figure it out 😩 

## Testing done
Repeatedly tried to add media, clicked "Reusable" a few times, added media with existing media, etc.  No automated tests yet.  I think that would be a good idea, so... forthcoming?

## Screenshots
Screen Shot 2020-12-02 at 12 38 13 PM
Screen Shot 2020-12-02 at 12 38 20 PM

## QA steps
As a user with the ability to edit nodes containing media content blocks:
1. Edit [a node that can have attached media](http://va-gov-cms.lndo.site/node/7016/edit)
1. Click "Add Content block".
1. Select either "Embedded Image" or "Link to File or Video".
1. Click "Add media".
1. Click "Choose file" and upload a file from your machine.
1. Then validate Acceptance Criteria from #3378:
- [ ] the "Save and Select" button should be visible when "Reusable" is checked
- [ ] the "Save and Select" button should be invisible when "Reusable" is unchecked

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `Needs PO review`

This is going to simplify the UX, so I assume this will need an announcement.

- [x] Yes, but it hasn't yet been written 
  - [x] Don't merge yet -- ping @ rachel-kauff to prompt her to write and queue content


docroot/modules/custom/va_gov_backend/js/media_library_add_form_upload.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @file
+ */
+
+(function ($, Drupal) {
+
+  Drupal.behaviors.vaGovMediaLibraryReusable = {
+    attach: function () {
+      $(document).ajaxComplete(function (event, xhr, settings) {
+        $('input.field_media_in_library[type=checkbox]:not(checked)').each(function () {
```
acrollet, line 10: The default state of the button is to be visible, right? Could we handle this in the form alter and hide the button with FAPI if the state of the checkbox is not checked? Or do the multiple layers of forms make that difficult? Just to avoid potential "flashing" during load...

docroot/modules/custom/va_gov_backend/js/media_library_add_form_upload.js
```diff
@@ -0,0 +1,30 @@
...
+          $('.ui-dialog-buttonpane button').first().hide();
+        });
+        $('input.field_media_in_library[type=checkbox]:checked').each(function () {
+          $('.ui-dialog-buttonpane button').first().show();
+        });
+        $('input.field_media_in_library')
+          .off('.vagov_media_library')
```
acrollet, line 17: could we avoid this by wrapping with [.once()](https://www.drupal.org/docs/drupal-apis/javascript-api/javascript-api-overview)? Or maybe there's a pre-existing event handler I don't know about :-)

docroot/modules/custom/va_gov_backend/js/media_library_add_form_upload.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @file
+ */
+
+(function ($, Drupal) {
+
+  Drupal.behaviors.vaGovMediaLibraryReusable = {
```
acrollet, line 7: maybe this could be more self-documenting? vaGovMediaHideSelectButtonWhenNotReusable? Or maybe that's just silly as I look at it :-P

acrollet (commented):
I left some picky comments but the overall approach is good by me. :-)

docroot/modules/custom/va_gov_backend/js/media_library_add_form_upload.js
ndouglas: That's worth a try.  I'll check it out.

ndouglas: So the reason for that code is that when the form submission fails validation, the form will be redrawn with visible buttons, regardless of it having been hidden by JS.  So if the checkbox is unchecked, it will be in an inconsistent state -- someone can fill out some fields, click "Save and select," and they'll be in the same mess they were in to begin with.

We don't need to worry about the case where the reusable box is checked, so this:

```javascript
        $('input.field_media_in_library[type=checkbox]:checked').each(function () {
          $('.ui-dialog-buttonpane button').first().show();
        });
```

is unnecessary.  I think I stuck that in on auto-pilot.

Based on my testing this morning, the form alter hook isn't invoked when the form fails validation (guess it's just validated client-side and redrawn, which makes sense now that I think about it).  So I think the JS solution is necessary.

So I'm removing that unnecessary snippet and I think otherwise not making any changes here, unless you see any flaws here.

docroot/modules/custom/va_gov_backend/js/media_library_add_form_upload.js
ndouglas: Wow, that's _exactly_ what I needed there.  ❤️

docroot/modules/custom/va_gov_backend/js/media_library_add_form_upload.js
ndouglas: Improved this a bit.  Possibly.

docroot/modules/custom/va_gov_backend/js/media_library_add_form_upload.js
acrollet: nope, looking good to me! I'll review on the PR env after my current review.

acrollet: annnnd devshop completely lost the env - I'll pull it local.

ndouglas: heh. devshop hadn't blown up on me this week, so I guess it was just a matter of time.

acrollet (approved):
Tested working locally, code looks good.

PR #3634, opened 2020-12-04
VACMS-000: Add clp edit page to toggle logic.

ethanteague:
## Description
We missed adding the edit page - only covered the clp add page


swirtSJW (approved):
Good catch

PR #3675, opened 2020-12-09
Add back code scanning 

indytechcook:
## Description

This is a follow up of #3620 and #3638 

cc @zkoppert

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/3675)


ndouglas (approved):
Checked against the usage doc, seems legit.

PR #3828, opened 2020-12-18
Add nvm and pin node version to be compatible with vets-website

indytechcook:
## Description

To keep our code compatible with vets-website, I'm adding the node version and a nvm file.

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/3828)


package.json
```diff
@@ -42,6 +42,9 @@
     "axe",
     "tests"
   ],
+  "engines": {
```
acrollet, line 45: Matches https://github.com/department-of-veterans-affairs/vets-website/blob/74d62bd5a9a732299ab51ae543641b0cc759ac14/package.json#L272

Version actually installed also matches:

Screen Shot 2020-12-21 at 8 45 38 AM

acrollet (approved):

Issue #3011, opened 2020-09-28
Optimize Composer Install


indytechcook:
- [ ] Composer install should run with the following arguments when building for production:

`composer install --prefer-dist --no-dev --optimize-autoloader --classmap-authoritative --apcu-autoloader`

ndouglas:
@indytechcook advised me to look into adding this to the `config` key in `composer.json`, as this seemed to be the cleanest way to make changes throughout Composer with less risk of encountering issues with Jeff Geerling's ansible stuff, etc.  The drawback of this is that some options, namely `classmap-authoritative`, might be unsuitable for non-BRD environments.

Per [this](https://getcomposer.org/doc/articles/autoloader-optimization.md), though, it seems that `classmap-authoritative` and `apcu-autoloader` should not be used together.  Since `apcu-autoloader` seems to be a good choice and I think APCu is available in all of our environments, I'm just going to add that and forget anyone ever mentioned `classmap-authoritative`.

indytechcook:
The php ini change has been made on devshop

`apc.enable_cli = 1` on /etc/php-7.1.d/50-apc.ini

PR #3855, opened 2020-12-23
VACMS-3011: Optimize Composer install

ndouglas:
## Description

See #3011. 

This could use some discussion, as it'll impact every environment and every operation involving Composer.

There's some discussion to be had about whether these flags are necessary or sufficient, whether this is the appropriate location for these changes, etc.

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `Needs PO review`


ndouglas:
@indytechcook @acrollet Any thoughts about this or my [rationale](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/3011#issuecomment-750441649)?

acrollet:
@ndouglas I agree that the settings are necessary, I can't find hard numbers with a quick search, but they should provide a performance improvement. I think we may need to turn off the default class loader to let apcu be used though: https://git.drupalcode.org/project/drupal/-/blob/8.9.x/sites/default/default.settings.php#L455

NB, the setting shouldn't be necessary once we're on D9: https://www.drupal.org/node/3116384

ndouglas:
>I think we may need to turn off the default class loader to let apcu be used though:

ugh, I came across that but misread it :facepalm: I saw "Symfony" and read "Composer" and drew the exact opposite conclusion.

acrollet:
> ugh, I came across that but misread it 🤦 I saw "Symfony" and read "Composer" and drew the exact opposite conclusion.

The docs on it are not super clear! It should probably be verified empirically...

ndouglas:
Yeah, I was just about to post and say I'm going to try and do some primitive benchmarking.  The fact that there are multiple competing APCu-optimized autoloaders and your reveal that this'll be moot in D9 dampens my ardor a bit.

acrollet:
tbf I'm not sure that the composer settings aren't necessary, it seems that the drupal config setting will no longer be necessary?

indytechcook (changes requested):
Go ahead and this to the `settings.php`

```
$settings['class_loader_auto_detect'] = FALSE;
```

ndouglas:
So I ran this benchmark script three times yesterday (disregard the "Minimum" figure):

1. Baseline from `master`:
3011-Baseline

2. With `composer.json` changes but no change to `settings.php`:
3011-Composer-No-Settings-Change

3. With `composer.json` changes and the change to `settings.php`:
Screen Shot 2020-12-30 at 9 14 12 AM

Not a substantial amount of difference, though it does seem that the change to `settings.php` makes a difference that's outside the margin of error.

As each test iteration (if you haven't read the script, it does a `drupal cr` and then a `lando --tags="content_editing"`) takes about 6 minutes, I think the overhead from Lando invocation, etc, should be fairly minimal.  I also think the test covers the basic editing tasks we'd be interested in optimizing.

It might be that the Behat part of the equation dilutes the effectiveness of the changes and that I should run, say, the PHPunit tests instead.

acrollet:
> It might be that the Behat part of the equation dilutes the effectiveness of the changes and that I should run, say, the PHPunit tests instead.

I would guess that's true, at least some CPU time will be spent on overhead and the browser process...

ndouglas:
That settings.php change seems to break PHPUnit entirely, so now I'm working to figure out why...

ndouglas:
The `settings.php` change broke PHPUnit because Lando's PHP config doesn't enable APC(u) for the CLI.  After doing that, we have a baseline of:
3011-2-Baseline

And an after-changes measurement of:
Screen Shot 2020-12-30 at 1 51 07 PM

So... worse.

I'm gonna recheck the baseline because there was a `lando rebuild` in there, so there were some other changes external to the `composer.json` and `settings.php` changes we're trying to measure.

ndouglas:
Recheck:

Without changes:
3011-3-without-changes

With changes:
3011-3-with-changes

I think a substantial part of the difference between the two configurations earlier was due to the anomalous 386-second run, which was also the first run after the `lando rebuild` -- that might've been longer for a few different reasons.

Anyway, I don't know if this change is going to rock our socks or not, and I'm not aware of a better proxy -- at least, a quick one -- to measure general performance than running a broad variety of tests and timing them, so 🤷

indytechcook (dismissed):

ndouglas:
@indytechcook Sorry, I should've removed the benchmark script clutter, forgot I'd added that 😩

.lando/zzz-lando-my-custom.ini
```diff
@@ -1,6 +1,7 @@
 [PHP]
 ; File is named zzz-lando-my-custom.ini because Lando renames on copy to /usr/local/etc/php/conf.d/
 
+apc.enable_cli = 1
```
indytechcook, line 4: Actually, do we need this in our other environments?

ndouglas: Looks like it's disabled in BRD, casual glance seems to indicate APCu isn't installed on Tugboat

ndouglas: So: yeah, lemme do Tugboat here and I'll open a PR on devops for BRD

ndouglas: I think I added this to Tugboat and enabled it in 4506cb7.

indytechcook: I enabled it in devshop.  I'm testing the cms-export then we can merge

indytechcook:
When I testing this locally there was on issue with getting a package via `dist` but composer sorted itself out 👍 

```
  - Installing weitzman/drupal-test-traits (dev-master 9234dcc): Downloading (failed)    Failed to download weitzman/drupal-test-traits from dist: The "https://gitlab.com/api/v4/projects/weitzman%2Fdrupal-test-traits/repository/archive.zip?sha=9234dcceff424180f013b66a9f0a430c985966d0" file could not be downloaded (HTTP/1.1 429 Too Many Requests)
    Now trying to download from source
  - Installing weitzman/drupal-test-traits (dev-master 9234dcc): Cloning 9234dcceff from cache
```

indytechcook:
cms export finished without any errors:

```
root@php:/var/lib/tugboat# drush va-gov-cms-export-all-content --process-count=8 --entity-count=500 --delete-existing
Exported config, content, and files.
```

Looking at the directory, the files have been updated with the correct date stamp:

```
   4 -rw-rw-r--  1 root www-data    3098 Jan  8 13:50 taxonomy_term.f4c11247-49d8-48e5-88eb-18a64f111be2.json
   4 -rw-rw-r--  1 root www-data    3222 Jan  8 13:50 taxonomy_term.f4e91e06-aeca-4008-9998-7e8d6d56406b.json
   4 -rw-rw-r--  1 root www-data    3011 Jan  8 13:50 taxonomy_term.f4eb738f-cafb-428b-9f32-160bfa083367.json
   4 -rw-rw-r--  1 root www-data    2988 Jan  8 13:50 taxonomy_term.f55a34b8-f0cc-41f1-a6a3-8375160003ab.json
   4 -rw-rw-r--  1 root www-data    2764 Jan  8 13:50 taxonomy_term.f6067c2a-a0c7-4512-aac0-ad865d127329.json
   4 -rw-rw-r--  1 root www-data    2988 Jan  8 13:50 taxonomy_term.f750bca5-c4d7-424b-af63-70de3479c24c.json
   4 -rw-rw-r--  1 root www-data    3168 Jan  8 13:50 taxonomy_term.f80a46ca-685c-4426-9496-947446be0965.json
   4 -rw-rw-r--  1 root www-data    3089 Jan  8 13:50 taxonomy_term.f89701d9-68e7-47ae-85e0-42b7ce40f188.json
   4 -rw-rw-r--  1 root www-data    2936 Jan  8 13:50 taxonomy_term.fa144f04-6600-4847-a4d0-3f8f38068e16.json
   4 -rw-rw-r--  1 root www-data    3094 Jan  8 13:50 taxonomy_term.faa562e5-c603-4674-aec8-49e376ae1de1.json
   4 -rw-rw-r--  1 root www-data    2828 Jan  8 13:50 taxonomy_term.fb7324c9-382a-4a85-8bed-8864273c9b54.json
   4 -rw-rw-r--  1 root www-data    3229 Jan  8 13:50 taxonomy_term.fc8527c3-7ded-47b3-a24c-d54821b23765.json
   4 -rw-rw-r--  1 root www-data    2992 Jan  8 13:50 taxonomy_term.fde46275-08f9-44ad-8ec3-d7cd8fbb746e.json
   4 -rw-rw-r--  1 root www-data    2920 Jan  8 13:50 taxonomy_term.ffa066cd-bfb6-43fb-be7f-a8e4ab911da7.json
root@php:/var/lib/tugboat/docroot/sites/default/files/cms-export-content#
```

indytechcook (approved):

Issue #3124, opened 2020-10-07
Have CMS notify slack if content release fails


swirtSJW:
## Description
Update failure logic here 
docroot/modules/custom/va_gov_build_trigger/src/Service/BuildFrontend.php
to put the error message as it does, but ALSO have it notify cms-team direclty using the slack API if this is on prod.  There is no reason a human has to notify us of this.

## Acceptance Criteria
- [ ]  When a failure to create a content release on prod happens, slack cms-team should be notified.
- [ ]  No slack notifications from lando, CI or dev, and staging.
- [ ] Content release error message that is displayed in Drupal CMS is updated to fit the new workflow
- [ ] content release error message that is sent to Slack is written to include relevant information

indytechcook:
The change requested here will not notify the cms channel if the content deploy job fails.  It will only notify slack if the CMS can't access the Jenkins API.

ndouglas:
Should this be moved to the devops repo then?

indytechcook:
> Should this be moved to the devops repo then?

We need both.  Probably a follow up issue for the devops team

ElijahLynn:
Moving back to "ready" column right now on a call with @indytechcook and @oksana-c. We can discuss more.

acrollet:
Closing in favor of #5050

PR #3862, opened 2020-12-28
VACMS 3124: Have CMS notify Slack if content release fails.

ndouglas:
## Description

See #3124.

I'm pretty tentative about this, because I haven't touched BRD before.

I have some questions about this:

1. #3124 mentions "notify cms-team", but AFAICT we don't have a specific webhook URL for posting to the #cms-team channel.  I'm just reusing the #cms-notifications channel webhook URL, but wanted to mention that this differs from the issue's ACs in case there's something we need to discuss there.
2. I looked and didn't find a better way of determining whether we're on prod, so I copied and modified some of the logic in docroot/modules/custom/va_gov_build_trigger/src/Form/PreviewForm.php .  I feel like this should be centralized somewhere, maybe in docroot/modules/custom/va_gov_build_trigger/src/Environment/EnvironmentDiscovery.php, and I actually started doing that, but wanted to get some feedback on that idea first.

## Testing done
None, lol.  Mostly just wanted to get eyes on this to see if everyone is okay with this approach.

## QA steps

As any developer:
1. Then validate Acceptance Criteria from #3124:
    - [ ]  When a failure to create a content release on prod happens, slack cms-team should be notified.
    - [ ]  No slack notifications from lando, CI or dev, and staging.
    - [ ] Content release error message that is displayed in Drupal CMS is updated to fit the new workflow
    - [ ] content release error message that is sent to Slack is written to include relevant information 

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `Needs PO review`


ndouglas:
@acrollet @swirtSJW If you guys have some free time, I'd appreciate some feedback on this approach.  I'd like to add some tests to this (which I think will involve some slight refactoring), but I wanted to get a little buy-in on whether it's heading in the right direction or if we need to have a wider discussion about how and to whom the notification is sent, how we determine prod, etc, or if I'm just flat-out ignorant about something.

acrollet:
As far as determining whether we're on prod goes, I'm seeing at least three approaches in use:

- Inspecting `$config['environment_indicator.indicator']['name']`
- hostname inspection
- The `CMS_ENVIRONMENT_TYPE` environment variable

For my $.02, I'd be in favor of standardizing on one approach and having something like a `isProduction()` method in `EnvironmentDiscovery`...

CC @indytechcook

ndouglas:
_Definitely_ think we should standardize on something.  And I'm happy to roll that job into this PR, of course.

docroot/modules/custom/va_gov_build_trigger/src/Plugin/VAGov/Environment/BRD.php
```diff
@@ -22,6 +24,12 @@
  */
 class BRD extends EnvironmentPluginBase implements ContainerFactoryPluginInterface {
 
+  // Hosts we associate with the Prod BRD environment.
+  public const VAGOV_PRODUCTION_HOSTS = [
```
indytechcook, line 28: If we are in the plugin then we are in BRD.  No need to add any more checks.

ndouglas: I don't understand.  This isn't to check whether we're in BRD, but to distinguish between BRD environments.  Am I lost?

indytechcook:
We should do another plugin for `Production`

docroot/modules/custom/va_gov_build_trigger/src/Plugin/VAGov/Environment/BRD.php
indytechcook: Add a plugin for `Production`

indytechcook:
> Inspecting $config['environment_indicator.indicator']['name']

This is just a label and not used anywhere else right?

> The CMS_ENVIRONMENT_TYPE environment variable

This is only used in `post_api` code now.  This should be deprecated in favor of the `EnvironmentDiscovery` code.

In the current design, the `EnvironmentDiscovery` plugins are more for the frontend web build.  I do like the idea of expanding the usage but we need to be careful about the plugin classes doing to much.  As the `Environment` plugin grows, it should start to become a collection of classes where it would have a `getWebBuilder` method and `getWebBuildForm` method.

docroot/modules/custom/va_gov_build_trigger/src/Plugin/VAGov/Environment/BRD.php
ndouglas: Gotcha.  I'll do it that way.

ndouglas:
Warning: code is untested.  LMK if this is more-or-less what you had in mind.  I read through the BRD code, contemplated a couple different ways of refactoring it, and, uh, did this.  If you wanna pick this apart on Twitch or whatever, that's fine with me -- there are some places with which I'm not entirely happy, but I didn't want to go _too_ wild.

docroot/modules/custom/va_gov_build_trigger/src/Environment/EnvironmentDiscovery.php
```diff
@@ -115,8 +142,10 @@ public function getEnvironment() : EnvironmentInterface {
     if ($this->environment) {
       return $this->environment;
     }
-
     $build_type = $this->getBuildTypeKey();
+    if ($build_type === 'brd' && $this->isProd()) {
```
indytechcook, line 146: the build type is set in the corresponding settings file, in this case `settings.prod.php`.   Let's not override it anywhere.

docroot/modules/custom/va_gov_build_trigger/src/Environment/EnvironmentDiscovery.php
```diff
@@ -82,6 +99,16 @@ public function isLocal() : bool {
...
   }
 
+  /**
+   * Is this on Production?
+   *
+   * @return bool
+   *   Is this on Production?
+   */
+  public function isProd() : bool {
+    return !empty($this->request) && in_array($this->request->getHost(), static::VAGOV_PRODUCTION_HOSTS);
```
indytechcook, line 109: Let the plugin define this. Add a `isProd() : bool` to the interface and default it to false in `EnvironmentPluginBase`

docroot/modules/custom/va_gov_build_trigger/src/Plugin/VAGov/Environment/BRD.php
```diff
@@ -68,96 +67,112 @@ public static function create(ContainerInterface $container, array $configuratio
...
+        CURLOPT_USERNAME => $va_cms_bot_github_username,
+        CURLOPT_PASSWORD => $va_cms_bot_jenkins_auth_token,
+        CURLOPT_FOLLOWLOCATION => 1,
+        CURLOPT_RETURNTRANSFER => 1,
+        CURLOPT_HEADER => 1,
+      ],
+    ];
+    // Setup the REQUEST retry logic w/lazy backoff.
+    // @see http://dawehner.github.io/php,/guzzle/2017/05/19/guzzle-retry.html
+    $handler_stack = HandlerStack::create();
```
indytechcook, line 100: Ideally, you would inject this also but we are looking a larger refactor of this logic later so this is fine for now.

indytechcook (changes requested):
While this code refactor/changes if fine and dandy, this not the correct place to handle this change.  The Jenkins job itself is where the content job runs.  This will only notify slack if Jenkins is down, not if the job fails.

docroot/modules/custom/va_gov_build_trigger/src/Environment/EnvironmentDiscovery.php
```diff
@@ -3,12 +3,19 @@
...
 
 /**
  * A class responsible for environment (dev/stage/prod) discovery.
  */
 class EnvironmentDiscovery {
 
+  // Hosts we associate with the Prod BRD environment.
+  public const VAGOV_PRODUCTION_HOSTS = [
+    'cms.va.gov',
+    'prod.cms.va.gov',
```
swirtSJW, line 16: Most of our other logic around the site uses the environment name.   This would be a switch to using host detection.  Can this be made to use environment names for consistency of approach?

swirtSJW: Ignore this... somehow I missed all the other discussion on this topic on this PR.

PR #3843, opened 2020-12-21
VACMS-3840: Add new leadin field to services.

ethanteague:
## Description
See #3840 

## FE Implementation
Attn: @mmiddaugh @kevwalsh @mpelzsherman @brangi 
- Add `fieldHealthServiceApptLeadin` to `src/site/stages/build/drupal/graphql/supportService.graphql.js`
- Update templates that use services data (e.g.: `src/site/facilities/health_service.drupal.liquid`, etc.) to include `fieldHealthServiceApptLeadin` - add check so that when field is empty, output this default text: `Contact us to schedule, reschedule, or cancel your appointment. If a referral is required, you’ll need to contact your primary care provider first.`

## Testing done
Visual/Behat

## Screenshots
image

## QA steps
As an administrator, go to `admin/structure/types/manage/support_service/form-display` and visually verify that `Appointment lead-in text` appears identical to screenshot (when cog is clicked)

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [ ] `Core Application Team`
- [x] `Product Support Team`

### Does this PR need review from a Product Owner

- [ ] `Needs PO review`

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change? 
- [ ] Yes, and it's written in issue ____ and queued for publication. 
  - [ ] Merge and ping @ rachel-kauff so she's ready to publish after deployment
- [ ] Yes, but it hasn't yet been written 
  - [ ] Don't merge yet -- ping @ rachel-kauff to prompt her to write and queue content
- [ ] No announcement is needed for this code change. 
  - [ ] Merge & carry on unburdened by announcements

PR #3863, opened 2020-12-28
Revert "VACMS-3840: Add new leadin field to services."

swirtSJW:
Reverts department-of-veterans-affairs/va.gov-cms#3843


swirtSJW:
image

Issue #3577, opened 2020-11-24
Content Release Job status


ElijahLynn:
## Description
Currently, when a user triggers a content deploy with the "Deploy content" button they just see a status message and need to refresh the page manually to see the results. 

We need to display real-time user feedback to the user notifying them of completion status and if there are any errors. 

## Acceptance Criteria
- [ ] User sees completion status of content deploy on PR & DEMO (Tugboat) environments (without having to refresh)
- [ ] User has a way to see logs from within the CMS (e.g. simple link to raw logs via Tugboat UI)
- [ ] When a web build status is trigger the job status should say `Pending`
- [ ] When the advanced queue job starts, the job should say `In Progress`
- [ ] When the job complete, the status should say `Sucess` if there are no error or `Error`.

Implementation notes:
- We are using advancedqueue to run the web build jobs in tugboat.  Create a block/view to show the status of the job entities from advanced queue.
- The advanced queue status block/view should auto refresh every 10 seconds
- Use a tugboat env variable to get the build ID which can be used for the link to the logs: https://docs.tugboat.qa/reference/environment-variables/

acrollet:
@indytechcook I can look at this if you're not currently working on it?

indytechcook:
@acrollet all you!

PR #3864, opened 2020-12-28
[DO NOT MERGE] VACMS-3791: built front-end PRs

acrollet:
## Description

See #3791 

## QA steps

As user _uid_ with _user_role_
1. Do this
1. Then that
1. Then validate Acceptance Criteria from issue
- [ ] This
- [ ] That
- [ ] The other thing

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

### Does this PR need review from a Product Owner

- [ ] `Needs PO review`

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change? 
- [ ] Yes, and it's written in issue ____ and queued for publication. 
  - [ ] Merge and ping @ rachel-kauff so she's ready to publish after deployment
- [ ] Yes, but it hasn't yet been written 
  - [ ] Don't merge yet -- ping @ rachel-kauff to prompt her to write and queue content
- [ ] No announcement is needed for this code change. 
  - [ ] Merge & carry on unburdened by announcements

PR #3940, opened 2021-01-05
VACMS-3577: content release job status block

acrollet:
## Description

See #3577 

## Screenshots

chrome-capture (2)

## QA steps

1. Do this
1. Then that
- [ ] This
- [ ] That
- [ ] The other thing

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


indytechcook:
Can you make this a PR against https://github.com/department-of-veterans-affairs/va.gov-cms/pull/3864?  Perhaps we can merge it since we can't merge 3864 until after the deploy today.

acrollet:
> Can you make this a PR against #3864? Perhaps we can merge it since we can't merge 3864 until after the deploy today.

will do

docroot/modules/custom/va_gov_build_trigger/src/Controller/ContentReleaseStatusBlockController.php
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Get content_release_status_block content.
+   *
+   * @return string
+   *   Rendered block.
+   */
+  public function getBlock() {
+    $block = $this->pluginManagerBlock->createInstance('content_release_status_block');
+    $output = render($block->build());
+    return new Response($output);
```
indytechcook, line 39:
```suggestion
    return Response::create($output);
```

As a general rule, try to use factory methods to create classes.

docroot/modules/custom/va_gov_build_trigger/src/Plugin/Block/ContentReleaseStatusBlock.php
```diff
@@ -0,0 +1,262 @@
...
+   */
+  private function getCommandRunnerJob() {
+    return $this->database->select('advancedqueue', 'aq')
+      ->condition('aq.queue_id', 'command_runner')
+      ->fields('aq')
+      ->range(0, 1)
+      ->orderBy('available', 'DESC')
+      ->execute()
+      ->fetchObject();
+  }
```
indytechcook, line 253:
```suggestion
  private function getCommandRunnerJob() : ?stdClass {
    return $this->database->select('advancedqueue', 'aq')
      ->condition('aq.queue_id', 'command_runner')
      ->fields('aq')
      ->range(0, 1)
      ->orderBy('available', 'DESC')
      ->execute()
      ->fetchObject() ?? NULL;
  }
```

Not really required just a thought

acrollet: ah cool, works for me :-)

acrollet:
> Can you make this a PR against #3864? Perhaps we can merge it since we can't merge 3864 until after the deploy today.

@indytechcook here's the PR: https://github.com/acrollet/va.gov-cms/pull/1

acrollet:
@indytechcook I've updated the PR with the items we discussed. I didn't switch to use the entity query manager because although queues are entities, jobs are not. (I also didn't find any API for querying jobs in advancedqueue, it just claims jobs one at a time...)

indytechcook:
> @indytechcook I've updated the PR with the items we discussed. I didn't switch to use the entity query manager because although queues are entities, jobs are not. (I also didn't find any API for querying jobs in advancedqueue, it just claims jobs one at a time...)

I stand corrected, though it does feel like a miss on their part.  I think it would be easier to read if we used `Drupal\advancedqueue\Job` with something like this https://www.drupal.org/docs/7/api/database-api/fetching-into-a-custom-class (if possible;f or just loading the class manually.  There is no other way than using `new Job` to create the object so that's ok.

indytechcook (approved):

Issue #2681, opened 2020-08-14
Redirect saving is broken


swirtSJW:
**Describe the defect**
If you edit a naturally occurring redirect in the UI, it goes from looking like this before save
![image](https://user-images.githubusercontent.com/5752113/90297138-8bf99000-de5b-11ea-92f3-954c2c4cce60.png)

To this after saving:
![image](https://user-images.githubusercontent.com/5752113/90297213-c8c58700-de5b-11ea-90c1-fe17e9da6d99.png)

There was some "custom" logic and handling that was added in this contrib module https://www.drupal.org/project/redirect_options that is likely at play here.   We should revisit what this module is doing and why.
Redirect type and Match Type  
![image](https://user-images.githubusercontent.com/5752113/90297396-50ab9100-de5c-11ea-9596-c769692ee2c9.png)
are added by this module and the purpose is not clear.

Also following the save, the redirect does not actually work.

This defect is a blocker to implementing out proposed redirect strategy with the front end.
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/2297

**Expected behavior**
When editing a redirect,and saving it should not say "Server" in the destination.  It should show the path like it did before editing.

An edited redirect should actually redirect.

ndouglas:
I haven't had any luck reproducing the issue with redirects working.  They continue to work for me, across cache clears, etc.  It seems like the title change behavior was a workaround for an issue with the GraphQL export, so I'm digging into that to understand it better.  I've opened [an issue](https://www.drupal.org/project/redirect_options/issues/3189372) for a (probably unrelated) warning displayed on save.

ndouglas:
Welcome back, @ethanteague !  

>It seems like the title change behavior was a workaround for an issue with the GraphQL export, so I'm digging into that to understand it better.

I've poked into this and `diff`ed a CMS export, and there _are_ changes:

<img width="940" alt="Screen Shot 2021-01-11 at 10 37 52 AM" src="https://user-images.githubusercontent.com/1318579/104202589-1c330580-53f9-11eb-81ff-cdf367eeaee3.png">

But I'm wondering how critical they are -- I've poked into the frontend code and didn't find any that seemed to deal with the internals of redirects, switched on whether they're Template or Server redirects, etc.  Do you remember any of this?

As I see it, my options are to:

1) simply remove the line of code that changes the title of the redirect, which might have unexpected effects
[...]

PR #3971, opened 2021-01-11
VACMS 2681: Alter Redirect view to work around redirect_options behavior.

ndouglas:
## Description

See #2681. 

This PR edits the displays of the Redirects view to display the redirect destination, as one would expect, rather than the "Server" or "Template" string.  I was not able to replicate the issue reported wherein edited redirects no longer worked.

## Testing done
Edited a few redirects, essentially at random, and verified that they continued to function and that the view displayed a linked destination path rather than a "Server" or "Template" string.  Also verified that edited redirects continued to function, even after clearing caches, etc.

## Screenshots
Screen Shot 2021-01-11 at 3 01 52 PM

## QA steps
As a user with admin access:
1. View the [admin redirects page](http://va-gov-cms.lndo.site/admin/config/search/redirect)
2. Then validate Acceptance Criteria from issue #2681:
    - [x] When editing a redirect,and saving it should not say "Server" in the destination. It should show the path like it did before editing.
    - [x] An edited redirect should actually redirect.

As a user with the 'Redirect Admin' role:
1. View the [non-admin redirects list](http://va-gov-cms.lndo.site/redirects-list)
2. Then validate Acceptance Criteria from issue #2681:
    - [x] When editing a redirect,and saving it should not say "Server" in the destination. It should show the path like it did before editing.

### Definition of Done

### Select Team for PR review
- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


ndouglas:
@acrollet I know you have a lot going on right now, but if you have a free minute I'd appreciate a quick glance at this -- it's functionally minor and should be pretty easy.

ndouglas:
After saving `/admin/config/search/redirect/edit/461` without changes, @acrollet reported that the redirect does not seem to work, so this means the original issue is now reproducible 🎉   And that I've wasted some time letting this sit around...

ndouglas:
I think that the redirects are not working _before_ they're saved and then continue to not work _after_ being saved.

It's a little difficult to confirm this, given that the only redirects that provably _haven't_ been saved with `redirect_options` active (i.e. `redirect_redirect__title` IS NULL) are also redirects that function correctly.

For example, take `locations/tobyhanna-va-clinic`, which currently works.  It will continue to work after saving.

Screen Shot 2021-01-26 at 3 16 34 PM

DB row prior:
Screen Shot 2021-01-26 at 3 24 40 PM

and after:
Screen Shot 2021-01-26 at 3 25 03 PM

The default sort of this View is problematic for testing, because it sorts ascending by date:
Screen Shot 2021-01-26 at 3 30 54 PM

But I think the main issue is that these path sources have a trailing slash.  If I remove the slash, everything works -- both here and on my ex-employer's CMS that doesn't have `redirect_options` installed.

If you can confirm that, I think we're still good to go.  Minus the fact that I just lost several hours of my life that I'll never get back because of trailing slashes.

acrollet:
I can confirm that with the trailing '/' removed, the redirect works with and without the trailing slash:

https://pr3971-0jzed8iphkaolhvtv4y3hkbcrtydwara.ci.cms.va.gov/employment/job-seekers/alternative/ and https://pr3971-0jzed8iphkaolhvtv4y3hkbcrtydwara.ci.cms.va.gov/employment/job-seekers/alternative both redirect to https://pr3971-0jzed8iphkaolhvtv4y3hkbcrtydwara.ci.cms.va.gov/careers-employment

acrollet (approved):

Issue #3923, opened 2021-01-03
Prevent user from exceeding limit of 10 FAQ paragraphs


kevwalsh:
## User Story or Problem Statement

As a user editing CLP, i want to be prevented from adding an 11th FAQ, so that i don't waste time creating one that is only disallowed later during node save.

This is the behavior established in other paragraph fields where number of values is limited, such as Link teasers in the Spotlight field.

## Acceptance Criteria
- [ ] Once a user has added 10 FAQs, the button to add another disappears. 

![Edit_Campaign_Landing_Page_TEST_Get_your_flu_shot___VA_gov_CMS.jpg](https://images.zenhubusercontent.com/5d6e57ad330f64311395f08b/2df2005f-3acb-40ea-954e-a44e3c0e7985)

acrollet:
@kevwalsh @indytechcook @ethanteague Since this is a pattern for us, should we look at whether [Field Limiter](https://www.drupal.org/project/field_limiter) would meet our use case? The module is simple, has clean code and would reduce our tech debt. (assuming it works the way we need)

PR #3985, opened 2021-01-12
VACMS-3923: Limit qa's in FAQ panel to 10.

ethanteague:
## Description
See #3923 

## Testing done
Visual / behat

## Screenshots
image

## QA steps
- As an administrator, go to `/node/15051/edit` and open the FAQs panel.
- Visually verify that the `Add Q&A` button does not appear.
- Remove one Q&A paragraph, and visually verify that the `Add Q&A` button does appear.

### Definition of Done

### Select Team for PR review

- [ ] `Core Application Team`
- [x] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


indytechcook:
Could this be done as a field config?

Hi ethan!

ethanteague:
> Could this be done as a field config?
> 
> Hi ethan!
Howdy @indytechcook ! Sadly, trying to do this with config results in a really gnarly layout - whenever you set a limit to a certain number, you end up with an initial display of the max number.

acrollet:
@kevwalsh @indytechcook @ethanteague Since this is a pattern for us, should we look at whether [Field Limiter](https://www.drupal.org/project/field_limiter) would meet our use case? The module is simple, has clean code and would reduce our tech debt. (assuming it works the way we need)

indytechcook:
I do like the approach of using field formatted as it's much more Drupal.  The downside is that it adds 2 dependant modules in Field Limiter and Field Formatter. But I think it will give us more flexibility going forward.

acrollet:
The other advantage is less custom js functionality to re-validate as we're looking to change themes in the relatively near future.

swirtSJW (dismissed):

swirtSJW:
@ethanteague  can you rework this with method @indytechcook  and @acrollet are suggesting.

ethanteague:
> @ethanteague can you rework this with method @indytechcook and @acrollet are suggesting.

:+1:

ethanteague:
@indytechcook @acrollet @swirtSJW field_limiter only works for node display output (module doesn't have logic for form mode). Unless we patch it, it won't meet our use case.

acrollet:
@ethanteague oh shoot, sorry for the rabbit trail :-\ I did just find this module which might do what we actually want, although I don't like the code quite as well: https://www.drupal.org/project/limited_field_widgets

Probably for expediency we should go with the js solution and create a follow-on to have a unified long-term solution for this use case.

ethanteague:
> @ethanteague oh shoot, sorry for the rabbit trail :-\ I did just find this module which might do what we actually want, although I don't like the code quite as well: https://www.drupal.org/project/limited_field_widgets
> 
> Probably for expediency we should go with the js solution and create a follow-on to have a unified long-term solution for this use case.

@acrollet limited_field_widget seems pretty legit to me - I just pushed a new commit that uses it / updated config in place of my previous js commit, as well as the other instance in clp where we were using js to limit adding more than 3 paragraphs in the spotlight panel.
cc: @indytechcook @swirtSJW

indytechcook (approved):
Looks Good.  I went to the node in question and wasn't able to add any more.  I remove one and the new button appeared and worked as expected.

swirtSJW (approved):
Nice pivot @ethanteague

PR #4005, opened 2021-01-13
VACMS-3824 Remove Pittsburgh from Covid Vaccine migration.

swirtSJW:
## Description

Pittsburgh needs to be removed from the migration since it is managed by a human now.

## QA steps

None as it is just a data change


ndouglas (approved):
I'll take your word for it 😄

Issue #2824, opened 2020-09-01
Lower flood control limit to 30 minutes


kevwalsh:
## User Story or Problem Statement

As someone who has failed to log in, i don't want to wait 6 hours to try again. 

https://www.drupal.org/project/flood_control

## Acceptance Criteria
* [ ] Users who have been added to flood table can try again after 30 minutes.

See also #2823

Issue #2823, opened 2020-09-01
Configure flood_unblock to expose flood table UI


kevwalsh:
## User Story or Problem Statement

As a user admin, i want to be able to unblock a user who has failed login 5 times. 

## Acceptance Criteria
- [ ] User admins can unblock IP addresses from the UI
- [ ] UI reviewed and demo'ed for @xiongjaneg and/or @rachel-kauff to support User Guide writing in #3938 and #3939

## Implementation steps
- [ ] Install and configure https://www.drupal.org/project/flood_unblock

PR #4038, opened 2021-01-19
VACMS-2824: Add patched flood_control module & configure user block expiration

acrollet:
## Description

See #2824 and #2823

## Testing done

Manual testing. 

## Screenshots

Screen Shot 2021-01-19 at 4 44 16 PM

## QA steps

Attempt to login with a valid username and incorrect password more than 5 times in a row. You should see a message that the account has been temporarily blocked. Then, review the [new documentation](https://github.com/acrollet/va.gov-cms/blob/VACMS-2824-2825-flood_control/READMES/user_management.md) and log in as a user with the `Administrator` role:

- [ ] Confirm that the user login flood entries are set to expire 30 minutes after the login attempt was made
- [ ] The blocked user account can be manually unblocked

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

### DIFF
--- a/READMES/user_management.md
+++ b/READMES/user_management.md
@@ -1,4 +1,16 @@
-# Bulk User Import Procedure
+# CMS User Management
+
+## Login Issues
+
+### Account temporarily blocked
+
+If a user makes too many attempts to log in with an incorrect password in a short time, their account will be temporarily blocked. The following message will be displayed:
+
+> There have been more than 5 failed login attempts for this account. It is temporarily blocked. Try again later or request a new password.
+
+To resolve this issue, log in as an administrator and visit `Home >> Administration >> People >> Flood Unblock`. Check the box next in each row with the user's username/email address, and click the `Remove selected items from the flood table` button.
+
+## Bulk User Import Procedure
 
 To import users:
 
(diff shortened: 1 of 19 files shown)


ndouglas:
Screen Shot 2021-01-20 at 11 57 25 AM

Curses!  Foiled again!

ndouglas (approved):
I'm not a big fan of Core's method of determining the source IP, and I'm concerned that as we expand to greater numbers of editors that we might end up inadvertently DoSing ourselves just by a comparatively small number of people each failing a login and getting the IP of the proxy banned.

Obv. that's not a concern with this PR but a concern with Drupal itself.  And the odds are low that it'd cause an issue with our specific setup, SOCKS and all.

So, I guess my only tentative suggestion would be to consider upping the number of failed logins permissible from a given IP within the time range.  Otherwise, works as intended.

acrollet:
> So, I guess my only tentative suggestion would be to consider upping the number of failed logins permissible from a given IP within the time range. Otherwise, works as intended.

I'd guess that this is unlikely to be an issue in practice because most logins on prod should happen via SSO. However this will give us an interface to easily see if we're approaching the limit...

PR #3996, opened 2021-01-13
Notifications and Events base PR

indytechcook:
## Description

#3737

This PR will provide the base tugboat environment and branch for the notifications/events.  This is necessary because the danse module requires PHP 7.2.  Once #3775 or #1004 is complete then this can be merged to master.

New PRs for events/notifications should be made to this branch.

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

### Does this PR need review from a Product Owner

- [ ] `Needs PO review`

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change? 
- [ ] Yes, and it's written in issue ____ and queued for publication. 
  - [ ] Merge and ping @ rachel-kauff so she's ready to publish after deployment
- [ ] Yes, but it hasn't yet been written 
  - [ ] Don't merge yet -- ping @ rachel-kauff to prompt her to write and queue content
- [x] No announcement is needed for this code change. 
  - [x] Merge & carry on unburdened by announcements

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/3996)

Issue #3772, opened 2020-12-16
A Flag should be created when a user edits a node.


indytechcook:
As a CMS user, I want to receive a notification, every time the node I ever edited is updated, so I can keep track of the content that I have edited in the past.

## Description

As part of the event/notifications systems, users need to be notified when changes are made to nodes that they are subscribed to.  
The flag module will be used to track when a user should get a notification for a node update.

## Acceptance Criteria
- [ ] Flag module is enabled
- [ ] When a user updates a node, a flag is created. The flag indicates that a user is subscribed to editorial workflow notifications about that specific node.
- [ ] A flag is created retroactively for all users that have edited nodes in the past.
- [ ] The flag/unflag link should not be shown to editors

PR #4043, opened 2021-01-20
VACMS-3772: A Flag should be created when a user edits a node.

ndouglas:
## Description

See #3772.  This is based off PR #3996. 

## Testing done
Adds a Behat test to confirm that the `edited` flag is set when a user updates or inserts a node.

## Screenshots
Forthcoming.

## QA steps

As a developer:
1. Validate Acceptance Criteria from issue #3772
    - [ ] When a user updates a node, a flag is set. The flag indicates that a user is subscribed to editorial workflow notifications about that specific node.
    - [ ] A flag is created retroactively for all users that have edited nodes in the past.
    - [ ] The flag/unflag link should not be shown to editors
 
### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1092,6 +1092,39 @@ function _va_gov_backend_smart_date_paragraph_formatting(array &$element) {
   unset($element['subform']['field_datetime_range_timezone']['widget'][0]['value']['#title']);
 }
 
+/**
+ * Implements hook_ENTITY_TYPE_update() for nodes.
+ */
+function va_gov_backend_node_update(NodeInterface $node) {
```
indytechcook, line 1098: Let's use the new events from `hook_event_dispatcher`

ndouglas: okey-doke

docroot/modules/custom/va_gov_backend/src/Event/Entity/EditedFlag.php
```diff
@@ -0,0 +1,90 @@
...
+    $this->flagService = $flagService;
+  }
+
+  /**
+   * Sets `edited` flag for current user for default revision of specified node.
+   *
+   * @param \Drupal\node\NodeInterface $node
+   *   The node on which the flag should be set.
+   */
+  public function setEditedFlag(NodeInterface $node): void {
```
indytechcook, line 40: Let's create a new Service that encapsulates the log around if a node should be flagged for a user.

I'm also thinking about which module to push this code, whether we create a new `va_gov_notificaions` or to put it in `va_gov_workflow_assignments`

ndouglas: Mmm.  I thought the scope of this issue was just to set this flag unconditionally.

If it's not really an `edited` flag but means something like "followed" or something then I should rename it.

I sometimes awake from dreams in which I've split `va_gov_backend` into about 30-40 different modules, so I'm all in favor of shifting this into a `va_gov_notifications` module.

ndouglas: Or do you mean just that we have a number of other considerations (user role, etc) that we want to take into account when deciding whether to set this flag, and that is complicated enough in a different class?

indytechcook: > Or do you mean just that we have a number of other considerations (user role, etc) that we want to take into account when deciding whether to set this flag, and that is complicated enough in a different class?

Yes, and there will be other considerations in the future.

ndouglas: okay, groovy

I'ma go ahead and split this off into `va_gov_notifications`.

ndouglas:
I added a service, `va_gov_notifications.flag_decisions`, `Drupal\va_gov_notifications\Service\FlagDecisions`.

The current interface is not great:

```php
  /**
   * Should we set the `edited` flag for this node and user?
   *
   * @param \Drupal\node\NodeInterface $node
   *   A node this user has ostensibly edited.
   * @param \Drupal\user\UserInterface $user
   *   A user who has ostensibly edited a node.
   *
   * @return bool
   *   TRUE if the flag can and should be set, otherwise FALSE.
   */
  public function shouldSetEditedFlag(NodeInterface $node, UserInterface $user);
```

I'd like to have a stable API obv, and I don't think this is a reasonable API to keep stable.

I'd prefer it to be something like:

`public function shouldSetFlag(FlagInterface $flag, <context>);`

But this brings up the question of what `<context>` would be.  This will naturally depend on the type of the flag (e.g. `personal`, `content` flags would consist of a `User` and `Node`) and possibly the flag itself and whatever event triggered it.

In the case of the flags we're discussing right now, that information is provided by the `Event`.  Both the `Node` and the `User` in question are derived from the event.

So I think a reasonable interface might be something like this:

`public function shouldSetFlagOnEvent(FlagInterface $flag, Event $event);`

We've briefly discussed a plugin-based approach to setting up the criteria, and I haven't thought too much about the specifics of how that would work, but I feel reasonably comfortable moving forward with this approach if it sounds reasonable.

I've considered a few other options too, so open to other possibilities obv.

indytechcook:
The patterns we want to use haven't quite fleshed out.  I agree this isn't' quite there yet but it least abstracts the code out of the Event.

The decisions will really be one the notification, not flag.  `Does this user have a notification active for this Node?` is one question it would answer.   But let's hold off on the refactor for now.

ndouglas:
Re-reading the issue description, I guess I interpreted the goal of this and the overall structure of the notification system incorrectly.

I think I was thinking this PR would operate like this:
- someone updates/inserts a node
- this triggers an update/insert event
- this event is received by a subscriber
- the subscriber sets the edited flag for that user on that node

And then the notification system as a whole would operate something like this:
- someone updates/inserts a node
- this triggers an update/insert event
- this event becomes a DANSE event
- these events are processed and the users fitting the criteria would receive a notification

That last part would be, basically, `!userIgnoredNode && (userEditedNode || userFollowedNode)`, and the list of criteria there could expand endlessly in all directions until the end of time.

So these sentences in #3772: 

>The flag module will be used to track when a user should get a notification for a node update.

>The flag indicates that a user is subscribed to editorial workflow notifications about that specific node.

I was interpreting these as just explanatory, like "this is the function that these things will ultimately serve."  I didn't interpret the original scope of this PR as the event subscriber actually having to _know_ anything about how the notification system operates, or even really know that it exists.  I feel like this event subscriber is, or should be, an extremely simple thing.

Am I stupid? 😕 

I'm not sure where exactly I went off the rails with this, but apparently it was some distance back 😟

indytechcook:
You are pretty much on point.

This went off the rails when I tried to have you do too much in this PR.   I was trying to pull in the wider notification services into this and it was too early.

There eventually will be a service that answers the question `Does this user have a notification active for this Node?` This will drive (amount other things) the button in the UI and if a notification entity should be created.

ndouglas:
understand-now-your-request-i-do

indytechcook (approved):

ndouglas:
hol' up, I was gonna back out the changes I made with the flag decisions service

indytechcook:
> hol' up, I was gonna back out the changes I made with the flag decisions service

Having a separate class handle the logic is good :) Just rename it to `FlaggingService` or something.

ndouglas:
Having rejoined consensus reality, I've made some changes:

- `EditedFlag` -- simplified and refactored, doesn't set flag directly
- `FlagDecisions` -> `Flagging`  -- sets the flag if appropriate
- `edited_flag.feature` -> `personal_content_flags.feature` -- renamed and changed the structure/wording to cover tests on any personal (as opposed to global) content flags.

ndouglas:
One more commit forthcoming to refactor the one-time migration script, once I've confirmed that it works...

ndouglas:
Script was refactored and still works.

indytechcook (approved):

PR #4181, opened 2021-02-02
Fix tablefield sanitize function for caption

indytechcook:
## Description

Fix tablefield sanitize function for caption

## Testing done
Saved node which has a paragraph with a table field with a caption.
Make sure the caption still exists on edit after save!
Did this on video call with @swirtSJW and @ndouglas

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/4181)


swirtSJW (approved):

swirtSJW:
Tested this and it works
image

Issue #4105, opened 2021-01-27
Render JSD widget on the Access Denied page in CMS


oksana-c:
## User Story

As a CMS User, I want to be able to get help quickly if I get an "Access Denied" message in the CMS while logged in.

## Acceptance Criteria
- [ ] JSD widget is rendered on the Access Denied page in CMS

## Implementation notes

* va_gov_help_center module controls which pages include JSD widget library.

Issue #4109, opened 2021-01-27
Spike: Testing Improvements


acrollet:
## User Story or Problem Statement

As a developer, I need tests to be:

- easy to develop and validate
- not affect content that CMS end users will see
- use a modern browser environment
- total test run time is as long as the longest test

### Participants
Required: @indytechcook @swirtSJW @kaise-lafrai @ndouglas @ethanteague @acrollet @ElijahLynn @olivereri 
Optional: @mchelen-gov 

- [ ] Spike is held and actionable next steps and implementation timeline are generated
- The following issues are addressed
  - tests are run against the same DB that users see
  - development workflow is improved
  - modern browser environment should be used for functional tests
- [ ] A process is defined for ensuring that new content types are tested for accessibility as they are created

## Implementation steps

Possible solutions to discuss:
- End to End/Functional Tests complete within 10 minutes of push
- Using cypress for all functional testing
- Test against an in-memory sqlite copy of the DB (perhaps purged of as much content as possible)
- When and under what conditions is test-generated content removed?  (e.g. at the end of successful tests, at the end of successful tests and when the tests fail, only on specific environments, never)
- should we prioritize moving tests to github actions? If so, how & when?
- is it worth using a different task runner suite? (go-task vs yaml-tasks, see comment below)

acrollet:
An alternative to forking `yaml-tests` or moving to `yaml-tests` is to use an entirely different test runner suite. There are not a lot of options out there for flexible test runners, but I believe I've found a good one: [task](https://taskfile.dev/#/usage?id=getting-started)

Positives:

- [actively maintained](https://github.com/go-task/task/commits/master)
- clear [yaml config file format](https://taskfile.dev/#/usage?id=getting-started)
- supports:
  - [parallelization](https://taskfile.dev/#/usage?id=task-dependencies)
  - [task dependencies](https://taskfile.dev/#/usage?id=task-dependencies)
  - [ignoring errors](https://taskfile.dev/#/usage?id=ignore-errors)
  - [.env files](https://taskfile.dev/#/usage?id=env-files)


Negatives:
- no native support for reporting check status/comments to the github api. We would need to use something like [reviewdog](https://github.com/reviewdog/reviewdog), [github-status-updater](https://github.com/cloudposse/github-status-updater), or [github-commenter](https://github.com/cloudposse/github-commenter) on each step or build a tool if that won't work
[...]

PR #4187, opened 2021-02-03
VACMS-4105: Render JSD Widget on Access-Denied pages

ndouglas:
## Description

See #4105. 

## Testing done
Clicked a couple admin pages as both a :godmode: admin and non-admin user.

Started writing tests but ran into issues with visibility, presumably because of the delay and JS involved with Behat.

## Screenshots
Normal User:
Screen Shot 2021-02-03 at 2 18 55 PM
Admin User:
Screen Shot 2021-02-03 at 2 19 17 PM

## QA steps
As a user with non-administrative privileges:
1. Visit some path to which you shouldn't have access, like `/admin/reports/status`
1. Then validate Acceptance Criteria from issue #4105:
- [ ] JSD widget is rendered on the Access Denied page in CMS

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


docroot/modules/custom/va_gov_help_center/va_gov_help_center.module
```diff
@@ -40,6 +40,10 @@ function va_gov_help_center_page_attachments(array &$attachments) {
     // Attach Jira Service Desk widget library to user login route.
     $attachments['#attached']['library'][] = 'va_gov_help_center/helpdesk-widget';
   }
+  elseif ($route_name === 'system.403') {
```
oksana-c, line 43: I may be totally out of style here 💃 🕺 , but can we add 403 route as OR condition in the if statement above? DRY(er)

oksana-c (commented):
not a requirement (I also forgot to do it myself when created `va_gov_help_center` module), but can we move `docroot/modules/custom/va_gov_backend/src/Plugin/Block/WhatsNew.php` to `va_gov_help_center` so as to keep all help center related functionality in place?

docroot/modules/custom/va_gov_help_center/va_gov_help_center.module
ndouglas: Sure.  I was thinking about writing some tests in Cypress and then refactoring to see if I could get that attachment in only one place 😄

ndouglas: I suspect we're going to add more cases as time goes by.

oksana-c: always possible that we may add more 👍

acrollet:
🌲 🌲 🌲

ndouglas:
(Don't worry, @acrollet, I know I'll probably be rewriting/refactoring/rm -rf'ing this after we have the spike 😁 )

acrollet:
> (Don't worry, @acrollet, I know I'll probably be rewriting/refactoring/rm -rf'ing this after we have the spike 😁 )

nah this just means you're in charge of the spike 🧐

ndouglas:
I still have a couple things to do on this:

- [ ] I'd like to add a couple more tests since I figured out the iframe was the issue
- [ ] move the thing to the module like Oksana said

ndouglas:
@acrollet I'd also dig your thoughts on this:

`I am at "<page>"` is not fault-tolerant.  So if you visit `/some-page-that-doesn't-exist`, you get a 404 error and Cypress kicks up an error.  Good.

The problem is that sometimes we want to visit a page and get an error.  For instance, a 403 error, so that we can see if the JSD widget pops up.

I used `I attempt to visit "<page>"` as a fault-tolerant version of that, but I feel like there should be a better phrasing, and I'd like to get that better phrasing in place before we write a bunch of tests :D  

Thoughts?

acrollet:
> Yeah, agreed, I'll toss that in.

acrollet:
> I used `I attempt to visit "<page>"` as a fault-tolerant version of that, but I feel like there should be a better phrasing, and I'd like to get that better phrasing in place before we write a bunch of tests :D

`And I navigate to "<page>"`???

tests/behavioral/cypress/integration/common/i_should_see_element.js
```diff
@@ -0,0 +1,5 @@
+import { Then } from "cypress-cucumber-preprocessor/steps";
+
+Then(`I should see element {string}`, (text) => {
```
acrollet: Maybe this could be something like "I should see the element with selector {string}" to make it more self-documenting?

ndouglas: Sure, sounds good to me.

docroot/modules/custom/va_gov_help_center/va_gov_help_center.module
```diff
@@ -35,26 +34,40 @@ function va_gov_help_center_toolbar() {
...
+    'user.login',
+  ];
+  if (in_array($route_name, $unconditional_routes)) {
+    return TRUE;
+  }
+  if ($route_name === 'entity.node.canonical' && !empty($route_parameters['node']) && $node = $route_parameters['node']) {
+    // Attach Jira Service Desk widget library to all nodes of type
+    // `documentation_page` with an exception of CMS Help Center homepage -
+    // NID 6454.
+    if ($node->getType() === 'documentation_page' && $node->id() != 6454) {
```
acrollet, line 68: Since we've already got a couple of mentions of 6454 in the module should we consider setting up a CMS_HELP_CENTER_HOMEPAGE_NID constant or some such?

ndouglas: Yeah, magic numbers suck.

tests/behavioral/cypress/integration/common/i_am_logged_in_as_a_user_with_the_role.js
```diff
@@ -0,0 +1,9 @@
+import { Then } from "cypress-cucumber-preprocessor/steps";
+import faker from "faker";
+
+Then(`I am logged in as a user with the {string} role`, (text) => {
+  let username = faker.internet.userName();
+  let password = faker.internet.password();
+  cy.drupalAddUserWithRole(text, username, password);
```
acrollet, line 7: I like this, is there anything happening to remove the created user after the test is complete?

tests/behavioral/cypress/integration/common/i_create_a_node.js
```diff
@@ -0,0 +1,18 @@
...
+    cy.scrollTo('top');
+    cy.findAllByLabelText('Page title').type(faker.lorem.sentence(), { force: true });
+    cy.findAllByLabelText('Page introduction').type(faker.lorem.sentence(), { force: true });
+    cy.get('form.node-form').find('input#edit-submit').click();
+  }
+};
+
+Given(`I create a {string} node`, (contentType) => {
+  let creator = creators[contentType];
+  assert.isNotNull(creator, `I do not know how to create ${contentType} nodes yet.  Please add a definition in ${__filename}.`);
```
acrollet, line 16: 💯

tests/behavioral/cypress/integration/common/i_am_logged_in_as_a_user_with_the_role.js
ndouglas: Sure.  Do you think we should not worry about removing them if a test fails (to allow some freedom in debugging) but remove them after a successful test?

acrollet: hmm, perhaps we could detect if the tests are being run interactively and not remove in that case? The concern is that leftover test content will be confusing for end users in tugboat environments and on dev/staging. If we can move to ephemeral DBs for testing then I wouldn't be concerned at all... LMK what you think.

ndouglas: Or just keep 'em on Lando, zap 'em elsewhere?

acrollet: That works for me - this would be a good thing to revisit with everyone when we have the testing spike I think.

ndouglas: Added that to #4109 😃

ndouglas: Did some reading about this and it seems to be a nightmare.

Cypress [warns against](https://docs.cypress.io/guides/references/best-practices.html#Using-after-or-afterEach-hooks) using the after/afterEach hooks for this.  So okay, we could use before/beforeEach.  

The best use of that would be to just set the DB back to a known state, I figure.  That also removes any work performed by users or any other test processes.  Guessing that's an issue on BRD.

But ignoring that, the process would seem to be either costly in terms of time (if we just reimported the DB at the end) or simply not possible (ideally, we could just start a transaction when tests are started and then cancel it when tests end, but MySQL doesn't support nested transactions [does anything?  probably PostgreSQL... [eh, not really](https://stackoverflow.com/questions/25425944/does-postgres-support-nested-or-autonomous-transactions)].

**TL;DR**: I was hoping to throw in a few lines that'd be 99% effective pending the testing spike.  Looks like that's not going to happen.

Given that we already create junk test data, I feel fairly comfortable doing diddly about this until the spike.

acrollet: Ok, per earlier discussions with @swirtSJW perhaps we could prefix the auto-generated usernames/titles with `test_` or some such to help make things clear to users?

ndouglas:
@acrollet eh?

acrollet:
I see the JSD widget on a 403 request as expected:

Screen Shot 2021-02-08 at 1 09 59 PM

Issue #3998, opened 2021-01-13
bulk CMS user password creation


xiongjaneg:
### User Story
As a CMS administrator, I want to be able to manually set the password for new users when importing via CSV at https://prod.cms.va.gov/migrate_source_ui

### Acceptance criteria

- [ ] Rows should be validated to confirm that the email address ends with `@va.gov`
  - If not, an on-screen message should be displayed to the user running the import: "The user with email address `foo@example.com` was not created, since it is not a va.gov email address."

- [ ] New user accounts should be configured to use SSO for login by default
- [ ] default password is set to `drupal8` if not on BRD + status message is displayed to user running the import "Password was set to `drupal8` for all imported users".
- [ ] Users accounts imported on non-BRD environments are imported as "Active"
   - [ ] a message is displayed to the user importing accounts on BRD "All accounts are created as active."
- [ ] User accounts imported on BRD are created as "Blocked"
    - [ ] a message is displayed to the user importing accounts on BRD "All accounts are created as blocked."

oksana-c:
@acrollet the last two ACs about the SSO and pass need to change based on the 16th min convo we've had this morning.

All users should have SSO enabled by default.

The part around passwords needs a bit more thinking.
We should  _not_ generate random passwords.
User accounts are created in bulk to support training, but also these accounts will stick around on prod.
On prod users should be logging in w/ PIV.
On training environments, passwords are reset anyway, so there is no need to set them at all.

So if we follow the workflow ->> `generate accounts on prod > propagate to training by rebuilding the env` then we should be good without default passwords.

the only need to have default passwords is when we cannot create user account on prod for whatever reason. 
where workflow is reversed ->> `generate accounts on training env > generate them again on prod once users are cleared w/ training` 
Then we probably need to have a column for default passwords where if @xiongjaneg  starts importing on training env instead on prod, she can assign `drupal8` to all imported users. 

The security constraint would be:
* do not set passwords on prod even if they exist in the import file + output the message  "passwords are not set - users will to log in with PIV" and
* allow passwords to be set only on non-prod environments

oksana-c:
@acrollet the last two ACs about the SSO and pass need to change based on the 16th min convo we've had this morning.

All users should have SSO enabled by default.

The part around passwords needs a bit more thinking.
We should  _not_ generate random passwords.
User accounts are created in bulk to support training, but also these accounts will stick around on prod.
On prod users should be logging in w/ PIV.
[...]

Issue #3072, opened 2020-09-30
Change default password on lower environments from "drupal8" to something less drupally, and easy to type 


ElijahLynn:
Change default password on local developer snapshot DB from "drupal8" to "1", or something where it is less keystrokes...

`va`?
`1`?
etc.

## ACs:
- [ ] Change Sanitize job to set default password to "_tbd_" (token) (devops repo)
- [ ] Update HTTP Auth for GraphQL queries from vets-website PRs (devops or vets-website repo)
- [ ] Update docs: https://github.com/department-of-veterans-affairs/va.gov-cms/search?q=drupal8
- [ ] Update references to 'drupal8' in `va_gov_user\EventSubscriber\UserImport` (cf: #3998) (cms repo)

## New Stories when committing:
- [ ] Training material updates
- [ ] Release note / communication to relevant stakeholders

Further discussion: (Internal) https://dsva.slack.com/archives/CT4GZBM8F/p1601495109277200

robbiethegeek:
WeAre1VA :)

ElijahLynn:
> WeAre1VA :)

Thanks, updated!

ElijahLynn:
Changed this to "1" now for maximum developer experience, as it seems easiest to type.

kevwalsh:
I've changed the title of this, i don't think "1" is ideal, and this merits some discussion.

ElijahLynn:
> I've changed the title of this, i don't think "1" is ideal, and this merits some discussion.

K, the goal in my eyes is for the fastest UX and that would be a single character for my vision. Definitely up for discussing!

acrollet:
@kevwalsh could you expand a bit and provide a suggestion? We're now at the point where `drupal8` is inaccurate :-)

TheBoatyMcBoatFace:
Not a priority right now. Reopen if this becomes an issue.

PR #4188, opened 2021-02-03
VACMS-3998: user import improvements

acrollet:
## Description

See #3998 

## Testing done

Manual testing on lando with default env and faked as BRD

## QA steps

As an admin user:
1. Go to `/migrate_source_ui`
1. Import a file with emails that are @va.gov and another domain
- [x] VA emails are imported
- [x] Non-VA emails are not imported and you see this message:
  - [x] "The user with email address foo@example.com was not created, since it is not a va.gov email address."
- [x] Password is set to `drupal8` and the user can log in with password
- [x] These messages are shown:
  - [x] "All accounts are created as active."
  - [x] "Password was set to drupal8 for all imported users"

Then, fake the environment you're using to test as BRD by editing its settings file:
- Change `$settings['va_gov_frontend_build_type']` to `brd`.
- Add this line: `settings['jenkins_build_env'] = 'test';`

- [x] VA emails are imported
- [x] Non-VA emails are not imported and you see this message:
  - [x] "The user with email address foo@example.com was not created, since it is not a va.gov email address."
- [x] New user accounts should be configured to use SSO for login
- [x] New user accounts are blocked
- [x] These messages are shown:
  - [x] "All accounts are created as blocked."
  - [x] "Password was set to drupal8 for all imported users"

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


docroot/modules/custom/va_gov_user/src/EventSubscriber/UserImport.php
```diff
@@ -0,0 +1,320 @@
...
+
+  /**
+   * Perform Post Import events.
+   *
+   * @param \Drupal\migrate\Event\MigrateImportEvent $event
+   *   Information about the event that triggered this function.
+   *
+   * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   * @throws \Drupal\migrate\MigrateException
```
ndouglas, line 126: I'm interested in why you're declaring these.  Genuinely interested, mind you, not that arms-crossed-across-my-chest-looking-down-my-nose-at-you-highly-skeptical-and-resistant-to-persuasion way that I so often have re: your pull requests.  I feel like I might really learn something here, so I don't wanna just wander past this.

As I see it, the main value of a @throws declaration is saying what exceptions might be thrown in this function so that some hapless schmuck calling the function will know, like, "okay, this can throw a FileNotFoundException because we're dealing with a calculated file path, but this is 1) Nate's backups so we want to rethrow that 2) Neil's backups so we want to fail silently 3) an SMB wrapper so we want to rethrow that as a different error" or something like that.  Except I don't really do it because I like to keep successive devs on their toes.

But in this case, since we're using an event subscriber, we're basically just documenting this into the void, right?

Or is this a matter of completeness?

Or if we want to test it by simulating some real-world nightmares?

docroot/modules/custom/va_gov_user/src/EventSubscriber/UserImport.php
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    $this->addUserToSections($row, $user);
+
+    if ($this->environmentDiscovery->isBRD()) {
+      $this->enableSamlAuth($user);
+      $user->block();
+    }
+    else {
+      $user->setPassword('drupal8');
```
ndouglas, line 183: Feel like we should use a static variable or something for the password value.  Or settings.  I want to avoid the case where we're doing a repo-wide search for `drupal8`.  Mebbe not necessary.

ndouglas:
Working so far with my super-secret CSV (one dupe caused a failure as expected)

Screen Shot 2021-02-03 at 3 26 12 PM

ndouglas:
With BRD-style settings:

Screen Shot 2021-02-04 at 2 46 24 PM
Screen Shot 2021-02-04 at 2 47 15 PM

ndouglas (approved):

docroot/modules/custom/va_gov_user/src/EventSubscriber/UserImport.php
acrollet: wow, I somehow never saw this. Honest answer - it was a copy-paste job and I never questioned the embodied assumptions! Cleaned up in #4358

docroot/modules/custom/va_gov_user/src/EventSubscriber/UserImport.php
acrollet: Also missed this - this is a good point, I had done a search and not seen any other references in the code, so I don't know if there's value to having a static var/constant used in exactly one place? I did add an AC to #3072 to update this as well.

Issue #4194, opened 2021-02-04
composer va:web:build fails on Jenkins/BRD only and not on PR or local environments


oksana-c:
## Problem statement

Issues with `vets-website` package break the build on BRD only and not on local or PR environments.
This causes unexpected delays in CMS deployment schedule.
Build breaking events like these should be caught in Tugboat / local environments, so that the team can react sooner.

Broken build sample: http://jenkins.vfs.va.gov/job/testing/job/cms-post-deploy-tests-staging/229/consoleFull 

`Error: Cannot find module 'chokidar'`

## Acceptance Criteria
- [ ] `composer va:web:build` fails / behaves in the same manner on all types of environments: BRD, Tugboat, local.

acrollet:
Ok, I believe I have the cause of this issue worked out.

`yarn install` is failing silently - compare the output in:
- http://jenkins.vfs.va.gov/job/testing/job/cms-post-deploy-tests-staging/227/consoleFull (successful)
- http://jenkins.vfs.va.gov/job/testing/job/cms-post-deploy-tests-staging/228/consoleFull (fails)

In the failing build, the output stops during step `[2/4] Fetching packages...`, without getting as far as `[3/4] Linking dependencies... `. However, the command fails silently with a successful exit status. I believe we're experiencing this bug:

https://github.com/yarnpkg/yarn/issues/6915

I highly suspect that it's related to the on-going issues with the production proxies.

This was fixed in https://github.com/yarnpkg/yarn/pull/6968, but the version of yarn we're using `v1.12.3` is older than that.

Working on a resolution...

indytechcook:
Great find @acrollet !

oksana-c:
@acrollet @indytechcook is this issue resolved?

acrollet:
I think so, @indytechcook have we seen this error again?

oksana-c:
@indytechcook please check comments above ^

indytechcook:
Yes we did.  My current theories are either related to the TIC dropping connections, a routing issue with the requests or how dependabot upgrades vets-website.

ElijahLynn:
Yeah, this isn't an issue anymore. We should close this and re-open if it comes up again.

indytechcook:
Reopening since it happened again.

http://jenkins.vfs.va.gov/job/testing/job/cms-post-deploy-tests-staging/268/consoleFull

```
[...]

PR #4197, opened 2021-02-04
VACMS-4194: Require installed yarn version to match vetsweb

acrollet:
## Description

See #4194 

## Screenshots

Screen Shot 2021-02-04 at 9 04 48 AM

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`


ndouglas (approved):

composer.json
```diff
@@ -140,7 +140,7 @@
         "knplabs/github-api": "^2.12.1",
         "michelf/php-markdown": "^1.8",
         "mouf/nodejs-installer": "^1.0",
-        "npm-asset/yarn": "1.12.3",
+        "npm-asset/yarn": "1.21.1",
```
ElijahLynn, line 143: Since beginning of the VA.gov-CMS 2+ years ago, this was only updated once, and that was just a month or so ago. Given that this is the second update in a month or so, it looks like it is going to be bumped more frequently now!

scripts/composer/check_yarn_version.sh
```diff
@@ -0,0 +1,17 @@
...
+# Allow script to be run anywhere in git repo
+cd "$(git rev-parse --show-toplevel)"
+
+YARN_VERSION_INSTALLED="$( yarn --version )"
+REQUIRED_YARN_VERSION=$( node --print --eval="require('./web/package.json').engines.yarn" | sed 's/[<>=~]*//' )
+if [ ! "$( printf '%s\n' "$REQUIRED_YARN_VERSION" "$YARN_VERSION_INSTALLED" | sort --version-sort | head --lines=1 )" = "$REQUIRED_YARN_VERSION" ]; then
+  echo "Installed yarn version ${YARN_VERSION_INSTALLED} is older than the required version: ${REQUIRED_YARN_VERSION}!"
+  exit 1
+else
+  echo "Installed yarn version ${YARN_VERSION_INSTALLED} is equal or greater to the required version ${REQUIRED_YARN_VERSION}."
```
ElijahLynn, line 16: I would like to see it only allow for a matching version so that there is more certainty in debugging issues.

ElijahLynn (changes requested):

scripts/composer/check_yarn_version.sh
acrollet: Updated!

ElijahLynn (approved):

PR #4343, opened 2021-02-16
FASTER VETS-WEBSITE (va-gov/web from 0.1.1303 to 0.1.1320)!!!!

indytechcook:
## Description

The upgrade!

---
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/4343)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -742,6 +742,16 @@
                 "stream_filter_append",
                 "stream_filter_register"
             ],
+            "funding": [
+                {
+                    "url": "https://clue.engineering/support",
+                    "type": "custom"
+                },
+                {
+                    "url": "https://github.com/clue",
+                    "type": "github"
+                }
+            ],
             "time": "2019-04-09T12:31:48+00:00"
         },
         {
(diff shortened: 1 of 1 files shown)


ndouglas (approved):

ElijahLynn:
After this is merged, let's keep an eye on the "Load" at https://tugboat.vfs.va.gov/admin. There are 36 cores available to to Tugboat right now with each service having access to all of them so seems like it should be fine.

ElijahLynn:
> After this is merged, let's keep an eye on the "Load" at https://tugboat.vfs.va.gov/admin. There are 36 cores available to to Tugboat right now with each service having access to all of them so seems like it should be fine.

It barely got touched, as the Drupal request was 30 seconds. This will be faster the MariaDB 10.5 upgrade in #4303 in a bit too.

indytechcook:
The commit passed test but this PR still has them pending: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/4343/commits/563e967dc0db8d62f547bd65efab13120b3a5294 so I'm merging

Issue #2743, opened 2020-08-21
Images derived from cropped image styles do not respect the crop. 


kevwalsh:
**Describe the defect**
Some cropped images do not respect the user-specified crops 

**To Reproduce**

I'm not able to reproduce this when uploading new images at /media/add/image, but i am able to find issues where existing images that are derivative of crops do not respect the crops. 

Try reproducing by uploading new photos, eg at /media/add/image. Crop at 2x1, and change the default crop. Derivative images of that crop (eg image styles `2_1_large` and `2_1_medium_thumbnail`) should be based on the crop you made. 

On staging and on local, this was true. However, on prod there are definitely examples of images where the derivatives do nto match the crop, see screenshots below:

Possible causes i can think of: 

* this is prod specific
* browser or operating system
* this occurs when adding images from /node/add/news_story (did not test), but not when adding at /media/add/image (did test)
* happens to older images (whether seconds old, or possibly much older)

![IMG_20200504_194411627_jpg__480×240__and_Edit_Image_IMG_20200504_194411627_jpg___VA_gov_CMS](https://user-images.githubusercontent.com/643678/90931573-e69a6b00-e3ca-11ea-88d3-c3a3c41055b7.jpg)
![Item-0_and_Item-0_and_Item-0_and_Item-0_and_com_bjango_istatmenus_combined_and_Item-0_and_Item-0_and_Item-0](https://user-images.githubusercontent.com/643678/90931597-f4e88700-e3ca-11ea-9cda-e3b2d19d715f.jpg)

VanessaLuxen:
This issue has resurfaced with Ryan Stubblebine reaching out via email. I hopped on a zoom call for him to show me exactly what he is continuing to see with cropping images. This is not stopping his work but is an ongoing frustration. He is also reporting that the image quality seems to be reduced in prod. 
![Image Editing 1](https://user-images.githubusercontent.com/73555217/104741477-43e6de00-5717-11eb-8ffc-7c2f2b2c5544.jpg)
![Image Editing 2](https://user-images.githubusercontent.com/73555217/104741483-46493800-5717-11eb-8007-d1ec70341a37.jpg)

ndouglas:
Unless I'm missing something, I _am_ able to reproduce this locally:

Image (reuploaded) with a very specific crop:

[...]

PR #4348, opened 2021-02-17
VACMS-2743: Fix image cropping issues.

ndouglas:
## Description

See #2743. 

The only way I can think of to effectively determine whether the platform is generating cropped images correctly is with visual regression testing.  So after some research and playing briefly with a couple of different options, and thinking about the options available in deployment (for instance, `imagick` doesn't seem to be installed on Tugboat, and it might not be present on BRD), I figured [cypress-visual-regression](https://www.npmjs.com/package/cypress-visual-regression) was likely to be the best path forward.

I'm going to write the tests first and then try to find the simplest change possible that will get the tests to pass.

There are several cases I'm concerned about:
1. when a new node is saved and the image is freshly uploaded and cropped
2. when the derivative is subsequently deleted
3. when a node is edited and the image is freshly uploaded and cropped
4. when a node is edited and the existing image is recropped

(I might think of more as I dig deeper)

Of these cases, I think the only one that does not currently work is # 1, but it's also the most important 😛  This will be confirmed or disproven with tests.

## Testing done
Forthcoming.

## Screenshots
Forthcoming.  Boy, are these forthcoming.

## QA steps
As a user with permission to create new image media:
1. Create a new image media (medium? 🤔 )
2. Crop the image in some subjectively satisfying way
3. Verify that:
    - [x] the image's `2_1_medium_thumbnail` derivative reflects the crop you've selected.

### Definition of Done

### Select Team for PR review
- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


ndouglas:
@acrollet I'd love to get your thoughts on this nightmare tomorrow.  I think there's some insight I'm missing into the DOM or event listening or something else and think you might be just the guy to inflict my neuroses on for a while.

acrollet:
@ndouglas 
[](https://media1.tenor.com/images/baaeeb9bc7c8215943fd04e8b1b50df4/tenor.gif?itemid=14186847)

ndouglas:
@acrollet I ended up figuring it out.  Thank you 😁

acrollet:
@ndouglas nice! Sleeping on it usually helps...

ndouglas:
So the first part of this ("tell cropper.js we're resizing the crop!") did not go as smoothly as I'd anticipated.

giphy (1)

I found a few issues with Cucumber, most importantly that regular expressions in step matching expressions don't seem to be working right now, hence one of the earlier commits reverting all of my triumphant regexes...

And I realized as I went further that I should use slightly different tools -- `cypress-visual-regression` is not appropriate in a case where I'm comparing two raw images, but we have `pixelmatch`...

And this really needs to be refactored and reworded, and I need to improve the way the ideas of these step definitions are expressed.

but given a test like this:

Screen Shot 2021-02-19 at 1 38 01 PM

https://user-images.githubusercontent.com/1318579/108546893-c420d600-72b7-11eb-9567-d9b8ae869587.mp4

We can get a diff image (should we care) and a raw number of pixels that differ.  (In this case, ~67K)
 
Manually cropping an image should lead to correctly cropped derivatives

So I'll write a few tests that branch off this for the other cases I'm worried about and then I can finally work on the solution.

ndouglas:
To succinctly give some background, we have two modules involved: `image_widget_crop`, which creates a crop configuration based on the user interface invoked when an image is uploaded, and `image_style_warmer`, which generates image derivatives when the entity is saved.

The issue is that `image_style_warmer`'s warming is invoked from `hook_ENTITY_TYPE_(insert|update)()`, for `file` entities, whereas `image_widget_crop`'s crop configurations are generated from `hook_entity_(insert|update)()` hooks.  This means that the image derivative is generated before and does not reflect the crop configuration.

`hook_ENTITY_TYPE_*()` hooks are invoked prior to `hook_entity_*()` hooks, unavoidably; changing the weights of these two modules should not ameliorate the bug (and, in testing, does not).

So we have a few possible solutions:
1.  An approach that appeals to me is to rewrite the `image_style_warmer` module to act on events rather than hooks.  If a list of styles and an affected entity are stored, then this list and entity can be retrieved later by a subscriber to the `TERMINATE` kernel event.  This would allow all hooks to modify the entity prior to the image derivative generation.  I haven't tested this approach, though, because ain't nobody got time for that.
2. As above, but enqueue all image derivative generation and run that in response to the `TERMINATE` event.
3. Add or modify an early-running module to hijack `image_widget_crop`'s crop configuration generation service and run it in `hook_ENTITY_TYPE_*()` implementations.  This was my idea from the other day, is the easiest to try, and, fortunately, we already have such an early-running module: `va_gov_content_export`.  And at least to me it seems an acceptable thing to shove into that module.  Unfortunately, this doesn't (and can't) work, as the crop settings don't really exist at the time that the file entity is created.

So we move on:

4.  Hack some extra settings into the `image_style_warmer` module's config form and use those settings for our own nefarious purposes 😈   Form looks like :peach:, but I think we can whip something up with some CSS to make it look more coherent:
 
Screen Shot 2021-02-22 at 3 43 26 PM

And it looks like the code has the desired effect:

Screen Shot 2021-02-22 at 3 44 44 PM
Screen Shot 2021-02-22 at 3 44 49 PM

:tada:

I wanted to get some feedback on this approach (now that I'm confident it works) and whether I should 1) turn it into a separate module, like `va_gov_image` or something, or 2) leave it where it is, or 3) open a PR with the `image_style_warmer` guy, either as an alternative to or replacement for the current behavior.  Or, 4) this approach flat-out sucks and I should fall on my sword.

Thoughts?

ndouglas:
Hey @oksana-c if you have a moment, what's the yuck factor of my approach here?  I was also thinking about editing the `image_style_warmer` form to just hide the "initial image styles" and replace that list with the one I'm using for this late generation.  So it would be a transparent change to the form.

oksana-c:
notes from our 🦆 session:
* let's move all form alters + helper function into `va_gov_media` (remember to mark it as D9 compatible)
* let's open an follow-up story to refactor the altered form + attached functionality once we have a queue runner

ndouglas:
Things to Do:
- disable queue image styles
- disable initial image styles
- move hooks to separate module
- write followup issue -- these hooks should be removed when we have a realtime queue runner; at that point we'll use the queue image styles and not need the initial image styles
- add comment in code referring to followup issue

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1883,3 +1884,63 @@ function _va_gov_backend_set_field_listing_value(EntityInterface &$entity, array
...
+      foreach ($entity->getFields() as $entity_fields) {
+        if ($entity_fields instanceof FileFieldItemList) {
+          foreach ($entity_fields->getValue() as $entity_file_field_value) {
+            \Drupal::service('image_style_warmer.warmer')
+              ->doWarmUp($entity_file_field_value->entity, array_keys($image_styles));
+          }
+        }
+      }
+    }
+  }
```
mchelen-gov, line 1945:
```suggestion
  if (
    !empty($image_styles)
    && isset($entity)
    && $entity instanceof FieldableEntityInterface
  ) {
    foreach ($entity->getFields() as $entity_fields) {
      if ($entity_fields instanceof FileFieldItemList) {
        foreach ($entity_fields->getValue() as $entity_file_field_value) {
          \Drupal::service('image_style_warmer.warmer')
            ->doWarmUp($entity_file_field_value->entity, array_keys($image_styles));
        }
      }
    }
  }
```

ndouglas: Sure!

ndouglas:
New form.  Other changes implemented as discussed in call 😃 

Screen Shot 2021-02-24 at 10 19 53 AM

tests/behavioral/cypress/integration/common/i_select_option_from_dropdown.js
```diff
@@ -0,0 +1,7 @@
+import { Given } from "cypress-cucumber-preprocessor/steps";
+
+Given(`I select option {string} from dropdown {string}`, (option, label) => {
```
acrollet, line 3: ❤️

tests/behavioral/cypress/integration/common/the_derivative_of_the_media_image_should_match_the_fixture.js
```diff
@@ -0,0 +1,31 @@
+import { Then } from "cypress-cucumber-preprocessor/steps";
+const pixelmatch = require('pixelmatch');
+const { PNG } = require('pngjs');
+const { expect } = require('chai');
+
+Then(`the {string} derivative of the media image should match the fixture {string}`, (derivative, fixture) => {
+  cy.window().then((window) => {
+    cy.get('@mediaImageUrl').then((url) => {
+      const derivativeUrl = url.replace('/files/', `/files/styles/${derivative}/public/`);
+      console.log('Derivative URL', derivativeUrl);
```
acrollet, line 10: testing code or useful for CI?

tests/behavioral/cypress/integration/common/the_derivative_of_the_media_image_should_match_the_fixture.js
```diff
@@ -0,0 +1,31 @@
...
+      cy.wrap(derivativeUrl).as('mediaImageDerivativeUrl');
+      cy.request({ url: derivativeUrl, encoding: 'binary' }).then((response) => {
+        const derivativeImage = PNG.sync.read(new Buffer(response.body, 'binary'));
+        const width = derivativeImage.width;
+        const height = derivativeImage.height;
+        console.log(derivativeImage);
```
acrollet, line 16:
```suggestion
```

tests/behavioral/cypress/integration/common/the_derivative_of_the_media_image_should_match_the_fixture.js
```diff
@@ -0,0 +1,31 @@
...
+        cy.fixture(fixture, 'binary').then((fixtureBody) => {
+          const fixtureImage = PNG.sync.read(new Buffer(fixtureBody, 'binary'));
+          const diff = new PNG({width, height});
+          const differences = pixelmatch(fixtureImage.data, derivativeImage.data, diff.data, width, height);
+          const diffData = PNG.sync.write(diff).toString('binary');
+          const path = `cypress/screenshots/pixelmatch_diffs/${cy.state('ctx').test.title}.png`;
```
acrollet, line 22:
```suggestion
          const path =
            `cypress/screenshots/pixelmatch_diffs/${cy.state('ctx').test.title}.png`;
```

tests/behavioral/cypress/integration/image_crop.feature
```diff
@@ -0,0 +1,17 @@
+Feature: CMS Users are able to manually crop images
+  In order to control the presentation of images
+  As anyone involved in the project
+  I need to be able to manually crop images
+
+  @image_crop
+  Scenario: Manually cropping an image should lead to correctly cropped derivatives
```
acrollet, line 7: awesome test

tests/behavioral/cypress/support/commands.js
```diff
@@ -39,6 +42,9 @@ Cypress.Commands.add('drupalLogout', () => {
 
 Cypress.Commands.add('drupalDrushCommand', (command) => {
   let cmd = 'drush %command';
+  if (Cypress.env('VAGOV_INTERACTIVE')) {
+    cmd = 'lando drush %command';
```
acrollet, line 46: yay

acrollet (changes requested):

docroot/modules/custom/va_gov_media/va_gov_media.module
```diff
@@ -0,0 +1,74 @@
+<?php
+
+/**
+ * @file
+ * Contains va_gov_migrate.module.
```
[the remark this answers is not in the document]
ndouglas: Kinda funny that it griped at me when I did't have the @file comment but didn't when I ineptly copypasta'ed it.

tests/behavioral/cypress/integration/common/i_delete_the_derivative_of_the_media_image.js
```diff
@@ -0,0 +1,16 @@
+import { Then } from "cypress-cucumber-preprocessor/steps";
+
+Then(`I delete the {string} derivative of the media image`, (derivative) => {
+  cy.window().then((window) => {
+    cy.get('@mediaImageUrl').then((url) => {
+      const derivativeUrl = url.replace('/files/', `/files/styles/${derivative}/public/`);
+      cy.exec('npm run build', { timeout: 20000 })
```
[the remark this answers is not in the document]
ndouglas: none at all, accidentally left behind reference code

tests/behavioral/cypress/integration/common/the_derivative_of_the_media_image_should_match_the_fixture.js
ndouglas: Not allowing commits with `console.log()` calls would be a nice hook

acrollet: I think eslint will catch that if we add the behavioral dir to it.

acrollet: let's not delay this PR on that though.

tests/behavioral/cypress/integration/image_crop.feature
ndouglas: Thanks... hope we can find a way to improve some of that phrasing.

acrollet: meh, it does what it says on the tin :-)

acrollet:
Crop UI:

Screen Shot 2021-02-24 at 1 19 59 PM

Generated Image:
Image from iOS-2

acrollet (approved):
Manual test passed, great new tests coming in.

Issue #4203, opened 2021-02-04
Refactor usages of computed field and remove Computed Field module


oksana-c:
## Problem statement

While value of Computed Field module is widely accepted by most site builders, there are many good reasons not to use it (out of scope of this issue).

Computed field add complexity and hides logic in unexpected places.

Current implementation of Computed Field module doesnt seem to be adding any value. All computed field values are currently set to empty.
`$value = '';` 

```
grep -ri computed *
core.extension.yml:  computed_field: 0
field.field.node.full_width_banner_alert.field_banner_alert_computdvalues.yml:    - computed_field
field.field.node.full_width_banner_alert.field_banner_alert_computdvalues.yml:label: 'Computed values for alerts'
field.field.node.full_width_banner_alert.field_banner_alert_computdvalues.yml:field_type: computed_string_long
field.storage.node.field_banner_alert_computdvalues.yml:    - computed_field
field.storage.node.field_banner_alert_computdvalues.yml:type: computed_string_long
field.storage.node.field_banner_alert_computdvalues.yml:module: computed_field
```

[Related internal Slack thread](https://dsva.slack.com/archives/CT4GZBM8F/p1612462925003200)

## Acceptance Criteria
- [ ] Current usage of Computed Fields is reviewed and the fact that is does nothing is validated
- [ ] All computed fields are converted to text fields or fields of appropriate type that matches established data model
- [ ] Change in data/content model is coordinated with appropriate FE team that is responsible for maintaining FE pages and transformers for identified fields
- [ ] Computed Field module is uninstalled and removed from the CMS 
- [ ] no DB tables left behind
- [ ] add tests if feasible

kevwalsh:
Seems like va_gov_govdelivery.module is setting a value for this field behind the scenes.

https://github.com/department-of-veterans-affairs/va.gov-cms/blob/e1bff4cfb9e28621a71a5e94a1dd6e184632248b/docroot/modules/custom/va_gov_govdelivery/va_gov_govdelivery.module#L96 

@swirtSJW can probably speak more to what's dependent on this, for sending emails out through govdelivery.

oksana-c:
@swirtSJW please add any details you have. The Core Application team is planning to work on this in Sprint 26. Thank you,

acrollet:
[...]

PR #4364, opened 2021-02-18
VACMS-4203: remove computed fields and dependent code

acrollet:
## Description

See #4203 

## Testing done

Updated and iteratively ran phpunit tests.

## QA steps

- [ ] phpunit test has been reviewed and is passing
- [ ] No computed field tables remain in the DB: `SHOW TABLES LIKE '%computd%';`
- [ ] banner alert nodes can be created & updated without errors
- [ ] the `Send email update on this situation` checkbox field is disabled for published banner alert nodes

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


acrollet:
@swirtSJW is there a way to manually test alert sending without spamming real people?

docroot/modules/custom/va_gov_govdelivery/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -0,0 +1,71 @@
...
+   *
+   * @param Drupal\Core\Entity\EntityInterface $entity
+   *   The entity being created or updated.
+   *
+   * @return bool
+   *   Whether or not this is an status update node.
+   */
+  protected function isStatusUpdateNode(EntityInterface $entity) : bool {
+    return $entity instanceof NodeInterface &&
+      $entity->getType() === 'full_width_banner_alert';
```
swirtSJW, line 68:
```suggestion
    return ($entity instanceof NodeInterface) &&
      ($entity->getType() === 'full_width_banner_alert');
```
Not necessary but more explicit?

docroot/modules/custom/va_gov_govdelivery/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -0,0 +1,71 @@
+<?php
```
swirtSJW, line 1: I REALLY like this architecture.   Thank you for implementing it.

One concern: with other facility types coming that will also be leveraging govdeliver  in different ways, should this 'EntityEventSubscriber' be namespaced more specifically to the content type?  Like 'EntityEventSubscriberStatusBulletins' or are you envisioning that we just add additional cases to 'processStatusBulletins()'?  If the latter then the naming of 'processStatusBulletins()' should probably become less specific like 'processStatusBulletins()'.
Seems like either the class needs to be more specific or the method needs to be less specific (heisenburg naming uncertainty principle :)  )

docroot/modules/custom/va_gov_govdelivery/src/Service/ProcessStatusBulletin.php
```diff
@@ -48,126 +109,183 @@ public function __construct() {
...
-      'theme_hook_original' => 'not-applicable',
+      '#alert_type' => $node->get("field_alert_type")->value,
+      '#theme' => 'va_gov_body_alert',
     ];
 
     if ($this->sendType === 'status alert') {
       $queue_id = "{$node->get('nid')->value}-alert";
-      // Pull the data from the alert fields.
-      $template_variables['message'] = $node->get('field_body')->value;
-      $template_variables['situation_update'] = FALSE;
```
swirtSJW, line 68: Intentionally removed?

swirtSJW (commented):
> @swirtSJW is there a way to manually test alert sending without spamming real people?

Two ways.

1.  Use one environment that does not have your code.   Edit a facility banner alert.   Look in the queue to get the data.   Repeat the same steps on an environment with your changes.  Compare the queued data.  There should be no diff.
2. Look in phpinfo on tugboat to see if it has govedelivery dev endpoint set.  If it does you could test on tugboat then have Elijah or I log into govedeliver dev and see if it sent.   If tugboat can't do it, we may have to have Elijah or Neil pull your PR to dev so that it can do a test send.

docroot/modules/custom/va_gov_govdelivery/va_gov_govdelivery.module
```diff
@@ -5,94 +5,43 @@
...
-    $node = $form_state->getFormObject()->getEntity();
-    if ($node->isPublished() || ($node->moderation_state->value === 'published')) {
-      // Disable send email field, this form will only send initial email once.
-      // @todo Check node revisions in addition to current state, disable this
-      // field if it was ever published.
-      $form['field_operating_status_sendemail']['widget']['#disabled'] = TRUE;
-    }
+function va_gov_govdelivery_form_alter(&$form, FormStateInterface $form_state, $form_id) : void {
+  if ($form_id !== 'node_full_width_banner_alert_edit_form') {
+    return;
```
swirtSJW, line 37: I am troubled by the use of this bail out early pattern on things like hook_form_alter because it does not encapsulate the code that is supposed to run if that is not the case.   It makes it risky for other developers' additions to the alter hook that might be unrelated to full width banners.  Ideally this could just be moved to another function  that gets called from within the alter.  Then it does its own defensive bail early check.

swirtSJW (commented):
Very awesome and thorough work.  I only had a few questions or concerns.

composer.lock
```diff
@@ -19444,6 +19496,78 @@
             ],
             "time": "2012-10-28T21:08:28+00:00"
         },
+        {
+            "name": "drupal/computed_field",
```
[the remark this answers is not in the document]
acrollet: brain fart, thank you!

docroot/modules/custom/va_gov_govdelivery/src/EventSubscriber/EntityEventSubscriber.php
acrollet: makes sense, updated

docroot/modules/custom/va_gov_govdelivery/src/Service/ProcessStatusBulletin.php
acrollet: good eye - I had set it to default to `FALSE` in the declaration of `$template` but that line went missing somehow. I put it back here since it doesn't really save anything to move it up.

docroot/modules/custom/va_gov_govdelivery/src/Service/ProcessStatusBulletin.php
```diff
@@ -48,126 +109,183 @@ public function __construct() {
...
+        ->referencedEntities();
+      $situationUpdatesLast = end($situationUpdatesList);
+
+      $situation_update_send =
+        !empty($situationUpdatesLast) ?
+        $situationUpdatesLast->get('field_send_email_to_subscribers')->value :
+        FALSE;
+
+      if (!$situation_update_send) {
+        return;
```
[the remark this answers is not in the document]
acrollet: You were right, this wanted a big re-factor ;-)

docroot/modules/custom/va_gov_govdelivery/va_gov_govdelivery.module
acrollet: good call, I switched to `hook_form_FORMID_alter` instead.

acrollet:
@swirtSJW the code is ready for another pass when you are!

acrollet:
The preview is failing, I'm seeing this: 
```
Next Drupal\Component\Plugin\Exception\PluginNotFoundException: Unable to determine class for field type 'computed_string_long' found in the 'field.storage.node.field_banner_alert_computdvalues' configuration in /var/lib/tugboat/docroot/core/modules/field/src/FieldStorageConfigStorage.php:174
```

Looks like we'll need a first PR to remove the config, and a follow-on to remove the module.

docroot/modules/custom/va_gov_govdelivery/va_gov_govdelivery.module
```diff
@@ -5,94 +5,46 @@
...
-    // Convert to string for field storage.
-    $storage = json_encode($computed_return);
-    $node->field_banner_alert_computdvalues->setValue(['value' => $storage]);
+function va_gov_govdelivery_form_node_full_width_banner_alert_edit_form_alter(
+  &$form,
+  FormStateInterface $form_state
+) : void {
+  $node = $form_state->getFormObject()->getEntity();
+  if (
+    $node->isPublished() || ($node->moderation_state->value === 'published')
```
indytechcook, line 44: Let's verify if checking the moderation state is necessary.

docroot/modules/custom/va_gov_govdelivery/src/Service/ProcessStatusBulletin.php
```diff
@@ -46,128 +87,273 @@ public function __construct() {
...
+   * @param string $send_type
+   *   The alert send type.
+   *
+   * @return string
+   *   The queue id.
+   */
+  protected function getQueueId(NodeInterface $node, $send_type) : string {
+    $queue_type = '';
+
+    if ($send_type === 'status alert') {
```
indytechcook, line 171: Let's use CONST for these

indytechcook (changes requested):

docroot/modules/custom/va_gov_govdelivery/va_gov_govdelivery.module
acrollet: good call, the banner alert node type is not moderated. Removed and also added logic to ensure the field is disabled if the node has ever been published.

acrollet:
Manual test 1: I created a new published bulletin node with sending enabled on a clone of master and this branch. I diffed the created queue items and found differences only in timestamps and IDs:

[branch_initial_publish.txt](https://github.com/department-of-veterans-affairs/va.gov-cms/files/6038122/branch_initial_publish.txt)
[master_initial_publish.txt](https://github.com/department-of-veterans-affairs/va.gov-cms/files/6038123/master_initial_publish.txt)

(Queue items retrieved with `drush sqlq "SELECT * FROM queue WHERE name = 'govdelivery_bulletins' ORDER BY item_id DESC LIMIT 1\G"`)

acrollet:
Manual test 2: I updated the previous bulletin node to add a situation update with sending enabled on a clone of master and this branch. I diffed the created queue items and again found differences only in timestamps and IDs:

[branch_situation_update.txt](https://github.com/department-of-veterans-affairs/va.gov-cms/files/6038145/branch_situation_update.txt)
[master_situation_update.txt](https://github.com/department-of-veterans-affairs/va.gov-cms/files/6038146/master_situation_update.txt)

acrollet:
@swirtSJW calling your attention to my previous comments, please take a look when you're able. Thanks! :-)

indytechcook (approved):

PR #4683, opened 2021-03-15
Add postStartCommand to start lando on resume

acrollet:
Automatically start lando when a codespace is resumed. Creating as draft until I have a chance to confirm that this works. (I don't see a way to force-suspend a codespaces instance)


acrollet:
Tests passed:
Screen Shot 2021-03-15 at 12 05 38 PM

Lando start was correctly run after a resume: (correctly did not run on create)
Screen Shot 2021-03-15 at 12 04 04 PM

Issue #4690, opened 2021-03-15
Clarify logic for frontend build trigger


mchelen-gov:
## Acceptance Criteria
- [ ] 👀 **Needs design review** because the work involves changes to labels or other design elements 
- [ ] 🙈 Can be implemented without consulting design team members
- [ ] Testable_Outcome_X
- [ ] Testable_Outcome_Y
- [ ] Testable_Outcome_Z

indytechcook:
@mchelen we have some tests already, what sort of tests were you talking about?

mchelen-gov:
@indytechcook oops i meant to update the title, mainly the goal was to refactor this a bit for readability and wanted to make sure it wasn't breaking anything

PR #4698, opened 2021-03-15
clarify frontend build conditional logic

mchelen-gov:
## Description

See https://github.com/department-of-veterans-affairs/va.gov-cms/issues/4690

## QA steps

1. Do this
1. Then that
- [ ] This
- [ ] That
- [ ] The other thing

### Definition of Done

### Select Team for PR review

- [ ] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


swirtSJW (approved):
Works for  me.

swirtSJW:
We could uses some QA steps to make sure this change works.
image

indytechcook (approved):
Tests? :)

mchelen-gov:
@swirtSJW @indytechcook that gets to my question about automated test coverage. looking at this feature for example, is it covered by Cypress, PHPUnit, Behat, etc?

ndouglas:
@mchelen I just started work on some PHPUnit kernel tests for this.  I don't believe there are any existing tests covering this class, though there are some for related classes and systems.

Part of the issue, I think, is that ~90% of the logic of this class is dealing with the complexity of nodes.  And not just that Node is complicated, and that all of its components are complicated, but that some of those complications (like `$node->original`) aren't really that OOP-friendly, aren't declared in NodeInterface, etc.  And that makes static analysis hard, and mocking hard, and then we have to use actual Node objects and that makes it slow and costly and tedious to write automated tests and then to execute those tests.

I think this is an interesting case... the code is solid and we know it works, but because of Drupal's IMHO incomplete commitment to OOP there are some gaps that make further improvement for testability a bit expensive and require refactoring.

In the interests of full disclosure, though, my first OOP language was Objective-C and I still find myself approaching solutions to problems through the lens of how I would approach them in ObjC, even though those same solutions wouldn't necessarily work in PHP.  For instance in Objective-C I could add a category on the Node class that would contain a method `-(bool)shouldTriggerFrontendBuild` that could cleanly encapsulate a lot of this logic and remove it from this class entirely.  Doing the same with a trait in PHP won't work (or won't work without reflection and other things I'd rather not consider).  

So then what, a service in `va_gov_build_trigger` to separate out the business logic associated with nodes?  Maybe another one to ease some of these operations (like dealing with changed fields) on nodes?

Asking honestly, because I feel like this is a place where I'm definitely lacking some insight into the best solution to this sort of problem.  I have in the past created substantial collections of services to encapsulate this sort of logic.  And it works -- I write a class, I write some unit tests, the tests pass (and quickly!), the class works, and I sleep better.  But I do question whether I took the best approach, particularly when I have to say "yeah, I took your class full of solid, known-working code and turned it into three classes and three test classes."  Maybe there's a better solution that involves adding just a trait.  Or maybe I'm missing some insight into testing that would make the existing class cleanly testable with no refactoring at all.

**TL;DR**: On an intellectual level I find testing, testability, and software design very interesting.  On a doughy-manboy-with-expensive-hobbies level, I'd like to be good at my job, so pls enlighten me.

mchelen-gov:
@ndouglas hmm that's a really good question, i'm not sure about the best practice for unit testing a class like this. splitting out `triggerFrontendBuildFromContentSave()` into a function that doesn't use `$this` seems fine for that particular function, but it doesn't really address the challenges with testing the class as a whole.

where exactly is the pain point in this case with mocking the constructor parameters, instantiating the class, and testing the behavior of public functions? i know some objects may use other drupal services but i'm not sure if that is occurring in this particular case

ndouglas:
>where exactly is the pain point in this case with mocking the constructor parameters, instantiating the class, and testing the behavior of public functions?

Nah, it's not about the services.  Mocking services is easy.  I've mocked them hundreds or thousands of times.  They adhere to an interface, by convention.  Node?  Node annihilates conventions.  It's Legionnaires' Disease in code form.

Dig this:

Screen Shot 2021-03-26 at 11 36 45 AM

Again, solid code, done in the Drupal way, and we know it works.

But it relies on eldritch abominations within Drupal.  Shambling horrors, which we should consign to the outer darkness where they'll thrash amid the pounding of drums and the hideous shrieking of accursed flutes.

The Shuggoth in question:
Screen Shot 2021-03-26 at 11 42 44 AM

Swirt knows what it is, I know what it is, you know what it is.  But static analysis can't know what it is because it's not declared in the [Interface](https://api.drupal.org/api/drupal/core%21modules%21node%21src%21NodeInterface.php/interface/NodeInterface/9.0.x) or even the [Class](https://api.drupal.org/api/drupal/core%21modules%21node%21src%21Entity%21Node.php/class/Node/9.2.x)!  (btw see John Carpenter's _In the Mouth of Madness_ (1994) for a _roman à clef_ of the typical progression of a Drupal developer's career.)

The function _says_ it operates on NodeInterface.  But it _lies_.  PHPStan would be appalled, as it should be.  And if we create a mock NodeInterface object in a test and pass it in, we can't cover the the full complement of test cases because the interface itself cannot represent them or fulfill the, uh, behavioral contract that a real Node object can.

And similarly, the rest of the line:
Screen Shot 2021-03-26 at 11 59 07 AM

So we'd transition to passing real Node objects instead, and that means involving fields (and again we can't just use interfaces because [FieldItemListInterface](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Field%21FieldItemListInterface.php/interface/FieldItemListInterface/9.2.x) doesn't declare a `value` property), we have to use real FieldItemLists, and real FieldItems, and a DB, and services, and this becomes a kernel test that's massively slower than a unit test.  

But that's not enough, because that doesn't model these specific content types and their specific fields, so we either have a substantial amount of setup code at the beginning of test classes:

Screen Shot 2021-03-26 at 12 31 41 PM

That's just to test one field!  Really!

Or we take the next step, which is moving to test classes that actually use the existing site as a base.  This works, is the fastest way of getting a test into the codebase, and is presumably faster than kernel tests because it doesn't involve the Drupal installation process.  But it causes issues since it's operating on the site database as deployed.

So, uh, to be clear I'm not complaining or saying this is impossible or anything like that, and I'm not taking a defensive or otherwise combative stance.  Hopefully this comes across in a positive, constructive, engaged way, because that's how I feel.  There's some issues about testing improvements assigned to me, and I wanted them, I'm enthusiastic about them, and I hope to complete them in a way that makes everyone happier and improves the project as a whole 😃 

But having developed very large projects with really high test coverage in Drupal and, comparatively, in NodeJS, I grok why we have a lot of test-less classes in this codebase and why the tests we have take forever to run.  I'm hoping I can make our testing experience more like the JS experience, which seems almost effortless and enjoyable to me, but JS really benefits from some architectural differences from PHP.

PR #4756, opened 2021-03-22
[DO NOT MERGE] Memecache POC

indytechcook:
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/4756)

### DIFF
--- a/docroot/sites/default/settings.php
+++ b/docroot/sites/default/settings.php
@@ -57,11 +57,16 @@ $settings['hash_salt'] = 'HJwuroKYGPRzhRHXnWR7H38RkH9rOkJ0WP8C5qD_pRStai8zvFX655
  * TRUE back to a FALSE!
  */
 $settings['update_free_access'] = FALSE;
+$env_type = getenv('CMS_ENVIRONMENT_TYPE') ?: 'ci';
 
 /**
  * Load services definition file.
  */
 $settings['container_yamls'][] = $app_root . '/' . $site_path . '/../default/services.yml';
+// Environment specific services
+if (file_exists($app_root . '/' . $site_path . '/services/services.' . $env_type . '.yml')) {
+  $settings['container_yamls'][] = $app_root . '/' . $site_path . '/services/services.' . $env_type . '.yml';
+}
 
 /**
  * The default list of directories that will be ignored by Drupal's file API.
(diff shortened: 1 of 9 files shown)


docroot/sites/default/settings.php
```diff
@@ -57,11 +57,16 @@
...
+$env_type = getenv('CMS_ENVIRONMENT_TYPE') ?: 'ci';
 
 /**
  * Load services definition file.
  */
 $settings['container_yamls'][] = $app_root . '/' . $site_path . '/../default/services.yml';
+// Environment specific services
+if (file_exists($app_root . '/' . $site_path . '/services/services.' . $env_type . '.yml')) {
+  $settings['container_yamls'][] = $app_root . '/' . $site_path . '/services/services.' . $env_type . '.yml';
+}
```
ndouglas, line 69: was actually thinking about doing this for non-memcached-related reasons so we could debug themes, etc easier

indytechcook: I'm not 100% about the exact pattern, I did something different here: https://github.com/department-of-veterans-affairs/va.gov-cms/commit/9f8f0e3f7efb55339db2b60b0e059334164604ab

Thoughts?

I was also thinking about a feature flag model where settings and/or services files are included based upon env variable like `CMS_CACHE_ENABLED`  This would make it easier to turn it on/off locally for testing.

ndouglas: I like the path pattern.  Changing environment variables sounds kinda irritating to me -- that'd require a restart of the PHP container in addition to clearing the caches, wouldn't it?

IMHO something like (untested):

```php
$services_dir = $app_root . '/' . $site_path . '/services';
$env_base = 'services.' . $env_type;

// services.lando.yml
$env_services = $services_dir . '/' . $env_base . '.yml';
if (file_exists($env_services)) {
  $settings['container_yamls'][] = $env_services;
}

// services.lando.local.yml
$env_local_services = $services_dir . '/' . $env_base . '.local.yml'; 
if (file_exists($env_local_services)) {
  $settings['container_yamls'][] = $env_local_services;
}
```

The most useful pattern for me personally would be to copy a `services.lando.yml` file to `services.lando.local.yml` post-install, and then have Lando commands that use `yq` or something to modify those settings in that local file and then clear caches as appropriate.  That way we can maintain sane defaults on a projectwide level but allow local modifications pretty easily.  Whenever I have to do Twig templating or something like that, I do `lando local-services "twig.config.debug: true"` or something like that.

I don't mess with services.yml all that much, but when I do I miss functionality like this.

indytechcook:
This doesn't quite work on tugboat yet but it does work on lando.

indytechcook:
The POC is complete.

Issue #4295, opened 2021-02-11
Upgrade Views Bulk Edit


oksana-c:
https://www.drupal.org/project/views_bulk_edit

minor version jump from 8.x-2.4 to 8.x-2.5

## Acceptance Criteria
- [ ] module is upgraded
- [ ] no regression in functionality - /admin/content/bulk Bulk Field Edit Action are executed without errors

PR #4770, opened 2021-03-22
VACMS-4296: update vbo

acrollet:
## Description

See #4296 

## Testing done

Manual testing of VBO views

## QA steps

As a user with the content admin role
1. Verify that:
- [ ] VBO views display the same available actions as on prod
- [ ]  /admin/content/bulk Bulk Actions are executed without errors
- [ ] overrides work correctly on /admin/content/facility-status (select a facility, click "Update operating status", confirm that only the Status/Details fields are shown)

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [x] `Core Application Team`
- [x] `Product Support Team`

### Is this PR blocked by another PR?

- [ ] `DO NOT MERGE`

### Does this PR need review from a Product Owner

- [ ] `Needs PO review`

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change? 
- [ ] Yes, and it's written in issue ____ and queued for publication. 
  - [ ] Merge and ping @ rachel-kauff so she's ready to publish after deployment
- [ ] Yes, but it hasn't yet been written 
  - [ ] Don't merge yet -- ping @ rachel-kauff to prompt her to write and queue content
- [x] No announcement is needed for this code change. 
  - [ ] Merge & carry on unburdened by announcements

PR #4779, opened 2021-03-23
VACMS-4295: Upgrade views_bulk_edit, apply paragraphs patch.

ndouglas:
## Description

See #4295.  This turned out to be an adventure because my cable modem is (I think) overheating wildly and causing my internet connection to drop pretty frequently, so everything took about six times longer than it should've.

I encountered the following error when creating services on facilities:

Screen Shot 2021-03-23 at 2 40 06 PM

I've encountered this issue before with `paragraphs`, but checked and found that it had been reported for `views_bulk_edit` @ 2.4 already, and the module maintainer had tracked it down and determined that it was `paragraphs`.  It's a longstanding issue and has a relatively up-to-date patch, so I installed it and confirmed that it fixed the issue:

Screen Shot 2021-03-23 at 2 37 26 PM

But why didn't this issue appear in 2.4?  I tested on staging and I couldn't trigger the bug.  However, it appears that I wasn't actually adding any paragraphs either!  I haven't dug too deeply to see whether it was just failing silently or if there's a more complicated defect conspiring against me.

Anyway, this seems to work now.  Possibly for the first time.

## Testing done
Clicked around and updated various fields on multiple items of different types.

## QA steps

As a developer:
1. Validate Acceptance Criteria from issue #4295:
- [ ] module is upgraded
- [ ] no regression in functionality - /admin/content/bulk Bulk Field Edit Action are executed without errors

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


acrollet:
So I'm trying to test this... I selected a couple of Q&A nodes, chose the "Need more help?" field, and selected the "VA burial benefits and memorial items" hub. Clicking "Apply" was a no-op, checked the js console and I see:

```
An invalid form control with name='node[q_a][field_related_information][0][subform][field_link][0][uri]' is not focusable.
An invalid form control with name='node[q_a][field_related_information][0][subform][field_link][0][title]' is not focusable.
```

I'm not sure if this is an interaction with our custom js or another one with the vbe/paragraphs interface. More broadly, I don't know that this is worth spending effort on - I just checked staging and see the same error. @oksana-c what is the main use-case for this page, do we need to be trying to chase down every possibility?

ndouglas:
That reminds me of the optional/hidden field issues that Kaise had to suffer through at some point, though that's just my 🕷️ sense tingling and not actually based on any rational thought on my part.  I'll poke around with the same field and see if I can figure out a root cause -- seems like it's worth opening a followup issue for if nothing else.

acrollet:
That makes sense to me, this PR seems good and I don't want to hold it up, but I'd also like to test a real-world use case if we know one.

ndouglas:
Also see this when modifying Q&A (but not the other types I tried 😩 ) `Notice: Undefined index: base_form_id in va_gov_backend_field_widget_moderation_state_default_form_alter() (line 562 of modules/custom/va_gov_backend/va_gov_backend.module).`

I think the PR I was thinking of was [this one](https://github.com/department-of-veterans-affairs/va.gov-cms/commit/7e78cdaccfcca71aeb4890c8da12083d7f8984c8) which definitely seems related.

```php

  // Add states + custom validation if the Standalone Page field is present.
  if (isset($form['field_standalone_page'])) {
    $form['field_primary_category']['widget']['#states']['required']['input[id="edit-field-standalone-page-value"]'] = ['checked' => TRUE];
    $form['field_tags']['widget'][0]['subform']['field_topics']['#states']['required']['input[id="edit-field-standalone-page-value"]'] = ['checked' => TRUE];

    // Change field uri widget to optional.
    $form['field_related_information']['widget'][0]['subform']['field_link']['widget'][0]['uri']['#required'] = FALSE;
```

So yeah, I guess it should be set optional in this context too.

acrollet:
FWIW I'm pretty sure I have that warning fixed in another PR

ndouglas:
That doesn't surprise me 😂

acrollet:
yeah, here: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/4770/commits/54189c93eb4e4a9b6e7e80f2d640166674e222fd - it was reverted yesterday tho because of the periodic tasks issue, looking at that now...

acrollet:
I see another js error when attempting to edit the event description field on event nodes:

```
An invalid form control with name='node[event][field_address][0][address][address_line1]' is not focusable.
An invalid form control with name='node[event][field_address][0][address][locality]' is not focusable.
An invalid form control with name='node[event][field_address][0][address][administrative_area]' is not focusable.
```

However this error also exists on staging, so it does not represent a regression and I don't think we should hold this PR on it. I added this case to #4796 

I was able to successfully modify field values on benefits detail pages, so the core functionality of this PR works.

acrollet (approved):
Approved pending tests, I think the current failure is a flicker.

PR #4788, opened 2021-03-24
[DO NOT MERGE] Redis POC

indytechcook:
This change is [Reviewable](https://reviewable.io/reviews/department-of-veterans-affairs/va.gov-cms/4788)


indytechcook:
Deleting since we are going to go with memecache for now.

Issue #4513, opened 2021-03-02
Make VHA service description plain text


kevwalsh:
### User Story or Problem Statement

VHA service taxonomy has a rich text Description field driven by the core taxonomy module. It should be plain text, to prevent editors from adding markup. 

Some background on a [drupal core issue](https://www.drupal.org/project/drupal/issues/569434) related to removing this field altogether

FE issue may be required when/if this changes to plain text.

### Affected users and stakeholders

* Editors of VHA services
* FE team 

## Possible implementations 

1. Change existing field
2. Migrate to new field and deprecate old one

Consideration
- an approach that may be useful for other vocabularies (do others face a similar problem with core taxonomy desription field?)

PR #4801, opened 2021-03-24
VACMS-4513: Disallow wysiwyg on core tax description field.

ethanteague:
## Description
See #4513

## Testing done
Visual / behat

## Screenshots
image

## QA steps

As an administrator, go to `taxonomy/term/94/edit` and visually verify that wysiwyg is no longer attached to description field.

### Definition of Done

### Select Team for PR review

- [ ] `Core Application Team`
- [x] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


kevwalsh (commented):
I'm wondering 

1. What may need to happen on vets-website before this gets merged (if anything)
2. What happens to the existing html tags in this field (like in that screenshot). Do we need a script to rip them out?

ethanteague:
> I'm wondering
> 
> 1. What may need to happen on vets-website before this gets merged (if anything)
> 2. What happens to the existing html tags in this field (like in that screenshot). Do we need a script to rip them out?

@kevwalsh 
1) The structure of the field is not changing (at least not the way it is fetched via graphql), so nothing to change on the frontend.
2) I don't think we need to strip out the p tags (it won't cause a ruckus in the build). I think we can let the user decide if they want to remove them. That said, it wouldn't be a big deal to do a strip_tags() on the field (just don't think it's necessary).

kevwalsh:
@ethanteague Will editors need to add `<p>` tags to get line breaks?

ethanteague:
> @ethanteague Will editors need to add `<p>` tags to get line breaks?

Nope, @kevwalsh

kevwalsh:
Per zoom: I think we want to consider stripping html from these (all `<p>` tags i believe) to reduce confusion for editors. I also want to manually QA that text breaks will make their way all to the FE. 

@swirtSJW

swirtSJW:
@kevwalsh @ethanteague    Yah I think, as it is, we are creating the same kind of situation that we had with the old description field that allowed html as plain text.   IF we are opposed to html then we should run it through the plain text filter.   Otherwise people could put html in that is not being cleaned up by the wysiwYg and create bigger issues.

swirtSJW (approved):

PR #4875, opened 2021-03-30
VACMS 4871: Switch from Yaml-Tasks to Task [DO NOT MERGE]

ndouglas:
## Description

See #4871.  At present this is just a PoC to see what breaks and where, and if there are any deficits in Task relative to Yaml-Tasks that might make a different course of action necessary.

## Testing done
Ran some Lando commands manually, saw expected things.

## QA steps

As a developer:
1. Validate Acceptance Criteria from issue #4871:
    - [ ] Marks commit statuses as pending
    - [ ] Marks corresponding commit status as fail/pass when test suite is finished

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

### Is this PR blocked by another PR?

- [ ] `DO NOT MERGE`

### Does this PR need review from a Product Owner

- [ ] `Needs PO review`

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change? 
- [ ] Yes, and it's written in issue ____ and queued for publication. 
  - [ ] Merge and ping @ rachel-kauff so she's ready to publish after deployment
- [ ] Yes, but it hasn't yet been written 
  - [ ] Don't merge yet -- ping @ rachel-kauff to prompt her to write and queue content
- [x] No announcement is needed for this code change. 
  - [x] Merge & carry on unburdened by announcements

Issue #4871, opened 2021-03-30
Yaml Tasks conflicts with Symfony/Process >= 4.0, preventing D9 upgrade.


ndouglas:
## Description
[Yaml Tasks](https://packagist.org/packages/devshop/yaml-tasks) depends heavily on [Power-Process](https://packagist.org/packages/devshop/power-process), which builds on Symfony Process.  However:

>The Symfony Process component has different typehints in the parameters of several APIs in Symfony 4 than Symfony 3 does. This is due to the fact that Symfony 4 requires PHP 7.1.3 or later, which supports a number of typehints not permitted in earlier PHP versions. This difference is minor for most clients, but presents problems for code that subclasses Process, as it is not possible to be compatible with both the Symofny 3 and the Symfony 4 typehints at the same time.

[`drupal/core`](https://packagist.org/packages/drupal/core) @9.anything requires Symfony Process >= 4.4.

A few possible solutions:

- Submit patches upstream for `devshop/power-process` to be compatible with the Symfony 4 typehints -- seems straightforward
- Replace Yaml Tasks
- Something else that I'm missing.

## Acceptance Criteria
- [ ] Yaml Tasks no longer blocks the Drupal 9 upgrade.

acrollet:
related: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/4109#issuecomment-788154979

ElijahLynn:
Summary from 16th minute:

* Composer-patch seems more difficult because it is a dependency resolution issue and the patches apply after dependencies have been resolved and installed, so it won't work. 
* ACs for replacing yaml-tasks with Task are:
  * Marks commit statuses as pending
  * Marks corresponding commit status as fail/pass when test suite is finished
* Benefits of moving forward towards Task are:
  * Significantly lower feedback loops for CI tests by parallizing some of the test suites
  * CD to DEV/STAGING can be significantly faster with parallelization
  * Moving to a tool with greater community support
  * Untying toolchain dependencies from application dependencies
* Moving to ReviewDog to get back commit-level test output GitHub comments (we broke this with the yaml-tests > yaml-tasks update and really need to get this back as it lowered the feedback loop (when it worked, which it didn't always for Behat because the output was too long and didn't fit inside the comment length limit, so we have room to improve that use case too). 

Unanimously decided to move forward with a PoC of Task with the intent of adding ReviewDog soon after, so that Drupal 9 can be unblocked.

PR #4880, opened 2021-03-31
VACMS-4871: Switch from Yaml Tasks to Task

ndouglas:
## Description

See #4871.  This replaces the original PR #4875 

## QA steps

As a developer
1. Validate Acceptance Criteria from issue #4871.
    - [x] Lando continues to function normally.
      - [x] `lando rebuild` installs [Task](https://github.com/go-task/task)
      - [x] `lando test` executes tests
      - [x] `lando phpunit` executes PHPUnit tests
      - [x] `lando task` executes Task and passes arguments appropriately
    - [ ] on Tugboat:
      - [ ]  Installs:
        - [ ] [`task`](https://github.com/go-task/task)
        - [ ] [`github-commenter`](https://github.com/cloudposse/github-commenter)
        - [ ] [`github-status-updater`](https://github.com/cloudposse/github-status-updater)
      - [ ] `task --taskfile=tests.yml tugboat-tests`
        - [ ] Marks GitHub checks as pending:
          - [ ] `va/tests/accessibility`
          - [ ] `va/tests/behat`
          - [ ] `va/tests/behavioral`
          - [ ] `va/tests/check-cer`
          - [ ] `va/tests/phplint`
          - [ ] `va/tests/phpunit`
          - [ ] `va/tests/revision-log`
          - [ ] `va/tests/status-error`
          - [ ] `va/web/build`
        - [ ] Runs `va/web/build`
        - [ ] Marks `va/web/build` as succeeded after completion
        - [ ] Runs `va/tests`
        - [ ] Marks tests as succeeded after completion:
          - [ ] `va/tests/accessibility`
          - [ ] `va/tests/behat`
          - [ ] `va/tests/behavioral`
          - [ ] `va/tests/check-cer`
          - [ ] `va/tests/phplint`
          - [ ] `va/tests/phpunit`
          - [ ] `va/tests/revision-log`
          - [ ] `va/tests/status-error`
        - [ ] Relevant context (i.e. not the entire output) is added as a comment on the PR for:
          - ~`va/tests/accessibility`~ will open up a followup issue, nontrivial
          - ~`va/tests/behat`~ will open up a followup issue, nontrivial
          - ~`va/tests/behavioral`~ will open up a followup issue, nontrivial
          - [ ] `va/tests/phplint`

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [x] `DO NOT MERGE`

- [ ] `Needs PO review`


ndouglas:
Today: 
- I followed Elijah's excellent instructions on how to build a preview off a branch and use that preview as a base preview for PRs based on that original branch.  I'm going to write up the instructions and shove 'em into a README so they're accessible, but it goes like this:
  1. Push the base branch (e.g. `VACMS-4871-Switch-from-Yaml-Tasks-to-Task` _`upstream`_, not just to `origin`).  This makes it usable as a Base Preview.
  2. In Tugboat, scroll to Branches and click "Build Preview."  This will create a new preview in the Base Previews section.  If you have also pushed to `origin`, **ignore that preview**.  It will only ~break your heart~ waste your time.
  3. Scroll down to the newly-created Base Preview and view it.  
  4. Click "Settings" and select "Use this Preview as a Base Preview", then select "Branch Base Preview."
  5. On your local machine, create a new branch based on the original branch, make changes, push them to `origin`, and when you make the PR, ensure that you are comparing against the base branch you created and not against `main` or `master`.

- I learned useful things about **Task**:
  1. It uses an internal shell written in `go` rather than a local shell so that it will work on Windows.  The tradeoff is that certain shell builtins and constructs won't work correctly, and there seems to be some weirdness with environment variables and variable expansion that's probably just a result of a small project interfacing with decades of accumulated UNIX nightmares.
  2. The YAML used by Task is not 1:1 with that we used with Yaml Tasks, but it's really close.  There's just a little bit of tweaking involved in most cases, but edge cases and obscure uses might cause some issues.

- I learned some useful things about **Tugboat**:
  1.  Apparently this is cached on some level, and it took me way too much time and stress to figure it out:
Screen Shot 2021-03-31 at 4 36 23 PM

- I learned some useful things about `Reviewdog`:
  1. It's proving difficult to get a "Hello, World" sort of thing going.  I'd hoped that I could basically just set up the environment variables and do `echo "test" | reviewdog` or something and get a comment posted in the thread, but that's not the case so far.  It might just be because I'm not posting files and line numbers relevant to the code changes in my test messages.
  2. There were a number of possible ways of implementing the GH integration, and I think only one ([this one](https://github.com/reviewdog/reviewdog#reporter-github-pullrequest-review-comment--reportergithub-pr-review)) is appropriate on a structural level.  Unfortunately, I don't think it's appropriate for all of our cases -- for instance, PHPUnit test failures should be PR comments and not PR review comments attached to specific changes, but I think we need to actually use the ReviewDog app integration for that, and that means using some guy's server as far as I can tell.  So this might be a partial solution at best.

So, all in all it was... an _educational_ day.

acrollet:
possible alternative to reviewdog: https://github.com/cloudposse/github-commenter

acrollet:
> So, all in all it was... an _educational_ day.

ndouglas:
yeah, `github-commenter` seems like a good fit for those more commit-wide or PR-wide comments we wanna do.  Good find!

ndouglas:
At my last job I made a Drupal module that checks for module updates and then makes GH issues (like [this](https://github.com/ainonline/test/issues/371)), so I'm reasonably familiar with the GH API and half of me wants to just do a custom solution, but the other half of me is sane and just wants this to be over with so I can get back to the sprint goals at hand.

Inside of me there are two wolves, etc 😩 

images

ndouglas:
@acrollet @ElijahLynn @oksana-c @indytechcook (Hi, Neil!):

I'm over the `5` estimate that I'd assumed this task would take, so here's an update: 

`reviewdog` is failing to do anything, and I'm not sure why -- fails silently.  🦆 'ed it with Adrian for about an hour just now and we've run several dozen variations of the command in question, used different GH tokens, etc, and it's apparently a no-go.  It's sending requests, but it seems like GH might be silently eating them.  Happy to investigate this further, but I feel like we've been pretty exhaustive.

Adrian identified two other tools, [github-commenter](https://github.com/cloudposse/github-commenter) and [github-status-updater](https://github.com/cloudposse/github-status-updater), which combined match the AC's of #4871 (comments and checks respectively).  `github-commenter` works with my personal GH API token (see test messages in #4881), and I have a script committed to install it, etc.  I haven't tried `github-status-updater`.

So my plan, unless there are objections, is to abandon `reviewdog` and shift to `github-commenter` and `github-status-updater`.

Another option is to fork Yaml-Tasks and Power-Process into this repository and patch them for compatibility with Drupal 9 and kick this can down the road a bit.  That seems faster, but... it's just kicking the can down the road.

Happy to rap about this further at any time, of course.  Except right now 😁

ElijahLynn:
> So my plan, unless there are objections, is to abandon `reviewdog` and shift to `github-commenter` and `github-status-updater`.

I think it makes sense to continue to explore this with the momentum that you have! A partially parallelized test suite solution here will have a huge ripple effect on our velocity, the value is very high.

.reviewdog.yml
```diff
@@ -0,0 +1,11 @@
+runner:
```
indytechcook, line 1: We are running phpcs and reviewdog with github actions.  Does this replace that? Or does this configure reviewdog which runs on Github Actions?

tests.yml
```diff
@@ -1,44 +1,204 @@
...
+      - |
+        {
+          ./tests/scripts/check-cer.sh && \
+            github-status-updater -action=update_state -state=success -context='va/tests/check-cer' -description='Success'
+        } || github-status-updater -action=update_state -state=failure -context='va/tests/check-cer' -description='Failed'
 
-va/tests/revision-log:
-  description: Ensure revision log field is present
-  command: ./tests/scripts/check-revision-logs.sh
+  va/tests/phpcs:
```
indytechcook, line 47: Do we need this since we are running this code view Github Actions?

indytechcook:
I think we should keep static code analysis 100% in Github Actions.  Thoughts?

tests.yml
ndouglas: Nah, it'd be redundant, I'll strip it out.

ndouglas:
I haven't worked with Actions before so I'm unfamiliar with the limitations... but if we can do it in Actions, I think that sounds _much_ better than Tugboat.

indytechcook:
Here is phpcs: https://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/.github/workflows/phpcs.yml

Here is what runs reviewdog (currently poorly named): https://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/.github/workflows/eslint.yml

ndouglas:
I feel like the main benefit of Reviewdog is its ability to do the line-level PR review comment/suggestions.  That should/can be doable with Actions, since there's no DB/file sync/LAMP infrastructure involved.  So really, there's no need to install Reviewdog in Tugboat.

What _would_ be useful would be to catch the Accessibility/Behat/PHPUnit/Build errors and report those as comments, right?

So maybe I was barking up the wrong tree thinking about/focusing on Reviewdog in the first place, and should just cut down these ACs or split those out for doing in Actions?

indytechcook:
> So maybe I was barking up the wrong tree thinking about/focusing on Reviewdog in the first place, and should just cut down these ACs or split those out for doing in Actions?

Keep the static code analysis in Github Actions.  We are already using reviewdog for one of them so maybe update all of them to use it?  I'll leave it to you.

> What would be useful would be to catch the Accessibility/Behat/PHPUnit/Build errors and report those as comments, right?

The ability to provide comments on commits from ansible is awesome and would be very beneficial.  

Also It what's the different between `github-status-updater` and reviewdog?

ndouglas:
For anyone interested, I'll open this for Review tomorrow but wanted to solicit any feedback on the ACs I've written above, any extra comments/suggestions/complaints/grievances, etc.

Stuff as Currently Implemented _and_ Tested:

- [ ] Lando continues to function normally.
  - [ ] `lando rebuild` installs Task
  - [ ] `lando test` executes tests
  - [x] `lando phpunit` executes PHPUnit tests
  - [x] `lando task` executes Task and passes arguments appropriately
- [ ] on Tugboat:
  - [x]  Installs:
    - [x] [`task`](https://github.com/go-task/task)
    - [x] [`github-commenter`](https://github.com/cloudposse/github-commenter)
    - [x] [`github-status-updater`](https://github.com/cloudposse/github-status-updater)
  - [x] `task --taskfile=tests.yml tugboat-tests`
    - [x] Marks GitHub checks as pending:
      - [x] `va/tests/accessibility`
      - [x] `va/tests/behat`
      - [x] `va/tests/behavioral`
      - [x] `va/tests/check-cer`
      - [x] `va/tests/phplint`
      - [x] `va/tests/phpunit`
      - [x] `va/tests/revision-log`
      - [x] `va/tests/status-error`
      - [x] `va/web/build`
    - [x] Runs `va/web/build`
    - [x] Marks `va/web/build` as succeeded after completion
    - [x] Runs `va/tests`
    - [ ] Marks tests as succeeded after completion:
      - [x] `va/tests/accessibility`
      - [x] `va/tests/behat`
      - [x] `va/tests/behavioral`
      - [x] `va/tests/check-cer`
      - [x] `va/tests/phplint`
      - [ ] `va/tests/phpunit`
      - [ ] `va/tests/revision-log`
      - [ ] `va/tests/status-error`
    - [ ] Relevant context (i.e. not the entire output) is added as a comment on the PR for:
      - [ ] `va/tests/accessibility`
      - [ ] `va/tests/behat`
      - [ ] `va/tests/behavioral`
      - [x] `va/tests/phplint`

ndouglas:
Actually, need to make sure devops is synced up with this first.

.tugboat/.env.j2
```diff
@@ -20,3 +20,11 @@ BEHAT_PARAMS='{"extensions":{"Behat\\MinkExtension":{"base_url":"{{ TUGBOAT_DEFA
 # Manually set column width to work around yaml-tests bug & Tugboat not having an interactive TTY when running tests.
 # TODO: explain bug
 COLUMNS=1000
+
+# GitHub-Commenter and GitHub-Status-Updater environment variables.
+GITHUB_ACTION=update_state
```
ElijahLynn, line 25: Discussed removing on a call.

tests.yml
```diff
@@ -1,44 +1,183 @@
...
+        {
+          ./tests/scripts/build-web.sh && \
+            github-status-updater -action=update_state -state=success -context='{{ .TASK }}' -description='Success'
+        } || github-status-updater -action=update_state -state=failure -context='{{ .TASK }}' -description='Failed'
+
+  set-check-status:
+    cmds:
+      - github-status-updater -action=update_state -state="{{ .STATUS }}" -context="{{ .CONTEXT }}" -description="{{ .DESCRIPTION }}"
+
+  set-all-tests-pending:
```
ElijahLynn, line 115:
```suggestion
  # Any changes to test names or additions or removals must be updated here as well for the Pending status to work. 
  set-all-tests-pending:
```

composer.json
```diff
@@ -138,11 +137,15 @@
...
         "knplabs/github-api": "^1.4",
+        "mglaman/phpstan-drupal": "^0.12.9",
+        "michelf/php-markdown": "^1.8",
         "mikey179/vfsstream": "^1.6",
         "mouf/nodejs-installer": "^1.0",
         "npm-asset/yarn": "1.21.1",
         "oomphinc/composer-installers-extender": "^2.0",
         "php-http/guzzle6-adapter": "^1.1",
+        "phpstan/phpstan": "^0.12.82",
+        "phpstan/phpstan-deprecation-rules": "^0.12.6",
```
ElijahLynn, line 148: We discussed moving these to a DEV dependency.

ElijahLynn (approved):

ElijahLynn:
Nice work Nate (and @acrollet)!!!!

We did discuss leaving DEV as a dry-run invocation in the other PR and not running tests there as we are going to be removing DEV altogether as we have separate cms-test STAGING and PROD to do actual development work on for devops and we have Tugboat for better dev environments for application engineers. 

There are/will be follow up issues for both the commenting part and parallelization.

.tugboat/config.yml
```diff
@@ -111,10 +120,11 @@ services:
       # and update steps, because the results of those are inherited
       # from the base preview.
       build:
-        # Currently resetting to the remote branch/PR SHA since yaml-tasks runs tests status against this commit.
+        # Currently resetting to the remote branch/PR SHA since Task runs tests status against this commit.
```
acrollet, line 123:
```suggestion
        # Currently resetting to the remote branch/PR SHA since [Task](https://github.com/go-task/task) runs tests status against this commit.
```

composer.json
```diff
@@ -10,7 +10,10 @@
         "drupal/console": "^1.9",
         "drupal/console-extend-plugin": "^0.9.5",
         "drupal/media_entity_generic": "^1.0",
-        "marcortola/behat-seo-contexts": "^3.1"
+        "marcortola/behat-seo-contexts": "^3.1",
+        "mglaman/phpstan-drupal": "^0.12.9",
+        "phpstan/phpstan": "^0.12.83",
+        "phpstan/phpstan-deprecation-rules": "^0.12.6"
```
acrollet, line 16: do we want to bring these deps in as part of this PR?

composer.json
```diff
@@ -137,6 +139,7 @@
         "easyrdf/easyrdf": "1.1.1 as 0.9.1",
         "http-interop/http-factory-guzzle": "^1.0",
         "knplabs/github-api": "^1.4",
+        "michelf/php-markdown": "^1.8",
```
acrollet, line 142: is this a dep of phpstan?

acrollet (dismissed):
Generally looks great, a very small suggestion and a couple questions

composer.json
ndouglas: wouldn't think so... I'll poke around.

acrollet (approved):
Approved pending confirmation of passing tests on tugboat from @ndouglas

acrollet (dismissed):

.lando.yml
```diff
@@ -141,11 +145,12 @@ tooling:
     cmd: "rm /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini && /etc/init.d/apache2 reload"
     user: root
 
-
   phpunit:
     service: appserver
     description: "Run VA PHPUnit tests found in va/tests/phpunit: lando phpunit. Syntax: lando phpunit"
-    cmd: "./bin/yaml-tasks --tasks-file=tests.yml va/tests/phpunit"
+    cmd: "./bin/task --taskfile=tests.yml va/tests/phpunit --"
```
indytechcook, line 151: What does the trailing `--` do?  Teach an old man

.tugboat/config.yml
```diff
@@ -139,7 +149,7 @@ services:
       online:
         # We only want tests running on PRs, not branches like the base preview build of master.
         # This sets the commit status checks to pending all at once.
-        - if [ "$TUGBOAT_PREVIEW_TYPE" = "pullrequest" ]; then ./bin/yaml-tasks --tasks-file=tests.yml va/web/build va/tests; fi
+        - if [ "$TUGBOAT_PREVIEW_TYPE" = "pullrequest" ]; then bash -lc 'task --taskfile=tests.yml'; fi
```
indytechcook, line 152: More learning for me... Does `bash -lc` do some extra logging?

.lando.yml
ndouglas: The `--` means that any additional CLI options will get passed to the process being called and not to Task itself.  TL;DR: tells `task` to ignore CLI options after that point.

.tugboat/config.yml
ndouglas: `bash -l` starts a login shell, which means it'll read from `.bashrc` and get the appropriate $PATH that we populated above.

tests.yml
```diff
@@ -1,44 +1,227 @@
...
-  description: Build VA.gov Front-end
-  command: ./tests/scripts/build-web.sh
+  va/tests/behat:
+    desc: Behat Tests
+    cmds:
+      - |
+        cd tests/behat
+        result="$(behat --colors)"
+        exit_code=$?
+        if [ "$SKIP_REPORTING" -ne 1 ]; then
```
indytechcook, line 33: I like the use of `$SKIP_REPORTING` throughout.  Much better the hard coding something.

.lando.yml
ndouglas: This actually isn't terribly useful anymore, since those passed arguments aren't getting referenced anywhere.  Back when the command being executed was `phpunit tests/phpunit`, it made sense.  Now that it's like 12 lines of shell hell, it makes less sense 😩

tests.yml
```diff
@@ -1,44 +1,227 @@
...
+          else
+            github-status-updater -action=update_state -state=failure -context='{{ .TASK }}' -description='Failed'
+            [ -z "${GITHUB_COMMENT_TYPE}" ] && github-commenter -delete-comment-regex="{{ .TASK }}" -comment="${result}"
+          fi
+        fi
 
-va/tests/check-cer:
-  description: Ensure cer fields exist
-  command: ./tests/scripts/check-cer.sh
+  va/tests/phplint:
```
indytechcook, line 69: I'd really like linting not to run in jenkins, we have https://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/.github/workflows/phpcs.yml.  Help me understand what this is doing.

ndouglas: `phplint` is running the built-in PHP linter (just `php -l`) on all our custom code.  It's not terribly slow.  I totally wanna move it to Actions too, but was trying to keep feature parity in this PR 😅

ndouglas: Re-read the code in question and realized that you were probably talking about everything _around_ `php -l` and not `php -l` itself.

It's catting a heredoc into bash because Task doesn't use a real shell like bash or even sh to execute commands -- it uses a built-in shell.  So some builtins and bash syntax won't work, and `trap` won't work -- which I use to simplify the logic for grabbing the output from the `find... php -l` command.

So it finds the .whatever files and passes them to `php -l`.  If it returns 0, then all the files are fine and we can move on with our lives.  If it returns non-0, then we append the `php -l` output to `$result` because it's relevant and we need to post it as a GH comment explaining why this test failed.

acrollet: > It's catting a heredoc into bash because Task doesn't use a real shell like bash or even sh to execute commands -- it uses a built-in shell. So some builtins and bash syntax won't work, and `trap` won't work -- which I use to simplify the logic for grabbing the output from the `find... php -l` command.

Use go-task, he said. It'll be easy, he said.

tests.yml
```diff
@@ -1,44 +1,227 @@
...
+            github-status-updater -action=update_state -state=success -context='{{ .TASK }}' -description='Success'
+          else
+            github-status-updater -action=update_state -state=failure -context='{{ .TASK }}' -description='Failed'
+          fi
+        fi
 
-va/tests/revision-log:
-  description: Ensure revision log field is present
-  command: ./tests/scripts/check-revision-logs.sh
+  va/tests/check-cer:
```
indytechcook, line 55: This one was a bit confusing to me so I looked at the script.  It's static code analysis (so we should some day move to github actions but NOT NOW).  Also, if we want to validate the fields exist should we be checking in Drupal instead of a file?  This is not to be addressed now, just something for use to follow up on.

tests.yml
ndouglas: "Nate's Task PR contains an ad hoc, informally-specified, bug-ridden, slow implementation of half of Yaml-Tasks."
- please don't say this

acrollet: /me won't

indytechcook: All good, let's keep it!

indytechcook (approved):

tests.yml
ndouglas: I prefer checking YAML to checking inside of a running Drupal system, personally.  I think that's more extensible, using `yq`, etc, than writing e.g. PHPUnit ExistingWebBase tests or something like that and having to write separate stuff for checking for the existence of certain fields, forms, content types, etc.  (Thinking about the intersection between this and Drupal-Spec-Tool)  And it'd be waaaaaaaaaaay faster and can run in parallel rather than adding additional functional tests that might have side effects.

indytechcook: Yes, I forgot about the spec tool.  

> I prefer checking YAML to checking inside of a running Drupal system, personally.

This assume the files might change.  Is that correct?

> run in parallel

Github actions run in parallel :). Though I assume you were talking about PHPUnit.

ndouglas: I assume anything could change.  Actually, I'm a little surprised that the Spec Tool doesn't cover this anyway 🤔  We already have a spike for the Spec Tool, so I can add this as a comment and either research and figure out why this exists separately or we can rap about it during the spike.

acrollet: I believe it's because the revision log isn't a regular DB field and doesn't get surfaced to the spec tool.

acrollet: Also to clarify, the script ensures that the revision log field is present in the right region of the form. (At the bottom of the main form rather than the default location in the sidebar)

ndouglas: wait, were we talking about `check-cer` or `revision-log`?

acrollet: whoops, good question!

ndouglas:
Forcefully merged because I had to resolve a composer.lock conflict 😩

Issue #4311, opened 2021-02-15
SPIKE: technical implementation of VAMC system menu IA governance


kevwalsh:
See #4144 for design/governance doc, and see [spreadsheet](https://docs.google.com/spreadsheets/d/1n6h0NUJmQcFsc90dy14FQrX_0wVPX0AM7NlZoHSFKaA/edit#gid=1058557895) with example paths, especially columns C,D and S and T. 

## User Story or Problem Statement

As a product owner (content_admin, or possibly an administrator only, TBD) 
* i can designate which parent items in VAMC system menus are eligible to have children menu items created by non-Content admins, 
* i can designate which paths should have "locked" menu items, meaning that non-content-admins would not be able to edit those menu items, across all VAMCs (such as pittsburgh-health-care/pharmacy AND erie-health-care/pharmacy, and all [health-care-system]/pharmacy)
* i can additionally designate individual URLs at certain VAMCs that should be locked (such as pittsburgh-health-care/dav-vans BUT NOT erie-health-care/dav-vans). 

As a content_editor, content_reviewer, or content_publisher with workbench access to a VAMC menu (eg [Lebanon Health Care menu](prod.cms.va.gov/admin/structure/menu/manage/va-lebanon)) : 

* i shouldn't be able to administer the menu itself (eg go to `admin/structure/menu/manage/va-lebanon` and drag and drop)
* i shouldn't be able to add or edit menu items for content types other than Detail Pages
* i should be able to place individual menu items of Detail Page content type, using the Menu settings in the right rail, but only with certain parents eligible. 
* i should be able to see the whole menu, with some items disabled (those that are not eligible, [see example screenshot](https://app.zenhub.com/workspaces/vagov-cms-team-5c0e7b864b5806bc2bfc2087/issues/department-of-veterans-affairs/va.gov-cms/4144#issuecomment-771067856)

The above behavior should only target VAMC menus (typically, those that have `-health-care` in their machine name, but some of the VISN 4 menus are missing that`) 

## Acceptance Criteria
- [x] We have identified a technical approach that will accomplish the above user stories for ~144 VAMC system menus. 

## Implementation steps
- [ ] TBD

shiragoodman:
@kevwalsh would it be okay to add my teams label to this so we can track the support Mikki is providing on our board?

kevwalsh:
> @kevwalsh would it be okay to add my teams label to this so we can track the support Mikki is providing on our board?

[...]

Issue #4847, opened 2021-03-29
Menu item should not have a default parent for any node type. 


kevwalsh:
**Describe the defect**

**To Reproduce**

As Victor.A.MCtest

1. Go to /node/add/event and enable menu item
2. note that the parent is set to Altoona

**Expected behavior**
There should be no default parent.

Also, grep the config/sync folder for `parent: '`, you'll find this is set on on a number of `node.type.*.yml`

**ACs**
- [ ] Menu items for all content types should always be specified manually.

PR #4894, opened 2021-04-01
VAGOV-4988: Menu governance for VAMC product.

ethanteague:
## Description
See #4311 and #4988

Due to not being able to roll out all of this  4311 and 4988 are not fully implemented.  They are intentionally short circuited.

However this is addressing :

- Remove ability for content_publishers to reorder their VAMC menus (only allow for content_admins) https://github.com/department-of-veterans-affairs/va.gov-cms/issues/5397
- Adding a menu item forces you to choose a parent.  There is no longer a default to the VAMC system root.

## Testing done
Visual / behat / phpunit

### Content Admin
**As a Lisa T (content_admin):**
- [x]  Go to `https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov/admin/config/content/va_gov_menu_access` and validate that you can access the page and see data in the two fields.
- [x] Go to https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov/admin/structure/menu/manage/pittsburgh-health-care and validate that you can see the menu items and move them around.
- Go to https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov/node/add/health_care_region_detail_page  Expand the "MENU Settings" widget and check , "provide a menu link"
   - [x] validate that `Parent item` has nothing preset and only says "- Select a value -"
image
- [x] Go to https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov/node/775/edit (an existing detail page) and validate that it is still assigned a parent of "---- About VA Pittsburgh" 
 image
- [x] Validate that when you click on the "Parent item" select list, you have access to all parents within  the VAMC universe.

### Content Editor
**As Ryan S (content _editor):**
- [x]  go to `https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov/admin/config/content/va_gov_menu_access` and validate you get access denied.
- [x] go to https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov/admin/structure/menu/manage/pittsburgh-health-care and validate you get access denied.
- go to https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov/node/add/health_care_region_detail_page  Expand the "MENU Settings" widget and check , "provide a menu link"
   - [x] validate that `Parent item` has nothing preset and only says "- Select a value -"
- [x] Go to https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov/node/775/edit (an existing detail page) and validate that it is still assigned a parent of "---- About VA Pittsburgh" 
 image
- [x] Validate that when you click on the "Parent item" select list, you have access to all parents within  the VAMC universe.

### Definition of Done

### Select Team for PR review

- [ ] `Core Application Team`
- [x] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


docroot/modules/custom/va_gov_menu_access/js/va_gov_menu_access.es6.js
```diff
@@ -0,0 +1,41 @@
...
+          "#edit-menu-menu-parent option"
+        );
+        if (
+          document.getElementById("edit-menu-title").hasAttribute("disabled")
+        ) {
+          document.getElementById("edit-menu-enabled").disabled = true;
+        }
+        parentOptions.forEach((opt) => {
+          // If an option doesn't pass the check, set it to disabled.
+          if (opt && opt.text.includes(" | Disabled")) {
```
kevwalsh, line 30: Looks like this can be customized, which we may want to do.

kevwalsh: Nice! 

Create_Detail_Page___VA_gov_CMS_and_iMac

kevwalsh: Actually, looks like the menu hierarchy is totally scattered there ^^^ . Items have retained the correct depth, but not the correct parents or weight.

kevwalsh: Related possibly to the order:

As vanessa.luxen@va.gov, i went to create an event. I checked "Provide a menu link"

The default menu item is `------ Health services | Disabled`, which should not be eligible at all (but appears enabled in my browser, see screenshot).  

Create_Event___VA_gov_CMS

When I saved the event, no menu item was apparently created (form validated fine), and i went back to edit, and saw this, which seems unusual.  

Edit_Event_asdf___VA_gov_CMS

kevwalsh (changes requested):
This looks like it's on the right track but there are a couple bugs and need some additional details. Possibly some UX writing attention. 

One more comment: i don't see anything about making specific rules for the `health_care_region_detail_page` ... one of the ACs is 

> i shouldn't be able to add or edit menu items for content types other than Detail Pages

As noted, as vanessa.luxen@va.gov, i was able to (sort of?) create a menu item for an event.

docroot/modules/custom/va_gov_menu_access/src/Form/VaGovMenuAccessConfigForm.php
```diff
@@ -0,0 +1,119 @@
...
+   *   Default form array structure.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   Object containing current form state.
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    $config = $this->config('va_gov_menu_access.settings');
+
+    // Menus.
+    $selected_menus = [];
+    $menus = !empty($form_state->getValue('menus')) ? $form_state->getValue('menus') : '';
```
swirtSJW, line 91:
```suggestion
    $menus = !empty($form_state->getValue('menus')) ? $form_state->getValue('menus') : [];
```

docroot/modules/custom/va_gov_menu_access/src/Form/VaGovMenuAccessConfigForm.php
```diff
@@ -0,0 +1,119 @@
...
+    if ($menus) {
```
swirtSJW, line 92:
```suggestion

```

docroot/modules/custom/va_gov_menu_access/src/Form/VaGovMenuAccessConfigForm.php
```diff
@@ -0,0 +1,119 @@
...
+      foreach ($form_state->getValue('menus') as $menu_key => $menu_val) {
```
swirtSJW, line 93:
```suggestion
      foreach ($menus as $menu_key => $menu_val) {
```

docroot/modules/custom/va_gov_menu_access/src/Form/VaGovMenuAccessConfigForm.php
```diff
@@ -0,0 +1,119 @@
...
+        if (!is_numeric($menu_val)) {
+          $selected_menus[$menu_key] = $menu_val;
```
swirtSJW, line 95: Could use a comment here to describe why we want to collect menus that are not numeric.

docroot/modules/custom/va_gov_menu_access/src/Routing/AdminRouteSubscriber.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Drupal\va_gov_menu_access\Routing;
+
+use Drupal\Core\Routing\RouteSubscriberBase;
+use Symfony\Component\Routing\RouteCollection;
+
+/**
+ * Limits access to admin paths by Extending RouteSubscriberbase.
```
swirtSJW, line 9: This seems to generic to be helpful.

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
```diff
@@ -0,0 +1,175 @@
...
+function _va_gov_menu_access_get_menu_access() {
+  // Get our menu form settings.
+  $menu_access_config = \Drupal::config('va_gov_menu_access.settings');
+
+  // Grab the allowed parents config.
+  $allowed_parents = $menu_access_config->get('va_gov_menu_access.paths');
+  $allowed_parents_array = explode("\n", $allowed_parents);
+  $universal_parents = [];
+  $single_parents = [];
+  // Sort parents into universal and single allowance arrays.
```
swirtSJW, line 37: Could use examples here of what makes something universal or single

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
```diff
@@ -0,0 +1,175 @@
...
+  $locked_paths_array = explode("\n", $locked_paths);
+  $universal_locked_paths = [];
+  $single_locked_paths = [];
+  // Sort locked paths into universal and single allowance arrays.
+  foreach ($locked_paths_array as $locked_path) {
+    if (strpos($locked_path, '*') !== FALSE) {
+      $universal_locked_raw = explode('*/', $locked_path);
+      $universal_locked_paths[] = trim(end($universal_locked_raw));
+    }
+
```
swirtSJW, line 59:
```suggestion
```

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
```diff
@@ -0,0 +1,175 @@
...
+        $form['menu']['link']['link_enabled']['#attributes']['disabled'] = TRUE;
+        $form['menu']['link']['weight']['#attributes']['disabled'] = TRUE;
+        return $form;
+      }
+    }
+
+    $universal_parents = $paths_config['universal_parents'];
+    $single_parents = $paths_config['single_parents'];
+
+    if (!empty($form['menu']['link']['menu_parent']['#options'])) {
```
swirtSJW, line 123: Could use a comment to explain what this is looking for or what it indicates to you if there are #options present.

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
```diff
@@ -0,0 +1,175 @@
...
+      $uuids_with_vals = [];
+      $uuids_only = [];
+
+      foreach ($parent_options as $key => $option) {
+        if (!empty(explode('menu_link_content:', $key))) {
+          // This is how we get our menu item uuid.
+          $menu_id_array = explode('menu_link_content:', $key);
+          $menu_id = !empty($menu_id_array[1]) ? $menu_id_array[1] : '';
+          // Load them into an array we can use for cherrypicking by keys.
+          $uuids_with_vals[$menu_id] = [
```
swirtSJW, line 139: IN line 137 you are giving the option for $menu_id to be an empty string.  Here you are creating an array with a key of an empty string.   You are going to get odd results since you are putting in numeric keys and I think empty strings are also going to generate numeric keys. and there is no allowance for when those numeric keys collide or get re-keyed.

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
```diff
@@ -0,0 +1,175 @@
...
+            'menu_key' => $key,
+            'option' => $option,
+          ];
+          // Flat array for multi menu loading.
+          $uuids_only[] = $menu_id;
```
swirtSJW, line 144: Here you could also end up with an empty value.

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
```diff
@@ -0,0 +1,175 @@
...
+      }
+
+      // Load up all our menu items at once.
+      $loaded_menu_items = \Drupal::entityTypeManager()->getStorage('menu_link_content')->loadByProperties(['uuid' => $uuids_only]);
+
+      // Loop through them, and decide whether it should be allowed in options.
+      foreach ($loaded_menu_items as $menu_item) {
+
+        // Get our alias, as well as last arg of alias for comparison.
+        $uri_parts = explode('node/', $menu_item->get('link')->uri);
```
swirtSJW, line 155: I am a little uneasy about this as there are 41 other words that end in node.  https://lotsofwords.com/*node that would give you unintended results if exploded like this.  It would be good to check that node/  appears as the first  5 chars first.

swirtSJW (commented):
Some thoughts

docroot/modules/custom/va_gov_menu_access/src/Routing/AdminRouteSubscriber.php
ethanteague: Lol - yeah.....I got pretty lazy on that one, @swirtSJW. I'll make this more meaningful.

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
ethanteague: Great catch / great tool, @swirtSJW ! Point taken - will tighten this up :+1:

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
ethanteague: ugh. Yep, @swirtSJW - :+1:

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
ethanteague: Yep, @swirtSJW - I'll fix this up :+1:

docroot/modules/custom/va_gov_menu_access/src/Form/VaGovMenuAccessConfigForm.php
ethanteague: Ugh - my bad, @swirtSJW. Will fix :+1:

kevwalsh:
@ethanteague i'm making this a draft because we don't want to merge this until we go through some design review.

docroot/modules/custom/va_gov_menu_access/js/va_gov_menu_access.es6.js
kevwalsh: > Looks like this can be customized, which we may want to do.

I had a look at this with @rachel-kauff. Decision was to just 

- [ ] remove the ` | Disabled` string altogether. 

Separately, we talked about removing the _other_ ` (disabled)` string that is appended to menu items that have menu items that don't appear in the left nav.  Can you look at the lift in removing this to? From an editor perspective, this isn't useful information because they can't control these things anyway. 

- [ ] Remove  ` (disabled)` string appended on menu items.

ethanteague: I'll just target the `Disabled` for removal in the js script that adds the disabled attribute, @kevwalsh - easy to do.

kevwalsh:
Moving this to #4988, for next sprint.

docroot/modules/custom/va_gov_menu_access/js/va_gov_menu_access.es6.js
ethanteague: > Related possibly to the order:
> 
> As [vanessa.luxen@va.gov](mailto:vanessa.luxen@va.gov), i went to create an event. I checked "Provide a menu link"
> 
> The default menu item is `------ Health services | Disabled`, which should not be eligible at all (but appears enabled in my browser, see screenshot).
> 

This will be resolved in https://github.com/department-of-veterans-affairs/va.gov-cms/pull/4911

kevwalsh (changes requested):
A few things. 

### 1. New spec for config

I've added new config spec on #4988, which should be added to our config

i also added some new ACs. 

### 2. Defect for parents

VAMC_Menu_Custom_Access___VA_gov_CMS


As admin, I entered the above config at `/admin/config/content/va_gov_menu_access` and cleared Cache, Then logged in as Vanessa, and the entire menu is disabled. 

- [x] I should be allowed to add the Locations as parents (eg for Facility maps).


Create_Detail_Page___VA_gov_CMS_and__1__AP__Solution_Brainstorm__OPEN__·_Issue__4809_·_department-of-veterans-affairs_va_gov-cms


### 3. Defect with default menu item parent value

1. As vanessa.luxen@va.gov, i created a detail page

2. Checked off "Provide a menu link" 

3. Default value for me is just the VA Butler health care homepage (which doesn't look disabled _until_ you click the dropdown, and i leave as is. (screenshot 1)

4. If i save as is, it saves properly, form validates, but without creating a menu item (node 18371). This is confusing UX. Not sure how to best handle the default value. 

Maybe a "- Select a value -" that doesn't validate. Ideas? 


Create_Detail_Page___VA_gov_CMS


### 4. Content admin followup to the above

5. If i go back to the same node 18371 above , the default value has _changed_ to `<Pittsburgh health care>`, which must be the default menu item spec'd in config for this content type (see #4847, which should be taking care of that). But regardless of 4847, there's some significnat risk that this could happen accidentally later. 

Edit_Detail_Page_Test_invalid_menu_item_at_root_VAMC_level___VA_gov_CMS

- [x] I think this situation can be avoided if we just don't let editors create "empty" menu items (see 4 above)

### 5. Content types with disabled menu items

- [x] Instead of disabling menu items for certain content types, we should hire this entire fieldset if possible. 

If there are issues, let's explore a different UX. 

Create_News_Release___VA_gov_CMS

### 6. Config menu item label

VAMC_Menu_Custom_Access___VA_gov_CMS

- [x] Menu item label should match h1 for the config page


### 7. Config field labels and help text

- [x] Use `URL` in both field labels
- [x] replace the ` | use` with `. Use`

VAMC_Menu_Custom_Access___VA_gov_CMS

### 8. Some new ACs that came up on Friday Apr 9 and added to #4988 

In the UI parent item selector, when creating a `health_care_region_detail_page`, i should be able to see:
- [x]  any menu item that is eligible to have children (and its children, which should also be eligible to have children)
- [x]  the ancestors of any menu item that is eligible to have children (_except the top level_, which is always the Health Care System, and is redundant)

I realize this last one complicates things and we should discuss feasibility / LOE.

ethanteague:
Back to you, @kevwalsh - more than happy to go over this with you via zoom.

kevwalsh:
@ethanteague as vanessa.luxen@va.gov i see no menu options at https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov/node/add/health_care_region_detail_page 

Create_Detail_Page___VA_gov_CMS

kevwalsh (changes requested):
Also some minor label changes to the config page 

`Prevent menu editing of "Detail Page" nodes for non-admin users at these URL paths`

VAMC_Menu_Custom_Access___VA_gov_CMS

ethanteague:
```
5. Content types with disabled menu items
 Instead of disabling menu items for certain content types, we should hire this entire fieldset if possible.
If there are issues, let's explore a different UX.
```
Can we do this in a different issue, @kevwalsh? It's a little more complicated to handle this than I anticipated, and hoping we can consider it as slightly beyond the original scope.

Updated to address other items in https://github.com/department-of-veterans-affairs/va.gov-cms/pull/4894#issuecomment-821163312 & https://github.com/department-of-veterans-affairs/va.gov-cms/pull/4894#pullrequestreview-637685000

kevwalsh:
Yeah we can handle that in a separate issue i suppose. 

I'm having a really hard time QAing this. As admin, i'm making sure that two pages with menu items exist that match the URL at Butler:

https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov//butler-health-care/programs
https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov//butler-health-care/internships-fellowships

I cleared cache. 

But as vanessa.luxen@va.gov i still don't see any menu options when i go to https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov/node/add/health_care_region_detail_page

The field briefly appears then disappears.

kevwalsh:
> 5. Content types with disabled menu items
> Instead of disabling menu items for certain content types, we should hire this entire fieldset if possible.
> If there are issues, let's explore a different UX.

FWIW, this fieldset is gone completely now from all node creation pages as vanessa.luxen@va.gov

ethanteague:
> > 1. Content types with disabled menu items
> >    Instead of disabling menu items for certain content types, we should hire this entire fieldset if possible.
> >    If there are issues, let's explore a different UX.
> 
> FWIW, this fieldset is gone completely now from all node creation pages as [vanessa.luxen@va.gov](mailto:vanessa.luxen@va.gov)

Apologies for the headache, @kevwalsh - looks like the build stalled, and didn't pick up my latest changes. I'm gonna rebuild the instance.

ethanteague:
Okay, @kevwalsh - should be working now.
image

kevwalsh:
@ethanteague Can we avoid this scenario somehow? Where if no eligible parents exist, like at Butler currently, no ancestors show up? 

Screen_Shot_2021-04-19_at_9_45_35_AM

kevwalsh:
Defect:

- [ ] As a Publisher at Pittsburgh, Research (/pittsburgh-health-care/research) is missing

Screen_Shot_2021-04-19_at_9_48_00_AM

kevwalsh (changes requested):
Defect

- [ ] as user 34 i'm going to https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov/node/add/health_care_region_detail_page and i can't successfully create a menu item for my detail page

Here's the node i was working with: https://pr4894-h9posaleh9umnvsbhxxfydmjbe3wrkcj.ci.cms.va.gov/node/18598

Node validates but breadcrumb is wrong and if i go back to edit the node, the menu item option is no longer checked. 


Create_Detail_Page___VA_gov_CMS

docroot/modules/custom/va_gov_menu_access/src/AccessChecks/RouteAccessChecks.php
```diff
@@ -0,0 +1,45 @@
...
+   *   Run access checks for this account.
+   */
+  public function access(Route $route, RouteMatchInterface $route_match, AccountInterface $account) {
+    // Restrict admin menu access.
+    $current_user_roles = $account->getRoles();
+    $admin_roles = [
+      'administrator',
+      'content_admin',
+    ];
+    $admin_role_count = count(array_intersect($admin_roles, $current_user_roles));
```
swirtSJW, line 33: This pattern we are using all over the place.   I would like to move this into the perms service as a public method  `->hasAdminRole()` that does this same logic and returns a bool.   That way we are defining in one spot what an admin role is.

docroot/modules/custom/va_gov_menu_access/src/Form/VaGovMenuAccessConfigForm.php
```diff
@@ -0,0 +1,109 @@
...
+   */
+  protected $cacheManager;
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
+    $instance = parent::create($container);
+    $instance->cacheManager = $container->get('cache.menu');
+
```
swirtSJW, line 26: This would be where you would load your user perms service and pass it in.

swirtSJW (commented):
Just added a couple of thoughts before I lose them.  It is an incomplete review as I need to get on the dashboard review.

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
```diff
@@ -119,9 +119,9 @@ function va_gov_menu_access_alter_menus(array $form, FormStateInterface $form_st
     // Unable to pass $form_id from parent function when using after_build.
     // Throws uneven arguments error.
     $form_id = $form_state->getFormObject()->getFormId();
-    // If (strpos($form_id, 'health_care_region_detail_page') === FALSE) {.
-    va_gov_menu_access_disable_parent_menu_options($form, $form_id);
-    // }
+    if (strpos($form_id, 'health_care_region_detail_page') === FALSE) {
```
ethanteague, line 122: @kevwalsh this was something I missed that I did when I was testing. This missed check was allowing the menu settings fields to be disabled on new Detail pages (node add). There is logic further down that adds this disabling to detail page _edit_ forms, which I thought was necessary (but I am happy to remove).

Anyway, fixing my conditional resolves the faulty page / breadcrumb creation issue:
image
image

docroot/modules/custom/va_gov_menu_access/src/AccessChecks/RouteAccessChecks.php
ethanteague: I'm having a challenge with this, @swirtSJW. I want to create a static method, so that I can call this new method without having to use DI (nervous about adding complexity without much time to resolve issues). Adding the static method is no problem, the issue I'm having is with accessing another method in the same class inside of my new static method. I want to get `currentUser` in my new `hasAdminRole`, but having trouble figuring this out.

docroot/modules/custom/va_gov_menu_access/src/Form/VaGovMenuAccessConfigForm.php
ethanteague: Apologies, @swirtSJW - I don't understand why I'm adding the user perms service, here.

kevwalsh (changes requested):
Defect

As an admin, i create a Detail page without a menu item, and i see this. 

Create_Detail_Page___VA_gov_CMS

kevwalsh (changes requested):
Related to the last issue

As admin, if i _add_ a menu item (with an eligible or non-eligible menu item, doesn't seem to matter), the Parent item value it doesn't "stick" and i basically get the same error as above, albeit with the menu item checkbox checked.

kevwalsh (changes requested):
Same issues as above for user 34, a content publisher with access to only one section. Can't save a detail page right now, with or without a menu item.

swirtSJW:
After looking at this with @kevwalsh there are two outstanding concepts we need to account for.
1) the addition of the default has to be done for all nodes where the menu setttings exist.
2) the reduction of the menu choices to only a set of items, has to be constrained to health_care_region_detail_page nodes.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -435,6 +435,10 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
   }
   // Require message on revision.
   _va_gov_vet_center_require_revision_message($form, $form_state, $form_id);
+
+  // Default owner field element handling.
+  _va_gov_backend_set_default_option_on_new_nodes($form, $form_state);
```
swirtSJW, line 440: Comment needs to go away and this needs to be renamed to reflect what it is doing.
```suggestion
  _va_gov_backend_set_menu_parent_to empty_default($form, $form_state);
```

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -591,6 +595,23 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
...
 
+/**
+ * Sets the owner field to a select option prompt.
+ *
+ * @param array $form
+ *   The exposed widget form array.
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
+ *   The form state.
+ */
+function _va_gov_backend_set_default_option_on_new_nodes(array &$form, FormStateInterface $form_state) {
```
swirtSJW, line 606:
```suggestion
function _va_gov_backend_set_menu_parent_to empty_default(array &$form, FormStateInterface $form_state) {
```

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -591,6 +595,23 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
 
 }
 
+/**
+ * Sets the owner field to a select option prompt.
```
swirtSJW, line 599: Description is not right.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -591,6 +595,23 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
...
+ *
+ * @param array $form
+ *   The exposed widget form array.
+ * @param \Drupal\Core\Form\FormStateInterface $form_state
+ *   The form state.
+ */
+function _va_gov_backend_set_default_option_on_new_nodes(array &$form, FormStateInterface $form_state) {
+  // Don't set parent menu item on new nodes.
```
swirtSJW, line 607: Comment no longer accurate

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -591,6 +595,23 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
...
+  $form_object = $form_state->getFormObject();
+  if ($form_state->getFormObject() instanceof EntityFormInterface
+    && $form_object->getEntity()->getEntityType()->id() === 'node') {
+    $form['#after_build'][] = '_va_gov_backend_alter_parent_menus';
```
swirtSJW, line 611: This method is not specific enough.  'alter ' is not meaningful.
```suggestion
    $form['#after_build'][] = '_va_gov_backend_set_menu_parent_requirement';
```

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -635,6 +656,42 @@ function _va_gov_backend_validate_required_revision_message(array $form, FormSta
...
+ *
+ * @param array $form
+ *   The node form array.
+ * @param Drupal\Core\Form\FormStateInterface $form_state
+ *   Instance of FormStateInterface.
+ *
+ * @return array
+ *   The altered form.
+ */
+function _va_gov_backend_alter_parent_menus(array $form, FormStateInterface $form_state) {
```
swirtSJW, line 670:
```suggestion
function _va_gov_backend_set_menu_parent_requirement(array $form, FormStateInterface $form_state) {
```

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -635,6 +656,42 @@ function _va_gov_backend_validate_required_revision_message(array $form, FormSta
   }
 }
 
+/**
+ * Returns form with empty option set for parent menu item.
```
swirtSJW, line 660: No longer accurate.

ethanteague:
> After looking at this with @kevwalsh there are two outstanding concepts we need to account for.
> 
> 1. the addition of the default has to be done for all nodes where the menu setttings exist.
> 2. the reduction of the menu choices to only a set of items, has to be constrained to health_care_region_detail_page nodes.

Addressed both in lates push, @swirtSJW

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
```diff
@@ -0,0 +1,366 @@
...
+      }
+
+      // Load up all our menu items at once.
+      $loaded_menu_items = \Drupal::entityTypeManager()
+        ->getStorage('menu_link_content')
+        ->loadByProperties(['uuid' => $uuids_only, 'enabled' => 1]);
+
+      $enabled_count = 0;
+
+      // Loop through them, and decide whether it should be allowed in options.
```
swirtSJW, line 265: From this point and below there is a whole lot of grinding going on that I am not sure is necessary.  I am so lost in all the loops and ifs leading up to this that I am a bit lost.   I am not sure if all the alias loading and juggling below is needed.  It certainly is not performant.   I think we need to look closer at this.

swirtSJW (changes requested):

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
```diff
@@ -0,0 +1,366 @@
...
+ * @param string $form_id
+ *   The form id.
+ *
+ * @return array
+ *   The form array.
+ */
+function va_gov_menu_access_disable_parent_menu_options(array &$form, string $form_id) {
+  $form_type_flag = 'detail-page';
+  if (strpos($form_id, 'vet_center') === TRUE) {
+    $form_type_flag = $form_type_flag;
```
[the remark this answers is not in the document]
ethanteague: Sorry, should be 'vet-center'. Changed.

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
```diff
@@ -0,0 +1,366 @@
...
+ *   The form array.
+ * @param string $form_id
+ *   The form id.
+ *
+ * @return array
+ *   The form array.
+ */
+function va_gov_menu_access_disable_parent_menu_options(array &$form, string $form_id) {
+  $form_type_flag = 'detail-page';
+  if (strpos($form_id, 'vet_center') === TRUE) {
```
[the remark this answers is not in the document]
ethanteague: I wanted to make sure that the menu disabling logic only targets VAMC's and can't spill over into vet center content types.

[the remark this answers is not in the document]
swirtSJW: My understanding is this is only supposed to be fore detail pages so there should be no other spill over.

swirtSJW:
I am working on this now.

swirtSJW:
Seems to be an issue with both the original and this one not grandfathering in existing patterns that may have been done by a content_admin.
image


An existing page with a path that does not match an allowed pattern, should bypass menu reduction.

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
```diff
@@ -0,0 +1,595 @@
...
+    $this->setEmptyMenuParentSelector($form);
+
+    // @todo Remove this nonsense check to activate the remainder of this logic
+    // when the other issues in #2427 have paved the way for adjusting and
+    // testing this logic.
+    if (TRUE) {
+      $this->nuke();
+      return;
+    }
+    // End of @todo removal.  Nothing below this point will run.
```
swirtSJW, line 138: Everything beyond this point can not get executed yet as stakeholders aren't ready for it, and some requirements may be changing as a result of other work that is being done in the full epic.

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
```diff
@@ -0,0 +1,595 @@
...
+  }
+
+  /**
+   * Checks if the menu has a parent set.
+   *
+   * @return bool
+   *   TRUE if the menu has a parent, FALSE otherwise.
+   */
+  protected function hasMenuParent() : bool {
+    return (!empty($this->formState->getValue(['menu', 'menu_parent']))) ? TRUE : FALSE;
```
indytechcook, line 171:
```suggestion
    return !empty($this->formState->getValue(['menu', 'menu_parent']));
```

This _should_ work...

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
```diff
@@ -0,0 +1,595 @@
...
+  }
+
+  /**
+   * Checks if the menu is enabled.
+   *
+   * @return bool
+   *   TRUE if the menu is enabled, FALSE otherwise.
+   */
+  protected function isMenuEnabled() : bool {
+    return ($this->formState->getValue(['menu', 'enabled']) === 1) ? TRUE : FALSE;
```
ethanteague, line 181:
```suggestion
    return $this->formState->getValue(['menu', 'enabled']);
```
Riffing on @indytechcook 's suggestion above, this should return a bool, so I don't _think_ you need the ternary.

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
```diff
@@ -0,0 +1,595 @@
...
+   * @param array $form
+   *   The form array by reference.
+   */
+  protected function applyVamcMenuRulesForDetailPage(array &$form) {
+    if ($this->nodeBundle !== 'health_care_region_detail_page') {
+      // Not our type, bail out.
+      return;
+    }
+    // List of menu items that aren't real links = faux menu items.
+    $allowed_separators = ['About'];
```
ethanteague, line 256: Wondering if this should live in config, @swirtSJW ? Seems like this is gonna be something we need to mess with on a regular basis.

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
```diff
@@ -0,0 +1,595 @@
...
+  protected function preventUnintendedChangeOfParent(array &$form) {
+    // Check to see if the current parent exists in the reduced form.
+    if (!in_array($this->currentMenuParent, $form['menu']['link']['menu_parent']['#options'])) {
+      // Parent does not exist in reduced form, Put the original parent options
+      // back to prevent data loss. The existing menu setting should not be
+      // allowed, but it exists, so allow it to persist. It may have been
+      // created and approved by a content_admin.
+      $form['menu']['link']['menu_parent']['#options'] = $this->originalMenuParentOptions;
+      // The menu form itself should be hidden for non-admins.
+      $form['menu']['#attributes'] = ['style' => 'display:none'];
```
ethanteague, line 562: @swirtSJW I'm curious why you are adding an inline style, in lieu of something like 
```suggestion
      $form['menu']['#attributes']['class'][] = 'element-invisible';
```

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/**
+ * Implements hook_form_BASE_FORM_ID_alter().
+ */
+function va_gov_menu_access_form_node_form_alter(&$form, &$form_state, $form_id) {
+  // Process menu options on all node forms.
+  $form['#attached']['library'][] = 'va_gov_menu_access/parent_section_select';
+  $form['#after_build'][] = '_va_gov_menu_access_alter_menus';
+  $form['#after_build'][] = '_va_gov_menu_access_set_menu_parent_requirement';
```
ethanteague, line 29: This is the kind of thing that I would have bungled, @swirtSJW - I wouldn't think to separate these concerns into two different functions. Nice.

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
```diff
@@ -0,0 +1,84 @@
...
+ * @param array $form
+ *   The node form array.
+ * @param Drupal\Core\Form\FormStateInterface $form_state
+ *   Instance of FormStateInterface.
+ */
+function _va_gov_menu_access_parent_menu_selected_validation(array $form, FormStateInterface $form_state) {
+  $menu_enabled = $form_state->getValue(['menu', 'enabled']);
+  $parent_empty = $form_state->isValueEmpty(['menu', 'menu_parent']);
+  if ($menu_enabled && $parent_empty) {
+    $form_state->setErrorByName('menu][menu_parent', t('Menu parent item cannot be empty.'));
```
ethanteague, line 63: Just for confirmation: this is setting an error on the appropriate thing in the parent menu widget, and not going nuclear and adding a bunch of red to every field in the widget, correct, @swirtSJW ?

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.services.yml
```diff
@@ -0,0 +1,13 @@
...
+    tags:
+      - { name: event_subscriber }
+  va_gov_menu_access.access:
+    class: Drupal\va_gov_menu_access\Access\RouteAccessChecks
+    arguments: ['@va_gov_user.user_perms']
+    tags:
+      - { name: access_check, applies_to: _menu_access_check }
+  va_gov_menu_access.menu_reduction_service:
+    class: Drupal\va_gov_menu_access\Service\MenuReductionService
+    arguments: ['@config.factory', '@entity_type.manager', '@path_alias.manager', '@va_gov_user.user_perms']
```
ethanteague, line 13: Gonna ask you to school me on arguments in services when you have some time, @swirtSJW

ethanteague (commented):
Ty +1000 for taking this on, @swirtSJW

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
swirtSJW: getValue returns NULL if it finds nothing.  Which then breaks the contract of returning a bool.  Or are you saying it is no longer needed because the return-type declaration of bool will actually cast it to a bool?

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
swirtSJW: Yah it is only here because that is where you had it and there is only 1.   Also I think as we roll menu goverance to other content types and other products, we will have too many variations to handle in config so some of this may move back into code. 
I think when we get all the aliases locked down to using the menu, then we can change much of the logic to not rely on aliases, but to go with more of an allowed menu pattern structure.

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
swirtSJW: Yah I hemmed and hawed on that one for a bit.  I mainly chose the inline style because I did not want a css snafu surfacing it.  I could be talked either way on it.  It felt dirty, but also felt right ;)

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
```diff
@@ -0,0 +1,595 @@
...
+   *
+   * @return string|null
+   *   The type of menu entry this would be based on path.
+   */
+  protected function getMenuType($alias) {
+    $type = NULL;
+    if ($alias) {
+      $type = $this->checkForTypeParentWithEnabledChildren($alias)
+      ?? $this->checkForTypeDisabledParent($alias)
+      ?? $this->checkForSimpleMatch($alias);
```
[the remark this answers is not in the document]
swirtSJW: Yep  Null coalescing operator is one of the cool new php items that makes this kind of thing so slick.

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.module
swirtSJW: Good question.  Just the appropriate amount of red.
image

You did it well.

docroot/modules/custom/va_gov_menu_access/va_gov_menu_access.services.yml
swirtSJW: Those get passed into the service constructor as though you were obtaining them through using the dependency injection create() method.

ethanteague:
QA pass with latest steps for Lisa T and Ryan S, @swirtSJW :+1:

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
swirtSJW: I had to prove it to myself.  Turns out the return type declaration works for everything but NULL
image

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
swirtSJW: Keeping as is because it does not cover the NULL option 
image

swirtSJW:
Updated the tests

swirtSJW:
image

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
indytechcook: No, but `empty` returns a boolean.

indytechcook: if you wanted to cast it `(bool)` is way faster also.

docroot/modules/custom/va_gov_menu_access/src/Service/MenuReductionService.php
indytechcook: You don't need the `TRUE : FALSE` part.  `($this->formState->getValue(['menu', 'enabled']) === 1)` returns a `bool`

Issue #4604, opened 2021-03-09
Owner should have no default value for any content type


kevwalsh:
**To Reproduce**
As an admin or content_admin
1. Go to https://prod.cms.va.gov/node/add/full_width_banner_alert
2. Default value for Owner is VA Pittsburgh health care

**Expected behavior**
There should be no default value (until we add some logic to make this smarter). 

**ACs**
- [ ]  Editors with only one workbench access section should see their section pre-populated in field_administration
- [ ] Editors with more than one workbench access section should be forced to choose

**Screenshot**
![Content___VA_gov_CMS](https://user-images.githubusercontent.com/643678/110403488-27965c00-804b-11eb-864a-0084a57380ce.png)

PR #4909, opened 2021-04-02
VACMS-4604: Change default field_administration to _none

ethanteague:
## Description
See #4604 
##AC's  
- [x] Editors with only one workbench access section should see their section pre-populated in field_administration
- [x] Editors with more than one workbench access section should be forced to choose

## Testing done
Visual

## QA steps
- [x] As an administrator, go to `user/34/roles` and assign the `Content admin` role
- [x] Login as user 34 (Ryan.Stubblebine@va.gov), and go to any node form add page.
- [x] Visually verify that the `field_administration` field is set to `Pittsburgh Health Care`
- [x] Change the `field_administration` value, complete the form and save.
- [ ] Visit the edit page for the node you just saved, and visually verify the `field_administration` you set appears as the default value.
- [ ] As an administrator, go to `user/34/workbench_access` and assign at least one more additional section.
- [ ]  Login as user 34 (Ryan.Stubblebine@va.gov), and go to any node form add page.
- [x] Visually verify that the `field_administration` field is set to `Select a value`
- [ ] Change the `field_administration` value, complete the form and save.
- [x] Visit the edit page for the node you just saved, and visually verify the `field_administration` value you set appears as the default value.

### Definition of Done

### Select Team for PR review

- [ ] `Core Application Team`
- [x] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/config/sync/field.field.node.full_width_banner_alert.field_administration.yml
+++ b/config/sync/field.field.node.full_width_banner_alert.field_administration.yml
@@ -6,8 +6,13 @@ dependencies:
     - field.storage.node.field_administration
     - node.type.full_width_banner_alert
     - taxonomy.vocabulary.administration
-  content:
-    - 'taxonomy_term:administration:87832236-1e54-4ce3-8141-8dec27c8a9a7'
+  module:
+    - entity_reference_validators
+third_party_settings:
+  entity_reference_validators:
+    circular_reference: false
+    circular_reference_deep: false
+    duplicate_reference: false
 id: node.full_width_banner_alert.field_administration
 field_name: field_administration
 entity_type: node
@@ -16,9 +21,7 @@ label: Owner
 description: ''
 required: true
 translatable: true
-default_value:
-  -
-    target_uuid: 87832236-1e54-4ce3-8141-8dec27c8a9a7
+default_value: {  }
 default_value_callback: ''
 settings:
   handler: 'default:taxonomy_term'
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -521,6 +521,8 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
 
   _va_gov_backend_dropdown_field_access($form, $targets);
 
+  _va_gov_backend_get_owner_section_default($form);
+
   if ($form_id === 'workbench_access_assign_user') {
     $form['#submit'][] = 'va_gov_backend_workbench_assign_user_form_submit';
   }
@@ -1189,6 +1191,31 @@ function _va_gov_backend_dropdown_field_access(array &$form, array $targets) {
             }
           }
         }
+      }
+    }
+  }
+}
+
+/**
+ * For setting the default owner section in dropdown.
+ *
+ * @param array $form
+ *   The entity form.
+ */
+function _va_gov_backend_get_owner_section_default(array &$form) {
+  if (!empty($form['field_administration']['widget'])) {
+    $user = \Drupal::currentUser();
+    $admin_roles = [
+      'administrator',
+      'content_admin',
+    ];
+    $role_count = array_intersect($admin_roles, $user->getRoles());
+    // User is admin, so try to set a default section in the dropdown.
+    if ($role_count > 0) {
+      $perms_service = \Drupal::service('va_gov_user.user_perms');
+      // If only 1, make default, otherwise defaults to Select prompt.
+      if ($default = $perms_service->getDefault($user)) {
+        $form['field_administration']['widget']['#default_value'] = $default;
       }
     }
   }
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -1204,19 +1204,10 @@ function _va_gov_backend_dropdown_field_access(array &$form, array $targets) {
  */
 function _va_gov_backend_get_owner_section_default(array &$form) {
   if (!empty($form['field_administration']['widget'])) {
-    $user = \Drupal::currentUser();
-    $admin_roles = [
-      'administrator',
-      'content_admin',
-    ];
-    $role_count = array_intersect($admin_roles, $user->getRoles());
-    // User is admin, so try to set a default section in the dropdown.
-    if ($role_count > 0) {
-      $perms_service = \Drupal::service('va_gov_user.user_perms');
-      // If only 1, make default, otherwise defaults to Select prompt.
-      if ($default = $perms_service->getDefault($user)) {
-        $form['field_administration']['widget']['#default_value'] = $default;
-      }
+    $perms_service = \Drupal::service('va_gov_user.user_perms');
+    // If only 1, make default, otherwise defaults to Select prompt.
+    if ($default = $perms_service->getDefault(TRUE)) {
+      $form['field_administration']['widget']['#default_value'] = $default;
     }
   }
 }
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -521,8 +521,6 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
 
   _va_gov_backend_dropdown_field_access($form, $targets);
 
-  _va_gov_backend_set_owner_section_default($form, $form_state);
-
   if ($form_id === 'workbench_access_assign_user') {
     $form['#submit'][] = 'va_gov_backend_workbench_assign_user_form_submit';
   }
@@ -1197,6 +1195,13 @@ function _va_gov_backend_dropdown_field_access(array &$form, array $targets) {
 }
 
 /**
+ * Implements hook_form_BASE_FORM_ID_alter().
+ */
+function va_gov_backend_form_node_form_alter(array &$form, FormStateInterface &$form_state) {
+  _va_gov_backend_set_owner_section_default($form, $form_state);
+}
+
+/**
  * For setting the default owner section in dropdown.
  *
  * @param array $form
@@ -1204,19 +1209,18 @@ function _va_gov_backend_dropdown_field_access(array &$form, array $targets) {
  * @param Drupal\Core\Form\FormStateInterface $form_state
  *   The entity form state.
  */
-function _va_gov_backend_set_owner_section_default(array &$form, FormStateInterface $form_state) {
+function _va_gov_backend_set_owner_section_default(array &$form, FormStateInterface &$form_state) {
   if (!empty($form['field_administration']['widget'])) {
-    // See if there is already a value set on the entity for this field.
-    $field_administration_value = NULL;
+    // No value set, so run default value set logic.
     if ($form_state->getFormObject() instanceof EntityFormInterface) {
-      $field_administration_value = $form_state->getFormObject()->getEntity()->field_administration->getString();
-    }
-    // There isn't a target id, so set a default using perms service.
-    if (!$field_administration_value) {
-      $perms_service = \Drupal::service('va_gov_user.user_perms');
-      // If only 1, make default, otherwise defaults to Select prompt.
-      if ($default = $perms_service->getDefault(TRUE)) {
-        $form['field_administration']['widget']['#default_value'] = $default;
+      $entity = $form_state->getFormObject()->getEntity();
+
+      if ($entity->isNew()) {
+        $perms_service = \Drupal::service('va_gov_user.user_perms');
+        // If only 1, make default, otherwise defaults to Select prompt.
+        if ($default = $perms_service->getDefault(TRUE)) {
+          $form['field_administration']['widget']['#default_value'] = $default;
+        }
       }
     }
   }
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -521,7 +521,7 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
 
   _va_gov_backend_dropdown_field_access($form, $targets);
 
-  _va_gov_backend_get_owner_section_default($form);
+  _va_gov_backend_set_owner_section_default($form);
 
   if ($form_id === 'workbench_access_assign_user') {
     $form['#submit'][] = 'va_gov_backend_workbench_assign_user_form_submit';
@@ -1202,7 +1202,7 @@ function _va_gov_backend_dropdown_field_access(array &$form, array $targets) {
  * @param array $form
  *   The entity form.
  */
-function _va_gov_backend_get_owner_section_default(array &$form) {
+function _va_gov_backend_set_owner_section_default(array &$form) {
   if (!empty($form['field_administration']['widget'])) {
     $perms_service = \Drupal::service('va_gov_user.user_perms');
     // If only 1, make default, otherwise defaults to Select prompt.
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -521,7 +521,7 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
 
   _va_gov_backend_dropdown_field_access($form, $targets);
 
-  _va_gov_backend_set_owner_section_default($form);
+  _va_gov_backend_set_owner_section_default($form, $form_state);
 
   if ($form_id === 'workbench_access_assign_user') {
     $form['#submit'][] = 'va_gov_backend_workbench_assign_user_form_submit';
@@ -1201,13 +1201,23 @@ function _va_gov_backend_dropdown_field_access(array &$form, array $targets) {
  *
  * @param array $form
  *   The entity form.
+ * @param Drupal\Core\Form\FormStateInterface $form_state
+ *   The entity form state.
  */
-function _va_gov_backend_set_owner_section_default(array &$form) {
+function _va_gov_backend_set_owner_section_default(array &$form, FormStateInterface $form_state) {
   if (!empty($form['field_administration']['widget'])) {
-    $perms_service = \Drupal::service('va_gov_user.user_perms');
-    // If only 1, make default, otherwise defaults to Select prompt.
-    if ($default = $perms_service->getDefault(TRUE)) {
-      $form['field_administration']['widget']['#default_value'] = $default;
+    // See if there is already a value set on the entity for this field.
+    $field_administration_value = NULL;
+    if ($form_state->getFormObject() instanceof EntityFormInterface) {
+      $field_administration_value = $form_state->getFormObject()->getEntity()->field_administration->getString();
+    }
+    // There isn't a target id, so set a default using perms service.
+    if (!$field_administration_value) {
+      $perms_service = \Drupal::service('va_gov_user.user_perms');
+      // If only 1, make default, otherwise defaults to Select prompt.
+      if ($default = $perms_service->getDefault(TRUE)) {
+        $form['field_administration']['widget']['#default_value'] = $default;
+      }
     }
   }
 }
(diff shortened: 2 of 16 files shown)


docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -568,6 +571,21 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
 
 }
 
+/**
+ * Sets empty option for owner field.
```
swirtSJW, line 575:
```suggestion
 * Sets empty default option for owner field.
```

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -568,6 +571,21 @@ function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form
...
+ *
+ * @param array $form
+ *   The node form.
+ * @param string $form_id
+ *   The form id.
+ */
+function _va_gov_backend_unset_default_owner(array &$form, $form_id) {
```
swirtSJW, line 582:
```suggestion
function _va_gov_backend_set_default_administration(array &$form, $form_id) {
```
Name of function is the opposite of what it does.  Also we want to address the field name rather than the label since the label may change.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -415,6 +415,9 @@ function _va_gov_backend_allowed_formats_remove_textarea_help(array $form_elemen
  * Implements hook_form_alter().
  */
 function va_gov_backend_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  // Don't set default owner on new nodes.
+  _va_gov_backend_unset_default_owner($form, $form_id);
```
swirtSJW, line 419:
```suggestion
  _va_gov_backend_add_default_administration($form, $form_id);
```

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/OwnerUndefined.php
```diff
@@ -0,0 +1,25 @@
...
+ * )
+ */
+class OwnerUndefined extends Constraint {
+
+  /**
+   * The message that will be shown if an owner is not chosen.
+   *
+   * @var \Drupal\va_gov_backend\Plugin\Validation\Constraint
+   */
+  public $notDefined = 'Owner must be defined';
```
swirtSJW, line 23:
```suggestion
  public $notDefined = 'Owner must be defined.';
```

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/OwnerUndefined.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Drupal\va_gov_backend\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Checks that the owner isn't set to none.
+ *
+ * @Constraint(
+ *   id = "OwnerUndefined",
```
swirtSJW, line 11:
```suggestion
 *   id = "FieldAdministrationUndefined",
```
field name is a more enduring source of truth.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/OwnerUndefined.php
```diff
@@ -0,0 +1,25 @@
...
+ *   label = @Translation("Owner Undefined", context = "Validation"),
+ *   type = "string"
+ * )
+ */
+class OwnerUndefined extends Constraint {
```
swirtSJW, line 16:
```suggestion
class FieldAdministrationUndefined extends Constraint {
```

The filename needs to be renamed to match.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/OwnerUndefinedValidator.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Drupal\va_gov_backend\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+/**
+ * Validates the OwnerUndefined constraint.
+ */
+class OwnerUndefinedValidator extends ConstraintValidator {
```
swirtSJW, line 11:
```suggestion
class FieldAdministationUndefinedValidator extends ConstraintValidator {
```
File needs to be renamed accordingly.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1386,6 +1404,10 @@ function va_gov_backend_inline_entity_form_entity_form_alter(&$entity_form, Form
  * Implements hook_entity_bundle_field_info_alter().
  */
 function va_gov_backend_entity_bundle_field_info_alter(&$fields, EntityTypeInterface $entity_type, $bundle) {
+  // Confirm owner is not set to ---none---.
+  if ($entity_type->id() === 'node' && isset($fields['field_administration'])) {
+    $fields['field_administration']->addConstraint('OwnerUndefined');
```
swirtSJW, line 1409:
```suggestion
    $fields['field_administration']->addConstraint('FieldAdministrationUndefined');
```

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/OwnerUndefinedValidator.php
```diff
@@ -0,0 +1,23 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function validate($item, Constraint $constraint) {
+    if ($item->getValue()[0]['target_id'] === '0') {
```
swirtSJW, line 17: I don't think we have to dig into the array here.  I am not clear what interface we have in $item but is it an object with a getValue method.   I would be surprised if it didn't have a isEmpty()   or a get('target_id`)  or something like that.  xdebug will tell you what the interface is or the class and you can see if there is a better method available.

swirtSJW (changes requested):
Sorry a bunch of picky renaming stuff junk.

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/OwnerUndefinedValidator.php
ethanteague: The problem with ->isEmpty() is that it returns false b/c there _is_ a target id on that empty string (gets set to 0 when ---none--- is added to the options array). I've not been able to find a method that gets me any closer to the target id than the get() for the field, @swirtSJW

kevwalsh:
Couple things: 

1. this is not really about field_office, more field_administration. (PR name is misleading for others coming back to this later). 
2. I was thinking that this was just config, there is a default value for field_administration set on a couple of content types. Could this just be done by setting config and some automated tests?

ethanteague:
> Couple things:
> 
> 1. this is not really about field_office, more field_administration. (PR name is misleading for others coming back to this later).
> 2. I was thinking that this was just config, there is a default value for field_administration set on a couple of content types. Could this just be done by setting config and some automated tests?

I updated this to use the existing _none field, @kevwalsh - sorry about the confusion.

swirtSJW:
conflicted

kevwalsh (changes requested):
There's some inconsistent behavior still.  As content admin, i would expect the default value to be "Select a value" any content type that i create, forcing me to choose. Instead, i'm now seeing "Veterans Affairs" as the default value on about half the content types, including, for example 

https://pr4909-6vtcmxuwrv3sd1pfglbk5bgq1hqvvbaz.ci.cms.va.gov/node/add/vet_center_outstation
and https://pr4909-6vtcmxuwrv3sd1pfglbk5bgq1hqvvbaz.ci.cms.va.gov/node/add/full_width_banner_alert

I think i need to be writing better ACs.

I still think this is a config problem, not one that needs to be solved in custom code. We have default values set in config in a few places, like this.  Instead of custom code that overrides this config setting, why not just fix the configuration? 


Owner_settings_for_VAMC_System_Banner_Alert_with_Situation_Updates___VA_gov_CMS

ethanteague:
> There's some inconsistent behavior still. As content admin, i would expect the default value to be "Select a value" any content type that i create, forcing me to choose. Instead, i'm now seeing "Veterans Affairs" as the default value on about half the content types, including, for example
> 
> https://pr4909-6vtcmxuwrv3sd1pfglbk5bgq1hqvvbaz.ci.cms.va.gov/node/add/vet_center_outstation
> and https://pr4909-6vtcmxuwrv3sd1pfglbk5bgq1hqvvbaz.ci.cms.va.gov/node/add/full_width_banner_alert
> 
> I think i need to be writing better ACs.
> 
> I still think this is a config problem, not one that needs to be solved in custom code. We have default values set in config in a few places, like this. Instead of custom code that overrides this config setting, why not just fix the configuration?
> 
> Owner_settings_for_VAMC_System_Banner_Alert_with_Situation_Updates___VA_gov_CMS

Apologies, @kevwalsh - I thought you and @swirtSJW  and I were in agreement on our Monday zoom call that this isn't possible to do via config (an empty owner option isn't available to content types that have available menus set on the content type form - setting it to empty in yaml breaks stuff). I'm glad to implement the same behavior on the admin form for the default value as I did on the node form, but I'm not sure if you are still okay with this approach (and that I'm not missing something). Just lemme know how you want me to proceed.

ethanteague:
My bad, @kevwalsh - you are right that this can be done in config. Looks like some of our content types already have this set, so I grepped on `config/sync` for `config/sync/field.field.node.*.field_administration.yml`, and added / changed the default to `default_value: {  }` on the items that were not set this way.

config/sync/field.field.node.full_width_banner_alert.field_administration.yml
```diff
@@ -6,8 +6,13 @@ dependencies:
     - field.storage.node.field_administration
     - node.type.full_width_banner_alert
     - taxonomy.vocabulary.administration
-  content:
-    - 'taxonomy_term:administration:87832236-1e54-4ce3-8141-8dec27c8a9a7'
+  module:
+    - entity_reference_validators
+third_party_settings:
```
kevwalsh, line 11: I assume this is intentional.

kevwalsh (commented):
@ethanteague cool, thanks! Just left one comment.

config/sync/field.field.node.full_width_banner_alert.field_administration.yml
ethanteague: I didn't change anything manually in config to add this, nor did I change anything in the ui - it was just picked up on export. Do you want me to remove, @kevwalsh ?

tests/behat/features/content_editing.feature
```diff
@@ -155,14 +155,15 @@ Feature: CMS Users may effectively create & edit content
     And I should see "City" in the "#edit-field-address-0" element
     And I should see "State" in the "#edit-field-address-0" element
 
-@content_editing
+@content_editing @et
```
swirtSJW, line 158: Phone home?

swirtSJW, line 158:
```suggestion
@content_editing
```

swirtSJW (approved):

swirtSJW (dismissed):

swirtSJW:
@kevwalsh I am wondering if this is scoped right?  
image
The issue makes it sound like it should only apply for content admin and admin.   But the AC makes it seem more broad.

If it is intended to be more broad, isn't that counter to our goal of having that field prepopulated based on what section an editor is a part of?

kevwalsh:
> @kevwalsh I am wondering if this is scoped right?
> image
> The issue makes it sound like it should only apply for content admin and admin. But the AC makes it seem more broad.
> 
> If it is intended to be more broad, isn't that counter to our goal of having that field prepopulated based on what section an editor is a part of?

I missed this comment the other day. 

This is one of those config things where we have defaults set, when we shouldn't. It's mainly a problem for content admins or editors who have access to more than one section. 

Here's another way of putting it: 
- [ ] Editors with only one workbench access section should see their section pre-populated 
- [ ] Editors with more than one workbench access section should be forced to choose. 

In the future, i think we'd like to find ways to pre-populate this field based on various contexts, for editors who have more than one section, but we're not there yet.

kevwalsh:
I've updated the ACs in the issue accordingly.

ethanteague:
@kevwalsh @swirtSJW I've updated this pr:
Now, when a user who has either the content admin or administrator role, but only one section selected they will see that section as the default owner. 
E.g.: 

- I assigned user 1730 the content admin role, here: `user/1730/roles` and then made sure only `VA Butler health care` section was assigned, here: `/user/1730/workbench_access` and then visited `node/add/full_width_banner_alert` and visually verified that the default `Owner` value was set to `VA Butler health care`.
- I then assigned an additional section to `/user/1730/workbench_access`, revisited `node/add/full_width_banner_alert` and visually verified that the `Owner` values was set to `Select a value`.

swirtSJW (changes requested):

docroot/modules/custom/va_gov_user/src/Service/UserPermsService.php
```diff
@@ -94,12 +96,13 @@ public function getSections(AccountInterface $user) {
...
     $query->fields('sa', ['section_id']);
     $results = $query->execute()->fetchCol();
-
-    if (($key = array_search('administration', $results)) !== FALSE || $user->hasPermission('bypass workbench access')) {
-      unset($results[$key]);
-      $tree = $entity_storage->loadTree('administration');
-      foreach ($tree as $term) {
-        $results[] = $term->tid;
+    if (!$admin_bypass) {
+      if (($key = array_search('administration', $results)) !== FALSE || $user->hasPermission('bypass workbench access')) {
```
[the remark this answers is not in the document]
ethanteague: Yeah, I hear you. I had something like you were suggesting initially, but I felt like it was a little bit harder to read than the extra nest I added. I'll change this, @swirtSJW

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1191,6 +1193,32 @@ function _va_gov_backend_dropdown_field_access(array &$form, array $targets) {
...
+      'administrator',
+      'content_admin',
+    ];
+    $role_count = array_intersect($admin_roles, $user->getRoles());
+    // User is admin, so try to set a default section in the dropdown.
+    if ($role_count > 0) {
+      $perms_service = \Drupal::service('va_gov_user.user_perms');
+      $user_sections = $perms_service->getSections($user, TRUE);
+      // If only 1, make default, otherwise defaults to Select prompt.
+      if (count($user_sections) === 1) {
```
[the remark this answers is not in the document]
ethanteague: Good point - will add, @swirtSJW :+1:

docroot/modules/custom/va_gov_user/src/Service/UserPermsService.php
```diff
@@ -77,14 +77,16 @@ private function getUser($user_id = NULL) {
    *
    * @param \Drupal\Core\Session\AccountInterface $user
    *   User object.
+   * @param bool $admin_bypass
+   *   Determines whether we get all sections or just ones selected on form.
```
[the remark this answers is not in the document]
ethanteague: That's right, @swirtSJW - I'll update this comment.

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1191,6 +1193,31 @@ function _va_gov_backend_dropdown_field_access(array &$form, array $targets) {
...
+    $admin_roles = [
+      'administrator',
+      'content_admin',
+    ];
+    $role_count = array_intersect($admin_roles, $user->getRoles());
+    // User is admin, so try to set a default section in the dropdown.
+    if ($role_count > 0) {
+      $perms_service = \Drupal::service('va_gov_user.user_perms');
+      // If only 1, make default, otherwise defaults to Select prompt.
+      if ($default = $perms_service->getDefault($user)) {
```
swirtSJW, line 1214: This approach seems a bit odd.  Your perms service already has the current user using proper dependency injection..  There is no need to pass it in.  You do have to pass in the admin bypas bool though.
image
 There is no need to load it `\Drupal::currentUser();`.   Lines 1204 - 1208 are not needed.

docroot/modules/custom/va_gov_user/src/Service/UserPermsService.php
```diff
@@ -111,6 +114,24 @@ public function getSections(AccountInterface $user) {
...
+   * Get the user's default workbench access section id.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $user
+   *   User object.
+   *
+   * @return mixed
+   *   Returns default user section key or NULL.
+   */
+  public function getDefault(AccountInterface $user) {
+    $user_sections = $this->getSections($user, TRUE);
```
swirtSJW, line 127:
```suggestion
  public function getDefault(bool $admin_bypass = FALSE) {
    $user_sections = $this->getSections($this->currentUser, $admin_bypass);
```

tests/behat/features/content_editing.feature
swirtSJW: No et.

swirtSJW (changes requested):

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1194,6 +1196,22 @@ function _va_gov_backend_dropdown_field_access(array &$form, array $targets) {
...
+ * For setting the default owner section in dropdown.
+ *
+ * @param array $form
+ *   The entity form.
+ */
+function _va_gov_backend_set_owner_section_default(array &$form) {
+  if (!empty($form['field_administration']['widget'])) {
+    $perms_service = \Drupal::service('va_gov_user.user_perms');
+    // If only 1, make default, otherwise defaults to Select prompt.
+    if ($default = $perms_service->getDefault(TRUE)) {
```
swirtSJW, line 1209: In looking at this closer, I think we are blindly overwriting if a value was already set.   We aren't restricting this just a node add.  I need to test this out to make sure.

swirtSJW (changes requested):
confirmed, We are not checking at all to see if the value was already set, we are just overwritting it with what the perms service says is the default.
image

docroot/modules/custom/va_gov_backend/va_gov_backend.module
ethanteague: :+1: updated, @swirtSJW

docroot/modules/custom/va_gov_backend/va_gov_backend.module
```diff
@@ -1194,6 +1196,32 @@ function _va_gov_backend_dropdown_field_access(array &$form, array $targets) {
...
+  if (!empty($form['field_administration']['widget'])) {
+    // See if there is already a value set on the entity for this field.
+    $field_administration_value = NULL;
+    if ($form_state->getFormObject() instanceof EntityFormInterface) {
+      $field_administration_value = $form_state->getFormObject()->getEntity()->field_administration->getString();
+    }
+    // There isn't a target id, so set a default using perms service.
+    if (!$field_administration_value) {
+      $perms_service = \Drupal::service('va_gov_user.user_perms');
+      // If only 1, make default, otherwise defaults to Select prompt.
```
swirtSJW, line 1217: This seems a little odd to dig into the entity that may not exist yet.  Also seems like a bit to much value juggling.
I didn't test this, but something like this "should" work 
```suggestion
   
    // There isn't a target id, so set a default using perms service.
    if ($form_state->isValueEmpty('field_administration') {
      $perms_service = \Drupal::service('va_gov_user.user_perms');
      // If only 1, make default, otherwise defaults to Select prompt.
```

ethanteague: Stumped on this one. $form_object values are all empty when I set breakpoints in this function.
image

swirtSJW (approved):
Code looks good.   Functional QA works for me too, but I also want @kevwalsh  to poke at it. and do functional QA

kevwalsh (changes requested):
[removed false alarm about a defect that was not a defect but an artifact of behat test that hadn't finished running]

swirtSJW:
image

Issue #4850, opened 2021-03-29
Vet Center MVP Dashboard implementation


kevwalsh:
## User Story or Problem Statement

As a Vet Center editor, i want a task-oriented dashboard so that i can edit content quickly.

## Acceptance Criteria

[The design](https://www.figma.com/file/885DMraAi9V2DLJnrIN7Wv/Vet-Centers?node-id=0%3A1) has been implemented as closely as possible in existing theme. 

## Implementation

(Possibly break this up into 3 issues)

Layout builder config for Vet Center dashboard page (@kevwalsh)
- [ ] #4855  

Dashboard switch
- [ ] `hook_entity_view_mode_alter()` to set the view mode based on a Sections product, traverse the parents and find the first parent with an assigned product 

Block types
- [ ] Card with one action, an edit link with anchor, that can detect context to build link
- [ ] Card with two actions View and Edit, that can can detect context to build links
- What to name these? "Content edit task"? 

![Vet_Centers_–_Figma](https://user-images.githubusercontent.com/643678/112855395-bf043300-907c-11eb-9acd-e570d2fa3efb.png)

## Assets

See icons for dashboard [vet-center-dashboard-icons.zip](https://github.com/department-of-veterans-affairs/va.gov-cms/files/6222006/vet-center-dashboard-icons.zip)

cmaeng:
the big picture is to get VC training completed by sprint 30. so if this work is finished by mid-sprint 30 we should be on track.

PR #4994, opened 2021-04-09
VACMS-4850 Vet Center dashboard MVP

ethanteague:
## Description
See #4850

## Testing done
Visual

## Screenshots
image
image
image

## QA steps
As an administrator, go to `/taxonomy/term/190/edit` and choose the product: `Vet Centers`.
- Clear cache, and go to `/section/veterans-health-administration/vet-centers/escanaba-vet-center`
- [ ] Visually verify that you see the new dashboard format (6 cards up top, listing down below)
image

- [ ]  verify view and edit links on the 6 cards work as expected.

- Go to /section/veterans-health-administration/daytona-beach-vet-center
- [ ] verify that only 5 cards appear as there is no locations page created yet for daytona. (fails gracefully)
image

-  Go to a vetcenter section that has no matching node like /section/veterans-health-administration/vet-centers-mvp
image

- [ ] validate that no cards appear and there are no errors  on the page (fails gracefully)
- [ ] Visit a health system section `/section/veterans-health-administration/vamc-facilities/va-salisbury-health-care` and verify that there is no dashboard and it functions as expected 
 image
- [ ] 
imageVisit another administration term page, and visually verify that the default view listing is displayed, e.g.: `/section/veterans-affairs`

CACHING Testing with xdebug:
- [x] Validate that the service is only called once during the initial sectionpage load
- [x] Validate that the block build() function is only called on the first page load.
- [x] Validate subsequent lods of the page do not run the service or the build(). (cached indefinitely)
- [x] Validate cache is only invalidated by cache flush OR editing he section and saving the section.

## Release instructions
- [ ] As an administrator, go to `/taxonomy/term/190/edit` and choose the product: `Vet Centers`.

### Definition of Done

### Select Team for PR review

- [ ] `Core Application Team`
- [x] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


docroot/modules/custom/va_gov_dashboards/src/Plugin/Derivative/VcDashboardsBlock.php
```diff
@@ -0,0 +1,133 @@
...
+        ];
+
+        $vc_locations_node_fetch = $this->entityTypeManager->getStorage('node')->loadByProperties([
+          'type' => 'vet_center_locations_list',
+          'field_administration' => $tid,
+        ]);
+        if (!empty($vc_locations_node_fetch)) {
+          $vc_locations_node = reset($vc_locations_node_fetch);
+          $vc_locations_node_id = $vc_locations_node->id();
+          $dash_panels['locations'] = [
```
swirtSJW, line 110: It would be great to consider how we could increase the re-useabiltiy  of this.
It could be as simple as moving the things like the specific term id and the content type to reference in vars at the top in one spot and then use the var and making  the vars uses less "vet center" specific.  Or it could be a little more broad by creating a trait that provides some helpers so we don't have to have so much duplicate code in each of these dashboards.

docroot/modules/custom/va_gov_dashboards/va_gov_dashboards.module
```diff
@@ -0,0 +1,83 @@
...
+      case '285':
+        $product = 'resources_and_support';
+        break;
+
+      case '289':
+        $product = 'vet_center_dashboard';
+        break;
+
+      case '290':
+        echo 'va_gov_homepage';
```
swirtSJW, line 74: ?

docroot/modules/custom/va_gov_dashboards/va_gov_dashboards.module
```diff
@@ -0,0 +1,83 @@
...
+ * Implements hook_entity_view_mode_alter().
+ */
+function va_gov_dashboards_entity_view_mode_alter(&$view_mode, EntityInterface $entity, $context) {
+  if ($entity->getEntityTypeId() === 'taxonomy_term' && $entity->bundle() === 'administration') {
+    // Check for a product id.
+    $product_id = $entity->field_product->getString() ? $entity->field_product->getString() : '';
+    // If we don't have one, traverse parents.
+    if (!empty($product_id)) {
+      $parents = \Drupal::service('entity_type.manager')->getStorage("taxonomy_term")->loadAllParents($entity->id());
+      foreach ($parents as $parent) {
```
swirtSJW, line 56: This assumes parents are returned in order of hierarchy.  Is that how they are returned?

docroot/modules/custom/va_gov_dashboards/va_gov_dashboards.module
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Implements hook_entity_view_mode_alter().
+ */
+function va_gov_dashboards_entity_view_mode_alter(&$view_mode, EntityInterface $entity, $context) {
+  if ($entity->getEntityTypeId() === 'taxonomy_term' && $entity->bundle() === 'administration') {
+    // Check for a product id.
+    $product_id = $entity->field_product->getString() ? $entity->field_product->getString() : '';
+    // If we don't have one, traverse parents.
+    if (!empty($product_id)) {
+      $parents = \Drupal::service('entity_type.manager')->getStorage("taxonomy_term")->loadAllParents($entity->id());
```
swirtSJW, line 55: I have some concerns about the performance of this on deep terms and how often people will be hitting the this dashboard.   This could be pretty costly.   We may want to kick around some ideas on how to cache this or create a lookup table.

swirtSJW (commented):
Just some thoughts.

docroot/modules/custom/va_gov_dashboards/va_gov_dashboards.module
ethanteague: Yeah, I was kinda feeling that as well, @swirtSJW . I like the idea of a some kind of a flat file / lookup table, but not sure how extensible this would be.

docroot/modules/custom/va_gov_dashboards/src/Plugin/Derivative/VcDashboardsBlock.php
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getDerivativeDefinitions($base_plugin_definition) {
+    $dash_panels = [];
+    if ($this->routeMatch->getRouteName() === 'entity.taxonomy_term.canonical' || $this->routeMatch->getRouteName() === 'layout_builder.defaults.taxonomy_term.view') {
+      $vc_node = '';
+      $tid = $this->routeMatch->getRawParameter('taxonomy_term') ? $this->routeMatch->getRawParameter('taxonomy_term') : '392';
```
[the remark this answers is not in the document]
kevwalsh: 392 is Escanaba Vet Center, one of 300 Vet Centers. Is that here as an example?  I didn't think TIDs would be in the code, since the trigger is in the content layer.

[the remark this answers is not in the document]
ethanteague: That's right, @kevwalsh.

docroot/modules/custom/va_gov_dashboards/va_gov_dashboards.module
ethanteague: D'oh! Thanks for catching that (typo), @swirtSJW !

docroot/modules/custom/va_gov_dashboards/va_gov_dashboards.module
ethanteague: Yep. I just tested by throwing a breakpoint in that foreach (after removing the break statement in the loop for testing), and checking the id returned at each iteration. Went like this: `392 (Escanaba) -> 699 (Vet Centers MVP) -> 190 (Vet Centers) -> 3 (Veterans Health Administration)`

docroot/modules/custom/va_gov_dashboards/va_gov_dashboards.module
ethanteague: Wondering if the lookup will happen every time the page is hit, @swirtSJW ? Won't the block cache that I have setup (using the term tag) serve this as static content?

docroot/modules/custom/va_gov_dashboards/src/Plugin/Derivative/VcDashboardsBlock.php
ethanteague: @swirtSJW | @kevwalsh 
Agreed - I would like to get a feel for what future dashboards are gonna look like. My thinking is, I would like to make a trait that is generic enough to allow for different panel implementations (fields, etc.). Do either of you all know what kind of stuff we are going to have in other dashboards? If this is still kinda fluid, I'm wondering if I can punt on this until we get other dash implementations a bit more nailed down?

docroot/modules/custom/va_gov_dashboards/src/Plugin/Derivative/VcDashboardsBlock.php
```diff
@@ -0,0 +1,164 @@
...
+   *   The term id.
+   *
+   * @return mixed
+   *   Node object or null if empty.
+   */
+  private function getCorrespondingNode(string $content_type, string $tid) {
+    $node = NULL;
+    $vc_node_fetch = $this->entityTypeManager->getStorage('node')->loadByProperties([
+      'type' => 'vet_center',
+      'field_administration' => $tid,
```
swirtSJW, line 65: On line 129 you create the possiblilty that $tid might be empty,   Here would be where you want to be defensive about what should or should not happen if it is empty.

docroot/modules/custom/va_gov_dashboards/src/Plugin/Derivative/VcDashboardsBlock.php
swirtSJW: Fair enough, lets make dashboard number 2 require  more abstraction of this.

swirtSJW (changes requested):
I am interrupted on my review so it is not complete, but wanted to get this in.

docroot/modules/custom/va_gov_dashboards/va_gov_dashboards.module
swirtSJW: It should, but we need to validate that hypothisis reloading the page with the debugger running and a breakpoint set in what should not be running repeatedly

docroot/modules/custom/va_gov_dashboards/src/Plugin/Derivative/VcDashboardsBlock.php
```diff
@@ -0,0 +1,167 @@
...
+   *   The bundle.
+   * @param string $tid
+   *   The term id.
+   *
+   * @return mixed
+   *   Node object or null if empty.
+   */
+  private function getCorrespondingNode(string $content_type, string $tid) {
+    if (empty($tid)) {
+      return;
```
swirtSJW, line 63: This has to come after where you set $node = null and then it has to return $node.

docroot/modules/custom/va_gov_dashboards/src/Plugin/Derivative/VcDashboardsBlock.php
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * {@inheritdoc}
```
swirtSJW, line 33: Because of what you did in the create, you are overriding the parameters from the base class, so you can't just inherit the doc.  So the docblock has to be explicit.

docroot/modules/custom/va_gov_dashboards/src/Plugin/Block/VcDashboardsBlock.php
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * The entity type manager.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
+   * {@inheritDoc}
```
swirtSJW, line 37: You are altering what gets passed in via the create, so you have to explicitly declare the docblock for the constructor. It can't be inherited.

docroot/modules/custom/va_gov_dashboards/src/Plugin/Derivative/VcDashboardsBlock.php
```diff
@@ -0,0 +1,167 @@
...
+   * @return mixed
+   *   Node object or null if empty.
+   */
+  private function getCorrespondingNode(string $content_type, string $tid) {
+    $node = NULL;
+    if (empty($tid)) {
+      return $node;
+    }
+    $vc_node_fetch = $this->entityTypeManager->getStorage('node')->loadByProperties([
+      'type' => 'vet_center',
```
swirtSJW, line 67:
```suggestion
      'type' => $content_type,
```

swirtSJW (changes requested):
Super minor stuff that needs resolving.

ethanteague:
@swirtSJW - derivatives are creating caching issues. If I move Nashville under Escanaba for the former to inherit the dashboard product treatment of the latter, the blocks created in the derivative get stuck on the term page first viewed after cache clear, so if you load Escanaba after viewing Nashville, you will see links to Nashville vc on Escanaba. 

From what I'm reading, there are a couple/few issues 
(e.g.: https://www.drupal.org/project/drupal/issues/2880682, https://www.drupal.org/project/drupal/issues/3001284)
with derivatives and how caching is handled by said api. To confirm that this behavior is associated with derivative controlled blocks, I created a new, singular block (not under purview of our derivative class), and added logic to output the vc nid associated with the current page term = correct nid is always returned.

So, I'm not sure about the best path forward. I realize that derivatives are the de facto solution for creating block multiples of similar content, but I'm not sure it's capable of handling our use case. I could pretty easily break each of these derivatives into their own block classes and place them in layout builder (probably an hour of work), but wanted to get your thoughts, first.

I updated this pr with Kaise's stuff, as well as the things you and I worked on, yesterday, and a couple of other issues I found when testing.

cc: @kevwalsh

swirtSJW:
@ethanteague  Interesting.  I just noticed we lost the dual view  and edit links.
spec
image


current
image


I don't think that was anything I touched, but I'll have to look into it.

swirtSJW:
I found the issue with the missing links.  They are restored now.

kevwalsh:
I did some clicking around from https://pr4994-ktrcnvwibjve1xd1v2cabeky6va0snm5.ci.cms.va.gov/section/veterans-health-administration/vet-centers-mvp and it looks good. 

It would be great to merge #5089 after this because it's dashboard related. 

We also spun out #5110 into its own issue. Not critical, but it's the next up that's related.

kevwalsh (approved):
QA pass!

acrollet:
minor note: I clicked to https://pr4994-ktrcnvwibjve1xd1v2cabeky6va0snm5.ci.cms.va.gov/section/veterans-health-administration/vet-centers-mvp and it's accessible to anonymous users - not sure if that's desired or not?

swirtSJW:
image

swirtSJW:
@acrollet Good spotting that issue.  It is not related to this PR as the sections taxonomy is available to anon users now.   I believe that is intentional, but I don't have the history on the rationale.    Here is an example from staging.
image

acrollet:
@swirtSJW good deal!

swirtSJW:
Will merge as soon as tests finish.

Issue #4926, opened 2021-04-05
Configure Lando for memcache


acrollet:
## Acceptance Criteria
- [ ] Memcache is configured to be on by default
- [ ] Create manual method to disable memcache for debugging (can be just docs)
- [ ] Process to enable/disable memcache is created & documented

indytechcook:
POC: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/4756

PR #5015, opened 2021-04-12
VACMS 4926: Enables memcache support on Lando.

ndouglas:
## Description

See #4926.  In accordance with updated ACs, this is enabled by default.  Opening a PR early to assist in work/debugging elsewhere, but it shouldn't be merged yet.  I assume (but haven't tested) that the config:import and possible composer installs will fail if the `memcache` module isn't in place at the time.

As far as I can tell, checking whether the `memcache` extension is loaded in `settings.lando.php` is sufficient for determining whether the cache should be in DB or memcache, so the `memcache-on` and `memcache-off` Lando commands can be fairly simple.  I threw a non-skippable cache-rebuild in there.  It might be worth throwing that into a bash script and adding an argument to allow skipping it, but given that this is a development environment I think the possible issues from _not_ rebuilding the cache after changing the location of the cache make this a reasonable decision.

To Do:
- [x] set up for locks
- [x] set up session management (this seems to not need to be handled explicitly because Drupal uses Symfony's session management, unless I'm missing something)
- [x] make sure installation works unattended (currently prompts for session support)

## Testing done
Rebuilding and clicking around.

## Screenshots
With the `memcache` module enabled (by `lando memcache-on`), indicating that information is being read from and written to memcache:
Screen Shot 2021-04-12 at 4 40 52 PM

**EDIT**: Originally included the wrong screenshot 😩 

With the `memcache` module disabled (by `lando memcache-off`), indicating that the site continues to function and that information is _not_ being read from or written to memcache:
Screen Shot 2021-04-12 at 4 42 17 PM

## QA steps

1. Do this
1. Then that
- [ ] This
- [ ] That
- [ ] The other thing

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [x] `DO NOT MERGE`

- [ ] `Needs PO review`


.lando.yml
```diff
@@ -36,6 +36,7 @@ services:
       # cypress-axe dependencies - https://docs.cypress.io/guides/getting-started/installing-cypress.html#System-requirements
       - "apt-get install libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify-dev libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb -y"
       - .lando/scripts/blackfire-init.sh
+      - pecl install memcache-4.0.5.2 && docker-php-ext-enable memcache
```
indytechcook, line 39: This shouldn't be needed

ndouglas: The `memcache` module needs to be installed.

indytechcook: I had it working on lando without it.

ndouglas: Screen Shot 2021-04-14 at 9 20 56 AM

ndouglas: Were you perhaps using `memcached`?

indytechcook: Anything is possible.

indytechcook:
Testing this local.  After a `lando composer nuke` and a `lando rebuild` and a `./scripts/sync_db.sh`:

image

Ran `lando memcache-off` and the cache was still enabled

image

I also monitored the number of commands that ran and and they did increase when memcache was supposed to be not be enabled.

ndouglas:
D'oh, it just falls back to using `memcached`, which Lando installs by default (as mentioned above) 🤦  I added a line to remove the `memcached` configuration and reload Apache, and now get appropriate errors.

indytechcook (approved):

Issue #5371, opened 2021-05-17
Test Failing: Duplicate Benefit Hub references should not be allowed


ElijahLynn:
```
Running:  entity_reference_validation.feature                                             (2 of 6)


  Entity Reference Validation
    1) Duplicate Benefit Hub references should not be allowed

  0 passing (46s)

  1) Entity Reference Validation
       Duplicate Benefit Hub references should not be allowed:
     AssertionError: Timed out retrying after 10000ms: Expected to find content: '1 error has been found' but never did.
      at Context.eval (https://pr5363-wzzeapb2fhjgbyb7dsiyua0xtvokozhg.ci.cms.va.gov/__cypress/tests?p=cypress/integration/entity_reference_validation.feature:174838:28)
      at Context.resolveAndRunStepDefinition (https://pr5363-wzzeapb2fhjgbyb7dsiyua0xtvokozhg.ci.cms.va.gov/__cypress/tests?p=cypress/integration/entity_reference_validation.feature:23780:9)
      at Context.eval (https://pr5363-wzzeapb2fhjgbyb7dsiyua0xtvokozhg.ci.cms.va.gov/__cypress/tests?p=cypress/integration/entity_reference_validation.feature:23121:35)
```

_Originally posted by @ElijahLynn in https://github.com/department-of-veterans-affairs/va.gov-cms/issues/5363#issuecomment-842578031_

PR #5398, opened 2021-05-19
VACMS-5371: Toughen entity reference validation behavioral test.

ndouglas:
## Description

See #5371.  I ran a total of about 120 trials of this and it failed twice in that time, for two different reasons -- one because the delay to wait for the iframe to complete loading was insufficient, and once because the delay between selecting benefit hubs wasn't long enough to allow Drupal to finish processing the side effects of the first selection.

I'd love to find an intelligent way of dealing with these issues, but there's a cost/benefit ratio to consider here.

## Testing done
Ran _a lot_ of tests.

## QA steps

1. Do this
1. Then that
- [ ] This
- [ ] That
- [ ] The other thing

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/tests/behavioral/cypress/integration/common/i_select_the_benefits_hub.js
+++ b/tests/behavioral/cypress/integration/common/i_select_the_benefits_hub.js
@@ -3,15 +3,16 @@ import { Given } from "cypress-cucumber-preprocessor/steps";
 Given(`I select the {string} benefits hub`, (text) => {
   cy.contains("Select Benefit Hub(s)").click({ force: true });
   cy.get(".entity-browser-modal iframe").should("exist");
-  cy.wait(2000);
+  cy.wait(3000);
 
   cy.get(".entity-browser-modal iframe").iframe().within(() => {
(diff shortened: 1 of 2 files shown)


ndouglas:
@indytechcook I was looking mostly at events re: the modal (and iframe) opening and closing, trying to solve this problem on a general level.  I could look at different events (e.g. form being populated after the modal is closed) and solve it more specifically.  My feeling after triggering the error intentionally by removing the delays entirely is that there's a race condition involved in the way Drupal handles inserting data into a form when you immediately open another modal to modify the same form.  There's definitely more to do in this area, and I'm happy to dig into it, but as mentioned during the call I started getting a little twitchy about the cost/benefit ratio.  Thoughts?

indytechcook (approved):
I think you've struck the right balance of effort and value here.  

There probably is more we could do with finding javascript evenst fired by the different components but there would probably still be strange edge cases.

Issue #5436, opened 2021-05-24
Sections field on media doesn't respect workbench access limitations in the media widget version of form


kevwalsh:
**Describe the defect**

As Austin.wilmarth@va.gov, sections field on media allows me to add media to any section at VA. I should only see Colorado Springs 

**To Reproduce**
1. Log in as Austin.wilmarth@va.gov
1. Go to /media/add/image
1. See that you only have access to one section (which should be pre-populated, but is not)
1. Go to https://staging.cms.va.gov/node/3792/edit and add a photo (upload from your computer)
1. Note the Sections field shows _all_sections. 

**Expected behavior**

I should only see the section i have access to (Colorado Springs Vet Center), like i do on the /media/add/image form.
Ideally, if i only have access to one section, it should be prepopulated

**Screenshots**

Current behavior on media widget version of image form

![Edit_Vet_Center_Colorado_Springs_Vet_Center___VA_gov_CMS](https://user-images.githubusercontent.com/643678/119399010-5a450d80-bca6-11eb-9814-99aa6a345685.png)

Comparison of two widgets, between the media widget (top) and /media/add/image (below)

![media-section-choice](https://user-images.githubusercontent.com/643678/119399119-819bda80-bca6-11eb-82b2-e09bd9f8787c.gif)

### CMS Team

Please leave only the team that will do this work selected. If you're not sure, it's fine to leave both selected.

- [x] `Core Application Team`
- [x] `Product Support Team`

ethanteague:
Per our scrum convo on 6/3, approach should be to resolve workbench bug that is preventing modal from being aware of current context. E.g.: don't extend the js approach used to winnow health service dropdowns.

ethanteague:
Gonna have to abandon this one for now - stepping through it in the debugger is yielding no clues.
@kevwalsh

EWashb:
I found this without the team label on it. Adding to our backlog @productmike @BerniXiongA6

PR #5482, opened 2021-05-28
VACMS-5436: Restrict field_media to field_administration value context on node forms.

ethanteague:
## Description
See #5436

## Testing done
Visual

## QA steps
- As a non administrator, go to `node/3731/edit` and click the `Add media` button - visually verify Section select in modal only contains values available to user from options selected in user profile workbench access sections.

### Definition of Done

### Select Team for PR review

- [ ] `Core Application Team`
- [x] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/config/sync/views.view.media_library.yml
+++ b/config/sync/views.view.media_library.yml
@@ -15,6 +15,7 @@ dependencies:
     - media_library
     - taxonomy
     - user
+    - workbench_access
 id: media_library
 label: 'Media library'
 module: views
@@ -497,6 +498,7 @@ display:
         - 'config:core.entity_view_display.media.video.default'
         - 'config:core.entity_view_display.media.video.embedded'
         - 'config:core.entity_view_display.media.video.thumbnail'
+        - 'config:core.entity_view_display.media.video.user_guides'
   page:
     display_plugin: page
     id: page
@@ -810,6 +812,7 @@ display:
         - 'config:core.entity_view_display.media.video.default'
         - 'config:core.entity_view_display.media.video.embedded'
         - 'config:core.entity_view_display.media.video.thumbnail'
+        - 'config:core.entity_view_display.media.video.user_guides'
   widget:
     display_plugin: page
     id: widget
@@ -928,6 +931,8 @@ display:
         arguments: false
         header: false
         css_class: false
+        relationships: false
+        exposed_form: false
       display_description: ''
       access:
         type: perm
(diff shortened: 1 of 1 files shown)


swirtSJW (changes requested):
The code looks good.  
The title on this PR is confusing because it says it is specific to vetcenter, but 
a) it does not match the original issue.
b) it does not contain logic to limit to vetcenter

So the title... which becomes the commit message when this gets merged is misleading.

Also Is this supposed to operate on paragraphs that use media?  It seems like it would not.

ethanteague:
> The code looks good.
> The title on this PR is confusing because it says it is specific to vetcenter, but
> a) it does not match the original issue.
> b) it does not contain logic to limit to vetcenter
> 
> So the title... which becomes the commit message when this gets merged is misleading.
> 
> Also Is this supposed to operate on paragraphs that use media? It seems like it would not.

It doesn't target paragraphs, @swirtSJW ......should it? Maybe more of a @kevwalsh q?

swirtSJW:
> The logic I have in place targets all instances of field_media on node forms (even if the field is in a paragraph on a node form), so this should cover all use cases for said field, @swirtSJW.

I don't think that is true.   The JS would find it regardless of paragraph or not, but because of this bit
```
if (!empty($form['field_media']) &amp;&amp; !empty($form['field_administration'])) {    $form['#attached']['library'][] = 'va_gov_backend/limit_modal_services_to_user_sections';  }
```
The JS would not be attached unless there was media specifically on the node.

ethanteague:
> > The logic I have in place targets all instances of field_media on node forms (even if the field is in a paragraph on a node form), so this should cover all use cases for said field, @swirtSJW.
> 
> I don't think that is true. The JS would find it regardless of paragraph or not, but because of this bit
> 
> ```
> if (!empty($form['field_media']) &amp;&amp; !empty($form['field_administration'])) {    $form['#attached']['library'][] = 'va_gov_backend/limit_modal_services_to_user_sections';  }
> ```
> 
> The JS would not be attached unless there was media specifically on the node.

Yeah, you are right - I was thinking field_media would be the thing that would act as the entity ref to a paragraph (got it twisted). That said......do we have media browsers in paragraphs? Sorry for what's probably a dumb q, just not aware of where (or if) this is an issue. Can you weigh in, @kevwalsh ?

Issue #4204, opened 2021-02-04
Prevent deletion or unpublishing of content fetched with EFF module 


kevwalsh:
As JHL or a VHA HQ editor of centralized content leveraged by VAMC content like Policies, i want to be prevented from deleting or archiving content that would break "downstream" content. 

ACs
- [ ] See https://www.drupal.org/project/entity_field_fetch/issues/3193585
- [ ] No one can delete node 15588 if paragraph in it is referenced
- [ ] A content_admin can delete node 15888 it if it contains no paragraphs referenced via EFF.

ethanteague:
Having a tough time making the paragraph constraint method work. Reader's Digest version is that I can't figure out how to target the triggering element (the paragraph delete button) in a constraint. So, what ends up happening is that on node save, validation runs on all paragraphs, and fails on the first on that is ref'd on the page (regardless of whether it was the subject), and throws a message for it. I pushed up all of this to: https://git.drupalcode.org/issue/entity_field_fetch-3193585/-/tree/3193585-prevent-deletion-of

Tagging this as blocked, and hoping that @swirtSJW might have some ideas.
cc: @kevwalsh

swirtSJW:
The EFF module has been updated. It prevents deletion of any node or term that is used as a source by EFF regardless of whether is it via the /delete form or a bulk operation or any other programmatic use of the API.  It will allow operations by CLI (drush) however as a way to resolve issues.

It does not cover prevention of paragraph removal.

PR #5743, opened 2021-06-29
VACMS-4204: Add patch to prevent deletion of fetched entities.

ethanteague:
## Description
See #4204 

## Testing done
Visual

## QA steps
As an administrator, go to `node/15588/delete` (an entity whose content has been fetched) and visually verify `Access denied` message.

## Screenshots
image

### Definition of Done

### Select Team for PR review

- [ ] `Core Application Team`
- [x] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -335,6 +335,9 @@
             "drupal/entity_clone": {
                 "Cloning child references to nodes has potential to cause memory resource issues": "https://www.drupal.org/files/issues/2020-02-10/entity_clone-child-node-system-drain-3112577-1.patch"
             },
+            "drupal/entity_field_fetch": {
+                "3193585 - Prevent deletion of fetched entities": "https://www.drupal.org/files/issues/2021-06-30/prevent-deletion-of-eff-3193585-2.patch"
+            },
             "drupal/entity_reference_revisions": {
                 "2984027 Incorrect manipulation of default revision flag": "https://www.drupal.org/files/issues/2020-09-03/entity_reference_revisions-default_revision-2984027-10.patch"
             },


ethanteague:
Current pr approach is flawed (querying for entity ref's instead of eff's). Working on new solution.

swirtSJW:
I'm going to close this for 2 reasons:
1. The patch is not complete (does not account for paragraph deletion)
2. We won't apply the patch... once I merge the patch upstream, we will just bump to the newer module release.

swirtSJW:
Looks like this got re-opened when the issue did,  converted to draft to avoid confusion.

swirtSJW:
In testing this patch out for node deletion.  It is only preventing deletion on the node delete form.   If does not prevent deletion by bulk operation.
image

image

image

swirtSJW:
I think access is still the right control, but it has to be on an operation, not a form.

swirtSJW:
I am going to putter with this a bit.

swirtSJW:
I worked on this quite a bit.  There just is no way to prevent deletions that are performed using the API which is what the bulk operation uses other than throwing an exception that would cause a whitescreen.

@ethanteague  I commited your patch but then followed it up with some changes to account for taxonomy terms and some other refinements.  Here are the changes I made if you want to see them. 
 https://git.drupalcode.org/project/entity_field_fetch/commit/eded92c

ethanteague:
> I worked on this quite a bit. There just is no way to prevent deletions that are performed using the API which is what the bulk operation uses other than throwing an exception that would cause a whitescreen.
> 
> @ethanteague I commited your patch but then followed it up with some changes to account for taxonomy terms and some other refinements. Here are the changes I made if you want to see them.
> https://git.drupalcode.org/project/entity_field_fetch/commit/eded92c

Really appreciate your time and hard work on this, @swirtSJW . Ty, buddy.

swirtSJW:
This is replaced by https://github.com/department-of-veterans-affairs/va.gov-cms/pull/6053

PR #6026, opened 2021-08-03
VACMS-4919: Static analysis and deprecated code checks.

### LINKED ISSUE #4919: Add PHPStan GitHub Action
## Description
It'd be very helpful to have some feedback from PHPStan to prevent some common programming bugs before they become an issue, e.g. undeclared interfaces because we forgot to add an appropriate `use` statement, etc.

This is substantial enough to merit an issue simply because PHPStan is greatly enhanced for use with Drupal by the `mglaman/phpstan-drupal` and `phpstan/phpstan-deprecation-rules` packages and a `phpstan.neon` file and possibly custom bootstrapping code, so it would presumably require a separate repo to configure this, much as we did with PHPCS.

At this time, I think it would be best to keep this in an advisory role and not block builds.  That will allow team members to assess its effectiveness and modify the rules and exclusions and so forth so that it serves as a benefit and not a hindrance to the development process.

## Acceptance Criteria
- [ ] PHPStan performs static analysis on PRs and makes suggestions (but does not block builds).

ndouglas:
## Description

Closes #4919.

This is pretty much just the code I had running before -- I think there's going to be some trial-and-error here so just don't look at this for a while.

I considered working with a self-hosted GitHub Actions runner.  I dug around quite a bit trying to understand how it would be accomplished, what it would involve, and why it would be done that way.  It seems to me a bit like swatting a fly with a Buick, especially given that the reason we're doing this in GitHub Actions instead of Tugboat is _because static analysis shouldn't require a lot of overhead_.

I haven't determined why I was encountering different message formats for certain issues.  I was thinking it might involve the autoloader and this might somehow be triggered slightly differently in an Action versus running it locally, but I have no real evidence for that and it seems kinda woo.  I'll continue fiddling with it and run a few searches, but it's not really a big deal and shouldn't block this PR getting merged; the best solution is probably just to **fix the danged error**.

## QA steps

1. Do this
1. Then that
- [ ] This
- [ ] That
- [ ] The other thing

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


ndouglas:
Fails when an error not in the baseline is introduced:

Local:
Screen Shot 2021-08-04 at 10 13 53 AM

Status:
Screen Shot 2021-08-04 at 10 13 04 AM

Review:
Screen Shot 2021-08-04 at 10 13 17 AM

ndouglas:
It looks like that if I do the `continue-on-error` thing, that the status will be marked passing even if an issue is found:

Screen Shot 2021-08-04 at 10 19 10 AM

Screen Shot 2021-08-04 at 10 21 01 AM

Assuming we're at least clicking the "Review" tab when we review a PR, this might be suitable for our purposes.

However, I think that use of deprecated code, or code that triggers the static analyzer in some other way, _should_ block the build and should be more aggressive.  Thoughts?

acrollet:
> However, I think that use of deprecated code, or code that triggers the static analyzer in some other way, _should_ block the build and should be more aggressive. Thoughts?

make it so

(IMO)

acrollet (approved):
This looks great! I agree on the approach vis a vis including the config in the repo.

Issue #4676, opened 2021-03-12
"No authentication credentials provided" during test run


ElijahLynn:
## Description

As a Drupal engineer, I want no php errors to be generated during test runs, so that new code that introduces errors may be prevented from being merged.

## Background

Somehow during our test runs, I think the Behat or va:web:build step, this PHP Error Exception is thrown:
```
Symfony\Component\HttpKernel\Exception\UnauthorizedHttpException: No authentication credentials provided. in Drupal\basic_auth\Authentication\Provider\BasicAuth->challengeException() (line 168 of /var/lib/tugboat/docroot/core/modules/basic_auth/src/Authentication/Provider/BasicAuth.php).
```
![image](https://user-images.githubusercontent.com/1504756/111009431-77756b80-8348-11eb-845d-5e7e6033f3f9.png)

## Acceptance Criteria
- [ ] A full test run does not trigger the "No authentication credentials provided" exception

PR #6089, opened 2021-08-11
VACMS-4676: Rewrites a PHPUnit test that creates some logspam.

ndouglas:
## Description

Relates to #4676.

This test doesn't really do what it's intended to do.  It will only fail if Drupal throws a non-401 exception.  If Drupal throws no exception at all, e.g. because someone accidentally disabled basic_auth on GraphQL, then it will just pass silently.  If we want to test whether the GraphQL endpoint is protected by basic authentication, I think we can use a different method.

And it creates some logspam when everything is configured correctly, so there's not much point in it.

I was curious about whether this 401 exception was unavoidable; some clients will attempt to connect without passing auth headers, then when they receive a 401 will retry with headers.  However, I don't think this is the case with the frontend project.  Even with a simple Node test script using `http`, the auth headers were supplied on the first connection.  So this should close the issue.

Fortunately, the culprit test was easy to find, as only PHPUnit tests call `/graphql?_format=json`.  The frontend doesn't, AFAICT.

I went ahead and rewrote the test.

Now looks like this:

Screen Shot 2021-08-11 at 3 47 46 PM

That PHP error's a different one:

Screen Shot 2021-08-11 at 3 19 55 PM

And triggered in `decoupled.feature`.  Followup issue created in #6093 .


ndouglas:
who luvs ya, @ElijahLynn ?

ElijahLynn:
Very nice Nate!! And thanks for creating the follow up issue too!

ElijahLynn (approved):

Issue #4830, opened 2021-03-26
Add automated tests for BuildFrontend.php


mchelen-gov:
## Description
add automated tests for `docroot/modules/custom/va_gov_build_trigger/src/Service/BuildFrontend.php`

## Acceptance Criteria
- [ ] Unit tests exist for each method of BuildFrontend.php

PR #6162, opened 2021-08-18
VACMS-4830: Add tests for BuildFrontend.

ndouglas:
## Description

Relates to #4830.  Had worked on this when @mchelen-gov brought it up, decided to finish the job.  Private methods and NodeInterface made this not terribly fun, but my heart will go on.

## Testing done
literally nothing but tests.  a landscape of tests and test data extending boundlessly in all directions.  אור אין סוף

## Screenshots
Screen Shot 2021-08-18 at 10 43 49 AM

## QA steps

1. Do this
1. Then that
- [ ] This
- [ ] That
- [ ] The other thing

### Definition of Done

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/tests/phpunit/FrontendBuild/BuildFrontendTest.php
+++ b/tests/phpunit/FrontendBuild/BuildFrontendTest.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace tests\phpunit\FrontendBuild;
+
+use Drupal\Component\Plugin\Exception\PluginException;
+use Drupal\Core\Logger\LoggerChannelFactoryInterface;
+use Drupal\Core\Logger\LoggerChannelInterface;
+use Drupal\Core\Messenger\MessengerInterface;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\va_gov_build_trigger\Service\BuildFrontend;
+use Drupal\va_gov_build_trigger\Environment\EnvironmentDiscovery;
+use Drupal\va_gov_build_trigger\WebBuildStatusInterface;
+use Prophecy\Argument;
+use weitzman\DrupalTestTraits\ExistingSiteBase;
+
+/**
+ * Functional test of the BuildFrontend class.
+ *
+ * @coversDefaultClass \Drupal\va_gov_build_trigger\Service\BuildFrontend
+ */
+class BuildFrontendTest extends ExistingSiteBase {
+
+  /**
+   * Test triggerFrontendBuild()
+   *
+   * @param string $env
+   *   Environment name, like 'lando', 'prod', etc.
+   * @param bool $permitted
+   *   Indicates whether the specified environment name is considered valid.
+   *
+   * @dataProvider triggerFrontendBuildDataProvider
+   */
+  public function testTriggerFrontendBuild(string $env, bool $permitted) {
+    $realPermitted = $this->container->get('plugin.manager.va_gov.environment')->hasDefinition($env);
+    $this->assertEquals($permitted, $realPermitted);
+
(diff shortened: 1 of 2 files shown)


tests/phpunit/FrontendBuild/BuildFrontendTest.php
```diff
@@ -0,0 +1,278 @@
...
+    $environmentDiscoveryProphecy = $this->prophesize(EnvironmentDiscovery::class);
+    $loggerProphecy = $this->prophesize(LoggerChannelInterface::class);
+    $messengerProphecy = $this->prophesize(MessengerInterface::class);
+    $webBuildStatusProphecy = $this->prophesize(WebBuildStatusInterface::class);
+    $nodeProphecy = $this->prophesize(NodeInterface::class);
+
+    $environmentDiscoveryProphecy->shouldTriggerFrontendBuild()->willReturn(FALSE);
+    $environmentDiscoveryProphecy->triggerFrontendBuild()->shouldNotBeCalled();
+
+    $nodeProphecy->isPublished()->willReturn(FALSE)->shouldNotBeCalled();
```
indytechcook, line 158: Does this say that the method `isPublished` will return FALSE and it runs a test to make sure the method isn't called?

indytechcook (approved):
Now that you have written tests for this class, how might have we make the class better?

tests/phpunit/FrontendBuild/BuildFrontendTest.php
ndouglas: Yeah, the `willReturn(FALSE)` could be removed without really hurting anything.  This and the other similar test function were factored out into two functions, and this is a vestige of that.  

`shouldNotBeCalled()` is probably unnecessary too, since if `$environmentDiscovery->triggerFrontendBuild()` is called then that's sufficient to know that someone done screwed up, and knowing that `$node->isPublished()` shouldn't be called is probably excessive knowledge about the internal workings of `triggerFrontendBuildFromContentSave()`.  

That said, I kept it in just because there are cases where `$node->isPublished()` might be called but  `$environmentDiscovery->triggerFrontendBuild()` might not be, whether that's improper knowledge or not.  

I think that the "comparing truth tables" nature of this test is a sign that the class could use a refactor, but we still could use that truth table while refactoring.

ndouglas:
Things like this are not trivial to refactor because of the dependency on `Node` magic properties and things like that.

The `getOriginalFieldValue` concept is useful, but it's not safe and depends on specific properties of the field.  I think that would need to be handled in Core to be 1) safe, 2) comprehensive, and 3) as useful as it should be.  We can do a thing, but Core support would just be better ultimately IMHO.

`changedStatus` is super specific to our business logic.  Again, with Core support for something like `getOriginalFieldValue` this could be three conditions OR'ed together somewhere else.  The method being `private` really is an admission that this code will never be useful anywhere else ever again.  But it's a method because it's too complicated to fit in a single expression, even a nicely-formatted-with-line-breaks one.

`triggerFrontendBuildFromContentSave()` should actually be its own class, I think.  I don't think it should call `$environmentDiscovery->shouldTriggerFrontendBuild()`.  I think it should be something along the lines of `shouldTriggerFrontendBuildFromContentSave()`.  The form's debatable, but I don't think it should check the environment first -- I think it should make decisions based on the content and operation alone.

Ideally all of that business logic would be somewhere outside of the codebase, in config or whatever.  Selecting content types to opt in and selecting fields where changes merit a content release (or "any") should be doable in YAML and with normal Drupal forms.

I think this module has a rather confused and confusing structure in general, just because of the sheer number of changes we've made.

PR #6265, opened 2021-08-27
Enable jsonapi modules and openapi endpoint

indytechcook:
This adds more structure to the JSON:API via the `jsonapi_extras` module.  THis removes some of the "Drupalness" of the API by changing them to more key:value.

This PR also adds open API functionality.


ndouglas:
but seriously, [this](https://www.drupal.org/project/subrequests) is essential

ndouglas:
I've been meaning to take a quick look at this and will do so soon.  The default response structure can definitely be impenetrable and counterintuitive.  I've seen a few solutions for this, but haven't actually used any so I'm curious about the changes here.

ndouglas:
[gists](https://gist.github.com/ndouglas/95440a8a73548aa186d83899c4df9c15) with results of some very simple queries. **TL;DR**: works as expected, actually don't notice much difference from stock JSON:API, but that might be just because I've forgotten the details.

Some things that are super useful to development:
  - some method of automatically/transparently/easily denormalizing the response.
  - having the UUIDs of entities very easily available in HTML source or JS(Drupal.settings?), or a tab that directly exposes the JSON:API version of the node

The former's a FE thing but should be done really early on, otherwise the suffering is substantial.

ndouglas (approved):

indytechcook:
I had to rebase to fix `composer.lock` conflict.  Assuming it works I'll merge fast so I don't get another one.

Issue #4103, opened 2021-01-27
Some redirects are apparently broken.


ndouglas:
## Description
While working on #3971 , I became aware of some broken redirects (when trying to test against them for QA :weary:).  They seem to be broken because the source URI has a trailing slash, which doesn't seem to work. 

 There are about 220, I think:

## To Reproduce
1. Run the following SQL query: `SELECT * FROM redirect WHERE redirect_source__path LIKE '%/';`

Now, there're a few things we might do to try to fix this:
- [ ] Better documentation for users so they don't create redirects with trailing slashes
- [ ] Validation for that field to ensure no trailing slashes are added
- [ ] Manually fixing the existing problematic redirects

These probably need to be discussed though, so I'm just going to nope out of here now.

## Acceptance Criteria
- [ ] Redirects with trailing slashes are fixed to remove the trailing slashes

acrollet:
We just checked current data, and there are still 220 redirects like this. This suggests that this is not an on-going issue, and we can simply fix the bad data and move on.

ndouglas:
Just need to run this query:

```
MariaDB [drupal8]> SELECT * FROM redirect WHERE redirect_source__path LIKE '%/';
...
220 rows in set (0.00 sec)

MariaDB [drupal8]> UPDATE redirect SET redirect_source__path = TRIM(TRAILING '/' FROM redirect_source__path);
Query OK, 220 rows affected (0.07 sec)
Rows matched: 16438  Changed: 220  Warnings: 0

MariaDB [drupal8]> SELECT * FROM redirect WHERE redirect_source__path LIKE '%/';
Empty set (0.03 sec)
```

I'm going to go ahead and open a PR for this, just because I'm hazy about how this data might be used in content-build and don't wanna #DoItLive and then break content builds.

PR #6366, opened 2021-09-09
VACMS-4103: Repairs broken redirects.

ndouglas:
## Description

Relates to #4103.  Created an update hook rather than just running the query live because I wanted to make sure it didn't break content builds.

## Testing done
this very pr

### Select Team for PR review

- [x] `Core Application Team`
- [ ] `Product Support Team`


acrollet (commented):
This is a nit but I think [hook_post_update_NAME](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21module.api.php/function/hook_post_update_NAME/9.2.x) is a bit more appropriate. We have a [post_update file](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/docroot/modules/custom/va_gov_db/va_gov_db.post_update.php) in va_gov_db.

docroot/modules/custom/va_gov_backend/va_gov_backend.install
```diff
@@ -424,3 +424,16 @@ function va_gov_backend_update_8010() {
...
 }
+
+/**
+ * Strip trailing slashes from redirects.
+ */
+function va_gov_backend_update_8011() {
+  $connection = \Drupal::database();
+  $connection->update('redirect')
+    ->fields([
+      'redirect_source__path' => 'TRIM(TRAILING \'/\' FROM redirect_source__path)',
```
acrollet, line 435: nifty query!

ndouglas:
>Executes an update which is intended to update data, like entities.

That's not a nit, you're absolutely right.  Honestly, I forgot about hook_post_update.  Thanks!

acrollet:
> > Executes an update which is intended to update data, like entities.
> 
> That's not a nit, you're absolutely right. Honestly, I forgot about hook_post_update. Thanks!

learned it from @indytechcook !

ndouglas:
- [Before Update, Selected Rows](https://gist.github.com/ndouglas/6aadd8cb6705e1f2057c46d4905c38b6)
- [Before Update, Specifically Affected Rows](https://gist.github.com/ndouglas/a42c10d1d412764e9756fd914304d838)
- [After Update, Selected Rows](https://gist.github.com/ndouglas/589ea3726578649bf6be4d69f030dad1)

acrollet (approved):
The code is concise, does the job and I like the use of the expression - really nice way to write a self-documenting query. I tested a few randomly selected redirects and they go where they're supposed to according to the table. Nice work!

Issue #3203, opened 2020-10-19
Spike: post deploy/live action mechanism


oksana-c:
## Owner

Things that would normally be in a hook_update() but take too long to run so we don't want them there. Current workaround is to do them manually by a DevOps engineer. 

## Acceptance Criteria

- [ ] we brainstormed all known and possible scenarios where we could utilize post deploy/live action
- [ ] we have an implementation plan
- [ ] Documentation has been written/updated to provide guidance to CMS engineers on when to use DB updates vs other approaches

Participants:

TBD

PR #6496, opened 2021-09-21
Vacms 3573 trim node titles on save

timcosgrove:
## Description

Relates to #_3573_. 

## QA steps

As a user with node edit permissions.
1. Create any node, giving it a title that has leading and/or trailing spaces in it, e.g. "   About Us  ", and save it.
1. Edit the node you just created, and examine the title.
- [ ] There should be no leading spaces
- [ ] There should be no trailing spaces
- [ ] The title is otherwise unaltered. 

1. Create a taxonomy term, giving it a name that has leading and/or trailing spaces in it, e.g. "   Orange  ", and save it.
1. Edit the term you just created, and examine the name.
- [ ] The title is unaltered, including the original spaces.

### Definition of Done

### Select Team for PR review

- [ ] `Core Application Team`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
+++ b/docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace Drupal\va_gov_backend\EventSubscriber;
+
+use Drupal\node\Entity\Node;
+use Drupal\core_event_dispatcher\Event\Entity\EntityPresaveEvent;
+use Drupal\hook_event_dispatcher\HookEventDispatcherInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * VA.gov Backend Entity Event Subscriber.
+ */
+class EntityEventSubscriber implements EventSubscriberInterface {
+
+  /**
+   * Entity create Event call.
+   *
+   * @param \Drupal\core_event_dispatcher\Event\Entity\EntityPresaveEvent $event
+   *   The event.
+   */
+  public function entityPresave(EntityPresaveEvent $event): void {
+    $entity = $event->getEntity();
+    if ($entity->getEntityTypeId() === 'node') {
+      $this->trimNodeTitleWhitespace($entity);
+    }
+  }
+
+  /**
+   * Trim any preceding and trailing whitespace on node titles.
+   *
+   * @param \Drupal\node\Entity\Node $node
+   *   The node to be modified.
+   */
+  private function trimNodeTitleWhitespace(Node $node) {
+    $title = $node->getTitle();
+    // Trim leading and then trailing separately to avoid a convoluted regex.
+    $title = preg_replace('/^\s+/', '', $title);
+    $title = preg_replace('/\s+$/', '', $title);
+    $node->setTitle($title);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents(): array {
+    return [
+      HookEventDispatcherInterface::ENTITY_PRE_SAVE => 'entityPresave',
+    ];
+  }
+
+}
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.install
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.install
@@ -428,7 +428,7 @@ function va_gov_backend_update_8010() {
 /**
  * Re-save all nodes to correct leading and trailing whitespace issues.
  */
-function va_gov_backend_update_8013(&$sandbox) {
+function va_gov_backend_update_8011(&$sandbox) {
   $node_storage = \Drupal::entityTypeManager()->getStorage('node');
 
   // Get the node count.
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.install
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.install
@@ -424,3 +424,92 @@ function va_gov_backend_update_8010() {
   }
   return "content_moderation_state_field_data and content_moderation_state_field_revision updated to workflow editor.";
 }
+
+/**
+ * Re-save all nodes to correct leading and trailing whitespace issues.
+ */
+function va_gov_backend_update_8013(&$sandbox) {
+  $node_storage = \Drupal::entityTypeManager()->getStorage('node');
+
+  // Get the node count.
+  // This runs only once.
+  if (!isset($sandbox['total'])) {
+    $query = $node_storage->getQuery();
+    $nids_to_update = $query->execute();
+    $result_count = count($nids_to_update);
+    $sandbox['total'] = $result_count;
+    $sandbox['current'] = 0;
+    $sandbox['nids_to_update'] = array_combine(
+            array_map('_va_gov_backend_stringifynid', array_values($nids_to_update)),
+            array_values($nids_to_update));
+  }
+
+  // Do not continue if no nodes are found.
+  if (empty($sandbox['total'])) {
+    $sandbox['#finished'] = 1;
+    return t('No nodes were found to be processed.');
+  }
+
+  $limit = 25;
+
+  // Load entities.
+  $node_ids = array_slice($sandbox['nids_to_update'], 0, $limit, TRUE);
+  $nodes = $node_storage->loadMultiple($node_ids);
+
+  foreach ($nodes as $node) {
+    // Save some time by only saving nodes that need adjustments.
+    if (
+        preg_match('/\s+$/', $node->getTitle())
+        ||
+        preg_match('/^\s+/', $node->getTitle())
+        ) {
+      // Make this change a new revision.
+      $node->setNewRevision(TRUE);
+
+      // Set revision author to uid 1317 (CMS Migrator user).
+      $node->setRevisionUserId(1317);
+      $node->setChangedTime(time());
+      $node->setRevisionCreationTime(time());
+
+      // Set revision log message.
+      $node->setRevisionLogMessage('Resaved node to update title.');
+      $node->save();
+    }
+
+    unset($sandbox['nids_to_update']["node_{$node->id()}"]);
+    $nids[] = $node->id();
+    $sandbox['current'] = $sandbox['total'] - count($sandbox['nids_to_update']);
+  }
+
+  // Log the processed nodes.
+  Drupal::logger('va_gov_backend')
+    ->log(LogLevel::INFO, '%current nodes saved to update the title & alias. Nodes processed: %nids', [
+      '%current' => $sandbox['current'],
+      '%nids' => implode(', ', $nids),
+    ]);
+
+  $sandbox['#finished'] = ($sandbox['current'] / $sandbox['total']);
+
+  // Log the all-finished notice.
+  if ($sandbox['#finished'] == 1) {
+    Drupal::logger('va_gov_backend')->log(LogLevel::INFO, 'RE-saving all %count nodes to correct title issues.', [
+      '%count' => $sandbox['total'],
+    ]);
+    return "Node re-saving complete. {$sandbox['current']} / {$sandbox['total']}";
+  }
+
+  return "Processing node titles...{$sandbox['current']} / {$sandbox['total']}";
+}
+
+/**
+ * Callback function to concat node ids with string.
+ *
+ * @param int $nid
+ *   The node id.
+ *
+ * @return string
+ *   The node id concatenated to the end o node_
+ */
+function _va_gov_backend_stringifynid($nid) {
+  return "node_$nid";
+}
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.services.yml
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.services.yml
@@ -9,3 +9,7 @@ services:
   va_gov_backend.va_gov_url:
     class: Drupal\va_gov_backend\Service\VaGovUrl
     arguments: ['@http_client', '@settings', '@va_gov.build_trigger.environment_discovery']
+  va_gov_backend.entity_event_subscriber:
+    class: Drupal\va_gov_backend\EventSubscriber\EntityEventSubscriber
+    tags:
+      - {name: event_subscriber }
--- a/tests/phpunit/Content/TrimNodeTitleWhitespaceTest.php
+++ b/tests/phpunit/Content/TrimNodeTitleWhitespaceTest.php
@@ -1,0 +1,138 @@
+<?php
+
+namespace tests\phpunit\Content;
+
+use weitzman\DrupalTestTraits\ExistingSiteBase;
+
+/**
+ * A test to confirm the proper functioning of TrimNodeTitleWhitespace.
+ *
+ * @group functional
+ * @group all
+ *
+ * @coversDefaultClass \Drupal\va_gov_backend\EventSubscriber\EntityEventSubscriber
+ */
+class TrimNodeTitleWhitespaceTest extends ExistingSiteBase {
+
+  /**
+   * Confirm that the trimmed title is as expected.
+   *
+   * @param string $titleInput
+   *   The node title to test.
+   * @param string $expectedTitleOutput
+   *   The expected node title.
+   *
+   * @dataProvider confirmExpectedOutputDataProvider
+   */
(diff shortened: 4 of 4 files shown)


tests/phpunit/Content/TrimNodeTitleWhitespaceTest.php
```diff
@@ -0,0 +1,138 @@
...
+   * @dataProvider confirmExpectedOutputDataProvider
+   */
+  public function testConfirmExpectedOutput($titleInput, $expectedTitleOutput) {
+    // Creates a user. Will be automatically cleaned up at the end of the test.
+    $author = $this->createUser();
+
+    // Create a node with the input title.
+    $entity = $this->createNode([
+      'title' => $titleInput,
+      'type' => 'health_care_region_page',
```
timcosgrove, line 34: Again, cribbed from other code, but it didn't seem like it would be an issue to use this content type.

docroot/modules/custom/va_gov_backend/va_gov_backend.install
```diff
@@ -424,3 +424,92 @@ function va_gov_backend_update_8010() {
...
+    if (
+        preg_match('/\s+$/', $node->getTitle())
+        ||
+        preg_match('/^\s+/', $node->getTitle())
+        ) {
+      // Make this change a new revision.
+      $node->setNewRevision(TRUE);
+
+      // Set revision author to uid 1317 (CMS Migrator user).
+      $node->setRevisionUserId(1317);
```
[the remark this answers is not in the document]
swirtSJW: That Id is right.

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -0,0 +1,51 @@
...
+   * Trim any preceding and trailing whitespace on node titles.
+   *
+   * @param \Drupal\node\Entity\Node $node
+   *   The node to be modified.
+   */
+  private function trimNodeTitleWhitespace(Node $node) {
+    $title = $node->getTitle();
+    // Trim leading and then trailing separately to avoid a convoluted regex.
+    $title = preg_replace('/^\s+/', '', $title);
+    $title = preg_replace('/\s+$/', '', $title);
```
ndouglas, line 38: Out of curiosity, why not `trim()`?

docroot/modules/custom/va_gov_backend/va_gov_backend.install
```diff
@@ -424,3 +425,96 @@ function va_gov_backend_update_8010() {
...
+  if (empty($sandbox['total'])) {
+    $sandbox['#finished'] = 1;
+    return t('No nodes were found to be processed.');
+  }
+
+  $limit = 25;
+
+  // Load entities.
+  $node_ids = array_slice($sandbox['nids_to_update'], 0, $limit, TRUE);
+  $nodes = $node_storage->loadMultiple($node_ids);
```
ndouglas, line 459: I'm not fully awake yet so I might be making a mistake here, but note that since we have content moderation involved for some of the content types, `loadMultiple()` will return the default revisions and not the latest revisions, so there's a possibility for data loss when creating a new revision based on the result of `loadMultiple()`.  Again, not fully awake and I rarely am, so I might be missing something.

docroot/modules/custom/va_gov_backend/va_gov_backend.install
```diff
@@ -424,3 +425,96 @@ function va_gov_backend_update_8010() {
...
+
+  foreach ($nodes as $node) {
+    if ($node instanceof NodeInterface) {
+      // Save some time by only saving nodes that need adjustments.
+      if (
+        preg_match('/\s+$/', $node->getTitle())
+        ||
+        preg_match('/^\s+/', $node->getTitle())
```
ndouglas, line 467: Might use `str_starts_with()` and `str_ends_with()`, but that's probably just premature optimization on my part.

ndouglas (commented):
Looks great so far!

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
timcosgrove: Because I didn't think of it :) I may change.

I guess my only thought would be that one can use consistent regex patterns in both `preg_replace` and `preg_match`, which means it can be used in control structures consistently (`if (preg_match() {}`). But regex does make it marginally harder to read.

docroot/modules/custom/va_gov_backend/va_gov_backend.install
timcosgrove: If I'm reading [str_starts_with()](https://www.php.net/manual/en/function.str-starts-with.php) etc correctly, those match exact strings. We don't know how much whitespace we're dealing with here, so we need a pattern.

ndouglas: It's PHP 8 😩

docroot/modules/custom/va_gov_backend/va_gov_backend.install
timcosgrove: This is good point. I'll take a closer look at how this interacts with draft revisions of published nodes. I hadn't considered that.

If it keeps the draft revision in place though, and the editor can go back to it, then we're fine, because even if the draft revision still has the trailing spaces and that gets publishing, the presave event handler here will kick and make the correction again.

timcosgrove: @ndouglas so I did a bit of testing with this and this is what I found:

1. Working with database prior to running this update.
2. Took an existing node, and added spaces to the title, to recreate the situation we're trying to fix. Saved as published, and confirmed the spaces are intact.
3. Edited the node and made a change elsewhere in the node, and saved a draft. Observed that the published node is still the prior version, with the spaces in the title but without this draft edit. The draft edit is the lead revision, but not the default.
4. Ran the update script.
5. Looking at the node, the published version is now the lead, with the corrected title.
6. Go to the revisions page, and select "revert to" for the draft.

Result: published node is still the correct node. The draft can still be worked on. The "new" copy of the draft also has the corrected title, since it was saved.

So, I'm torn. On one hand, this creates work for editors that they might not expect, and it might require them to have a better understanding of how to interact with revisions. OTOH, the other way to do it (load and save the same revision) doesn't create a change record, which I think is problematic.

ndouglas: I saw this before your messages in Slack, but couldn't respond at the moment.  Having caught up in Slack, I think your solution's as good as it gets 😅

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -0,0 +1,51 @@
+<?php
```
ndouglas, line 1: Thanks for doing this the effort-intensive new best practices way and not just jumping to the hook implementation ❤️

ndouglas (approved):
Code looks great, very thorough, excellent tests, adheres to our best practices.  One suggestion I'd make is to keep the php extension on the script just for the sake of syntax highlighting.

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
timcosgrove: There's a hook implementation? 😃

timcosgrove: Oh there *is* a hook implementation. No thank you.

scripts/content/vacms-3573/NodeTitleWhitespaceTrimmer.php
```diff
@@ -0,0 +1,248 @@
+<?php
```
timcosgrove, line 1: Despite the conversation (and also how I feel about these sort of ad-hoc human-run scripts), I did implement this this way. I'd still like us to look into a more robust cron setup so that we can start using workers to keep things tidy on an ongoing basis, rather than running scripts manually.

docroot/modules/custom/va_gov_backend/va_gov_backend.services.yml
```diff
@@ -9,3 +9,7 @@ services:
   va_gov_backend.va_gov_url:
     class: Drupal\va_gov_backend\Service\VaGovUrl
     arguments: ['@http_client', '@settings', '@va_gov.build_trigger.environment_discovery']
+  va_gov_backend.entity_event_subscriber:
```
swirtSJW, line 12: minor:  Can we alphasort this with the other services?

scripts/content/vacms-3573/NodeTitleWhitespaceTrimmer.php
swirtSJW: Agreed.  Not much of a stub, but there is a stub here https://github.com/department-of-veterans-affairs/va.gov-cms/issues/3203

scripts/content/vacms-3573/NodeTitleWhitespaceTrimmer.php
```diff
@@ -0,0 +1,248 @@
...
+
+            // Set revision author to uid 1317 (CMS Migrator user).
+            $node->setRevisionUserId($this::CMS_MIGRATOR_ID);
+            $node->setChangedTime(time());
+            $node->setRevisionCreationTime(time());
+            $title_field_label = $node->getFieldDefinition('title')->getLabel();
+            $substitutions = [
+              '%title_field_label' => $title_field_label,
+              '%original_title' => $node->getTitle(),
+              '%new_title' => trim($node->getTitle()),
```
swirtSJW, line 125: minor: This reporting method differs from the method being used on save in the event subscriber (trim vs dual regex).  Three is a pretty minor risk but one is reporting something that the other is not doing.

swirtSJW (approved):
Tim this looks really good.  Nice thoughtful work.
The two minor things I raised are not required fixes.  (I typically call things out as I go then decide whether there is anything that is a must fix).

swirtSJW:
Can you run script on tugboat so we can see the log?
And maybe run it a second time to show there are no items to fix on the second run?

docroot/modules/custom/va_gov_backend/va_gov_backend.services.yml
timcosgrove: Good call; I'll update this.

scripts/content/vacms-3573/NodeTitleWhitespaceTrimmer.php
timcosgrove: No, this is a good call; it smelled funny to me as I wrote it. Will update.

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * Entity presave Event call.
+   *
+   * @param \Drupal\core_event_dispatcher\Event\Entity\EntityPresaveEvent $event
+   *   The event.
+   */
+  public function entityPresave(EntityPresaveEvent $event): void {
+    $entity = $event->getEntity();
+    if ($entity->getEntityTypeId() === 'node') {
```
indytechcook, line 23: `$enttiy typeOf NodeInterface`

scripts/content/vacms-3573/NodeTitleWhitespaceTrimmer.php
```diff
@@ -0,0 +1,250 @@
...
+    $this->nodeStorage = $nodeStorage;
+    $this->moderationInformation = $moderationInformation;
+    $this->lockService = $lockService;
+    $this->logger = $logger;
+  }
+
+  /**
+   * Runs the node update process.
+   */
+  public function run() {
```
indytechcook, line 85: Not for now, but I'd love to abstract this logic as we have done it a few times.

scripts/content/vacms-3573/NodeTitleWhitespaceTrimmer.php
```diff
@@ -0,0 +1,250 @@
...
+    $this->logMessage('Found %total_nodes nodes to process.', ['%total_nodes' => $nids_to_process_count]);
+
+    while (count($nids_to_process)) {
+      $this->logMessage('%remaining_count nodes left to process.', ['%remaining_count' => count($nids_to_process)]);
+      // Load entities.
+      $modified_nids = [];
+      $node_ids = array_slice($nids_to_process, 0, $query_limit, TRUE);
+      $nodes = $this->nodeStorage->loadMultiple($node_ids);
+      foreach ($nodes as $node) {
+        if ($node instanceof NodeInterface) {
```
indytechcook, line 100: Do we need to do this check since the query is only looking at nodes?

scripts/content/vacms-3573/NodeTitleWhitespaceTrimmer.php
```diff
@@ -0,0 +1,250 @@
...
+            &&
+            !$this->moderationInformation->hasPendingRevision($node)
+            &&
+            $this->lockNode($node->id())
+          ) {
+            // Make this change a new revision.
+            $node->setNewRevision(TRUE);
+
+            // Set revision author to uid 1317 (CMS Migrator user).
+            $node->setRevisionUserId($this::CMS_MIGRATOR_ID);
```
indytechcook, line 118:
```suggestion
            $node->setRevisionUserId(static::CMS_MIGRATOR_ID);
```

scripts/content/vacms-3573/NodeTitleWhitespaceTrimmer.php
```diff
@@ -0,0 +1,250 @@
...
+    print strtr($message, $context) . PHP_EOL;
+  }
+
+  /**
+   * Lock the specified node.
+   *
+   * @param int $nid
+   *   The node ID.
+   */
+  private function lockNode(int $nid) {
```
indytechcook, line 195:
```suggestion
  private function lockNode(int $nid) : bool {
```

scripts/content/vacms-3573/NodeTitleWhitespaceTrimmer.php
```diff
@@ -0,0 +1,250 @@
...
+  /**
+   * Callback function to concat node ids with string.
+   *
+   * @param int $nid
+   *   The node id.
+   *
+   * @return string
+   *   The node id concatenated to the end o node_
+   */
+  private function stringifynid($nid) {
```
indytechcook, line 237:
```suggestion
  private function stringifynid($nid) : string {
```

indytechcook (dismissed):
Small, unblocking comments

scripts/content/vacms-3573/NodeTitleWhitespaceTrimmer.php
timcosgrove: Me too, but I'd also like it to be run as a different mechanism. We'll get there.

scripts/content/vacms-3573/NodeTitleWhitespaceTrimmer.php
timcosgrove: This is here primarily to allow php scan tests to pass. `nodeStorage->loadMultiple()` returns `EntityInterface` objects, not `NodeInterface`, so all the node specific methods can't be found.

indytechcook: It's not a big deal but we could type hint it with `/** @var \Drupal\node\NodeInterface[] $nodes */`

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
timcosgrove: `$entity instanceof NodeInterface` in PHP :trollface: will do

timcosgrove:
Script running first time:

scripts/content/vacms-3573/NodeTitleWhitespaceTrimmer.php
timcosgrove: Done. Resolving.

swirtSJW (approved):
Logs look good to me.   Nice work Tim.

Issue #6363, opened 2021-09-08
Make staff profile bio and intro text field required, but only show them based on a boolean 


swirtSJW:
For refining in Sprint 41

## Description

A change was made to the FE that does not create  a staff profile (and a link to it) if the page has no "bio First sentence  text AND Bio.
![image](https://user-images.githubusercontent.com/5752113/132560039-b5fe61cc-41ea-41e7-ba73-47eebffb58ae.png)

https://github.com/department-of-veterans-affairs/va.gov-team/issues/29096

https://dsva.slack.com/archives/C01SR56755H/p1631029274004700 and PR >>> https://github.com/department-of-veterans-affairs/content-build/pull/501/f

The help text needs to be updated to indicate that is the case .   

## Acceptance Criteria

- [ ] Boolean field with label = "Create profile page with biography."
- [ ] Hide bio fields when boolean is unchecked
- [ ] Make first sentence and body text fields required

![image](https://user-images.githubusercontent.com/643678/133656015-34879a2f-266d-42a1-a083-279787390233.png)

## Implementation TBD 

Add boolean along the lines of "Create a page for this profile" which would then require First sentence and Body. 

### CMS Team

Please leave only the team that will do this work selected. If you're not sure, it's fine to leave both selected.

- [ ] `Core Application Team`
- [x] `Product Support Team`

rachel-kauff:
Nothing revolutionary to add here! My recommendation/validation of what's suggested in this ticket:
- Yes to the boolean
- Hide bio fields when boolean is unchecked
- Boolean label = "Create profile page with biography." 
- Make first sentence and body text fields required

<img width="802" alt="Screen Shot 2021-09-13 at 2 34 12 PM" src="https://user-images.githubusercontent.com/41017988/133159465-660f84e0-91d8-444b-a92d-728dc561ef5e.png">

PR #6503, opened 2021-09-22
VACMS-6363: Add bio fields toggle.

ethanteague:
## Description
Relates to #6363 

## Testing done
Visual / Behat

## Screenshots
image
image
image
image
image

## QA steps
 - [x] As an administrator, go to `/node/35031/edit` and visually verify that the `Create profile page with biography` boolean appears in the bio fieldset, and checking it on/off toggles the bio fieldset fields.
 - [x] Visually verify intro and body fields are required
 - [x] Empty both fields, submit the form, and visually verify the page doesn't save and error messages appear with jump links
 - [x] Add value to first sentence and verify the page does not save and provides immediate feedback. (HTML 5 error pre-save)
 - [x] Remove the text from the first sentence but add text to body field.  verify the page does not save and provides error at top of page, with a working ump link to the actual error. (field validation error after save)
 - [x] Add values to the fields, and confirm that the page saves without errors.
 - [x] Collapse the bios fieldset by unchecking the boolean, and confirm the form saves without errors.
 - [x] Empty the fields, collapse the bios fieldset by unchecking the boolean, and confirm the form saves without errors.

Go to terminal on tugboat
  - run `drush scr ./scripts/content/VACMS-6363-set-bio-booleans-that-have-content.php`
  - [x] validate something like  is returned (count may vary slightly)`Turned bio display boolean on in '702' nodes.`
  - [x]  validated a bunch of nodes were that were updated correctly
  ```
  https://pr6503-wmjbgqepacaadrsr6r83inmmsh4gso25.ci.cms.va.gov/node/35917/edit
https://pr6503-wmjbgqepacaadrsr6r83inmmsh4gso25.ci.cms.va.gov/node/35894/edit
https://pr6503-wmjbgqepacaadrsr6r83inmmsh4gso25.ci.cms.va.gov/node/35329/edit
https://pr6503-wmjbgqepacaadrsr6r83inmmsh4gso25.ci.cms.va.gov/node/33597/edit
https://pr6503-wmjbgqepacaadrsr6r83inmmsh4gso25.ci.cms.va.gov/node/33029/edit
https://pr6503-wmjbgqepacaadrsr6r83inmmsh4gso25.ci.cms.va.gov/node/19636/edit
https://pr6503-wmjbgqepacaadrsr6r83inmmsh4gso25.ci.cms.va.gov/node/18486/edit
https://pr6503-wmjbgqepacaadrsr6r83inmmsh4gso25.ci.cms.va.gov/node/17631/edit
  ```
- [x] validate revision log messages look correct 
- [x]  validate original node author is reflected as the author.


kevwalsh (changes requested):
The unhidden fields should be required  (AC buried in a comment, sorry)

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -110,6 +110,42 @@ public function alterFullWidthBannerNodeForm(FormIdAlterEvent $event): void {
     $form['#attached']['library'][] = 'va_gov_vamc/limit_vamcs_to_workbench';
   }
 
+  /**
```
swirtSJW, line 113: I don't think we want this in va_gov_vamc because person_profiles are not just a VAMC thing.  Any product can use them.

ethanteague: Makes sense. Where do you think it should live, @swirtSJW ?

swirtSJW: va_gov_backend is our more generic / multi product module.

ethanteague: Agreed, but I was thinking we were wanting to avoid putting much more into it. I guess it's alright since most of the bloat is in the .module file, and we are moving away from the old hook system and using event dispatcher hooks.

swirtSJW: It where it belongs.... so it is not like we are just walling that module off and not using it anymore.  We just want to be careful to check to see if there is a more appropriate spot.... which in this case there is not.

I dealy over time we will want to look at what is in that module that is specific to other products and move it out of there.

ethanteague:
Updated, @kevwalsh @swirtSJW

kevwalsh (changes requested):
@ethanteague @swirtSJW didn't think about this earlier, and this may blow the estimate out, but i think we need to do some database updates before this gets merged. Thinking about this as a new AC

- [ ] Existing staff profiles that have content in both  `field_body ` and `field_intro_text` have their  `field_complete_biography_create` populated as TRUE

ethanteague:
> @ethanteague @swirtSJW didn't think about this earlier, and this may blow the estimate out, but i think we need to do some database updates before this gets merged. Thinking about this as a new AC
> 
> * [ ]  Existing staff profiles that have content in both  `field_body ` and `field_intro_text` have their  `field_complete_biography_create` populated as TRUE

Good thinking, @kevwalsh - I'll work on this.

swirtSJW:
@kevwalsh  and @ethanteague  Hmmm I was thinking the update script was not needed, because this is a UI thing only.    BUT to make this truthful with the front end, we need the front end to actually use the value of this checkbox to decide not just the presence of the fields values the way it is now.

Otherwise, someone can fill out the fields but then uncheck the box and be confused that the bio page is still created.

ethanteague:
> @kevwalsh and @ethanteague Hmmm I was thinking the update script was not needed, because this is a UI thing only. BUT to make this truthful with the front end, we need the front end to actually use the value of this checkbox to decide not just the presence of the fields values the way it is now.
> 
> Otherwise, someone can fill out the fields but then uncheck the box and be confused that the bio page is still created.

Oh......I was thinking @kevwalsh 's concern was about the content entry experience on the node form, @swirtSJW . If an editor goes to update a staff profile, and there is content in the bio fields, they won't be aware of it.....granted, there aren't any other fields where they could enter bio type text (just some name / contact type fields).

Y'all just lemme know how you want me to proceed.

ethanteague:
Update script added to this pr.

scripts/content/VACMS-6363-set-bio-booleans-that-have-content.php
```diff
@@ -0,0 +1,44 @@
...
+    $node->setOwnerId(1317);
+    // Turn on our bio fieldset.
+    $node->get('field_complete_biography_create')->setValue(1);
+    // Set revision log message.
+    $node->setRevisionLogMessage('Resaved with new "Create profile page with biography" boolean checked.');
+    $node->save();
+    $count++;
+  }
+}
+Drupal::logger('va_gov_db')->log(LogLevel::INFO, 'Turned bio display boolean on for %count nodes in va_gov_db_update_8033.', [
```
swirtSJW, line 41: Logger not speaking truth.
```suggestion
Drupal::logger('va_gov_db')->log(LogLevel::INFO, 'Turned bio display boolean on for %count nodes via set-bio-booleans script.', [
```

swirtSJW (changes requested):
One minor change

scripts/content/VACMS-6363-set-bio-booleans-that-have-content.php
```diff
@@ -0,0 +1,44 @@
...
+  if ($node instanceof NodeInterface) {
+    $node->get('moderation_state')->setValue($current_moderation_status);
+    // Make this change a new revision.
+    $node->setNewRevision(TRUE);
+    // Set revision author to uid 1317 (CMS Migrator user).
+    $node->setRevisionUserId(1317);
+    $node->setChangedTime(time());
+    $node->isDefaultRevision(TRUE);
+    $node->setRevisionCreationTime(time());
+    $node->setOwnerId(1317);
```
swirtSJW, line 32: I know you got this from a previously used pattern but I think we are making a mistake setting the owner.
It makes it looks like cms migrator was the original author.

image

swirtSJW: I validated that  the display in that screen cap is indeed showing the owner (original author) not just the last person to save the revision.

swirtSJW: I updated the testing steps for checking the script run. and the bit about the author.

ethanteague: Thanks @swirtSJW :+1: - updated.

swirtSJW (approved):

swirtSJW:
We'll need these deploy steps run after it deploys today
- [ ] on prod, run `drush scr ./scripts/content/VACMS-6363-set-bio-booleans-that-have-content.php`
- [ ] Copy and paste terminal output in a comment on this PR.

ndouglas:
```bash
[cms@ip-10-247-33-17 cms]$ drush scr ./scripts/content/VACMS-6363-set-bio-booleans-that-have-content.php
Turned bio display boolean on in '714' nodes.[cms@ip-10-247-33-17 cms]$ 
```

ethanteague:
> ```bash
> 
> [cms@ip-10-247-33-17 cms]$ drush scr ./scripts/content/VACMS-6363-set-bio-booleans-that-have-content.php
> 
> Turned bio display boolean on in '714' nodes.[cms@ip-10-247-33-17 cms]$ 
> 
> ```

Thanks buddy!

PR #6538, opened 2021-09-24
VACMS-6363 bio updates hook

ethanteague:
## Description
Relates to #6363
**MERGE AFTER https://github.com/department-of-veterans-affairs/va.gov-cms/pull/6503 is deployed to prod.**
Update hook that turns on the bio fieldset for staff profiles that have both intro and body field content.

@kevwalsh @swirtSJW - I can do this via an update script that we can have the Platform team fire if you all prefer, but wanted to get in a pr to get the approach convo rolling

## Testing done
`admin/content` & dblog

## QA steps
 - [ ] As an administrator, go to `/admin/content` and review a handful of staff profiles that have been updated by CMS Migrator.
 - [ ] Visually verify that revision moderation status matches the moderation status of the last revision.
 - [ ] Go to `/admin/reports/dblog` and visually verify a message that states `Turned bio display boolean on for 697 nodes in va_gov_db_update_8033` exists.


swirtSJW:
@ethanteague  I think we do want this to go as part of 6503 as a script not a hook.  That way we minimize the window of time where the toggle is out of sync with the content intent.  Otherwise things are out of sync for 24 hrs.

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -646,3 +646,38 @@ function va_gov_db_update_8032(&$sandbox) {
...
+function va_gov_db_update_8033() {
+  $query = \Drupal::entityQuery('node')
+    ->condition('type', 'person_profile')
+    ->exists('field_intro_text')
+    ->exists('field_body');
+  $entity_ids = $query->execute();
+  $nodes = \Drupal::entityTypeManager()->getStorage('node')->loadMultiple($entity_ids);
+  $count = 0;
+  foreach ($nodes as $node) {
+    // Get current revision mod status to plugin to new revision.
```
swirtSJW, line 662: I think we want an inline var comment here declaring $node as a node interface which should make these phpstan errors go away.

docroot/modules/custom/va_gov_db/va_gov_db.install
```diff
@@ -646,3 +646,38 @@ function va_gov_db_update_8032(&$sandbox) {
...
+    $current_moderation_status = $node->get('moderation_state')->getString();
+    $node->get('moderation_state')->setValue($current_moderation_status);
+    // Make this change a new revision.
+    $node->setNewRevision(TRUE);
+    // Set revision author to uid 1317 (CMS Migrator user).
+    $node->setRevisionUserId(1317);
```
swirtSJW, line 668: minor: Since you are using this value in multiple places, set it as a var and re-use the var.

swirtSJW (changes requested):
The update hook on this looks good, but I think we want to move it to a script so we can run it after the deploy is complete.

Also do you know how many this actually operates on? (ballpark)

ethanteague:
> The update hook on this looks good, but I think we want to move it to a script so we can run it after the deploy is complete.
> 
> Also do you know how many this actually operates on? (ballpark)

6

> The update hook on this looks good, but I think we want to move it to a script so we can run it after the deploy is complete.
> 
> Also do you know how many this actually operates on? (ballpark)

697

Issue #6372, opened 2021-09-09
Remove "Nearby Vet Centers and Outstations" field and entity browser


kevwalsh:
## User Story or Problem Statement
See #6371 

## Acceptance Criteria
- [ ] Field has been removed
- [ ] Entity browser and any associated config (views) have been removed. 

## Implementation steps
Config

### CMS Team

Please leave only the team that will do this work selected. If you're not sure, it's fine to leave both selected.

- [ ] `Core Application Team`
- [x] `Product Support Team`

PR #6606, opened 2021-10-01
VACMS-6372: Replace nearby vcs field with fieldset filled with help text

ethanteague:
## Description
Relates to #6372

## Testing done
Visual / behat

## Screenshots
image

## QA steps
 - [ ] As an administrator, go to `/node/17949/edit` and visually verify presence of new `NEARBY VET CENTERS AND OUTSTATIONS` fieldset, replacing old
 - [ ] Text should match fieldset
 - [ ] First link should open in new window (preview link, so may not be available in tugboat).
 - [ ] Second link should open in same window, and lead to help desk form
 - [ ] Go to `admin/structure/types/manage/vet_center_locations_list/fields` and visually verify `Nearby Vet Centers and Outstations` field is not present
 - [ ] Go to `admin/config/content/entity_browser` and visually verify Vet center browser is not present
 - [ ] Go to `admin/structure/views/view/vet_centers` and visually verify Vet Center entity browser is not present, and only one display remains.


kevwalsh:
@rachel-kauff can you do a design review?

rachel-kauff:
@kevwalsh @ethanteague  Sorry if this is a little late in the game, but I think the "contact the CMS Helpdesk" link should also open in a new tab. I know for accessibility we should default to opening in the same window, but the big exception is forms, where navigating away from the page risks losing any changes the user may have made. This form is on the lighter side, but 'tis a form nonetheless.

ethanteague:
> @kevwalsh @ethanteague Sorry if this is a little late in the game, but I think the "contact the CMS Helpdesk" link should also open in a new tab. I know for accessibility we should default to opening in the same window, but the big exception is forms, where navigating away from the page risks losing any changes the user may have made. This form is on the lighter side, but 'tis a form nonetheless.
> 
> I also realize this prob should have been in the design spec. Do either of you have thoughts on this?

:100: @rachel-kauff

rachel-kauff:
Yay! One last thing @ethanteague, could the linked text also include the parenthetical (opens in a new tab)? Like so:

>  [contact the CMS Helpdesk (opens in a new tab)](https://va-gov.atlassian.net/servicedesk/customer/portal/3/group/8/create/26)

ethanteague:
Updated, @swirtSJW

tests/behat/drupal-spec-tool/content_model_vet_center_content_type_fields.feature
```diff
@@ -46,8 +46,8 @@ Feature: Content model: Vet Center Content Type fields
 | Content type | Vet Center - Locations List | Page introduction | field_intro_text | Text (plain, long) | Required | 1 | Text area (multiple rows) | Translatable |
 | Content type | Vet Center - Locations List | Display Additional Mobile Vet Centers | field_nearby_mobile_vet_centers | Entity reference |  | Unlimited | Entity Browser - Table |  |
 | Content type | Vet Center - Locations List | Enforce unique combo | field_enforce_unique_combo | Allow Only One |  | 1 | Allow Only One widget | Translatable |
-| Content type | Vet Center - Locations List | Nearby Vet Centers and Outstations | field_nearby_vet_centers | Entity reference |  | Unlimited | Entity Browser - Table |  |
 | Content type | Vet Center - Locations List | Vet Center | field_office | Entity reference | Required | 1 | Select list | Translatable |
+| Content type | Vet Center - Locations List | Nearby Vet Centers and Outstations | field_nearby_vet_centers | Entity reference |  | Unlimited | -- Disabled -- |  |
```
kevwalsh, line 50: The original AC is to remove the field, but i'm OK with splitting that out into a separate issue. Is there still a FE dependency on it?

swirtSJW (commented):
Code looks good.  But to Kev's question, why aren't we removing the field?

ethanteague:
> Code looks good. But to Kev's question, why aren't we removing the field?

It's still in the content-build graphql query.

swirtSJW (approved):
QA checks out except the removal part which has been moved to another ticket.

I did notice that our targeting the #id does not function because that id is not on the page when the page loads.  It comes in later.   That's an FE issue.
image

Issue #6602, opened 2021-10-01
CMS User Report Last Accessed in Fixed Time


chriscraiga6:
## User Story or Problem Statement

As a member of the helpdesk, I need the CMS user report to generate the last accessed date in a fixed time instead of relative so I can more efficiently manage user accounts.

## Acceptance Criteria
- [ ] When the CSV button is pressed, the output displays last accessed field in fixed time (Ex: 13-SEP-21 10:16 AM)

## Implementation steps
<optional>

## Design principles

Which [CMS design principle](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/cms/cms-product-design-principles.md) is at play?
- [ ] Purpose-driven: Creates an opportunity to involve the editor community in VA’s mission and content strategy goals.
- [x] Efficient: Removes distractions and create clear, straightforward paths to get the job done.
- [ ] Approachable: Offers friendly guidance over authoritative instruction.
- [ ] Consistent: Reduce user’s mental load by allowing them to fall back on pattern recognition to complete tasks.
- [x] Empowering: Provide clear information to help editors make decisions about their work.

### CMS Team

Please leave only the team that will do this work selected. If you're not sure, it's fine to leave both selected.

- [ ] `Platform CMS Team`
- [x] `Sitewide CMS Team`

PR #6610, opened 2021-10-02
VACMS-6602: Set login time from last to time.

ethanteague:
## Description
Relates to #6602

## Testing done
Visual / behat

## Screenshots
image

## QA steps
 - [ ] As an administrator, go to `admin/people` and visually verify last column reads `LAST ACCESS (USA NYC TIME)`
 - [ ] Visually verify last column login times are in `30-Sep-21 7:37pm` format
 - [ ] Download csv from link at bottom of page, and visually verify time format for last login is `30-Sep-21 7:37pm`

### DIFF
--- a/config/sync/views.view.user_admin_people.yml
+++ b/config/sync/views.view.user_admin_people.yml
@@ -676,11 +676,22 @@ display:
           hide_empty: false
           empty_zero: false
           hide_alter_empty: true
-          type: timestamp_ago
+          click_sort_column: value
+          type: timestamp
           settings:
-            future_format: '@interval hence'
-            past_format: '@interval ago'
-            granularity: 2
+            date_format: custom
+            custom_date_format: 'j-M-y g:ia'
+            timezone: ''
+          group_column: value
+          group_columns: {  }
+          group_rows: true
+          delta_limit: 0
+          delta_offset: 0
+          delta_reversed: false
+          delta_first_last: false
+          multi_type: separator
+          separator: ', '
+          field_api_classes: false
           plugin_id: field
           entity_type: user
           entity_field: access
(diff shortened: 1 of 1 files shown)


config/sync/views.view.user_admin_people.yml
```diff
@@ -634,7 +635,7 @@ display:
           relationship: none
           group_type: group
           admin_label: ''
-          label: 'Last access'
+          label: 'Last access (USA NYC time)'
```
swirtSJW, line 638: Are we sure of it being for NY?  Doesn't it use the users local time?

swirtSJW: Confirmed.  It does NOT use NY.   It uses time relative to logged in user viewing the View
image

Pacific time on the left, NY time on the right.

swirtSJW (changes requested):
We should take out the reference to the time zone on the header.

kevwalsh (changes requested):
This is great and has unblocked some of the user audit work. 

A few changes i'd like to request

- [x]  Revert to "time ago" dates for Last Access on the /admin/people page 
- [x]  In the CSV output
  - [x] switch to non-relative dates for both the Member Since AND the Last Access fields
  - [ ] Use the date format "October 5, 2021, 5:21pm ET" on the CSV
- [ ] What Steve said (revert to the old column field labels, in both the CSV and the view screen)
- [ ] Stretch: on the CSV output: add a new column with an "Edit link" column label, and output a raw link, eg, https://prod.cms.va.gov/user/UID/edit

ethanteague:
Updated @swirtSJW @kevwalsh :+1:

kevwalsh (approved):
Looks good to me. Sent to Chris and Stef and Jane for review.

swirtSJW (approved):
Looks great.  Thanks Ethan.
image


image

PR #6613, opened 2021-10-04
[docs] Facility migrations documentation to include settings.local.php

JayDarnell:
## Description

Updated facility migrations documentation to include details for the required settings.local.php file.


swirtSJW (approved):
Very thoughtful contribution Jay.  Thank you.

PR #6706, opened 2021-10-15
VACMS-6454: remove title from vet center migration - copy from field_official_name

### LINKED ISSUE #6454: Update Vet Center migration to only populate Official name, but not node title
ACs
- [ ] Update migration again to only populate field_official_name.

JayDarnell:
## Description

closes #6454 

## Testing done

Visual testing

## Screenshots

Checked the content page `/admin/content` and filtered by Content type = Vet Center to see how many items there are:

Screen Shot 2021-10-15 at 1 50 00 PM

From the terminal I then ran `lando drush mr va_node_facility_vet_centers` to wipe the migration map table

Next I ran the migration fresh with `lando drush migrate:import va_node_facility_vet_centers --update`

I then checked the content page `/admin/content` and filtered by Content type = Vet Center again to see the item count. Because the migration map table was wiped out we now have duplicates to test. I then searched for a specific Vet Center to be sure there were two copies and that the new one has a title:

Screen Shot 2021-10-15 at 1 58 34 PM

I then edited the new item's title and ran the migration again. To edit the title you need to make the field visible (It is labeled: Name of Facility) here: `/admin/structure/types/manage/vet_center/form-display`. You will also need to comment out the va_gov_consumers_form_alter function on line 14 of `docroot/modules/custom/va_gov_consumers/va_gov_consumers.module`. Once the migration finishes running I checked the edited Vet Center to make sure the page title change I just made wasn't reverted.

Screen Shot 2021-10-15 at 2 41 59 PM

## QA steps

As an admin user
1. [x] Check the content page `/admin/content` and filter by Content type = Vet Center to see how many items there are
2. [x] From the terminal run `lando drush mr va_node_facility_vet_centers` to wipe the migration map table
3. [x] From the terminal run `lando drush migrate:import va_node_facility_vet_centers --update`
4. [x] Visit `/admin/structure/types/manage/vet_center/form-display` and set the **Name of Facility** field to visible
5. [x] In your IDE comment out the va_gov_consumers_form_alter() function in va_gov_consumers.module
6. [x] Check the content page `/admin/content` and search for a specific Vet Center (you should see two entries)
7. [x] Edit the most recently migrated copy and change the **Name of Facility** (note the new name)
8. [x] From the terminal run `lando drush migrate:import va_node_facility_vet_centers --update` again
9. [x] Refresh the `/admin/content` page and filter by the new name you created in step 7

If the new name you created was not overwritten by the migration running again then you're good to go.


config/sync/core.entity_form_display.node.vet_center_locations_list.default.yml
```diff
@@ -112,6 +112,7 @@ content:
     third_party_settings: {  }
 hidden:
   created: true
+  field_nearby_vet_centers: true
```
swirtSJW, line 115: These were carryovers from a config change yesterday.  They are out of place, but fine and expected.

docroot/modules/custom/va_gov_migrate/va_gov_migrate.module
```diff
@@ -12,7 +12,15 @@ use Drupal\Core\Routing\RouteMatchInterface;
  * Implements hook_entity_presave().
  */
 function va_gov_migrate_entity_presave(EntityInterface $entity) {
-  _va_gov_migrate_process_va_form($entity);
+  if ($entity->bundle() == 'va_form') {
+    _va_gov_migrate_process_va_form($entity);
```
swirtSJW, line 16: _va_gov_migrate_process_va_form() is self checking (it has the bundle check in the function)  so there is no need for this to be wrapped in the bundle check.

docroot/modules/custom/va_gov_migrate/va_gov_migrate.module
```diff
@@ -22,10 +30,6 @@ function va_gov_migrate_entity_presave(EntityInterface $entity) {
  *   Entity.
  */
 function _va_gov_migrate_process_va_form(EntityInterface &$entity) {
-  if ($entity->bundle() !== 'va_form') {
```
swirtSJW, line 25: Lets keep this here so the function remains self checking.  It is the only thing that has to know what it is operating on.   That way no external safety checks are needed before anyone calls it.

docroot/modules/custom/va_gov_migrate/va_gov_migrate.module
```diff
@@ -12,7 +12,15 @@ use Drupal\Core\Routing\RouteMatchInterface;
  * Implements hook_entity_presave().
  */
 function va_gov_migrate_entity_presave(EntityInterface $entity) {
-  _va_gov_migrate_process_va_form($entity);
+  if ($entity->bundle() == 'va_form') {
+    _va_gov_migrate_process_va_form($entity);
+  }
+  elseif ($entity->bundle() == 'vet_center') {
```
swirtSJW, line 18: Lets move this check to inside (_va_gov_migrate_process_vet_center)   so it is self checking.

The self checking pattern has a couple benefits:
1. It keeps our hooks easy to read because there is not a lot of gatekeeper logic to read through.
2. It reduces the risk of someone calling the function to act on something it was not intended to act on.  If someone calls it inappropriately, no harm done.

docroot/modules/custom/va_gov_migrate/va_gov_migrate.module
```diff
@@ -12,7 +12,15 @@ use Drupal\Core\Routing\RouteMatchInterface;
...
+    _va_gov_migrate_process_vet_center($entity);
+  }
+  else {
```
swirtSJW, line 21: Even with the gatekeeper logic, I don't think we need this else.  hook_entity_presave does not have a return

docroot/modules/custom/va_gov_migrate/va_gov_migrate.module
```diff
@@ -50,6 +54,19 @@ function _va_gov_migrate_process_va_form(EntityInterface &$entity) {
   }
 }
 
+/**
+ * Process Vet Center.
+ *
+ * @param \Drupal\Core\Entity\EntityInterface $entity
+ *   Entity.
+ */
+function _va_gov_migrate_process_vet_center(EntityInterface &$entity) {
```
swirtSJW, line 63: On the va_form example you were copying, the word "process" was generic enough to cover all that was going on.  In this case this function is only doing one thing so it is better to make it speak its own truth. Something like `_va_gov_migrate_set_default_vet_center_title` speaks the truth and as a result is a lot more readable and understandable (clean)

In general your spidey sense should tingle whenever you are using generic verbs or nouns in your code.   Sometimes it is unavoidable but usually when it is used, it is a sure sign the "generic function" has too broad a "concern" and  should be split into smaller functions.

docroot/modules/custom/va_gov_migrate/va_gov_migrate.module
```diff
@@ -50,6 +54,19 @@ function _va_gov_migrate_process_va_form(EntityInterface &$entity) {
...
+  // If a Vet Center title is empty copy the value from field_official_name.
+  if (empty($entity->getTitle())) {
+    $entity->setTitle($entity->get('field_official_name')->value);
```
swirtSJW, line 66:
```suggestion
    $default_title = $entity->get('field_official_name')->value ?: '- none -';
    $entity->setTitle($default_title);
```

Just in case for some reason the migration brings in nothing for the official name.  It should not happen.... but I have learned to not trust the facility API... it sometimes goes wonky.  This at least protects against ending up with a titleless node.

swirtSJW (changes requested):
Thanks Jay.  I left some suggestions.  Let me know if you have any questions.

docroot/modules/custom/va_gov_migrate/va_gov_migrate.module
swirtSJW: Looks like this check still needs to be put back Jay.

JayDarnell: Strange, I thought for sure I had updated this. I'll check again.

JayDarnell: This is what my local shows...
Screen Shot 2021-10-20 at 1 31 23 PM

Issue #6449, opened 2021-09-16
Implement editorial experience design for Official name


kevwalsh:
See also #6955, which is higher priority, working on the same field. 

AC
- [ ] Implementing design from #6447
- [ ] PR must go through design review (with @EWashb or @suzanne-gray)
- [ ] PR must go through change management review (with @xiongjaneg) to determine if an announcement will be needed, and/or if KB article should be updated.

rachel-kauff:
@ethanteague I've wrapped up design thoughts in #6447 so hopefully that unblocks this one now!.

ethanteague:
> @ethanteague I've wrapped up design thoughts in #6447 so hopefully that unblocks this one now!.

TY @rachel-kauff !

PR #6961, opened 2021-11-12
VACMS-6955: Reveal title field for admins.

### LINKED ISSUE #6955: Unlock title field on vet centers
## Description

We need to change [Kailua-Kona Vet Center](https://prod.cms.va.gov/kailua-kona-vet-center) to Kona Vet Center, but we can't because the field is not editable, even by admins. 

This work could get folded into #6449, but is higher priority. 

Design_editorial_experience_for_Vet_Center_official_name_field_·_Issue__6447_·_department-of-veterans-affairs_va_gov-cms

</details>

## Acceptance Criteria
- [ ] Vet Center Node title field should be editable by admins, but not by non-admins. 
- [ ] Once this is on prod, unleash the request fulfillment process mentioned in the [Facilities sync notes from Nov 10](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/cms/product-team-support/product-team-weekly-syncs/facilities-cms-weekly-sync.md) 

### CMS Team

Please leave only the team that will do this work selected. If you're not sure, it's fine to leave both selected.

- [ ] `Platform CMS Team`
- [x] `Sitewide CMS Team`

ethanteague:
## Description
closes #6955

## Testing done
Visual / manual / behat

## Screenshots
image

## QA steps
 - [ ] As an administrator, go to `node/3840/edit` and visually verify presence of Official name field.
 - [ ] Change the value in Official name field, and confirm it is reflected in the page title on the view page and the node form.
 - [ ] Login as any user who is not an admin and has access to node 3840, and go to `node/3840/edit` and visually verify absence of the Official name field.

### DIFF
--- a/config/sync/node.type.vet_center.yml
+++ b/config/sync/node.type.vet_center.yml
@@ -3,6 +3,7 @@ langcode: en
 status: true
 dependencies:
   module:
+    - menu_force
     - menu_ui
     - node_revision_delete
     - node_title_help_text
@@ -16,6 +17,9 @@ third_party_settings:
     minimum_revisions_to_keep: 500
     minimum_age_to_delete: 0
     when_to_delete: 0
+  menu_force:
+    menu_force: 0
+    menu_force_parent: 0
 name: 'Vet Center'
 type: vet_center
 description: 'Location and page content for community-based counseling centers.'
--- a/config/sync/node.type.vet_center.yml
+++ b/config/sync/node.type.vet_center.yml
@@ -3,7 +3,6 @@ langcode: en
 status: true
 dependencies:
   module:
-    - menu_force
     - menu_ui
     - node_revision_delete
     - node_title_help_text
@@ -17,9 +16,6 @@ third_party_settings:
     minimum_revisions_to_keep: 500
     minimum_age_to_delete: 0
     when_to_delete: 0
-  menu_force:
-    menu_force: 0
-    menu_force_parent: 0
 name: 'Vet Center'
 type: vet_center
 description: 'Location and page content for community-based counseling centers.'
--- a/docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
+++ b/docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
@@ -84,11 +84,18 @@ function _vagov_consumers_disable_non_vha_facilities_api_form_fields(array &$for
       // Prepoulate the title to warn it will be overwritten by api data.
       $form['title']['widget'][0]['value']['#default_value'] = t('May be overwritten by Facility API Data');
     }
-    elseif (!Drupal::service('va_gov_user.user_perms')->hasAdminRole()) {
-      // Disable the title, it has content already, but allow
-      // admins discretion to change if necessary.
+    // Only allow admins to alter vc titles.
+    elseif (Drupal::service('va_gov_user.user_perms')->hasAdminRole() && in_array($form_id, [
+      'node_vet_center_edit_form',
+      'node_vet_center_form',
+    ])) {
+      $form['title']['widget'][0]['value']['#attributes']['disabled'] = FALSE;
+    }
+    // Lockdown titles for all roles for all other types.
+    else {
       $form['title']['widget'][0]['value']['#attributes']['disabled'] = TRUE;
     }
+
     // Disable the facility locator API field if it is populated because it
     // should not be altered, or it will deviate from the API.
     if (!empty($form['field_facility_locator_api_id']['widget'][0]['value']['#default_value'])) {
--- a/docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
+++ b/docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
@@ -84,8 +84,9 @@ function _vagov_consumers_disable_non_vha_facilities_api_form_fields(array &$for
       // Prepoulate the title to warn it will be overwritten by api data.
       $form['title']['widget'][0]['value']['#default_value'] = t('May be overwritten by Facility API Data');
     }
-    else {
-      // Disable the title, it has content already.
+    elseif (!Drupal::service('va_gov_user.user_perms')->hasAdminRole()) {
+      // Disable the title, it has content already, but allow
+      // admins discretion to change if necessary.
       $form['title']['widget'][0]['value']['#attributes']['disabled'] = TRUE;
     }
     // Disable the facility locator API field if it is populated because it
--- a/docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
+++ b/docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
@@ -58,7 +58,7 @@ class EntityEventSubscriber implements EventSubscriberInterface {
   public function unsetNameFieldForNonAdmins(FormIdAlterEvent $event): void {
     if (!$this->userPermsService->hasAdminRole()) {
       $form = &$event->getForm();
-      unset($form['field_official_name']);
+      $form['field_official_name']['#disabled'] = TRUE;
     }
   }
 
--- a/docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
+++ b/docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
@@ -10,6 +10,7 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
 use Drupal\Core\StringTranslation\TranslationInterface;
 use Drupal\hook_event_dispatcher\HookEventDispatcherInterface;
+use Drupal\va_gov_user\Service\UserPermsService;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
 /**
@@ -19,13 +20,23 @@ class EntityEventSubscriber implements EventSubscriberInterface {
   use StringTranslationTrait;
 
   /**
+   * UserPerms Service.
+   *
+   * @var \Drupal\va_gov_user\Service\UserPermsService
+   */
+  protected $userPermsService;
+
+  /**
    * Constructs a EntityEventSubscriber object.
    *
    * @param \Drupal\Core\StringTranslation\TranslationInterface $string_translation
    *   The string translation service.
+   * @param \Drupal\va_gov_user\Service\UserPermsService $user_perms_service
+   *   The string translation service.
    */
-  public function __construct(TranslationInterface $string_translation) {
+  public function __construct(TranslationInterface $string_translation, UserPermsService $user_perms_service) {
     $this->stringTranslation = $string_translation;
+    $this->userPermsService = $user_perms_service;
   }
 
   /**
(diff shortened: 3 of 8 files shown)


kevwalsh (changes requested):
This should remain hidden on the form until #6449 (next up!). This issue, #6955 should just be allowing admins to edit the node title (the "common name", not the "official name")

 Looks like this code will be helpful for #6449 so keep it around!

ethanteague:
> This should remain hidden on the form until #6449 (next up!). This issue, #6955 should just be allowing admins to edit the node title (the "common name", not the "official name")
> 
> Looks like this code will be helpful for #6449 so keep it around!

@kevwalsh looks like we are using the official name to populate the common name (node title) when a migration doesn't have a title for the center. I don't think we've ever used a drupal entity title for vc's.....have we? I didn't see it anywhere in our codebase.

kevwalsh:
@ethanteague looks like it's called "Name of faciliity", but wlil be renamed to "Common name" in #6449.  

From https://pr6961-clpzzkqy6xmupbqs2rshkm9dte4aw4p3.ci.cms.va.gov/admin/structure/types/manage/vet_center/form-display


Manage_form_display___VA_gov_CMS


</details>

We may want to go ahead and rename the title field now, and enable it for admins, hide it for others. 

From https://pr6961-clpzzkqy6xmupbqs2rshkm9dte4aw4p3.ci.cms.va.gov/admin/structure/types/manage/vet_center


Edit_Vet_Center_content_type___VA_gov_CMS

</details>


Others can currently see the node titile, but no one can edit it. 

Edit_Vet_Center_Kailua-Kona_Vet_Center___VA_gov_CMS



</details>

netboss (dismissed):
@ethanteague This looks good. The only one thing I might suggest is to highlight the page title text either with `<b>`/`<strong>` or maybe put it in `<h*>` tags, but I guess this depends on approved designs.

docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -38,6 +49,19 @@ public function alterVetCenterLocationsListNodeEditForm(FormIdAlterEvent $event)
...
+  /**
+   * Removes official name field on vc forms when user is non admin.
+   *
+   * @param \Drupal\core_event_dispatcher\Event\Form\FormIdAlterEvent $event
+   *   The event.
+   */
+  public function unsetNameFieldForNonAdmins(FormIdAlterEvent $event): void {
+    if (!$this->userPermsService->hasAdminRole()) {
+      $form = &$event->getForm();
+      unset($form['field_official_name']);
```
swirtSJW, line 61: Are you sure we don't want to make this disabled?   

The AC does not mention hidden or removal.
image

docroot/modules/custom/va_gov_vet_center/va_gov_vet_center.services.yml
```diff
@@ -1,6 +1,6 @@
 services:
   va_gov_vet_center.entity_event_subscriber:
     class: Drupal\va_gov_vet_center\EventSubscriber\EntityEventSubscriber
-    arguments: ['@string_translation', '@url_generator']
+    arguments: ['@string_translation', '@va_gov_user.user_perms']
```
swirtSJW, line 4: Weird.... we were passing in an arg that was not being caught... I would have expected some php warnings over that one.  Good catch.

swirtSJW (commented):
I had some questions about this approach.  It does not seem to match the AC's.   There will be more coming later related to conditional visibility but I don't think unsetting the title field from the form is the interim approach we want to take.

docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
ethanteague: Yeah, I'll make it disabled.

ethanteague:
Updated

config/sync/node.type.vet_center.yml
```diff
@@ -3,6 +3,7 @@ langcode: en
 status: true
 dependencies:
   module:
+    - menu_force
```
swirtSJW, line 6: Confused  about this addition.   It seems completely unrelated to revealing the title field?

docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
```diff
@@ -84,8 +84,9 @@ function _vagov_consumers_disable_non_vha_facilities_api_form_fields(array &$for
       // Prepoulate the title to warn it will be overwritten by api data.
       $form['title']['widget'][0]['value']['#default_value'] = t('May be overwritten by Facility API Data');
     }
-    else {
-      // Disable the title, it has content already.
+    elseif (!Drupal::service('va_gov_user.user_perms')->hasAdminRole()) {
```
swirtSJW, line 87: This should apply only to vet centers.  By putting it here is seems like it is applying to all non vha facilities.

swirtSJW (commented):
Seeing a couple changes I am not following the need for:

config/sync/node.type.vet_center.yml
ethanteague: Yeah, I noticed this, and wondered if maybe it held some necessary magic. No biggie to pull out (doesn't break anything when I test locally).

docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
ethanteague: Wouldn't we want to allow for admins to have the same level of control over api populated titles for temp / on the fly stuff across all types? I guess I don't understand why only vc's? I'm changing....just letting you know my thought process.

swirtSJW: Because vet center titles are  not populated by the facility API any more.  They have the new Official name field which gets populated by the migration.   The node title is now considered the plain language name and we have to be able to edit it (via change management request when needed)

Now that being said.  All the facility api fields should be unlocked for admins and are currently across the VAMC facilities.
image
They are also unlocked for vetcenters too
image


The problem is, the title is not located in that field group so is missing the logic to unlock it.  It somehow got a snowflake treatment that put it someplace without it being treated as a field 
image


It probably didn't need a special case treatment for dissabling if it was just moved back the facilityAPI box, but now it actually would be mislieading to put in faclity API box since it no longer is controlled by that.

swirtSJW (changes requested):
I don't think this logic is working correctly.
I logged in as me and could edit... which is good
image


But then I logged in as a vetcenter editor Sharran.Chambers-Murphy@va.gov (not content admin)
image
 
and I could still edit the title field (which is bad).
image

docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
ethanteague: > I don't think this logic is working correctly. I logged in as me and could edit... which is good image
> 
> But then I logged in as a vetcenter editor [Sharran.Chambers-Murphy@va.gov](mailto:Sharran.Chambers-Murphy@va.gov) (not content admin) image
> 
> and I could still edit the title field (which is bad). image

Dangit. Sorry about that, @swirtSJW - updated.

docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -38,6 +49,19 @@ public function alterVetCenterLocationsListNodeEditForm(FormIdAlterEvent $event)
     $this->buildNearbyVCandOutStationFieldsetContent($event);
   }
 
+  /**
+   * Removes official name field on vc forms when user is non admin.
+   *
+   * @param \Drupal\core_event_dispatcher\Event\Form\FormIdAlterEvent $event
+   *   The event.
+   */
+  public function unsetNameFieldForNonAdmins(FormIdAlterEvent $event): void {
```
swirtSJW, line 58: Name and comment don
t agree with what is happening here.
**doc**: remove
**function**: unset
**code**: disable

Only the code is speaking the truth.  The other two are sowing seeds of confusion.

docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
```diff
@@ -84,10 +84,21 @@ function _vagov_consumers_disable_non_vha_facilities_api_form_fields(array &$for
...
-      // Disable the title, it has content already.
+    // Only allow admins to alter vc titles.
+    elseif (!Drupal::service('va_gov_user.user_perms')->hasAdminRole() && in_array($form_id, [
+      'node_vet_center_edit_form',
+      'node_vet_center_form',
+    ])) {
       $form['title']['widget'][0]['value']['#attributes']['disabled'] = TRUE;
     }
+    // Lockdown titles for all roles for all other types.
+    elseif (Drupal::service('va_gov_user.user_perms')->hasAdminRole() && !in_array($form_id, [
```
swirtSJW, line 95: This comment does not agree with what is happening here.  it says for all roles, but then it is only locking it down if the user as an admin role.    Also does not make sense to lock it down for only admins.

In the larger picture, something seems odd here.  You replaced a single else (generic catch-all) with two very specific else_ifs.  There is no longer a catch-all.  I think you are broadly impacting a lot with this change.
I could see an addition of a single else_if... but I would still expect it to end with an else.

docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -218,8 +242,9 @@ public function requireRevisionMessage(array &$form, FormStateInterface &$form_s
   public static function getSubscribedEvents(): array {
     return [
       HookEventDispatcherInterface::FORM_ALTER => 'formAlter',
-      // React on Vet center locations list edit form.
       'hook_event_dispatcher.form_node_vet_center_locations_list_edit_form.alter' => 'alterVetCenterLocationsListNodeEditForm',
+      'hook_event_dispatcher.form_node_vet_center_edit_form.alter' => 'unsetNameFieldForNonAdmins',
+      'hook_event_dispatcher.form_node_vet_center_form.alter' => 'unsetNameFieldForNonAdmins',
```
swirtSJW, line 247: This pattern prevents these form.alter from doing anything else.   They can only do `unsetNameFieldForNonAdmins`
A better pattern would be something like
```suggestion
      'hook_event_dispatcher.form_node_vet_center_form.alter' => vetcenterFormAlter',
```

Then inside that function it would call `unsetNameFieldForNonAdmins` and any other methods we need to call.

swirtSJW (changes requested):

docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
ethanteague: image

docroot/modules/custom/va_gov_consumers/va_gov_consumers.module
ethanteague: Yeah, I got too cute on this one - cleaned it up, @swirtSJW :+1:

docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
ethanteague: Yep, agreed. Got lazy - updated, @swirtSJW :+1:

ethanteague:
@kevwalsh - I agree with you that this work has been subsumed by https://github.com/department-of-veterans-affairs/va.gov-cms/pull/7045. Closing this one down. 
cc: @swirtSJW

Issue #6941, opened 2021-11-09
VAMC banner alerts should only be placeable on VAMC systems that user has section access to


kevwalsh:
## User Story or Problem Statement

It is currently possible for VAMC editors to place banners on VAMC systems where they are not section members.

VAMC system editors have to look through a long list of systems before finding their own. 

<details><summary>Screenshot</summary>

<img width="1274" alt="Create_VAMC_System_Banner_Alert_with_Situation_Updates___VA_gov_CMS" src="https://user-images.githubusercontent.com/643678/140859155-bf5254de-800d-4b62-8d77-bf1ffe5650d6.png">

</details>
## Acceptance Criteria

On /node/add/full_width_banner_alert or /node/6922/edit, in the  "Where will this banner appear?" field
- [ ] As user `ryan.stubblebine@va.gov`  i should only see "VA Pittsburgh health care" 
- [ ] As user `louis.scavnicky@va.gov` i should only be able place VAMCs in VISN 4 (VA Pittsburgh, VA Erie, VA Butler, VA Philadelphia, etc). 

## Implementation steps
- Must the Section be selected first for anything to appear in the list of available VAMC systems?

## Design principles

Veteran-centered
- [ ] `Single source of truth`: Increase reliability and consistency of content on VA.gov by providing a single source of truth.
- [ ] `Accessible, plain language`: Provide guardrails and guidelines to ensure content quality.
- [ ] `Purposely structured content`: Ensure Content API can deliver content whose meaning matches its structure.
- [x] `Content lifecycle governance`: Produce tools, processes and policies to maintain content quality throughout its lifecycle.

Editor-centered
- [ ] `Purpose-driven`: Create an opportunity to involve the editor community in VA’s mission and content strategy goals.
- [x] `Efficient`: Remove distractions and create clear, straightforward paths to get the job done.
- [ ] `Approachable`: Offer friendly guidance over authoritative instruction.
- [x] `Consistent`: Reduce user’s mental load by allowing them to fall back on pattern recognition to complete tasks.
- [ ] `Empowering`: Provide clear information to help editors make decisions about their work.

### CMS Team

Please leave only the team that will do this work selected. If you're not sure, it's fine to leave both selected.

- [ ] `Platform CMS Team`
- [x] `Sitewide CMS Team`

PR #6987, opened 2021-11-17
VACMS-6941: Winnow list of VAMCs on VAMC banner form to those user has access to.

ethanteague:
## Description
Relates to #6941 

## Testing done
Visual

## QA steps
 - [ ] As user `Ryan.Stubblebine@va.gov` go to `node/add/full_width_banner_alert` and visually verify systems checkboxes only shows Pittsburgh.


kevwalsh (approved):
QA pass!  I tested as jorge.cardich-fuster@va.gov who has access to multiple VAMCs, and created a little followup to sort whatever is listed alphabetically, in #6995 

image

</details>

swirtSJW:
Conflicted

ethanteague:
> I think we need some more QA stesp around testing what happends if someone with access to multiple creates it, then a person with acces to only one, edits it and saves it. We want to make sure we do not lose the original systems that the second editor does not have access to.

Good call, @swirtSJW - your suggestion resulted in a fix :+1:

swirtSJW (changes requested):
image

I had to add Jorge to pittsburgh health care.
image
Added test.
As admin I could see and edit all.
image
As RS I could see and edit my checkbox as well as remove it and re-add it (as long as the alert was still in my section)

swirtSJW (approved):
Sorry selected the wrong option.  This is approved.

Issue #6334, opened 2021-09-02
Drupal errors not logging to sentry


indytechcook:
## Description

Some Drupal / PHP errors are not sending to sentry

Exampels:

```
	RuntimeException: Failed to start the session because headers have already been sent by "/var/www/cms/docroot/vendor/symfony/http-foundation/Response.php" at line 1224. in Symfony\Component\HttpFoundation\Session\Storage\NativeSessionStorage->start() (line 152 of /var/www/cms/docroot/vendor/symfony/http-foundation/Session/Storage/NativeSessionStorage.php)
#0 /var/www/cms/docroot/core/lib/Drupal/Core/Session/SessionManager.php(162): Symfony\Component\HttpFoundation\Session\Storage\NativeSessionStorage->start()
#1 /var/www/cms/docroot/core/lib/Drupal/Core/Session/SessionManager.php(193): Drupal\Core\Session\SessionManager->startNow()
#2 /var/www/cms/docroot/vendor/symfony/http-foundation/Session/Session.php(189): Drupal\Core\Session\SessionManager->save()
#3 /var/www/cms/docroot/core/lib/Drupal/Core/StackMiddleware/Session.php(60): Symfony\Component\HttpFoundation\Session\Session->save()
#4 /var/www/cms/docroot/core/lib/Drupal/Core/StackMiddleware/KernelPreHandle.php(47): Drupal\Core\StackMiddleware\Session->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#5 /var/www/cms/docroot/core/modules/page_cache/src/StackMiddleware/PageCache.php(106): Drupal\Core\StackMiddleware\KernelPreHandle->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#6 /var/www/cms/docroot/core/modules/page_cache/src/StackMiddleware/PageCache.php(85): Drupal\page_cache\StackMiddleware\PageCache->pass(Object(Symfony\Component\HttpFoundation\Request), 1, true)
#7 /var/www/cms/docroot/core/lib/Drupal/Core/StackMiddleware/ReverseProxyMiddleware.php(47): Drupal\page_cache\StackMiddleware\PageCache->handle(Object(Symfony\Component\HttpFoundation\Request), 1, true)
[...]

PR #7057, opened 2021-11-24
VACMS-6334 Add test page to log different types of errors [DO NOT UPDATE]

indytechcook:
Just a PR to find which error types are not making it to sentry.

#6334

### DIFF
--- a/docroot/modules/custom/va_gov_backend/src/Controller/ErrorTestController.php
+++ b/docroot/modules/custom/va_gov_backend/src/Controller/ErrorTestController.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Drupal\va_gov_backend\Controller;
+
+use Drupal\Core\Controller\ControllerBase;
+use Drupal\Core\Logger\LoggerChannelFactoryInterface;
+
+class ErrorTestController extends ControllerBase {
+
+  /**
+   * The logger service.
+   *
+   * @var \Drupal\Core\Logger\LoggerChannelInterface
+   */
+  protected $logger;
+
+  public function __construct(LoggerChannelFactoryInterface $logger_factory) {
+    $this->logger = $logger_factory->get('va_sentry_test');
+  }
+
+  public function errorTestPage($errorType) : array {
+
+    switch ($errorType) {
+      case 'uncaught':
+        throw new \Exception('Uncaught Exception');
+        break;
+
+      case 'error':
+        trigger_error('Trigger an E_ERROR manually', E_USER_ERROR);
+        break;
+
+      case 'caught-nothing':
+        try {
+          throw new \Exception('Caught but not logged');
+        } catch (\Exception $e) {
+          // do nothing
+        }
+        break;
+
+      case 'caught-watchdog-exception':
+        try {
+          throw new \Exception('Caught and sent to watchdog exception');
+        } catch (\Exception $e) {
+          watchdog_exception('va_sentry_test', $e);
+        }
+        break;
+
+      case 'caught-and-logged':
+        try {
+          throw new \Exception('Caught and sent to logger');
+        } catch (\Exception $e) {
+          $this->logger->error('Caught and sent to logger class');
+        }
+    }
+
+    return ['#markup' => 'Look in Sentry'];
+  }
+
+}
--- a/docroot/modules/custom/va_gov_backend/src/Controller/ErrorTestController.php
+++ b/docroot/modules/custom/va_gov_backend/src/Controller/ErrorTestController.php
@@ -3,27 +3,21 @@
 namespace Drupal\va_gov_backend\Controller;
 
 use Drupal\Core\Controller\ControllerBase;
-use Drupal\Core\Logger\LoggerChannelFactoryInterface;
 
+/**
+ * Controller class for testing errors.
+ */
 class ErrorTestController extends ControllerBase {
 
   /**
-   * The logger service.
-   *
-   * @var \Drupal\Core\Logger\LoggerChannelInterface
+   * {@inheritdoc}
    */
-  protected $logger;
-
-  public function __construct(LoggerChannelFactoryInterface $logger_factory) {
-    $this->logger = $logger_factory->get('va_sentry_test');
-  }
-
   public function errorTestPage($errorType) : array {
 
     switch ($errorType) {
+
       case 'uncaught':
         throw new \Exception('Uncaught Exception');
-        break;
 
       case 'error':
         trigger_error('Trigger an E_ERROR manually', E_USER_ERROR);
@@ -32,15 +26,17 @@ class ErrorTestController extends ControllerBase {
       case 'caught-nothing':
         try {
           throw new \Exception('Caught but not logged');
-        } catch (\Exception $e) {
-          // do nothing
+        }
+        catch (\Exception $e) {
+          // Do nothing.
         }
         break;
 
       case 'caught-watchdog-exception':
         try {
           throw new \Exception('Caught and sent to watchdog exception');
-        } catch (\Exception $e) {
+        }
+        catch (\Exception $e) {
           watchdog_exception('va_sentry_test', $e);
         }
         break;
@@ -48,9 +44,11 @@ class ErrorTestController extends ControllerBase {
       case 'caught-and-logged':
         try {
           throw new \Exception('Caught and sent to logger');
-        } catch (\Exception $e) {
-          $this->logger->error('Caught and sent to logger class');
         }
+        catch (\Exception $e) {
+          $this->getLogger('va_sentry_test')->error('Caught and sent to logger class');
+        }
+
     }
 
     return ['#markup' => 'Look in Sentry'];
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.routing.yml
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.routing.yml
@@ -15,3 +15,14 @@ va_gov_backend.content_release_status:
     _title: 'Last Content Release'
   requirements:
     _permission: 'access content'
+
+va_gov_backend.error_test:
+  path: '/admin/config/development/logging/test/{errorType}'
+  defaults:
+    _controller: '\Drupal\va_gov_backend\Controller\ErrorTestController::errorTestPage'
+    _title: 'Sentry Testing'
+  requirements:
+    _permission: 'administer site configuration'
+  options:
+    _admin_route: true
+


ndouglas:
- `uncaught` 
Screen Shot 2021-11-29 at 11 08 43 AM

- `error`
Screen Shot 2021-11-29 at 11 08 21 AM

- `caught-nothing` – nothing, as expected

- `caught-watchdog-exception`
Screen Shot 2021-11-29 at 11 07 23 AM

- `caught-and-logged`
Screen Shot 2021-11-29 at 11 06 42 AM

So within this specific set of circumstances, everything appears to be functioning as expected.  

@indytechcook mentioned that the following didn't appear in Sentry, but I saw it:

Screen Shot 2021-11-29 at 10 48 51 AM

I revisited my `drush eval` diddlings from the other day:

```bash
drush eval 'throw new \Exception("nd-Uncaught Exception");'
```

Screen Shot 2021-11-29 at 11 13 12 AM

```bash
drush eval 'trigger_error("nd-Trigger an E_ERROR manually", E_USER_ERROR);'
```

**(diddly -- but this should make it to Sentry)**

```bash
drush eval 'try { throw new \Exception("nd-Caught but not logged"); } catch (\Exception $e) { }'
```

(diddly, as expected)

```bash
drush eval 'try { throw new \Exception("nd-Caught and sent to watchdog exception"); } catch (\Exception $e) { watchdog_exception("va_sentry_test", $e); }'
```

Screen Shot 2021-11-29 at 11 17 43 AM

```bash
drush eval 'try { throw new \Exception("nd-Caught and sent to logger"); } catch (\Exception $e) { \Drupal::logger("va_sentry_test")->error("nd-Caught and sent to logger class"); }'
```

Screen Shot 2021-11-29 at 11 20 26 AM

So, weirdly, `drush eval 'throw new \Exception("nd-Uncaught Exception");'` is working today on Tugboat but didn't appear to a couple of weeks ago on CMS-Test prod.  I'm going to do some checks to see if this is consistent or if there might be a random issue.

ndouglas:
I checked and verified that CMS-Test Staging (updated yesterday) matches the Tugboat environment in behavior for the `drush` commands above (it doesn't have the controller installed, natch).  All seem to function as expected except for triggered errors.

Now I'm trying to trigger specific errors that have been identified as not making it to Sentry.

I triggered a RuntimeException via Drush on CMS-Test Staging:

Screen Shot 2021-12-02 at 8 38 24 AM

I added `<html>` before the opening PHP tag in `settings.php`, and:

Screen Shot 2021-12-02 at 8 36 29 AM

So it would appear that any issue in delivery of these specific exceptions is not tied to the source or specific type of exception.

Throwing the SimpleSaml exception through two separate methods worked as well...

- directly
```bash
drush eval 'throw new \Drupal\simplesamlphp_auth\Exception\SimplesamlphpAttributeException(sprintf("Error in simplesamlphp_auth.module: no valid \"%s\" attribute set.", "adUPN"))'
```

- more closely simulating the original traceback:
```bash
drush eval '\Drupal::service("simplesamlphp_auth.manager")->getAttribute("adUPN");'
```

Screen Shot 2021-12-02 at 8 50 44 AM

I've been trying to diagnose the failure in things like the eval'ed call to `trigger_error()`, and I think it comes down to how Drush works.  Drush uses its own error handler (see docroot/vendor/drush/drush/src/Runtime/ErrorHandler.php), which will then get replaced by Drupal's error handler.  Drush's error handler does not itself emit any error events, which means that Raven apparently won't observe error events emitted at certain points in command execution.

So that explains _part_ of the puzzle, namely the part that I can reproduce.  However, it doesn't explain the parts that I can't reproduce, or have experienced only intermittently, which are those issues referred to in the original issue.

My current hypothesis is that outbound requests to Sentry fail under certain circumstances, e.g. because the body is too large or because Drupal fails while trying to construct the body.  That's all too easy to do with Drupal, given circular dependencies in the object graphs and so on, and our options currently are pretty verbose:

Screen Shot 2021-12-02 at 11 06 37 AM

There's nothing there to suggest that the backtrace includes parameters (something that's often a kiss of death when trying to debug an error in Drupal).  So I gotta keep digging.

indytechcook:
> My current hypothesis is that outbound requests to Sentry fail under certain circumstances, e.g. because the body is too large or because Drupal fails while trying to construct the body. That's all too easy to do with Drupal, given circular dependencies in the object graphs and so on, and our options currently are pretty verbose:

Maybe use `hook_raven_filter_alter` to limit the size?

ndouglas:
Looking at the backtrace of a couple of these errors, I see that they happen really early in the Drupal lifecycle:

```
[Thu Dec 02 18:06:33.174251 2021] [php7:notice] [pid 26290] [client 10.247.33.23:21812] RuntimeException: Failed to start the session because headers have already been sent by "/var/www/cms/docroot/vendor/symfony/http-foundation/Response.php" at line 1224. 
	in /var/www/cms/docroot/vendor/symfony/http-foundation/Session/Storage/NativeSessionStorage.php on line 152 
#0 /var/www/cms/docroot/core/lib/Drupal/Core/Session/SessionManager.php(162): Symfony\\Component\\HttpFoundation\\Session\\Storage\\NativeSessionStorage->start()\n
#1 /var/www/cms/docroot/core/lib/Drupal/Core/Session/SessionManager.php(193): Drupal\\Core\\Session\\SessionManager->startNow()\n
#2 /var/www/cms/docroot/vendor/symfony/http-foundation/Session/Session.php(189): Drupal\\Core\\Session\\SessionManager->save()\n
#3 /var/www/cms/docroot/core/lib/Drupal/Core/StackMiddleware/Session.php(60): Symfony\\Component\\HttpFoundation\\Session\\Session->save()\n
#4 /var/www/cms/docroot/core/lib/Drupal/Core/StackMiddleware/KernelPreHandle.php(47): Drupal\\Core\\StackMiddleware\\Session->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)\n
#5 /var/www/cms/docroot/core/modules/page_cache/src/StackMiddleware/PageCache.php(106): Drupal\\Core\\StackMiddleware\\KernelPreHandle->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)\n
#6 /var/www/cms/docroot/core/modules/page_cache/src/StackMiddleware/PageCache.php(85): Drupal\\page_cache\\StackMiddleware\\PageCache->pass(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)\n
#7 /var/www/cms/docroot/core/lib/Drupal/Core/StackMiddleware/ReverseProxyMiddleware.php(47): Drupal\\page_cache\\StackMiddleware\\PageCache->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)\n
#8 /var/www/cms/docroot/core/lib/Drupal/Core/StackMiddleware/NegotiationMiddleware.php(52): Drupal\\Core\\StackMiddleware\\ReverseProxyMiddleware->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)\n
#9 /var/www/cms/docroot/vendor/stack/builder/src/Stack/StackedHttpKernel.php(23): Drupal\\Core\\StackMiddleware\\NegotiationMiddleware->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)\n
#10 /var/www/cms/docroot/core/lib/Drupal/Core/DrupalKernel.php(717): Stack\\StackedHttpKernel->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)\n
#11 /var/www/cms/docroot/index.php(19): Drupal\\Core\\DrupalKernel->handle(Object(Symfony\\Component\\HttpFoundation\\Request))\n
#12 {main}, referer: https://prod.cms.va.gov/user/2147
```

Looking at the code in the neighborhood of DrupalKernel line 17, `handle()`:

Screen Shot 2021-12-02 at 3 04 28 PM

So we see that `bootEnvironment()` has already been called.  That, among other things, registers the Drupal error and exception handlers:

Screen Shot 2021-12-02 at 3 06 54 PM

And prior to line 717, where `handle()` is called and we continue up the callstack, we see that `boot()` is called; within that method, the service container is constructed.  So we would assume that Raven is able to process and forward this information at that point.

And yet, these errors are not making it to Sentry.

So I think I'm going to travel down that callstack tomorrow and try to find where the tipping point is (if there is one) where errors fail to be reported and then are successfully reported.

ndouglas:
That didn't produce any useful insights.

I walked all the way up the above backtrace, inserting exceptions, but each and every one was reported.  I finally threw an exception at line 151 of `docroot/vendor/symfony/http-foundation/Session/Storage/NativeSessionStorage.php`, but it was reported successfully:

Screen Shot 2021-12-03 at 9 11 43 AM

(This included truncating the sessions table, clearing cache, etc.)

So I think my hypothesis that this was something that could happen at certain phases in the request lifecycle is scuttled.  Or at least I can't reproduce it.

I was hoping that the 502s in Sentry might coincide with the errors in the Drupal log, but no, it doesn't even appear close.

So this is going to take more time.  I think I might need to fallback to doing some sort of reporting so I can see where exactly in the Sentry process this is failing, but that's going to take some doing.

ndouglas:
Things I ruled out today:

- supplying parameter arguments.  Even passing a $node, etc gets processed successfully.
- issues with the other options -- I disabled a couple on prod to see if I could catch a change in behavior but nothing has happened yet.
- a specific path through the Drupal error handler -- I tried a few combinations and all of them showed up on Sentry
- I _thought_ for a second I could reproduce this locally/Tugboat because I noticed that the editors experiencing one kind of error message were one-section editors, so I got really excited and logged in as a one-section editor on Tugboat.  The exception _was_ thrown, but it showed up on Sentry.  I'll look around for a couple other places where we do a similar short-circuit -- it's possible that we're breaking Drupal's lifecycle and therefore the Raven logger isn't flush()ing its cache.

Things I know:
- All errors that _don't_ seem to be showing up in Sentry appear as `php` in DBLog.  This might be necessary to cause the Sentry failure, but it is not sufficient.
- The two I see most often with this problem are SimpleSaml adUPN exceptions and RuntimeExceptions from headers being sent twice.  I've not seen a SimpleSaml adUPN exception make it to Sentry except the one I attempted to trigger manually.  I've triggered plenty of RuntimeExceptions through different methods and they've all made it.
- I've seen a MySQL deadlock error make it to Sentry.
- I think every other error I've seen _not_ make it to Sentry, I've seen make it to Sentry in other cases.

I'm going to look deeper into where we might be causing the RuntimeException for sending headers twice and learn a bit more about SimpleSaml and see if I can trigger a more authentic exception.  It's tough, though, since none of the SimpleSaml errors provide a backtrace!

ndouglas:
I was able to reproduce this locally, finally, by inserting `\Drupal::service('plugin.manager.field.widget')->createInstance('hidden');` in `va_gov_backend_entity_presave()`.  I shoved a lot of debugging code into the various error handler functions, etc, and verified that at some point the exception just disappeared without getting processed correctly.  I followed it into the actual Raven logger plugin, where I found that processing the exception context seemed to make the exception not get reported.  I disabled "reflection tracing in stacktraces" and the exception was correctly delivered to Sentry.

However, this doesn't help, since I'd realized already that it might be a problem and disabled it on production.  However, SimpleSaml errors keep occurring on production and not getting reported to Sentry, so it appears that I've uncovered _one_ failure mode but not all of them.

ndouglas:
After a couple of hours of awesome, informative debugging with @ElijahLynn and @cweagans, it appears that the issue might be due to a problem with reading the `SENTRY_DSN` environment variable.  Intentionally populating the corresponding config field results in sending at least another class of errors to Sentry, at least in testing with Staging.  I've explicitly set that value in the production DB, but if it works, we will need to figure out a permanent solution.  It's arguable as to whether this value is a secret, but currently all VFS teams seem to be treating it as one.

ndouglas:
[Works](https://dsva.slack.com/archives/CJT90C0UT/p1638913001207200).

PR #7065, opened 2021-11-25
VACMS-6640: Show what Vet Center services are required in Vet Center form and prevent removal.

### LINKED ISSUE #6640: Indicate to Vet Center editors when a service is required, and prevent archiving.
<details><summary>Design spec</summary>

From https://www.figma.com/file/QUZpQV5Nf0LhtsrnqqVHas/Vet-Centers-forms?node-id=0%3A1 

Vet_Centers_forms_–_Figma

</details>

## ACs
- [ ] Implement design from above
  - [ ] Stretch: Use the Vet Center name in the first column's header. Only go ahead with this if it's an easily reusable approach to "tokenizing" labels in forms.
- [ ] Required services lose their Remove button on the IEF form.
- [ ] Required Vet Center services may not be archived except by Content Admin (we may want to spin this into a separate ticket)
- [ ] Let KB editors know that this change is incoming. 
- [ ] Design review

ethanteague:
## Description
Closes #6640

## Testing done
Visual / behat

## Screenshots
image

## QA steps
 - [x] As an administrator, go to `node/3840/edit` and visually verify table has new Required/Optional column.
 - [x] Visually verify items marked as required do not have delete buttons
 - [x] Visually verify table headers match screenshot
 - [x] Visually verify services are alpha sorted
 - [x] Visually verify table drag and drop column is not present
 - [x] Edit a service, save, and confirm the change is saved and errors are not thrown.
 - [x] ADDED  Confirm edit button on service works
 - [x] ADDED  Confirm remove button on service works


docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -212,6 +225,86 @@ public function requireRevisionMessage(array &$form, FormStateInterface &$form_s
...
+  /**
+   * Checks if service is required.
+   *
+   * @param string $service
+   *   The service name.
+   *
+   * @return bool
+   *   True is service is required.
+   */
+  public function checkIfServiceRequired($service) {
```
kevwalsh, line 291: i don't like the idea of hard-coding this, because the single source of truth should be the Service taxonomy term, which has a required boolean.

docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -212,6 +225,86 @@ public function requireRevisionMessage(array &$form, FormStateInterface &$form_s
...
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The form state.
+   */
+  public function modifyIefServicesFormDisplay(array &$form, FormStateInterface $form_state) {
+    $form_object = $form_state->getFormObject();
+    if ($form_object instanceof EntityFormInterface) {
+      /** @var \Drupal\node\NodeInterface $node */
+      $node = $form_object->getEntity();
+      $node_title = $node->getTitle();
+      $form['field_health_services']['widget']['entities']['#table_fields']['label']['label'] = $this->t('Services offered at :title', [':title' => $node_title]);
```
kevwalsh, line 260: This is great, but i guess i wonder how scaleable this is. 

Say we want to drop the node title into other labels, in other places. Will we need this much field instance and widget-specific code? Is it worth this much custom code right now, if (and it's a big if) there's an opportunity to do this more generically later?

EWashb:
@ethanteague everything looks like the figma design...great work! However, I was not able to edit the service. "Edit" wasn't clickable for me, so I couldn't do the last QA step.

ethanteague:
> @ethanteague everything looks like the figma design...great work! However, I was not able to edit the service. "Edit" wasn't clickable for me, so I couldn't do the last QA step.

You probably were viewing a locked page - looks like Kev was testing right before you, so you will need to click the break lock link on the top of the form. You should be able to edit after that, @EWashb

docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
ethanteague: Meh. Not too worried about it. If we wanna reuse this, no biggie to create a reusable method that we call as needed. That said, this is a pretty common pattern that we use / have in use all over the place.

docroot/modules/custom/va_gov_vet_center/src/EventSubscriber/EntityEventSubscriber.php
ethanteague: That's my bad. Forgot about that bool. Updated.

kevwalsh (changes requested):
Two things i saw after comparing to the design.

- [ ] Sort by Required, then by alpha. 
- [ ] Remove the `"[Vet Center name] - "` from the first column. We want to keep these service node titles as they are, but have the node titles appear simpler in this context, with just the taxonomy term name. 

Edit_Vet_Center_Nashville_Vet_Center___VA_gov_CMS

ethanteague:
> Two things i saw after comparing to the design.
> 
> * [ ]  Sort by Required, then by alpha.
> * [ ]  Remove the `"[Vet Center name] - "` from the first column. We want to keep these service node titles as they are, but have the node titles appear simpler in this context, with just the taxonomy term name.
> 

Thanks for catching these, @kevwalsh. Updated :+1:

swirtSJW (changes requested):
Seems like this may have broken the winno-ing logic as I can still select a service that is already present and required.   And after adding it, the service shows as optional rather than required.
image

ethanteague:
> Seems like this may have broken the winno-ing logic as I can still select a service that is already present and required. And after adding it, the service shows as optional rather than required. 

Thank you for catching that, @swirtSJW ! Updated :+1:

swirtSJW (commented):
Since this is a sprint goal.   Lets chat

docroot/themes/custom/vagovadmin/vagovadmin.theme
```diff
@@ -86,3 +87,145 @@ function _vagovadmin_get_node_parent(Paragraph $paragraph) {
...
+ * Prepares variables for inline_entity_form_entity_table form templates.
+ *
+ * Default template: inline-entity-form-entity-table.html.twig.
+ *
+ * @param array $variables
+ *   An associative array containing:
+ *   - form: A render element representing the form.
+ */
+function vagovadmin_preprocess_inline_entity_form_entity_table(array &$variables) {
+  $form = $variables['form'];
```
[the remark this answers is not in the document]
swirtSJW: @ethanteague  I added a commit to move the long bunch of code to a function.   It should really be broken down further, but we can come back to it.

swirtSJW (changes requested):
Adding a new service works.
Winnoing logic works.
However, the edit and remove buttons don't do anything.
image

swirtSJW:
I added testing the edit and remove buttons to the QA steps.

swirtSJW:
@ethanteague  @kevwalsh  Is it desired/expected that the the edit window jumps to the tip of the list rather than being inline?
image

swirtSJW:
If I edit two services simultaneously, canceling one, will cancel all.   This results in data loss.
image

swirtSJW (changes requested):
Edit and cancel now work.   But cancel, cancels all edits... leading to data loss.

swirtSJW:
@EWashb @kevwalsh   I have some  concerns about this design that may not have to block this PR but we might want to address soon.  

1 . I can not tell from this display which services are published and which are not.
image

2.  When I add a service, it is not published and I can only publish it through my section dashboard
image
3.  If I make an inline edit to a published service but I save the vetccenter changes as draft.... I have no idea what I just did to revisions and I have no idea what is now published within the service edit. 
Looks like it goes straight through to published 
image
4.  Revision log message is really confusing because it did not go through a proper save so the log message is reused from previous.
image

I suggest the edit link does not take to inline edit but takes to service node edit.

ethanteague:
> If I edit two services simultaneously, canceling one, will cancel all.   This results in data loss.
> 
> image
> 
> 
> 
> 

This will take a while to figure out, @swirtSJW @kevwalsh - seems like there is a lot here that needs to be addressed, so gonna leave this one alone until we discuss further.

EWashb:
I agree- this seems to be adding more complex issues and I wouldn't be comfortable introducing those for the sake of the intended design. Definitely warrants more discussion @ethanteague @swirtSJW @kevwalsh

kevwalsh:
> @ethanteague @kevwalsh Is it desired/expected that the the edit window jumps to the tip of the list rather than being inline?

Yeah this is definitely not ideal.

> If I edit two services simultaneously, canceling one, will cancel all. This results in data loss.

I just reproduced this on staging, not a regression introduced by this PR.  I'm curious is this is specific to our configuration of IEF, or an IEF problem more generally. 

> When I add a service, it is not published and I can only publish it through my section dashboard

I just reproduced this on staging, not a regression introduced by this PR.

Turns out this change happened in an earlier iteration, but hasn't been caught/realized, probably because of the nature of the launch runbooks for each Vet Center. At some point, the content moderation field got dropped from the inline form, but the ramifications of that were not caught. Either we need to:

1. bring the content moderation field to the IEF form, or
2. we need to make these IEF nodes more "paragraph-like", where their published status is somehow inherited from the parent node's published status. 

Number 1 is easier, but I like #2. Still, it probably belongs in a separate issue. 

====

My sense is that if we can address the window jumping around, we can still get this done this sprint. But we should really prioritize addressing these two other issues, particularly the content moderation one.

ethanteague:
@kevwalsh the jump is being caused by the alpha sort logic introduced in this task. Another tough one to figure out. Apologies for being a downer, just don't know how to prevent the jump when we need to sort services alphabetically.

config/sync/user.role.content_editor.yml
```diff
@@ -29,6 +29,7 @@ permissions:
   - 'create media'
   - 'create promo block content'
   - 'create video media'
+  - 'delete any vet_center_facility_health_servi content'
```
kevwalsh, line 32: i think this permission, if it's needed (see https://github.com/department-of-veterans-affairs/va.gov-cms/issues/7144), should be moved to user.role.vet_center_content_creator.yml instead. Not a big deal, since it's all workbench access controlled anyway. but it's a more product-oriented role.

kevwalsh:
> @kevwalsh the jump is being caused by the alpha sort logic introduced in this task. Another tough one to figure out. Apologies for being a downer, just don't know how to prevent the jump when we need to sort services alphabetically.

what if this opened in a modal? modals create their own challenges, but that was in an earlier "blue sky" design by Kauff.

ethanteague:
> > @kevwalsh the jump is being caused by the alpha sort logic introduced in this task. Another tough one to figure out. Apologies for being a downer, just don't know how to prevent the jump when we need to sort services alphabetically.
> 
> 
> 
> what if this opened in a modal? modals create their own challenges, but that was in an earlier "blue sky" design by Kauff. 

That's an interesting idea, @kevwalsh - I'll give that a whirl in the am 👍

kevwalsh:
I created https://github.com/department-of-veterans-affairs/va.gov-cms/issues/7144 and https://github.com/department-of-veterans-affairs/va.gov-cms/issues/7145 for the two issues Steve found, that are not regressions.

ethanteague:
@kevwalsh @swirtSJW - I figured out a way to stop the form from not loading inline, and updated the pr with said logic.

swirtSJW (approved):
QA passed.

Nice work @ethanteague

Issue #7146, opened 2021-12-03
Screen reader does not read tooltip content on field.


swirtSJW:
## Describe the defect
The tooltip field formatter introduced on https://github.com/department-of-veterans-affairs/va.gov-cms/issues/6449
does not meet accessibility standards because the tooltip content is not available to screen readers.

The tooltip implementation used on our fieldset does get read appropriately by a screen reader

https://github.com/department-of-veterans-affairs/va.gov-cms/issues/6449

This is not caught by AXE tests, but was caught by an actual screen reader.

## Expected behavior
Screen reader will have be able to read tooltip content.

## Screenshots 
![image](https://user-images.githubusercontent.com/5752113/144546965-cdb21737-5a9e-48a4-9e97-561900bc6308.png)

 ## Additional context
Add any other context about the problem here. Reach out to the Product Managers to determine if it should be escalated as critical (prevents users from accomplishing their work with no known workaround and needs to be addressed within 2 business days).

 
## Labels
(You can delete this section once it's complete)
- [x] Issue type (red) (defaults to "Defect")
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [x] CMS workstream (orange) (not needed for bug tickets)
- [ ] CMS-supported product (black)

### CMS Team

Please leave only the team that will do this work selected. If you're not sure, it's fine to leave both selected.

- [ ] `Platform CMS Team`
- [x] `Sitewide CMS Team`

PR #7164, opened 2021-12-05
VACMS-7146: Make tooltips screenreader accessible.

ethanteague:
## Description
Relates to #7146
@swirtSJW not sure which screenreader you used. I don't know how to make any of the screenreaders I installed in a couple of different browsers (or the one that comes out of the box with Ubuntu) read the entire page - it captures most of it, but doesn't read any of the tooltips (the old implementation or the new). That said, I think changing role to "tooltip" and adding a css rule that shows the content on focus should resolve this. I visually verified that on tab, the new tooltips have `data-focus-visible-added` attribute added and are now visible (they weren't visible on tab, before) like the existing tippy.js implementation.

## Testing done
Visual with chromevox screenreader

## Screenshots
Used this extension in Chrome:
image

First tooltip with new fieldformatter and css-tooltip (focus is captured when tabbed):
image

Second tooltip with new fieldformatter and css-tooltip (focus is captured when tabbed):
image

Identical experience with Tooltip.js:
image

## QA steps
 - [ ] Using Chrome / Chromium, install chromevox screenreader and enable
 - [ ] As an administrator, go to `/node/3634/edit`
 - [ ] Using tab key, navigate page
 - [ ] Visually verify focus is captured on tooltips in `Name of Vet Center` and `Common name` fields and tooltips are displayed
 - [ ] In inspector, highlight one of the new tooltip buttons, and visually verify that on tab focus, `data-focus-visible-added` attribute is added


swirtSJW (approved):
Nice solution.  I can confirm that I heard what I saw.
There is one minor thing I think due to our JS-ery

The tab order is flawed.   It goes backwards on this element 

image
 

But that is unrelated to this issue.

swirtSJW:
image

Issue #7217, opened 2021-12-07
Vet Center service description help text is repeated


suzanne-gray:
## Describe the defect
On the Edit Vet Center page, the help text under `general service description` appears twice. 

## To Reproduce

View example at: https://prod.cms.va.gov/node/3597/edit#edit-field-health-services

also in node view: https://staging.cms.va.gov/node/41246/edit

## Expected behavior
The help text appears once. 

## Screenshots 

Edit page: 
![help text on Vet Center edit page](https://user-images.githubusercontent.com/90172187/145124626-ce6f0507-76be-4d7e-9f13-e7c430ddd788.png)

Node:
![Vet Center help text duplication](https://user-images.githubusercontent.com/90172187/145124638-981946a6-6edf-4544-b7a4-5e2112429a23.png)

 

## Labels
(You can delete this section once it's complete)
- [x] Issue type (red) (defaults to "Defect")
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [x] CMS workstream (orange) (not needed for bug tickets)
- [ ] CMS-supported product (black)

### CMS Team

Please leave only the team that will do this work selected. If you're not sure, it's fine to leave both selected.

- [ ] `Platform CMS Team`
- [x] `Sitewide CMS Team`

PR #7223, opened 2021-12-08
VACMS-7217: Fix duplicate help text below textarea fields.

ethanteague:
## Description
See #7217

## Testing done
Visual / behat

## Screenshots
image

## QA steps
 - [ ] As an administrator, go to `/node/3597/edit` and open up multiple existing services via the edit button.
 - [ ] Visually verify help text appears below title, and is not duplicated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -136,6 +136,7 @@
         "drupal/taxonomy_entity_index": "^1.1",
         "drupal/taxonomy_menu": "3.x-dev#e89d1e8",
         "drupal/textfield_counter": "^2.0",
+        "drupal/themable_forms": "^1.0@beta",
         "drupal/toolbar_menu": "^2.1",
         "drupal/user_history": "^1.0@alpha",
         "drupal/video_embed_media": "^2.2",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "ef57d436f973df829eeb9e004246f710",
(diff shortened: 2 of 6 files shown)


swirtSJW:
@ethanteague Is there  a future use/need  for themable_forms?

If No, is there another solution that does not involve adding a new module?

ethanteague:
> @ethanteague Is there a future use/need for themable_forms?
> 
> If No, is there another solution that does not involve adding a new module?

Asked myself the same question, and realistically, I think the answer is no. To do it without this module requires me adding some custom code, ala: a theme suggestions hook and then passing some variables to the template in a theme function. I can do this, but was thinking that adding a contrib module was kind of the harm reduction approach. I could be persuaded otherwise - just lemme know if you want me to go the other way, @swirtSJW

suzanne-gray:
Looks good to me @ethanteague ! Thank you!!

swirtSJW:
@ethanteague  That's a tough call.  It's a single use module AND it has no full release (security coverage).  And its requiring a local template.   We have other JS coming in as part of #7243   Can we just add to that JS to just eat the duplicate?


current PR is conflicted.

ethanteague:
> @ethanteague That's a tough call. It's a single use module AND it has no full release (security coverage). And its requiring a local template. We have other JS coming in as part of #7243 Can we just add to that JS to just eat the duplicate?
> 
> current PR is conflicted.

Yeah, I'll kill this one, @swirtSJW, and I'll get rid of this module / go the other way. Also agree with you about https://github.com/department-of-veterans-affairs/va.gov-cms/pull/7243#discussion_r766832507 , so I can address it will I'm in there.

ethanteague:
Closing in favor of resolution in https://github.com/department-of-veterans-affairs/va.gov-cms/pull/7243

Issue #7246, opened 2021-12-09
Remove facility status endpoints for sprint 47


swirtSJW:
## Description
Remove any endpoints for sites that have gone into full production

12/9
https://www.losangeles.va.gov/locations/facilities.asp

- [ ] `Platform CMS Team`
- [x] `Sitewide CMS Team`

PR #7247, opened 2021-12-09
VACMS-7246 Remove Los Angeles status endpoint.

swirtSJW:
## Description

Relates to #7246

## QA Steps
- [ ]  Validate the removal of Los Angeles from the two ymls
- [ ] Validate CMS build ran without config import error (tests pass)


JayDarnell (approved):
All tests have passed and losangeles has been removed from the migrations.

Issue #7081, opened 2021-11-29
Update OpenAPI spec to include entity- and product-based APIs for banners


jkalexander7:
## User Story or Problem Statement

As a CMS engineer, I need to update the OpenAPI spec to include both entity-based and product-based APIs within the same spec, so that I can set up dev environments for proceeding with POCs.

## Acceptance Criteria
- [x] Update YAML file
- [x] Share updated YAML file with Lighthouse team
- [x] Update AWS Gateway ticket (https://github.com/department-of-veterans-affairs/va.gov-cms/issues/7102) with link to file

## Implementation
- [x] Include all endpoints for banner use case

The API endpoint(s) should conform to the OpenAPI Spec from https://github.com/department-of-veterans-affairs/va.gov-cms/issues/6500; the current implementation has been posted here: https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/cms/accelerated_publishing/sample_openapi_spec.yml

The discovery and base implementation for the POC API was done as part of https://github.com/department-of-veterans-affairs/va.gov-cms/issues/6491; this can be used as a reference.

See https://vfs.atlassian.net/wiki/spaces/PCMS/pages/1923088577/API+Proof+of+concept+use+cases for details of the banner use case.

Pleasetake note note cases where the current OpenAPI spec does not match JSON:API; in particular, the conversion under individual entities/items of `data.properties` to `data.attributes.properties`

### CMS Team

Please leave only the team that will do this work selected. If you're not sure, it's fine to leave both selected.

- [x] `Platform CMS Team`
- [ ] `Sitewide CMS Team`

timcosgrove:
For review: 
https://github.com/department-of-veterans-affairs/va.gov-cms/blob/aa0341e1c40cd97ee0fbdf177567c2e876d62c67/docroot/openapi_no_relationships.yaml
https://github.com/department-of-veterans-affairs/va.gov-cms/blob/aa0341e1c40cd97ee0fbdf177567c2e876d62c67/docroot/openapi_relationships.yaml

mchelen-gov:
From followup convo:
- Validates successfully!
- Try to avoid using Git commit URLs, using Git branch URLs is fine
- Document difference between YAML files
- Link to when/where submitted to Lighthouse

timcosgrove:
https://github.com/department-of-veterans-affairs/va.gov-cms/blob/content_api_poc/READMES/openapi/openapi_no_relationships.yaml
[...]

PR #6811, opened 2021-10-27
PR for Content API POC

indytechcook:
## Description

This is a PR, not an issue. It is open to create a Tugboat environment for this branch, so that we can evaluate the Content API POC.

It should not be added to a sprint, assigned points, or closed.

Replacing https://github.com/department-of-veterans-affairs/va.gov-cms/pull/6557

PR #7308, opened 2021-12-16
Vacms 7081 updated poc openapi

timcosgrove:
See #7081 .

This contains everything that was in #6811 that we wanted to keep.


timcosgrove:
Merging to `content_api_poc` branch.

Issue #7082, opened 2021-11-29
Update Drupal to match OpenAPI spec for product-based API


jkalexander7:
## User Story or Problem Statement

As a CMS engineer, I need to update Drupal for the product-based API, so that it will match the OpenAPI spec.

## Acceptance Criteria
- [x] Complete updates for product-based API POC (https://github.com/department-of-veterans-affairs/va.gov-cms/issues/6500)
- [x] Ensure entity-based API requires no updates

### CMS Team

Please leave only the team that will do this work selected. If you're not sure, it's fine to leave both selected.

- [x] `Platform CMS Team`
- [ ] `Sitewide CMS Team`

timcosgrove:
If I look at this, the work here includes adding remaining product API endpoints which are not yet defined. This may need to roll over.

mchelen-gov:
@timcosgrove does that affect the scope or sizing of this issue?

PR #7518, opened 2022-01-07
Vacms 7082 product api additions

timcosgrove:
## Description

Relates to #7082


docroot/modules/custom/va_gov_api/src/Resource/ResourcesByTag.php
```diff
@@ -0,0 +1,132 @@
...
+use Drupal\jsonapi\ResourceType\ResourceType;
+use Drupal\node\NodeInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Resource for returning a collection of 'resource' product nodes.
+ *
+ * Apologies for the naming collision. Currently limited to q_a nodes.
+ */
+class ResourcesByTag extends VaGovApiEntityResourceBase {
```
timcosgrove, line 16: This one is in process, don't review it yet.

docroot/modules/custom/va_gov_api/src/Resource/BannerAlerts.php
```diff
@@ -2,155 +2,261 @@
...
-   * The entity type manager.
-   *
-   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   * {@inheritDoc}
    */
-  protected $entityTypeManager;
+  protected function collectResourceData(Request $request, ResourceType $resource_type) {
+    // Doing this as a case/switch for now. There are almost certainly better
+    // ways to do this.
+    switch ($resource_type->getTypeName()) {
```
indytechcook, line 23: > / Doing this as a case/switch for now. There are almost certainly better
> // ways to do this.

Yep, by using Plugins!

indytechcook:
Just some thoughts

- Make sure to type hint returns and make the responses as consistent as possible
- Separate out the builder of the Resource from from the actual Resource object.
- Create a consistent interface to pull the data from different resource types
- Use the Drupal Plugin API for the different builders for the resource types
- The plugin used will depend on the resource type.
- The plugin could also provide the path as part of it's metadata instead of a bunch of routing entries. https://www.drupal.org/node/2122201
- Try to follow a Composition pattern over inheritance (Single resource class with the different plugins to get the data is one example)

Ideally we need someway to integrate with the params to build the query using the filters, pages, etc in a way which is reused across all request.  I would assume this exists already since the resource module does this.

docroot/modules/custom/va_gov_api/src/Resource/BannerAlerts.php
timcosgrove: Want to chat about this at some point.

docroot/modules/custom/va_gov_api/src/Resource/QandA.php
```diff
@@ -3,83 +3,87 @@
 namespace Drupal\va_gov_api\Resource;
```
timcosgrove, line 3: This whole thing - retrieving an entity with a custom endpoint - is almost certainly not something we would actually do.

indytechcook (approved):

Issue #7102, opened 2021-11-30
Set up Tugboat instance with AWS Gateway


jkalexander7:
## User Story or Problem Statement

As a CMS engineer, I want to set up a Tugboat instance with AWS Gateway, so that I can complete a POC for determining the most viable solution to exposing the API to the public. 

## Acceptance Criteria
- [x] Data can be retrieved via AWS Gateway for use cases defined by the [API POC use cases](https://vfs.atlassian.net/wiki/spaces/PCMS/pages/1923088577/API+Proof+of+concept+use+cases)
    Endpoints: 
      * https://0exmeqltwi.execute-api.us-gov-west-1.amazonaws.com/test/
      * https://0exmeqltwi.execute-api.us-gov-west-1.amazonaws.com/poc/

Documentation about experience: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/7102#issuecomment-1012299654

## Resources
There are two OpenAPI specs that represent all the endpoints needed to retrieve all entities for our API POC use cases, one which describes the relationship endpoints that JSON API requires, and one which does not. The relationship endpoints add considerable size and complexity to the spec, and we may not make use of them.

https://github.com/department-of-veterans-affairs/va.gov-cms/blob/e571767199ceb2436c4d41b5da97f4208359f023/docroot/openapi_no_relationships.yaml
https://github.com/department-of-veterans-affairs/va.gov-cms/blob/e571767199ceb2436c4d41b5da97f4208359f023/docroot/openapi_relationships.yaml

### CMS Team

Please leave only the team that will do this work selected. If you're not sure, it's fine to leave both selected.

- [x] `Platform CMS Team`
- [ ] `Sitewide CMS Team`

indytechcook:
I tested importing the spec above and the following errors occurred:

```
Unable to create model for 'block_content--alert': Model name must be alphanumeric: block_content--alert
Unable to create model for 'block_content--promo': Model name must be alphanumeric: block_content--promo
Unable to create model for 'file--file': Model name must be alphanumeric: file--file
Unable to create model for 'links_metadata': Model name must be alphanumeric: links_metadata
Unable to create model for 'list_metadata': Model name must be alphanumeric: list_metadata
Unable to create model for 'media--image': Model name must be alphanumeric: media--image
Unable to create model for 'node--banner': Model name must be alphanumeric: node--banner
Unable to create model for 'node--checklist': Model name must be alphanumeric: node--checklist
[...]

PR #7533, opened 2022-01-10
VACMS-7102 AWS Gateway POC Work

indytechcook:
## Description

Relates to #7102


READMES/openapi/aws/lambda.js
```diff
@@ -0,0 +1,72 @@
+const https = require('https');
+
+function getRequest(path, queryParams) {
+  const options = {
+    hostname: 'content-api-poc-trbtppyqohmbnfdh8faiwyiic9hby79m.demo.cms.va.gov',
+    port: 443,
+    path: `/jsonapi${path}`,
+    query: queryParams,
+    method: 'GET',
+    rejectUnauthorized: false,
```
indytechcook, line 10: This is only for the POC, this file isn't used in anything.  Follow-up work will be using the real certs.

Issue #6920, opened 2021-11-05
Rename VA North Florida health care to VA North Florida/South Georgia health care


kevwalsh:
## User Story or Problem Statement

As a Veteran that searches for South Georgia OR North Florida, i want to find my health care system using a search engine. 

## Acceptance Criteria
- [x] h1 and breadcrumb for https://www.va.gov/north-florida-health-care/ has been changed to VA North Florida/South Georgia health care
- [x] No URLs are updated 
- [x] VAMC system health service nodes (and any other content that uses the VAMC system node title in its autoentity label) have been updated to reflect this 
- [ ] Internal documentation has been updated for "How to update the name of a VAMC system". This can be sent to Help desk team for their documentation. It should live along side the ["How to update the name of a VAMC facility"](https://dsva.slack.com/archives/CDHBKAL9W/p1636151114050100?thread_ts=1636149011.046700&cid=CDHBKAL9W) instructions, which were noted in this slack thread on Fri Nov 5.

> 1. Fix the menu name on the facility and check the box to "Generate automatic URL alias"  then save the facility.
> 2. Go to bulk edit and find all the facility services under the old name and use the option to "update url alias"
> 3. return to bulk edit and find all the facility services under the old name and use the "Resave content" option to cause the titles of the service to pick up the new name.

## Implementation steps

On tugboat
1. Edit node and menu item for the system
2. VBO to update VAMC system health services
3. Content release
4. Smoke test on Tugboat FE
5. Future Points of failure testing Tugboat
   1.    Rename a facility with the N. Florida health care system : /node/1573 (change title, menu title, click the box to create new alias)
   2.   Bulk operation update alias for all facility health services for "Waycross VA clinic"
   3. Bulk save the all facility health services for "Waycross VA clinic" so the title gets picked up.
   4. - [ ] Validate that none of new aliases contain `VA North Florida/South Georgia`
   5. Rebuild tugboat FE.   Check Waycross VA clinic path on FE.
6. Write KB article about what you did
7. Repeat steps 1-2 on prod. 

## Design principles

Veteran-centered
- [x] `Single source of truth`: Increase reliability and consistency of content on VA.gov by providing a single source of truth.
- [x] `Accessible, plain language`: Provide guardrails and guidelines to ensure content quality.
- [ ] `Purposely structured content`: Ensure Content API can deliver content whose meaning matches its structure.
[...]

PR #7613, opened 2022-01-14
VACMS-6920: update label for north florida/south georgia menu

JayDarnell:
## Description

Relates to #6920

## Testing done

Visual testing

## Purpose

This change captures the menu label in config to match the changes manually applied for ticket 6920


swirtSJW:
I am going to merge this because it is going to fail tugboat testing since tugboat is using old data that will not match this name.

PR #7655, opened 2022-01-19
VACMS-0000: yarn version check not necessarily repo-installed version

timcosgrove:
## Description

`composer va:web:install` contains a script which attempts to check that the installed yarn version matches the version specified by composer. However, the script checks globally installed yarn, which is not guaranteed to match what is installed by this repository.


swirtSJW:
Hi @timcosgrove   How can I help move this along?   I'd be happy to help, but without testing steps, I am not sure what I am looking for.

timcosgrove:
This has been sitting for too long; no longer believe this is necessary.

Issue #6057, opened 2021-08-09
Editor experience for repeating / recurring events


kevwalsh:
## User Story or Problem Statement

1. As an event editor, i should only be able to create a recurring event based on a simple pattern (like once a week, or 1st Tuesday of the month, or every M/W/F) 

2. As an event editor, i should only be able to create a single event date, or a single event date repeating pattern. I should not be able to create two "custom" dates, or two custom repeating patterns.

 - Smart date recurring module requires the date field to be multivalue. By default, this adds a "Add another item" button that is confusing. 

<details><summary>Screenshots</summary>

### New config for field_datetime_range_timezone	

From https://master-dzqiprnzpsoj5ifjekygsan9jad6kjdo.demo.cms.va.gov/admin/structure/types/manage/event/fields/node.event.field_datetime_range_timezone

![Date_and_time_settings_for_Event___VA_gov_CMS](https://user-images.githubusercontent.com/643678/128700945-679e9a27-08f8-4293-8aa4-580f209cb95b.png)

### Confusing multivalue UX

![Create_Event___VA_gov_CMS](https://user-images.githubusercontent.com/643678/128697958-7cbc3343-3ca7-4d8c-8127-a0096641a8f6.png)

### Desired permissions

![Permissions_filtered_by_modules___VA_gov_CMS](https://user-images.githubusercontent.com/643678/128700508-3ca3dae5-08e7-4c4b-9553-4dbdad3de57e.png)

</details>

## Acceptance Criteria
- [x] Recurring events are enabled set to max out at 12 months (setting on field)
- [x] Field help text is removed (for now, at least)
- [x] Enabled time zones match the time zones currently seen at /node/add/event, _plus_ Europe/Berlin
- [x] Permissions are set so that content content editors/reviewers/publishers and content admins are able to use the module 
- [x] "Add another item" is hidden on the event form when smart_date_recur is enabled
- [x] If #6055 has shown that we can change cardinality of field_datetime_range_timezone  without impacting existing graphql or FE templates, this can potentially be merged
- [ ] If #6055 has shown that changing cardinality of field_datetime_range_timezone will break the content API, then just create a PR assuming that's been done elsewhere and mark as DO NOT MERGE. 

## Design principles

Which [CMS design principle](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/cms/cms-product-design-principles.md) is at play?
[...]

Issue #2460, opened 2020-07-28
Add conditional fields to event content type


rachel-kauff:
## Description
Lack of conditional fields on the event content type form, especially for the address field group, is making it really confusing to use.

As a user, I only want to see fields that are relevant to my previous choices on the form.

## Acceptance Criteria
* [ ] If `Location type` = `VA facility`, then display `Facility location` field and make it required
* [ ] If `Location type` = `VA facility`, then display the field formerly known as`A human readable label...` soon to be changed to `Building, floor, or room` 
* [ ] If `Location type` = `Online`, display `Location URL` field
- [ ] If `Location` = `Online`, then don't display the `Address` field group and don't make it required.
- [ ] If `Location` = `VA facility` then don't require fields in the `Address` field group (this assumes that the system is managing to provide an address automatically based on the VA facility selected). 
- [ ] If `Location` = `VA facility` and the user opts to add a custom address, use that instead of the one automatically supplied by the facilities locator api.

### Implementation recommendations
1. We have a bunch of work on the horizon related to events that I would like to contain it in a new module `va_gov_events`.   It should be created as a result of this issue and related code should live there.

## Aritfacts
Annotated MURAL for event form help text:
<img width="950" alt="Screen Shot 2020-07-28 at 11 26 18 AM" src="https://user-images.githubusercontent.com/41017988/88706040-2bc9c680-d0c5-11ea-9b1e-1f56a05f7aa9.png">
- [MURAL link](https://app.mural.co/t/vagov6717/m/vagov6717/1595867163702/318273f5fb1166d8fb771da1c63ffbd16ae7073d)

ethanteague:
https://www.figma.com/file/3ysYRmbHbJ2d0auvFDiBgF/Event?node-id=0%3A1 makes this task obsolete.

kevwalsh:
I'm going to reopen this for the work in PR #7305, to keep #7305 separate from #6057

PR #7713, opened 2022-01-22
VACMS-6057 recurring events instances modal logic

ethanteague:
## Description
#6057

## Testing done
Visual / behat

## Screenshots
image

## QA steps
 - Will be collaborative / in real time with Rachel and Kev


docroot/modules/custom/va_gov_events/src/Controller/Instances.php
```diff
@@ -0,0 +1,548 @@
...
+   *
+   * @return array
+   *   A render array structure of fields for this entity.
+   *
+   * @see \Drupal\Core\Entity\EntityListBuilder::render()
+   */
+  public function buildRow(array $instance) {
+    // Get format settings.
+    // @todo make the choice of format configurable?
+    $format = \Drupal::getContainer()
```
ethanteague, line 196: The files in this dir are copypastas from smart_date_recur, which doesn't use DI. I changed logic as necessary to point to changes to text, etc. in va_gov_events, and applied phpcs changes. We could use DI, but imo, this will lead to potential headaches in the future if/when we need to update these files to track changes made to smart_date_recur.
@swirtSJW et.al.^^^^^^^

kevwalsh:
We should meet to go over this later, but this screenshot captures most of my concerns


**Placement of Manage event series and Time zone in vagovadmin (Seven) theme**

Edit_Event_asdf___VA_gov_CMS



</details>

1. Putting "Edit event series" (and Time Zone for that matter) way over to the right is not a common pattern for fields or buttons like this. I know i'm late in the game to say, but seeing it in action on my big monitor (or even at a common size for our editors, 20-24" makes it them harder to see)
2. At the very least, the modal loading icon needs to be connected to the button.
Suggestion: we solve these issues with Claro-specific theming (see below) 


**Unchecked "Make recurring" checkbox is confusing for existing events that are already reccurring**
1. Creating an event that's every week
2. Go back to edit, 
3. See that the Make recurring is unchecked. (which is on its own a bit confusing)
4. Check it again, and make it recur every day
5. Save
6. See that the changes you made to make it recur every day didn't do anything

Suggestion: we either
1. Fix it so that changes to recurring event patterns that are made to existing recurring events actually work, and the checkbox would be checked by default. (idealy but probably difficult)
2. Just hide the Make recurring checkbox for existing events that are already recurring. 
If we go with 2, i wonder if it makes sense to float the "Manage event series" back to the left. 

**Claro**
I just enabled Claro on [this tugboat environment](https://pr7713-faszuoll2jsttdybs8hfteejfge0kcfp.ci.cms.va.gov/node/43232/edit) to see what would happen. Things are kind of broken, but looks like minor fixes
* Start and end field labels are overlapping with their fields
* when selecting weekly pattern, "on these days" fields  are overlapping with the other recurring fields
* when selecting monthly pattern, "on the" fields fields are overlapping with the other recurring fields
* Loading icon is not attached to Manage series button
* "Show row weights" is visible but doesn't do anything because we only allow one pattern, and should be hidden
* Make recurring checkbox has a bad margin


<details><summary>Screenshots</summary>
Edit_Event_asdf___VA_gov_CMS

Edit_Event_asdf___VA_gov_CMS


</details>

rachel-kauff:
> Putting "Edit event series" (and Time Zone for that matter) way over to the right is not a common pattern for fields or buttons like this. I know i'm late in the game to say, but seeing it in action on my big monitor (or even at a common size for our editors, 20-24" makes it them harder to see)

I hadn't considered the screen size issue this might cause, and it certainly makes sense to move those righthand fields back for now, but this does make me think about how to think about the width of forms in future iterations of claro/design sys. I don't think the width should scale infinitely.

rachel-kauff:
Screen Shot 2022-01-26 at 7 53 18 AM

rachel-kauff:
Re: "Make recurring checkbox is unchecked when you go back to edit a recurring event."
I want to advocate hard for this one:

> Fix it so that changes to recurring event patterns that are made to existing recurring events actually work, and the checkbox would be checked by default. (idealy but probably difficult)

It's really confusing to not be able to see the changes I just made. What if I want to adjust what I just did to make my event repeat every two weeks instead of every one? The interface I had oriented myself to suddenly is gone and I start to lose trust that the content I had previously saved actually worked.

Screen Shot 2022-01-26 at 8 04 15 AM

rachel-kauff:
A few language changes to the buttons in this modal (What's in the PR still uses some of the OOB labels, but I also had some second thoughts about what was in the Figma.) 

Here's what's there:
Screen Shot 2022-01-26 at 8 43 09 AM

I think these button labels clarify a little more:
Screen Shot 2022-01-26 at 8 43 23 AM

ethanteague:
@kevwalsh @rachel-kauff will need to get with @tjheffner & @swirtSJW for an assist on sorting out the behavioral failures (there's some environment specific stuff that needs to be sorted out / doesn't occur in local instance), but I've applied the latest version of smart_date and made recurring logic & the rest of the events page compatible with Claro.

image

image

config/sync/system.theme.yml
```diff
@@ -1,4 +1,4 @@
-admin: vagovadmin
-default: vagovadmin
+admin: vagovclaro
+default: vagovclaro
```
tjheffner, line 2: we don't want to commit this change yet. it'll turn on claro theme for everyone

ethanteague: Yep, with you there, @tjheffner - was thinking this wouldn't be merged until claro was released.

ethanteague: turned off in pr

tests/behat/features/save_and_continue.feature
```diff
@@ -44,8 +44,8 @@ Feature: Save and continue button works as expected.
 
     # Confirm that the moderation history and state are shown correctly.
     And I should see "Draft" in the ".views-field-moderation-state" element
-    Then I visit the "moderation-history" page for a node with the title "Behat save and continue new test - edited"
-    And I should see "Set to Draft on " in the ".views-field-moderation-state" element
```
swirtSJW, line 48: Why is this changing with events?

tests/behat/features/user.feature
```diff
@@ -5,7 +5,6 @@ Feature: The VA Website is accessible to users.
     Given I am on the homepage
     Then I should see "Log in"
     And I should see "Veterans Affairs" in the "title" element
-    And I should see "Login with PIV or other Smartcard." in the "#edit-simplesamlphp-auth-login-link" element
```
swirtSJW, line 8: events?

tests/behat/features/content_editing.feature
```diff
@@ -72,8 +72,8 @@ Feature: CMS Users may effectively create & edit content
     And I press "Save"
 
     # Confirm that the va.gov url is not shown for nodes without a published revision.
-    Then I should see "Content Type: Office" in the "#block-entitymetadisplay" element
-    And I should not see "VA.gov URL" in the "#block-entitymetadisplay" element
+    Then I should see "Content Type: Office" in the "#block-vagovclaro-entitymetadisplay" element
```
swirtSJW, line 75: ALl these tests changes seem unrelated to evetns.

swirtSJW (commented):
Seeing some changes that don't seem related to events

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -74,6 +74,16 @@ public function __construct(
...
   }
 
+  /**
+   * Alteration to entity view pages.
+   *
+   * @param \Drupal\core_event_dispatcher\Event\Entity\EntityViewAlterEvent $event
+   *   The entity view alter service.
+   */
+  public function entityViewAlter(EntityViewAlterEvent $event):void {
+    $this->appendHealthServiceTermDescriptionToVetCenter($event);
```
swirtSJW, line 84: Is this related?   It seems like this whole file should be removed from this PR.   All the changes seem odd... stuff is jumping around and unrelated.

config/sync/field.field.node.event.field_address.yml
```diff
@@ -44,11 +44,19 @@ settings:
...
+      override: optional
+    addressLine2:
+      override: optional
     postalCode:
       override: hidden
     sortingCode:
       override: hidden
     dependentLocality:
       override: hidden
+    locality:
```
swirtSJW, line 57: related?

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -218,7 +232,7 @@ public function getProductTypeTermFields(array &$form, FormStateInterface $form_
   public function appendHealthServiceTermDescriptionToVetCenter(EntityViewAlterEvent $event):void {
     if ($event->getDisplay()->getTargetBundle() === 'vet_center') {
       $build = &$event->getBuild();
-      $services = isset($build['field_health_services']) ? $build['field_health_services'] : [];
```
swirtSJW, line 221: related?

docroot/modules/custom/va_gov_events/src/Controller/Instances.php
```diff
@@ -0,0 +1,552 @@
...
+  protected $formBuilder;
+
+  /**
+   * Indicating whether current controller instance uses Ajax.
+   *
+   * @var bool
+   */
+  private $useAjax;
+
+  /**
```
swirtSJW, line 68: can't inherit when you override with create()

docroot/modules/custom/va_gov_events/src/Controller/Instances.php
```diff
@@ -0,0 +1,552 @@
...
+        }
+        $instance['class'] = $override_type;
+        $instance['override'] = $override;
+      }
+      else {
+      }
+      $instance['rrule'] = $this->rrule->id();
+      $instance['rrule_index'] = $index;
+
+    }
```
swirtSJW, line 148: need an unset here to break the reference on $instance

swirtSJW (commented):
Ok This should be the last round of review.   I found a couple things.... I think

config/sync/views.view.non_clinical_services.yml
```diff
@@ -1124,7 +1124,7 @@ display:
           empty: true
           tokenize: false
           content:
-            value: 'You have not added a register for care location yet. '
```
swirtSJW, line 1127: Should this file be in this pr

docroot/modules/custom/va_gov_events/va_gov_events.services.yml
```diff
@@ -0,0 +1,22 @@
+services:
+  va_gov_events.entity_event_subscriber:
+    class: Drupal\va_gov_events\EventSubscriber\EntityEventSubscriber
+    arguments: ['@string_translation']
+    tags:
+      - { name: event_subscriber }
+  va_gov_events.route_subscriber:
+    class: Drupal\va_gov_events\Routing\RouteSubscriber
+    tags:
```
swirtSJW, line 9: The arguments that you were loading with create... should actually be defined here.  Then the create is not needed.

config/sync/field.field.node.event.field_address.yml
ethanteague: Yep, @swirtSJW  - this is events form logic from https://github.com/department-of-veterans-affairs/va.gov-cms/issues/2460 that we bundled with recurring.

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
ethanteague: Agreed - removed, @swirtSJW

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
ethanteague: removed

ethanteague: Actually, we need this b/c PHPstan was complaining:  `Use null coalesce operator instead of ternary operator.`

docroot/modules/custom/va_gov_events/src/Controller/Instances.php
ethanteague: Don't understand what you mean here, @swirtSJW

config/sync/views.view.non_clinical_services.yml
ethanteague: Removed

docroot/modules/custom/va_gov_events/src/Controller/Instances.php
ethanteague: @swirtSJW straightened me out in slack

docroot/modules/custom/va_gov_events/src/Controller/Instances.php
ethanteague: Was reading the wrong line - resolved.

tasks-periodic.yml
```diff
@@ -94,7 +94,7 @@ tasks:
   va/background/daily/migrate/flag_removed_facilities:
     desc: These flagging process is to run once nightly.
     cmds:
-      - drush $DRUSH_ALIAS drush va_gov_migrate:flag-missing-facilities
+      - drush $DRUSH_ALIAS va_gov_migrate:flag-missing-facilities
```
swirtSJW, line 97: merge bubble  This should not be here

docroot/modules/custom/va_gov_events/src/Controller/Instances.php
```diff
@@ -0,0 +1,557 @@
...
+        }
+        $instance['class'] = $override_type;
+        $instance['override'] = $override;
+      }
+      else {
+      }
+      $instance['rrule'] = $this->rrule->id();
+      $instance['rrule_index'] = $index;
+      unset($instance);
+    }
```
swirtSJW, line 153:
```suggestion
    }
    https://github.com/department-of-veterans-affairs/va.gov-cms/pull/7713
```

swirtSJW (approved):
Herculean effort brother.  Well done.

tests/behat/features/content_editing.feature
```diff
@@ -83,31 +84,32 @@ Feature: CMS Users may effectively create & edit content
...
     Then I visit the "edit" page for a node with the title "Test Office - BeHaT 404"
     And I select "Archived" from "edit-moderation-state-0-state"
     And I fill in "Revision log message" with "Test archiving"
     And I press "Save"
-    Then I should see "Test Office - BeHaT 404\u003C\/a\u003E\u003C\/em\u003E has been updated."
-    And I should see "Content Type: Office" in the "#block-entitymetadisplay" element
-    And I should not see "VA.gov URL" in the "#block-entitymetadisplay" element
+    Then I should see "has been updated."
+    And I should see "Content Type: Office" in the "#block-vagovclaro-entitymetadisplay" element
+    And I should not see "VA.gov URL" in the "#block-vagovclaro-entitymetadisplay" element
```
tjheffner, line 112: so i was fixing this in https://github.com/department-of-veterans-affairs/va.gov-cms/pull/7953 in a way that works in both themes... i'm worried this will fail tests regardless of which theme is built now

Issue #7735, opened 2022-01-25
Cannot edit menu link content


timcosgrove:
## Describe the defect
Attempting to edit a menu link results in an error.

## To Reproduce
1. Visit https://prod.cms.va.gov/admin/structure/menu/manage/education-benefits-hub for example.
2. Attempt to edit GI Bill School comparison

## Expected behavior
You should see an edit form for making adjustments to the menu link.

## Screenshots 
![image.png](https://images.zenhubusercontent.com/614213ae32396d2bc1205abd/b307f064-9914-4fe6-b18c-4ba582a25943)

 ## Additional context
In looking at the error logs, workbench_menu_access.module attempts to call a function `accessCheck()` on the menu links access control handler to determine if a user should have access to edit a given menu link.
https://git.drupalcode.org/project/workbench_menu_access/-/blob/8.x-1.x/workbench_menu_access.module#L54

`accessCheck()` is not the correct method name to call on a class that extends EntityAccessControlHandler: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityAccessControlHandler.php/function/EntityAccessControlHandler%3A%3AcheckAccess/9.2.x

It is very possible these checks have never worked. 

### CMS Team

Please leave only the team that will do this work selected.

- [x] `Platform CMS Team`
- [x] `Sitewide CMS Team`
  - [ ] `⭐️ Content ops`
  - [ ] `⭐️ CMS experience`
  - [ ] `⭐️ Offices`
  - [ ] `⭐️ Product support`
  - [ ] `⭐️ User support`

swirtSJW:
I am not following this error.  I am able to edit links at https://prod.cms.va.gov/admin/structure/menu/manage/education-benefits-hub and I am able to edit links. 
![image](https://user-images.githubusercontent.com/5752113/150900131-77fa52f2-94a3-406f-84a3-71009fc96067.png)
Are you encountering this error as a specific user?

swirtSJW:
Also if workbench menu access is calling the wrong method, shouldn't this  be an issue  with a patch, instead of overriding it locally?

swirtSJW:
This actually seems something specific to https://prod.cms.va.gov/admin/structure/menu/item/686/  I can edit the others.
Makes me suspect that menus with $ in them are messed up
[...]

PR #7736, opened 2022-01-25
VACMS-7735: add method "alias" to allow workbench_menu_access check t…

timcosgrove:
See #7735

### DIFF
--- a/docroot/modules/custom/va_gov_graphql/src/Access/MenuLinkContentAccessHandler.php
+++ b/docroot/modules/custom/va_gov_graphql/src/Access/MenuLinkContentAccessHandler.php
@@ -3,6 +3,7 @@
 namespace Drupal\va_gov_graphql\Access;
 
 use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Access\AccessResultInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\menu_link_content\MenuLinkContentAccessControlHandler;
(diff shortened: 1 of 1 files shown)


indytechcook:
Looks like Worbench should be called `access` since that's what is on the interface.  `access` does call `checkAccess` but all public interaction with the `AccessHandlerInterface` should go through `access`.

```
  /**
   * Checks access to an operation on a given entity or entity translation.
   *
   * Use \Drupal\Core\Entity\EntityAccessControlHandlerInterface::createAccess()
   * to check access to create an entity.
   *
   * @param \Drupal\Core\Entity\EntityInterface $entity
   *   The entity for which to check access.
   * @param string $operation
   *   The operation access should be checked for.
   *   Usually one of "view", "view label", "update" or "delete".
   * @param \Drupal\Core\Session\AccountInterface $account
   *   (optional) The user session for which to check access, or NULL to check
   *   access for the current user. Defaults to NULL.
   * @param bool $return_as_object
   *   (optional) Defaults to FALSE.
   *
   * @return bool|\Drupal\Core\Access\AccessResultInterface
   *   The access result. Returns a boolean if $return_as_object is FALSE (this
   *   is the default) and otherwise an AccessResultInterface object.
   *   When a boolean is returned, the result of AccessInterface::isAllowed() is
   *   returned, i.e. TRUE means access is explicitly allowed, FALSE means
   *   access is either explicitly forbidden or "no opinion".
   */
  public function access(EntityInterface $entity, $operation, AccountInterface $account = NULL, $return_as_object = FALSE);
```

docroot/modules/custom/va_gov_graphql/src/Access/MenuLinkContentAccessHandler.php
```diff
@@ -37,4 +38,21 @@ protected function checkAccess(EntityInterface $entity, $operation, AccountInter
...
+   * @param string $operation
+   *   The operation being performed.
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The user account for which access is being checked.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   */
+  public function accessCheck(EntityInterface $entity, $operation, AccountInterface $account) : AccessResultInterface {
+    return $this->checkAccess($entity, $operation, $account);
```
indytechcook, line 55:
```suggestion
    return $this->access($entity, $operation, $account);
```

swirtSJW (commented):
Left some comments in the issue.  This approach doesn't seem right to me.

ndouglas:
Gonna close this since the OP issue was closed.

PR #7821, opened 2022-01-31
VACMS-830: Add defensive timestamp check and fallback.

### LINKED ISSUE #830: Taxonomy revisions UI broken for existing taxonomy terms
## Background

#829 / #7725 added revision UI for all entity types other than node.  

Per @swirtSJW 's [comment](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/7775#issuecomment-1023628871), looking at revisions for existing terms results in an error. Creating a new term, then looking at revisions, works fine. 

See also [this comment](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/830#issuecomment-1023664939) containing a hunch for how to fix this. 

## Old issue description 

~In D8.7, taxonomy terms became revisionable. But they are not ready for content moderation yet, until [this issue](https://www.drupal.org/project/drupal/issues/3047110) gets completed:~

~As a manager of a taxonomy (Sections, VHA health services, etc), i want to see who made edits to a taxonomy term, so that i can _blame_ them.~ 

## AC: 
 - [ ] As an content_admin  ~editor~ of a taxonomy (Health services taxonomy, VBA services taxonomy, Sections taxonomy etc) i want to 
    - [ ] see the content editing and moderation history 
    - [ ] revert if needed to older versions, just like nodes.
~On cms.va.gov/section/veterans-affairs i see a History and Revisions tab, just like on a node view.~
- [ ]  content_admin needs to be able see and revert revisions on all taxonomies.
- [ ] content_publisher should nto be able to edit, delete, or create health service taxonomy terms (from comment below)

## No longer required
- [ ] ~On taxonomy term revisions pages, a block appears~
```
<h3>Known issues with revisions for taxonomy terms</h3>
<ul>
  <li>Not all revisions created prior to February 2022 will appear here.</li>
  <li>Some older revisions may be missing their timestamp (the date the term was edited).</li> 
  <li>A revision without a timestamp can not compared (diff’ed) to a newer revision.</li>
  <li>It is not possible to revert to revisions that do not have a timestamp.</li>
</ul> 
```
- [ ] ~Design review on taxonomy/term/{tid}/revisions tab~

ethanteague:
## Description
As @swirtSJW mentioned, the timestamp error we are seeing is the result of entity_diff_ui assuming term revision will have a timestamp. This pr adds a patch to fallback to current unix timestamp when revision doesn't have one.

closes #830

## Testing done
Visual

## Screenshots
**NO**
image

**YES**
image

## QA steps
 - [ ] As an administrator, go to `/taxonomy/term/184/revisions` and visually verify the presence of a diff ui page


swirtSJW (changes requested):
The patch looks good.  Sorry @ethanteague,  This needs some perms so that editors and higher can view revisions and revert.  I should have carried that forward from the other ticket.  I just updated it on #830 but I think it should not be content_editor  I think who has these perms on taxonomies needs some discussion, because I am not sure the current perms are correct.  Let me sort this out with Kev before proceeding on this.

swirtSJW (changes requested):
@ethanteague   This is the refined ACs around perms that need to be added.
https://dsva.slack.com/archives/CT4GZBM8F/p1643741393594629
Sorry about the swirl.

swirtSJW:
We might need more than just the patch as is.   Trying to diff the revision between the original and the new throws an error.
image

On trying to diff 

image


This might be resolveable... since there is no legit revision there.

swirtSJW:
When creating that link, it doesn't have a revision number to use to it just uses the term id.
image

swirtSJW:
@ethanteague  unfortunately I think the patch is just hiding the cause.  There are no revisions for the exisiting taxonomy.  Adding a bunch of checks for date time all over the place will cover the problem but not really provide the solution.   Sorry for steering you down the wrong path.    I think the patch needs to consist  of a hook_install (not hook_update) that loads up each term and creates a proper revision for it.  otherwise diff and revert are going to keep breaking when trying to look at the current state to new revision.

kevwalsh:
How/when will that script be run on the Tugboat environment for QA?

ethanteague:
> How/when will that script be run on the Tugboat environment for QA?

Left you hanging. Just did it, @kevwalsh

swirtSJW (approved):
Great work on this

swirtSJW:
image

swirtSJW:
script has been run on prod.
we are all good

PR #7848, opened 2022-02-02
VACMS-6057_7285 timezone no null

### LINKED ISSUE #7285: Make timezone data more rigid
## User Story or Problem Statement

As a a content API user I would like there to be consistency in the data provided by the smartdate field in regards to timezone so that I know how to handle the times I am getting from the API.

Currently, if the timezone field is left on the default entry, no data is provided for the timezone property
image
So even though the default is NY, the timezone is empty.

Upon a follow-up save of the node, or specifically setting the timezone to NY, the timezone is provided.
image

In either case the time provided is for NY, but the API consumer has no way to know that from the data.

Two possible approaches
  A)  A data resolver that sets "timezone" = "America/New_York"  whenever timezone is empty.  
  pro: This would account for all current instances in the db and would fix it across multiple content types.
  con: a GraphQL resolver would have no impact on the new JSON API so it might need its own treatment. 

B) JS that removes the "default entry" and forces the actual selection of NY.  This would fix any future uses.
pro  Makes sure the data goes in right.
con  Would need a database update process that fixes any existing revisions that have no timezone set.
  

## Acceptance Criteria
- [ ] Testable_Outcome_X
- [ ] Testable_Outcome_Y
- [ ] Testable_Outcome_Z
- [ ] Requires design review
- [ ] Requires KB article update

## Implementation steps
<optional>

## Design principles

Veteran-centered
- [ ] `Single source of truth`: Increase reliability and consistency of content on VA.gov by providing a single source of truth.
- [ ] `Accessible, plain language`: Provide guardrails and guidelines to ensure content quality.
- [ ] `Purposely structured content`: Ensure Content API can deliver content whose meaning matches its structure.
- [ ] `Content lifecycle governance`: Produce tools, processes and policies to maintain content quality throughout its lifecycle.

Editor-centered
- [ ] `Purpose-driven`: Create an opportunity to involve the editor community in VA’s mission and content strategy goals.
- [ ] `Efficient`: Remove distractions and create clear, straightforward paths to get the job done.
- [ ] `Approachable`: Offer friendly guidance over authoritative instruction.
- [ ] `Consistent`: Reduce user’s mental load by allowing them to fall back on pattern recognition to complete tasks.
- [ ] `Empowering`: Provide clear information to help editors make decisions about their work.

### CMS Team

Please leave only the team that will do this work selected. If you're not sure, it's fine to leave both selected.

- [x] `Platform CMS Team`
- [x] `Sitewide CMS Team`

ethanteague:
## Description
closes #7285

## Testing done
`lando drush scr ./scripts/content/VACMS-7825_add_timezone_value_to_empty_db_column.php`

## QA steps
 - [ ] As an admin, go to `/node/add/event` and in timezone, visually verify that the ` - default America/New_York - ` option is disabled.
 - [ ] Run `lando drush scr ./scripts/content/VACMS-7825_add_timezone_value_to_empty_db_column.php` and wait for process to complete
 - [ ] Go to log and make sure message appear as expected
 - [ ] Run `lando drush sqlc` then from mysql prompt run `SELECT * FROM node__field_datetime_range_timezone WHERE field_datetime_range_timezone_timezone='';` visually verifying no results are returned.


swirtSJW (changes requested):
This looks like it had a git catastrophe.  Has way more in it than I expected.

swirtSJW (commented):
Sorry just realized this was all blocked by recurring ....  I was just expecting option A for a data resolver that adjusts it on the way out.

ethanteague:
> Sorry just realized this was all blocked by recurring .... I was just expecting option A for a data resolver that adjusts it on the way out.

@swirtSJW I don't think the resolver approach is gonna work. Seems like this came up before: with fields that are already defined / have schemas created by other modules, we can't use any of the Deriver methods found in `Drupal\graphql_core\Plugin\Deriver\Fields\` namespace. We need to be able to use the `Drupal\graphql_core\Plugin\Deriver\Fields\EntityFieldPropertyDeriver` like you are in EFF, but can't get into it without overwriting the field such that it's properties would no longer be available in a graphQL query - we would end up with one big dump, without the option to pick the value, or just the rrule, etc.

ethanteague:
If I had my druthers, I would just call this one done with current approach (it's working). Wrt derivers, I've put in a pretty decent amount of time trying to solve, but the documentation is pretty thin. I don't know the correct annotation to use to make this work (or if there is a way for a multi valued `SmartDateFieldItemList` to work with `EntityFieldPropertyDeriver`). If we do want to pursue this path, I'm gonna need an assist. Here's what I'm doing, in `docroot/modules/custom/va_gov_graphql/src/Plugin/GraphQL/Fields/FieldDatetimeRangeTimezone.php` @swirtSJW :
```
<?php

namespace Drupal\va_gov_graphql\Plugin\GraphQL\Fields;

use Drupal\Core\Entity\FieldableEntityInterface;
use Drupal\graphql\GraphQL\Execution\ResolveContext;
use Drupal\graphql\Plugin\GraphQL\Fields\FieldPluginBase;
use GraphQL\Type\Definition\ResolveInfo;

/**
 * Operates on fieldDatetimeRangeTimezone recurring fields.
 *
 * @GraphQLField(
 *   id = "smartdate",
 *   secure = true,
 *   name = "smartdate",
 *   type = "SmartDateFieldItemList",
 *   field_types = {"date"},
 *   multi = true,
 *   parent = {"Entity"},
 *   deriver = "Drupal\graphql_core\Plugin\Deriver\Fields\EntityFieldPropertyDeriver"
 * )
 */
class FieldDatetimeRangeTimezone extends FieldPluginBase {

  /**
   * {@inheritdoc}
   */
  public function resolveValues($value, array $args, ResolveContext $context, ResolveInfo $info) {
    $blah = '';
    if ($value instanceof FieldableEntityInterface && $value->hasField('field_datetime_range_timezone')) {
      $dates = array_filter($value->field_datetime_range_timezone->view(), function ($key) {
        return is_numeric($key);
      }, ARRAY_FILTER_USE_KEY);
      foreach ($dates as $key => $date) {
        yield [
          $key => $date,
        ];
      }
    }
  }

}
```

PR #7887, opened 2022-02-03
VACMS-0000: Various updates to NPM packages. [DO NOT MERGE]

ndouglas:
Purely experimental.


package.json
```diff
@@ -57,24 +57,24 @@
...
+    "eslint-plugin-react": "^7.28.0",
+    "eslint-plugin-react-hooks": "^4.3.0",
     "json-schema": ">=0.4.0",
-    "prettier": "^2.2.1",
-    "stylelint": "^13.8.0",
-    "stylelint-config-sass-guidelines": "^7.1.0",
-    "stylelint-order": "^4.1.0"
+    "prettier": "^2.5.1",
+    "stylelint": "^14.3.0",
+    "stylelint-config-sass-guidelines": "^9.0.1",
```
tjheffner, line 76: love these dep updates! i'm curious if any of the stylelint updates (particularly this one, that our .stylelintrc rules extend) raise errors with how we have written styles on the project to this point? if `composer va:test:lint-css-all` comes back clean i have no issues.

ndouglas: It's breaking a few different things, I think.  It's causing issues with behavioral tests too, and it failed with the command you mentioned above (we should probably add a test for that).  I think I'm gonna bail on this experiment.

PR #7961, opened 2022-02-10
VACMS-0000: Cleans up a config discrepancy.

ndouglas:
@ryguyk was complaining about this.  Just cruft that wasn't in config but will show up until someone does this.


swirtSJW:
Nice fix Nate.  Too bad dependabot can't run config export  and then commit the changes.

swirtSJW (approved):

Issue #8014, opened 2022-02-14
PDF Audit: List shows where PDFs appear


msbtterswrth:
## Description
Stretch task of #6980 
- I want to know what pages these PDFs have been added to, so that i can understand the context

## Acceptance Criteria
- [ ] List shows where these PDFs appear.

## Implementation Ideas
There are two modules that may help us build something: https://www.drupal.org/project/entity_usage and possibly https://www.drupal.org/project/entity_usage_views_field

### CMS Team
Please check the team(s) that will do this work. 

- [ ] `CMS Program`
- [ ] `Platform CMS Team`
- [ ] `Sitewide CMS Team ` (leave Sitewide unchecked and check the specific team instead)
  - [x] `⭐️ Content ops`
  - [ ] `⭐️ CMS experience`
  - [ ] `⭐️ Offices`
  - [ ] `⭐️ Product support`
  - [ ] `⭐️ User support`

kevwalsh:
We may want to convert this into a broader spike ticket to evaluate all “library” user stories. Like

- how many times is an alert block used (i know one alert block is reused 150+ times)
- how many times a taxonomy term is referenced (like at /vha-health-services )
- how many times a Q&A is reused.
- how many times a VA Form is referenced (see our other thread from today about the PDF audit, and #3232 for related user stories.
- etc etc

The module has some other nice features like providing warnings to editors who are editing reused content. Would be nice to potentially extend this for the “Archive” action, and let the editor know, eg “There are 25 published facility services that reference this service. You must archive those first”. (epic #2833)

<details><summary>Screenshot</summary>

![image](https://user-images.githubusercontent.com/643678/154184698-0342e1f6-b9c8-438a-94bb-f7a80e6666c9.png)


</details>

Then, we should add an AC for this spike to evaluate/measure performance costs. 
- Does this slow down every node save? 
- How often will need to run the batch update? 
- etc

rachel-kauff:
@msbtterswrth Is it safe to assume that this work and anything related to it (say, showing where images appear) is fairly far off? It would add depth to my alt text audit, but I sense it's prob blocked / a big lift?

rachel-kauff:
Oh great! Prob not in time for this current audit, but means we can iterate later on.

mgifford:
[...]

PR #8037, opened 2022-02-15
VACMS-8014: Add Entity Usage module.

msbtterswrth:
## Description
Adds Entity Usage to the platform.

What it does:
It tracks references across entities, and allows one to view (in a local task/tab) where the current entity is being used. This is helpful in content governance where a user might need to know where an entity is being used as a means to determine if un-publishing a node is possible based on what links to elsewhere in the CMS, for instance.

How it works:
Entity Usage maintains usage statistics in a dedicated database table, `entity_usage`. Ongoing tracking takes place during entity save operations. However, there is a one-time initial scan of all (configured) entities to seed the database. Because of the intense nature of scanning all nodes in the database, this process can take several hours or longer to complete.

Initial seeding:
Seeding can done either via a button on the Entity Usage settings form (/admin/config/entity-usage/settings), or via a Drush command.

PHP timeouts will occur if one attempts to use the UI for seeding the usage table, even despite the fact that it is using Drupal batch operations.

Fortunately the module ships with a Drush command for seeding the usage statistics `entity-usage:recreate`. However, using that command without options will also timeout after 4 hours (the PHP script timeout). The recommended approach is to use the `--use-queue` option when initially seeding. This will cause the queue to be populated with the items to track. This is a cron queue implementation with a 60 second timeout. Each cron run will process thousands of nodes (~11,000 locally). The queue will have a few million records once fully populated, so there could be hundreds of cron runs necessary to fully populate the usage statistics table.

Relates to #8014 .

## Screenshots
Screen Shot 2022-06-09 at 3 44 43 PM
Screen Shot 2022-06-09 at 3 45 46 PM

## QA steps

As an editor
1. Edit this node /pension/aid-attendance-housebound
2. Re-save it as Published
3. Visit  /node/588/usage (this node is linked from the first node via a rich text field on a Paragraph reference field on the node.
4. Verify the node is being tracked (should look like screenshot 1 above)

### Definition of Done

### Select Team for PR review

- [ ] `Platform CMS Team`
- [x] `Sitewide CMS Team`
  - [ ] `⭐️ Content ops`
  - [ ] `⭐️ CMS Experience`
  - [ ] `⭐️ Offices`
  - [ ] `⭐️ Product Support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


msbtterswrth:
FYI I'm handing this off to @dsasser to do some configuring/performance improvements perhaps so we can start using this for audits and other things

config/sync/entity_usage.settings.yml
```diff
@@ -0,0 +1,22 @@
...
+_core:
+  default_config_hash: AXUBMTvVtpcNp0jKjDaWjSHqMy6HjuUf7j4yVbXgazI
+track_enabled_source_entity_types:
+  - node
+  - menu_link_content
+  - danse_event
+  - paragraph
+  - site_alert
+track_enabled_target_entity_types:
+  - media
```
dsasser, line 14: @msbtterswrth this would prevent tracking references from nodes to other nodes. Is that intentional?

config/sync/entity_usage.settings.yml
```diff
@@ -0,0 +1,22 @@
...
+edit_warning_message_entity_types: {  }
+delete_warning_message_entity_types: {  }
+track_enabled_plugins:
+  - linkit
```
dsasser, line 18: I would suggest we enable all the trackers, unless there is a compelling reason to not do so?

config/sync/node_link_report.settings.yml
```diff
@@ -22,6 +22,10 @@ domains_to_skip: |-
   https://acemapp.org/
   https://fcp.vetpro.org
   imablefoundation.org
+  militaryonesource.mil
```
dsasser, line 25: Are these changes expected in this PR? If so, what do these changes do (pardon my naiveness)?

composer.json
```diff
@@ -64,6 +64,7 @@
         "drupal/entity_reference_revisions": "^1.7",
         "drupal/entity_reference_unpublished": "^1.1",
         "drupal/entity_reference_validators": "^1.0@alpha",
+        "drupal/entity_usage": "^2.0@beta",
```
dsasser, line 67: Not necessarily related to this line of code, but what user roles should be given entity usage tab permissions? Entity usage provides three permissions, so identifying which roles should be given which permissions would be helpful. Permissions available:

```
access entity usage statistics
Users with this permission will be able to see statistics of entity usage.

perform batch updates entity usage
Users with this permission will be able to reset/relaunch the statistics counts on all existing entities in the system.

administer entity usage
Users with this permission will be able to administer the Entity usage module settings.
```

config/sync/entity_usage.settings.yml
msbtterswrth: nope! def want to track node -> node references

dsasser:
This command will display how many items are left in the Entity Usage processing queue:
```
drush queue:list --filter=entity_usage_regenerate_queue --format=string --fields=items
```

ndouglas:
Closed in favor of #9575.

Issue #8046, opened 2022-02-16
GraphQL module throws log notice on any query


timcosgrove:
## Description
Every GraphQL query writes a notice to logs. This has been noted as an issue on the GraphQL module issue queue:
https://github.com/drupal-graphql/graphql/issues/1229, and fixed with a simple patch in https://github.com/drupal-graphql/graphql/issues/1230.

The notices show up in both dblog and Grafana, and make it difficult to parse for more useful information. Every single GraphQL query causes this notice to be logged, which means any content build can cause the message to be logged hundreds of times.

## Acceptance Criteria
- [ ] GraphQL queries will no longer cause notices to be written to logs
- [ ] Logs will be easier to parse and examine, and the volume of text logged and transferred will be greatly reduced.

## Testing
**prod.cms.va.gov (this shows the current state):**
1. Visit https://prod.cms.va.gov/admin/reports/dblog and note the state of logged messages.
2. In a separate tab, visit https://prod.cms.va.gov/graphql/explorer, and make any GraphQL query. The following is fine:
```
query MyQuery {
  nodeQuery {
    count
  }
}
```
3. Reload the log page. You should see a message "Notice: Trying to access array offset on value of type bool in ....", with your user 

<img width="1687" alt="image" src="https://user-images.githubusercontent.com/203623/154183875-ce315668-674f-460f-9b66-d66f531f6bc1.png">

**Pull request environment for this ticket:**
1. Visit https://cms-of3cmpeckdhn8oupo0xxucrxglmn0hxa.ci.cms.va.gov/admin/reports/dblog and note the state of logged messages.
2. In a separate tab, visit https://cms-of3cmpeckdhn8oupo0xxucrxglmn0hxa.ci.cms.va.gov/graphql/explorer, and make any GraphQL query. The following is fine:
```
query MyQuery {
  nodeQuery {
    count
  }
}
```
3. Reload the log page. You should not see any new messages.
4. Run a content release, on Tugboat.
5. Reload the log page. You should not see any new messages of the form "Notice: Trying to access array offset on value of type bool in ....".

<img width="1655" alt="image" src="https://user-images.githubusercontent.com/203623/154184045-80db3b18-abc0-4831-b124-30c61f91f756.png">

## Screenshots
dblog:
![image.png](https://images.zenhubusercontent.com/614213ae32396d2bc1205abd/ff093e66-2f19-456b-a3cc-d25ec9123770)

grafana:
[...]

PR #8047, opened 2022-02-16
VACMS-8046: add patch to stop graphql throwing notices.

timcosgrove:
## Description

Relates to #8046. See that issue for tests, screenshots, etc.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -405,7 +405,8 @@
                 "1482958-47 Allow field_group with empty fields to be displayed via setting": "https://www.drupal.org/files/issues/2020-06-24/field_group-empty-config-1482958-47.patch"
             },
             "drupal/graphql": {
-                "Fix Explorer with Layout Builder enabled - https://github.com/drupal-graphql/graphql/pull/1144": "patches/graphql_8x3x_layout_builder_incompat.patch"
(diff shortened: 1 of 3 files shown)


ndouglas:
```
-    // By default we assume a 'single' request. This is going to fail in the
-    // graphql processor due to a missing query string but at least provides
-    // the right format for the client to act upon.
```

ndouglas (dismissed):
GraphQL's definitely much, much quieter, and tests passed.  LGTM.

Screen Shot 2022-02-16 at 7 09 22 AM

ndouglas (approved):

amponce:
Do we know what the original intention of using _controller for single requests was supposed to be?

ndouglas:
@amponce It seems pointless to me:

Use this request controller: https://github.com/drupal-graphql/graphql/blob/8.x-3.x/src/Routing/QueryRoutes.php#L57-L58

which handles single or multiple requests:
https://github.com/drupal-graphql/graphql/blob/8418bb7f6a869d0482eb45bfe22efa2f817d7efd/src/Controller/RequestController.php#L65

Maybe that line was vestigial from some prior implementation of the request controller.

PR #8054, opened 2022-02-16
[docs] Refine steps for VAMC - name change

JayDarnell:
The URL alias for the facility itself must be updated first or the bulk updates to facility service nodes will not reflect the new URL.


swirtSJW (approved):
Looks good.

Issue #7894, opened 2022-02-03
CMS events that trigger Content release should not trigger unnecessary workflow runs


timcosgrove:
## Description
Currently, when a CMS event happens that triggers a Content release - i.e., saving a facility alert - the Content release is triggered immediately, one for every event.

When many such events happen in a short period, it triggers one Content release per event. This mostly causes visual clutter, since all but the last Content release will be cancelled. Since this also creates a large number of immediately cancelled runs, it skews metric averages and counts for workflow runs unnecessarily.

These triggering events should only result in as many content release requests as is necessary. If there is a content release pending (i.e. requested but not yet started), there is no need to make an additional request. 

Once the content release is in progress, a new content release should be allowed to queue.

## Acceptance Criteria
- [ ] Unnecessary content release requests and cancellations are eliminated
- [ ] When a CMS event would trigger a content release, if there is no pending release, a content release is triggered
- [ ] When a CMS event would trigger a content release, if there is a pending release, no content release is triggered.

## Implementation details
When the current system was implemented, the situation was not noticed because we were using Jenkins for content release. Jenkins automatically does grouping and debouncing. Github Actions does not, and so we are seeing many superfluous triggers of Content releases.

The specifics of the implementation are up to the engineer who does this work. 

### CMS Team

Please leave only the team that will do this work selected. 

- [x] `Platform CMS Team`
- [ ] `Sitewide CMS Team`
  - [ ] `⭐️ Content ops`
  - [ ] `⭐️ CMS experience`
  - [ ] `⭐️ Offices`
  - [ ] `⭐️ Product support`
  - [ ] `⭐️ User support`

timcosgrove:
> What I do not like about this approach is "grouped and sent out on an agreed upon interval (5 minutes is a suggestion)." it delays content release by up to another 5 minutes. Roughly 15% increase in time for content to go out. We are trying to move the needle in the other direction.

@swirtSJW  I really appreciate this feedback. I've modified the description and AC to be more explicitly something like a debounce pattern. In the absence of recent triggers, a triggering event will immediately send a content release request. Following requests will be held in queue, subject to a cooldown interval. At the end of the cooldown interval, any number of queued requests will result in a single further content release request.

It is not an actual debounce. Debounce typically is:
* initial event comes in and is honored
* for a time period after the initial event, subsequent events are ignored
* once that time period has passed, events will be honored again

Our approach honors events that come in during the cooldown period; it just treats them as a single event.

[...]

PR #8095, opened 2022-02-18
VACMS-7894: do not trigger a workflow run if one is already pending.

timcosgrove:
## Description

Relates to #7894


docroot/modules/custom/va_gov_build_trigger/src/Plugin/VAGov/Environment/BRDGHA.php
```diff
@@ -82,18 +82,36 @@ public static function create(ContainerInterface $container, array $configuratio
...
       ];
-      $message = $this->t('Site rebuild request has been triggered. Please visit <a href="@job_link">@job_link</a> to see status.', $vars);
+      $workflow_runs = $this->githubAdapter->listWorkflowRuns('content-release.yml', $workflow_run_params);
+      if (!empty($workflow_runs['total_count']) && $workflow_runs['total_count'] > 0) {
+        $vars = [
+          '@job_link' => 'https://github.com/department-of-veterans-affairs/content-build/actions/workflows/content-release.yml',
+        ];
+        // This message is intentional. We are reporting that the request has
+        // been processed, which is true.
+        $message = $this->t('Site rebuild request has been processed. Please visit <a href="@job_link">@job_link</a> to see status.', $vars);
```
swirtSJW, line 103: This message seems a bit misleading. 
Must fix:  For editors we use the term "Content release".   Site rebuild is for us nerds. I'm not sure how that original wording snuck in.  
For the rest of it, I'd like to keep the message honest.  Something like.
"This content release request will be covered by one that is already in progress.  Please visit ..."

swirtSJW (commented):
Overall this looks good.  I made a few suggestions.

One point of failure that I see with this that would be covered by CMS just keeping state, or queueing is that if the content build that is being relied upon as already existing, or newly triggered fails,   it will not try again... it will fail and nothing will go out.

If we kept state in the CMS and the time gap was too big, it would make another request.  .... .... ... until it has success.

docroot/modules/custom/va_gov_consumers/src/Git/GithubAdapter.php
```diff
@@ -69,7 +69,7 @@ public function searchPullRequests(string $search_on, int $count = 10): array {
    * {@inheritdoc}
    */
   public function triggerWorkflow(string $action_name, string $ref, array $params = []): void {
-    list($user, $repo) = explode('/', $this->repositoryPath);
+    [$user, $repo] = explode('/', $this->repositoryPath);
```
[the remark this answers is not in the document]
timcosgrove: It failed githooks; wouldn't let me commit it like this.

docroot/modules/custom/va_gov_build_trigger/src/Plugin/VAGov/Environment/BRDGHA.php
```diff
@@ -82,23 +82,31 @@ public static function create(ContainerInterface $container, array $configuratio
...
     try {
-      $this->githubAdapter->triggerWorkflow('content-release.yml', $front_end_git_ref, [
-        'release_wait' => "0",
-        'deploy_environment' => $this->settings->get('github_actions_deploy_env'),
-      ]);
-      $vars = [
-        '@job_link' => 'https://github.com/department-of-veterans-affairs/content-build/actions/workflows/content-release.yml',
-      ];
-      $message = $this->t('Site rebuild request has been triggered. Please visit <a href="@job_link">@job_link</a> to see status.', $vars);
+      if ($this->pendingWorkflowRunExists()) {
```
timcosgrove, line 89: put this into a method as suggested.

docroot/modules/custom/va_gov_build_trigger/src/Plugin/VAGov/Environment/BRDGHA.php
```diff
@@ -82,23 +82,31 @@ public static function create(ContainerInterface $container, array $configuratio
...
+        $vars = [
+          '@job_link' => 'https://github.com/department-of-veterans-affairs/content-build/actions/workflows/content-release.yml',
+        ];
+        $message = $this->t('An upcoming content release is already pending and will include these changes. Please visit <a href="@job_link">@job_link</a> to see content release status.', $vars);
```
timcosgrove, line 93: I don't love this language because it will show any time this situation arises, including when a truly manual build request arises and there are therefore no 'these changes'. But, we're erring on the side of assuring the CMS editor here, so hopefully this is OK.

docroot/modules/custom/va_gov_build_trigger/src/Plugin/VAGov/Environment/BRDGHA.php
timcosgrove: Language updated w/ input from @rachel-kauff .

docroot/modules/custom/va_gov_build_trigger/src/Plugin/VAGov/Environment/BRDGHA.php
```diff
@@ -82,18 +82,36 @@ public static function create(ContainerInterface $container, array $configuratio
...
+      }
+      else {
+        $this->githubAdapter->triggerWorkflow('content-release.yml', $front_end_git_ref, [
+          'release_wait' => "0",
+          'deploy_environment' => $this->settings->get('github_actions_deploy_env'),
+        ]);
+        $vars = [
+          '@job_link' => 'https://github.com/department-of-veterans-affairs/content-build/actions/workflows/content-release.yml',
+        ];
+        $message = $this->t('Site rebuild request has been triggered. Please visit <a href="@job_link">@job_link</a> to see status.', $vars);
```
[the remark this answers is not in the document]
timcosgrove: language updated w/ input from @rachel-kauff

docroot/modules/custom/va_gov_build_trigger/src/Service/BuildFrontend.php
```diff
@@ -178,10 +178,11 @@ public function triggerFrontendBuildFromContentSave(NodeInterface $node) {
         return;
       }
       $this->triggerFrontendBuild();
+      // We previously also showed this message to the user, but the build
+      // trigger mechanism also needs to show a message, so this was redundant.
+      // We still send logs as they may be useful for diagnostics.
```
swirtSJW, line 183: minor:  I'd prefer  to let the commits tell the story of history of anything no longer in code.   Leaving a bunch of grave markers in the code is going to get messy and wordy over time.

docroot/modules/custom/va_gov_build_trigger/src/Plugin/VAGov/Environment/BRDGHA.php
```diff
@@ -123,4 +124,43 @@ public function getBuildTriggerFormClass() : string {
     return BrdBuildTriggerForm::class;
   }
 
+  /**
+   * Check for a pending content-release workflow run.
+   */
+  private function pendingWorkflowRunExists() : bool {
+    try {
+      // We limit our check to today, as some jobs long past report as being
```
swirtSJW, line 132: This comment is great and I appreciate it being here.   The ramifications of it scare me a bit.   Are you saying that when you query GHA that we can get counts that do not reflect what are actually going to run?   Can we not trust this count?  

Two concerns:
1) throwing away the previous day for a  12:30AM request is going to give unexpected results.
2) If we can't trust the pending jobs count  from back farther than midnight, how can we trust them for 12 or even 6 hours before?   If "stuck" job from 8AM prevents a request from 11AM from being made, that is an assumption that is going to fail and leave us looking bad.  As it is, this is not a self-correcting system, and it needs to be.

docroot/modules/custom/va_gov_build_trigger/src/Service/BuildFrontend.php
timcosgrove: The commits don't tell the story because we squash and all useful intermediary commit messages are lost. We'll have no history of why this was removed. I think squashing is good, mind you, but it makes code spelunking more difficult. 

This comment is there to explain the contrast to the if condition just above this, where we *do* show a message. If I were looking at this naively, I'd wonder why we set a message when a build isn't triggered, but we don't set one when we do trigger.

docroot/modules/custom/va_gov_build_trigger/src/Plugin/VAGov/Environment/BRDGHA.php
timcosgrove: > 1. throwing away the previous day for a 12:30AM request is going to give unexpected results.

In this case, a content release would be triggered. The system errs on the side of triggering a content release unless it gets clear indication that it shouldn't. So in this specific case, we might see a 'duplicative' content release. I'm ok with that. 

> 2. If we can't trust the pending jobs count from back farther than midnight, how can we trust them for 12 or even 6 hours before?

This is a good point. I did date-without-time because it was fairly straight-forward to do without getting into math. It was more elegant / lazier, take your pick :) 

This can be shifted to checking for the last 3 hours. Would that make you more comfortable?

timcosgrove: Also, shifting to 3 hours prior would alleviate your concerns in point 1.

docroot/modules/custom/va_gov_build_trigger/src/Service/BuildFrontend.php
swirtSJW: They do still tell the story because we squash and merge, not fixup and merge  No merge commit comments long or short are lost in the process.
image

swirtSJW:
```suggestion
      // Since a build requested message is already displayed, just log what triggered it.
```
This explains why, without tales of yesteryear ;)

docroot/modules/custom/va_gov_build_trigger/src/Plugin/VAGov/Environment/BRDGHA.php
swirtSJW: I think a shift to three hours before, reduces the window of risk. A shift to 2... better yet.... Not sure where the minimum threshold ought to be.   I am thinking 1 hour might even work.

timcosgrove: shifted to 2 hours.

swirtSJW (approved):
Nice work @timcosgrove   This looks solid.

PR #8117, opened 2022-02-23
Removes platform-overrides section from composer.lock

ndouglas:
## Description

This is causing PHPStan failures and Tugboat deployment failures when packages require PHP 7.4.


swirtSJW (approved):

swirtSJW:
image

PR #8263, opened 2022-03-04
VACMS-8165: Update admin experience for ehr on form and node pages.

### LINKED ISSUE #8165: Enhance UI guidance for admins on VA Health Connect phone number and Cerner/VistA content
## Background

See design implemented for #6572. 

As a CMS admin, I would like to see guidance on the content for the VA Health Connect phone number and the Cerner/VistA strategy. Structuring these fields with more guidance would allow admins to understand where the data is coming from and where it is going. 

Current design uses tooltip for editors, but this may not be the right use case for that component for admins. Discovery should include if this admin view would be better served with help text vs. the tooltip or something else. Also, we need to confirm the correct style for the phone number field from the VA.gov design system.

</details>

## ACs
- [x] Create guidance on data for admins through help text or something else
- [x] Create null field text for editors to see nothing exists yet
- [x] Confirm correct style for phone numbers in the VA.gov design system.
- [x] Dev feasibility review with Steve and Ethan

### LINKED ISSUE #8190: Ensure all question mark icons are transparent on tooltips
## Describe the defect

Some of the tooltips use question marks that have a grey background, which looks funky on a couple places, on white or on various greys. 

Screenshot_2022-02-28__1_32_PM

## To Reproduce
1. As admin, go to https://staging.cms.va.gov/node/1928/edit
2. See the grey square, which should be just a black circle. 

There are other places where this is visible, grey on grey. 

## Expected behavior
The icon should always be trnasparent

## Additional context
Add any other context about the problem here. Reach out to the Product Managers to determine if it should be escalated as critical (prevents users from accomplishing their work with no known workaround and needs to be addressed within 2 business days).

 
## Labels
- [ ] CMS subsystem (green)

### CMS Team

- [ ] `CMS Program`
- [ ] `Platform CMS Team`
  - [ ] `⭐️ Content ops`
  - [ ] `⭐️ CMS experience`
  - [ ] `⭐️ Offices`
  - [ ] `⭐️ Product support`
  - [ ] `⭐️ User support`

ethanteague:
## Description
closes #8165
closes #8190

## Testing done
Visual / behat

## Screenshots
image
image

## QA steps
 - [ ] As an administrator, go to `node/318/edit` and visually verify EHR fieldset now has tooltip and beige background color. and tooltip doesn't have a different colored background
 - [ ] Try to add a phone number with a goofy format, like `123456-7890` and verify a form validation error on submit
 - [ ] Save a published version of the form without a phone number, and verify `Undefined` text appears on node view


EWashb:
Everything looks good to me @ethanteague. I don't see anything that needs to be changed. Good work!

EWashb (approved):

config/sync/node_link_report.settings.yml
```diff
@@ -47,6 +48,7 @@ domains_to_skip: |-
   www.ridelats.com
   www.sirweb.org
   www.spiritrock.org
+  www.teamusa.org
```
swirtSJW, line 51: USA USA USA   :)

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -218,7 +220,7 @@ public function getProductTypeTermFields(array &$form, FormStateInterface $form_
   public function appendHealthServiceTermDescriptionToVetCenter(EntityViewAlterEvent $event):void {
     if ($event->getDisplay()->getTargetBundle() === 'vet_center') {
       $build = &$event->getBuild();
-      $services = isset($build['field_health_services']) ? $build['field_health_services'] : [];
+      $services = isset($build['field_health_services']) ?? $build['field_health_services'];
```
swirtSJW, line 223: major: I don't think this is equivalent.   The original  makes sure services is always an array so it does not throw a notice in the foreach.
The new line makes it TRUE or nothing.

image


Null Coalescing operator only works as expected with NULL or undefined, not with bool.  And it will never result in an empty array like the first one.

swirtSJW (changes requested):
Something pretty major.

swirtSJW:
Also... conflicts.

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
ethanteague: Amazed you were able to suss that out, @swirtSJW :heart: - my sniffer was beefing about me using a ternary and phpcbf was saying that I should use coalesce. I updated to set it as an array by default, and then set services in a conditional.

docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -221,7 +222,10 @@ public function getProductTypeTermFields(array &$form, FormStateInterface $form_
   public function appendHealthServiceTermDescriptionToVetCenter(EntityViewAlterEvent $event):void {
     if ($event->getDisplay()->getTargetBundle() === 'vet_center') {
       $build = &$event->getBuild();
-      $services = isset($build['field_health_services']) ? $build['field_health_services'] : [];
```
swirtSJW, line 224: Sorry.  This seems off.   I think the recommendation to use null coalesce was to do this:

`$services = $build['field_health_services'] ?? [];`

which is equivalent to this

`$services = isset($build['field_health_services']) ? $build['field_health_services'] : [];`

but is not equivalent to this 
`$services = isset($build['field_health_services']) ?? $build['field_health_services'];`

swirtSJW (approved):
Fully tested  QA passes.   Nice work Ethan.

PR #8303, opened 2022-03-09
VACMS-8135: Flag theming breaks some views config

### LINKED ISSUE #8135: Flag theming breaks some views config
## To Reproduce
1. Edit any view based on nodes (eg /admin/content)
2. Click the relationships tab
3. See that the table is out of whack
4. See that the `tr.flag` is the one that is breaking the layout

My guess is that this may be affecting other Config UI. 

## Expected behavior
View UI isn't 
## Screenshots 

Cursor_and_Top_task_pages__Content____VA_gov_CMS

 ## Additional context

Seems like we need to be a bit more specific about theming the flags on the /admin/content/flagged view
 

### CMS Team

- [ ] `CMS Program`
- [ ] `Platform CMS Team`
  - [x] `⭐️ Content ops`
  - [ ] `⭐️ CMS experience`
  - [ ] `⭐️ Offices`
  - [x] `⭐️ Product support`
  - [ ] `⭐️ User support`

netboss:
- Updated _views.scss and added CSS specificity to override default "flag" CSS in Views.

## Description

closes #8135 

## Testing done
Passes all tests

## Screenshots
VACMS-8135-flag-row

VACMS-8135-flag-row-hover

VACMS-8135-flagged-content

## QA steps

As user administrator
1. Go to `/admin/structure/views/view/content` (or any node related view):
    - [x] In the right column under "Advanced" click the "Add" button next to "Relationships" and scroll down until you find the "Content flag" and visually inspect to verify and confirm that the table row ins't broken any more and that the hover and focus effects works the same as other table rows
2. Go to `/admin/content/flagged` (or in `/admin/content` click on the "Flagged" tab):
    - [x] Verify and confirm that the "flags" CSS styling isn't broken or odd looking

### Definition of Done

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide CMS Team`
  - [ ] `⭐️ Content ops`
  - [ ] `⭐️ CMS Experience`
  - [ ] `⭐️ Offices`
  - [ ] `⭐️ Product Support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/docroot/themes/custom/vagovclaro/assets/scss/components/_flags.scss
+++ b/docroot/themes/custom/vagovclaro/assets/scss/components/_flags.scss
@@ -1,28 +1,29 @@
-//Styling for Flags
-.flag {
-  background: var(--va-gray-lighter);
-  border-radius: 30px;
-  display: inline-block;
-  padding: 4px 16px;
-  white-space: nowrap;
+// Styling for Flags
+.view-flagged-content {
+  .flag {
+    background: var(--va-gray-lighter);
+    border-radius: 30px;
+    display: inline-block;
+    padding: 4px 16px;
+    white-space: nowrap;
 
-  &.new {
-    background: var(--va-blue-light);
+    &.new {
+      background: var(--va-blue-light);
+    }
+
+    &.changed_filename,
+    &.changed_title,
+    &.changed {
+      background: var(--va-gold-lightest);
+    }
+
+    &.deleted,
+    &.removed-from-source {
+      background: var(--va-red-lightest);
+    }
+
+    &.removal-complete {
+      background: var(--va-green-lightest);
+    }
   }
-
-  &.changed_filename,
-  &.changed_title,
-  &.changed {
-    background: var(--va-gold-lightest);
-  }
-
-  &.deleted,
-  &.removed-from-source {
-    background: var(--va-red-lightest);
-  }
-
-  &.removal-complete {
-    background: var(--va-green-lightest);
-  }
-
 }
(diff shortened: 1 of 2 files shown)


swirtSJW (changes requested):
I don't think this is the right approach.   We have formatting that should only be impacting the flagged view, but it is not properly limited and is leaking out and affecting other places.  I think it would be better to prevent it leaking out, than to create special cases of overriding it at other places.   

Kev offered this in the ticket which pointed at a better solution 
"Seems like we need to be a bit more specific about theming the flags on the /admin/content/flagged view"

netboss:
> I don't think this is the right approach. We have formatting that should only be impacting the flagged view, but it is not properly limited and is leaking out and affecting other places. I think it would be better to prevent it leaking out, than to create special cases of overriding it at other places.
> 
> Kev offered this in the ticket which pointed at a better solution "Seems like we need to be a bit more specific about theming the flags on the /admin/content/flagged view"

@swirtSJW 
This is ready for review when you have a chance.

swirtSJW:
The code looks good.   Let's consider the QA steps to not only include checking the previously broken views admin, but it should also check the view that may have been changed.

netboss:
> The code looks good. Let's consider the QA steps to not only include checking the previously broken views admin, but it should also check the view that may have been changed.

@swirtSJW 
Updated QA steps to include the "Flagged" content and added screen shot of it showing the "flags" correctly styled.

swirtSJW (approved):
QA tests out good.  Nice work.

PR #8314, opened 2022-03-10
VACMS-0000: Ignores AWS PHP SDK patch updates.

ndouglas:
We get a large amount of PRs relating to the AWS PHP SDK.  

It's a good idea to stay updated, of course, but it's definitely not worth the amount of churn to update on every patch update, especially considering that the PHP SDK covers every AWS service and we only use a couple of them.

This PR instructs Dependabot to only open PRs for minor version updates, not patch updates, which should cut the noise down substantially.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,9 @@ updates:
   allow:
   - dependency-type: direct
   ignore:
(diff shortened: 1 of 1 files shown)


swirtSJW (approved):
That's an awesome solution Nate.   Thank you.

PR #8369, opened 2022-03-16
VACMS-0000: Updates Drupal to 9.3.8.

ndouglas:
https://www.drupal.org/project/drupal/releases/9.3.8

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -25187,5 +25187,5 @@
     "platform-overrides": {
         "php": "7.4"
     },
-    "plugin-api-version": "2.0.0"
+    "plugin-api-version": "2.2.0"
 }
--- a/composer.lock
+++ b/composer.lock
@@ -3841,16 +3841,16 @@
         },
         {
             "name": "drupal/core",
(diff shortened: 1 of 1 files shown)


composer.lock
```diff
@@ -25187,5 +25187,5 @@
     "platform-overrides": {
         "php": "7.4"
     },
-    "plugin-api-version": "2.2.0"
+    "plugin-api-version": "2.0.0"
```
tjheffner, line 25190: just verifying that this bit is intentional... everything else seems to move up a version save for this one

ndouglas: I think Dependabot tends to bump this up to 2.2.0 and most of us use 2.0.0.  It's not a separate package and has no effect  I can discern.

cweagans (dismissed):

tjheffner (dismissed):

ndouglas:
Thanks ❤️

ndouglas:
Merging because it was already approved, just GH fail.

PR #8379, opened 2022-03-17
VACMS-000: Remove phone widget number validation.

ethanteague:
## Description
Relates to #8165
Per convo in slack, we need to remove phone number field format constraint until we have figured out how to handle legacy extension entries in fields that have them.
This pr removes the validation logic that was added in https://github.com/department-of-veterans-affairs/va.gov-cms/pull/8263

## Testing done
Visual

## QA steps
 - [ ] As an administrator, go to node/318/edit and add a phone number with a goofy format, like 123456-7890 and verify that on form submit, an HTML5 phone field validation message is not thrown.


msbtterswrth (approved):
looks good! confirmed that html validation is not present for phone numbers.

PR #8488, opened 2022-03-28
VACMS-8467: Update button mixin to accept arguments.

### LINKED ISSUE #8467: Update button mixin to accept args
## Description
Currently VA design system buttons in the CMS are formatted like the following 
```
usa-button,
  .usa-button-primary {
    @include va-buttons;
    background-color: var(--va-blue-med);

    &:hover {
      background-color: var(--va-blue-darker);
    }
  }
```
We can update this to be more efficient be including arguments in the mixing like `@include va-buttons(primary, primary-dark)`

## Acceptance Criteria
- [ ] Mixin is updated to include arguments for background and background hover

### CMS Team

- [ ] `CMS Program`
- [ ] `Platform CMS Team`
  - [ ] `⭐️ Content ops`
  - [x] `⭐️ CMS experience`
  - [ ] `⭐️ Offices`
  - [ ] `⭐️ Product support`
  - [ ] `⭐️ User support`

msbtterswrth:
## Description
Closes #8467 .
 
## Testing done
Edited a node and navigated to an audit view and confirmed button styles were still functional.

## Screenshots
image
image

## QA steps

As CMS user
1. Edit a node (`/node/35055/edit`) and open the rich text editor
   - [x] Validate that button styles still work as expected
2. Then view the node and
   - [x] Validate that button styles still work as expected
3. Navigate to a view (`/admin/content/media/pdfs`) and
   - [x] Validate that button styles still work as expected
4. Then validate Acceptance Criteria from issue
   - [x] Mixin is updated to include arguments for background and background hover
   - [x] Button styles are unchanged on the frontend of the CMS

### Definition of Done

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide CMS Team`
  - [ ] `⭐️ Content ops`
  - [x] `⭐️ CMS Experience`
  - [ ] `⭐️ Offices`
  - [ ] `⭐️ Product Support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


tjheffner:
the mixin itself looks fine to me, just two nits: 
- the csv button is a VA.gov front-end style button that editors only see in wysiwygs / proof content, versus the cms button style that is elsewhere (button--primary / secondary etc)
- we're duplicating some mixins between `_mixins.scss` and `wysiwyg.scss`. can we simplify this? maybe move the the two that are duplicated into mixins() (va-buttons, wysiwyg-action-links) and then import mixins() into wysiwyg. i assume we have them in both at the moment so that wysiwyg content still gets the right button styles on node view pages

msbtterswrth:
Good point about that cvs button, i can switch that to a cms button for sure. 

as for the duplication, that's how I found these two files, and i'm not sure why they're duplicated, I was just going with the flow 😅. Definitely agree with you, I think we could probably do a fair bit of improvement in how this older scss was organized. If you're cool with it, I can take a stab at making that a bit smoother!

tjheffner:
> @tjheffner Good point about that cvs button, i can switch that to a cms button for sure.
> 
> as for the duplication, that's how I found these two files, and i'm not sure why they're duplicated, I was just going with the flow 😅. Definitely agree with you, I think we could probably do a fair bit of improvement in how this older scss was organized. If you're cool with it, I can take a stab at making that a bit smoother!

please do :) I cleaned up some tech debt in the switch between themes, but a lot carried over for the sake of not breaking stuff. This is a perfect time to do some cleanup imo

msbtterswrth:
ok @tjheffner I think that's much better now. I didn't want to get too deep into the other files just yet, because I think once we start getting the design system implemented we're just gonna be re-writing a bunch of this in a more extensible manner.

docroot/themes/custom/vagovclaro/assets/scss/components/_buttons.scss
```diff
@@ -1,3 +1,5 @@
+@import 'tokens/mixins';
```
tjheffner, line 1: don't need to import this here, the theme's gulp setup will handle it thanks to the import order in styles.scss / other top-level scss files (not starting with a `_`)

docroot/themes/custom/vagovclaro/assets/scss/components/_buttons.scss
```diff
@@ -76,3 +78,29 @@ table .button:not(:focus) {
 .js.no-touchevents .dropbutton--multiple.dropbutton--extrasmall .dropbutton__item:first-of-type {
   white-space: nowrap;
 }
+
+
+//Claro buttons
+.button--primary {
```
tjheffner, line 84: Do we actually need any of these in our stylesheet? Are there any updates here with our updated mixin from what claro's stylesheet is giving us by default? e.g. by adding `button button--primary`. 
I don't think we need to bloat our stylesheet here personally

docroot/themes/custom/vagovclaro/assets/scss/wysiwyg.scss
```diff
@@ -1,242 +1,6 @@
...
-  line-height: 1.5;
-  margin-bottom: 1em;
-}
-
-/*
- A small subset of Formation for html tags enabled in rich text editor used in
- wysiwyg editor and node views.
- */
-
-@mixin wysiwyg-va-basic-styling {
```
tjheffner, line 59: I would double check that this mixin isn't different from the other va-basic-styling one. I think I got dinged for not including it separately before as part of the airtable issues... which is why i only mentioned the other two in my original comment. I may be totally wrong here/we're fine to delete this but confirmation would be good.

docroot/themes/custom/vagovclaro/assets/scss/components/_views.scss
```diff
@@ -105,12 +105,9 @@
 
 //fixes export button
 .csv-feed .feed-icon {
-  @include va-buttons;
-  background-color: var(--va-blue-med);
+  @include button(var(--button-bg-color--primary), var(--button-fg-color--primary), var(--button--hover-bg-color--primary), var(--button--active-bg-color--primary));
```
tjheffner, line 108: this is great, thank you

tjheffner (changes requested):

docroot/themes/custom/vagovclaro/assets/scss/components/_buttons.scss
msbtterswrth: I added it here so that we can easily use it when we need to restyle buttons that don't have the appropriate classes. I could see us overriding a fair amount of contrib modules someday and i'd hate to write preprocesses for them all, but I'll let you make that call!

docroot/themes/custom/vagovclaro/assets/scss/wysiwyg.scss
msbtterswrth: I did test this one and didn't see any issues, but we should double check for sure. Do you recall what that old issue was?

docroot/themes/custom/vagovclaro/assets/scss/wysiwyg.scss
```diff
@@ -1,242 +1,6 @@
...
-      background-color: transparent;
-      color: #f9c642;
-      
-      &::before {
-        color: #f9c642;
-      }
-    }
-  }
-}
+@import 'tokens/variables';
```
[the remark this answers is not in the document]
msbtterswrth: I tried taking this out but the colors on the buttons are unset in the editor.  I feel like the wysiwyg is compiled separately from the rest of things but I'm not sure... 
image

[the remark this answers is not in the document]
tjheffner: yeah it is compiled separately, and gets included into the ckeditor iframe as its own stylesheet. thanks for checking, we can leave the other imports here so they are set :+1:

docroot/themes/custom/vagovclaro/assets/scss/wysiwyg.scss
tjheffner: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/8121 was me reverting using the same mixin before. what the exact difference was i wish i could tell you, but i forget

msbtterswrth: I wonder if it's the font family line; I left that with the literal fonts rather than the font-variable. It could also be that the variables weren't compiling bc we didn't import them, since this file is compiled differently than the others.
 
```
.cke_editable {
  background: #fff;
  color: #323a45;
  font-family: 'Source Sans Pro', Sans-Serif;
  font-size: 15px;
  margin: 0.5em; /* to provide padding within ckeditor */
  max-width: 700px;

  @include va-basic-styling;
}
```

tjheffner: works for me, let's rock with it and if it does get reported as a regression we can revert again :shrug:

docroot/themes/custom/vagovclaro/assets/scss/components/_buttons.scss
msbtterswrth: AH I totally misunderstood what you were saying here... fixed!

tjheffner (approved):

swirtSJW (approved):
Nice work.  
+ great QA steps
+ much appreciated involving @tjheffner  in the review and appreciate all the back and forth between the two of you.  Great model.

Issue #8310, opened 2022-03-09
Decouple content release from daily production deploy


cweagans:
The content-build daily production deploy is largely the same as a content release with a couple of exceptions:

1. There's a Slack notification to prompt people to go check their stuff in staging
2. There's a 1 hour wait
3. A tag is created in the content-build repo

Unfortunately, since both the daily production deploy workflow and the content release workflow are doing content releases, they have to coordinate to make sure that they aren't deploying content at the same time. Additionally, since the daily production deploy is currently considered more important than the content release (which results in any requested or scheduled content releases being queued until the daily prod deploy is done), there is a ~90-120 minute window in the middle of the day where content releases simply do not happen. This is a problem because our current agreement with editors is that releases go out roughly once per hour.

To remedy this, we can decouple these jobs from each other and let them run completely independently.

1. Create a new workflow that runs 60 minutes before the scheduled daily deployment time that sends a notification to Slack prompting developers to go check their work on staging.
2. Create a new workflow that runs at the scheduled daily deployment time that creates the tag in the content build repo.
3. Delete the daily production deploy workflow.
4. Remove inter-workflow communication mechanisms from content release.

The result of this change should be as follows:

1. The dead period in the middle of the day is eliminated.
2. Content releases can be packed into the day much tighter, resulting in a lower mean-time-between-deploys for editors #acceleratedPublishing
3. The content release that runs immediately after the workflow that creates a tag in the content build repo would be the one that deploys new code (this also has the side effect of making the exact deploy time somewhat indeterminate, but in speaking with the release tools team, it sounded like this was an acceptable compromise)
4. Rollbacks would be very straightforward - simply commit a change in the content build repo that reverts whatever change, create and push a new tag, and then either wait for a content release or trigger one manually.
5. The CMS will no longer need to know about nor schedule around the daily production deploy job when kicking off content releases.

cweagans:
Notes from content release engineering sync today:

* Content release success message in slack should include tag for visibility
* Update pre-deploy message to "Last call for changes, prod deploy in 60 minutes, check your stuff in changing" (or something along those lines)
* The notification that the daily deploy has completed is important (so that e.g. devs can go notify stakeholders that something is on prod)
* Probably need to notify consumers of content build of this change: a11y, facilities, etc. Look at committers and notify.

Thank you @alexandec and @rmessina1010 for your input on this! Super helpful!

cweagans:
[...]

PR #8521, opened 2022-03-29
VACMS-0000: remove parameters to GHA content-release workflow request that are no longer expected.

timcosgrove:
## Description

Relates to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8310

In #8310, we added a change to the content-release workflow that removed environment and time to wait until release as parameters for that workflow. However, the Github adaptor in the CMS that was sending release requests to the GHA workflow was still passing those parameters, which caused an invalid API response.


ndouglas (approved):

timcosgrove:
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8522

PR #8744, opened 2022-04-13
Disables behavioral test that is flaky on Tugboat.

ndouglas:
This test only tests its step definitions at this time.  However, for some reason -- likely the specifics of the field it's testing -- it appears to be quite flaky on Tugboat.


swirtSJW (approved):
This works for me.  The test is nice but I would not cry if it had to go away completely.

ndouglas:
@swirtSJW That test will get replaced as part of #8746.  It exists right now only to test the step definition. 😃

Issue #8780, opened 2022-04-15
Next-Drupal: Drupal setup


timcosgrove:
## Description
Drupal should be updated to support Next & Next-Drupal. These updates should only apply to dev, Tugboat, and local environments (additional modules & patches can be installed and not enabled on prod & stage). 

Where required configuration is in content-like config rather than actual config, the necessary entities should check if they're in the system and recreate if not, as post-update actions. This will allow continued operation while this functionality is not active on prod and staging.

## Acceptance Criteria
- [ ] Testable_Outcome_X
- [ ] Testable_Outcome_Y
- [ ] Testable_Outcome_Z
- [ ] Requires design review

### CMS Team
Please check the team(s) that will do this work.

- [ ] `Sitewide program`
- [ ] `Platform CMS Team`
- [ ] `Sitewide crew ` (leave Sitewide unchecked and check the specific team instead)
  - [ ] `⭐️ Content ops`
  - [ ] `⭐️ CMS experience`
  - [ ] `⭐️ Public websites`
  - [ ] `⭐️ Facilities`
  - [ ] `⭐️ User support`

PR #8781, opened 2022-04-15
Vacms 8780 next setup

timcosgrove:
## Description

See #8780.

* adds drupal/next as a requirement
* adds a couple patches
* enables drupal/next and next-jsonapi and dependencies
* adds a `nextjs_authenticator` role and assigns a few necessary permissions to it
* adds a Nextjs 'site' definition. The function of this is to direct the CMS what URLs to use for previewing.
* adds config associating entities with the Nextjs site, so they can be previewed.

Things outside scope:
* There will be URLs specific to each environment for the Nextjs site definition, and secrets will be pulled from the appropriate locations for those environments. What is committed here is not intended for production. This will be taken care of when setting up preview. https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8785
* `next_entity_type_config` entities should be generated automatically for all bundles of selected types. That is, all node types should have one of these config entities, and if a new node type is created, a config entity should be created for it automatically. https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8786

## QA suggestions
These changes will mostly only provide functionality in the context of the CMS as a provider of content to a Next-Drupal site, which is not yet available in a testable form. As such, QA of this consists of:

1. Ensure that normal functioning of the site continues and does not throw new warnings or errors. Automated testing against the environment should provide a set of 'normal functions', as does running a content-build against this environment. See https://cms-ezfyrhzpzziky2vsiaoem3qw1f4j0gys.ci.cms.va.gov/admin/reports/dblog to examine output.

2. Ensure that the modules, role, site, and entity config that are part of the PR are active. These are:
  * drupal/next and dependencies: https://cms-ezfyrhzpzziky2vsiaoem3qw1f4j0gys.ci.cms.va.gov/admin/modules
  * role/permissions: https://cms-ezfyrhzpzziky2vsiaoem3qw1f4j0gys.ci.cms.va.gov/admin/people/permissions
  * Next site: https://cms-ezfyrhzpzziky2vsiaoem3qw1f4j0gys.ci.cms.va.gov/admin/config/services/next
  * Entity configuration for next-drupal: https://cms-ezfyrhzpzziky2vsiaoem3qw1f4j0gys.ci.cms.va.gov/admin/config/services/next/entity-types


config/sync/next.next_site.va_gov_nextjs.yml
```diff
@@ -0,0 +1,9 @@
+uuid: 865647c3-24f3-49d0-9a89-1db8a122d474
+langcode: en
+status: true
+dependencies: {  }
+id: va_gov_nextjs
+label: 'VA.gov nextjs'
+base_url: 'http://localhost:3000'
+preview_url: 'http://localhost:3000/api/preview'
+preview_secret: nextjs-preview
```
timcosgrove, line 9: These will be overridden per-environment.

composer.lock
```diff
@@ -25280,5 +26170,5 @@
     "platform-overrides": {
         "php": "7.4"
     },
-    "plugin-api-version": "2.3.0"
+    "plugin-api-version": "2.2.0"
```
swirtSJW, line 26173: Interesting downgrade.

config/sync/user.role.nextjs_authenticator.yml
```diff
@@ -0,0 +1,19 @@
+uuid: 6ef91ed1-a9bd-4e11-9ee7-b0fc702d5eb9
```
swirtSJW, line 1: Should this role be added to our phpunit access checks for test coverage?

swirtSJW (commented):
Overall looks good.   Do we have a rudimentary set of QA steps so we can evaluate that it is doing something or not adding new notices?

composer.lock
timcosgrove: Will check this; might just be me doing composer.

timcosgrove: Indeed, this was just down to my local composer version (was 2.2.9 before I updated it via `ddev composer self-update --2`, which brought it to 2.3.5). This value will probably bounce back and forth based on whatever a given developer has as their local composer version, and would most likely have no negative effect. It's more reporting what version of Composer generated that copmoser.lock file.

Nevertheless, updated my composer, and ran `ddev composer update drupal/next` again, at which point this was repaired.

tests/phpunit/Security/RolesPermissionsTest.php
```diff
@@ -44,6 +44,36 @@ public function testSecurityRolesPermissions($roleMatch, $expectedPerms) {
    */
   public function expectedPerms() {
     return [
+      [
```
timcosgrove, line 47: Most of this is just re-ordering what was already there, for consistency.

tests/phpunit/Security/RolesPermissionsTest.php
```diff
@@ -760,34 +790,13 @@ public function expectedPerms() {
         ],
       ],
       [
-        'admnistrator_users',
+        'nextjs_authenticator',
```
timcosgrove, line 793: this is new

tests/phpunit/Security/RolesPermissionsTest.php
```diff
@@ -799,6 +808,26 @@ public function expectedPerms() {
           'view node link report',
         ],
       ],
+      [
```
timcosgrove, line 811: This was not present previously; added it for completeness.

swirtSJW: Thanks for catching this and adding it.

tests/phpunit/Security/RolesPermissionsTest.php
swirtSJW: Thanks for the alpha sort. Consistency is nice, though it does blow out the history of all these lines and attribute them all to 8780.

composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "ebda0a9b5b719b8b5557a61dcd6b9afb",
+    "content-hash": "25a236b6ce92db8bbba97479db37f3d7",
```
swirtSJW, line 7: We've had a couple of composer merges so we'll have to update this hash one last time in order to push this through.

swirtSJW (approved):
This looks great Tim.  
Other than refreshing the composer lock hash, we should be good.

swirtSJW:
Noting the passing tests so that we can just merge this after the composer lock hash gets updated. 
image

composer.lock
timcosgrove: > Thanks for the alpha sort. Consistency is nice, though it does blow out the history of all these lines and attribute them all to 8780.

The past cannot hold the present and future hostage.

PR #9001, opened 2022-05-09
VACMS-8988: Disable status id field for non admins.

### LINKED ISSUE #8988: Lock down supplemental status term field_status_id
On Facility Supplemental status terms, the field_status_id should not be editable by anyone other than true admins.
It should be locked down the same way that health_care_service_taxonomy->field_health_service_api_id is locked down.

This field should not change once it is establshed.  It should be as immutable as the term id (tid) is.

~Add "Allow only one" to this field do that it could not be duplicated either.~

This is NOT critical for initial launch.

## ACs
- [ ] on Supplemental status terms, field_status_id should not be editable by anyone other than admin.
- [ ] treatment/logic for disabling the field is identical to health_care_service_taxonomy->field_health_service_api_id
- [x] Allow only one field is added to the Facility Supplemental status vocabulary so that field_status_id can not be duplicated.  ---> this was already added to https://github.com/department-of-veterans-affairs/va.gov-cms/pull/8963

ethanteague:
## Description
closes #8988

## Testing done
Visual / behat

## Screenshots
Logged in as user selina.cooper@va.gov (content admin)
image

Logged in as ethan.teague (administrator)
image

## QA steps
 - [ ] As an administrator, go to `admin/structure/taxonomy/manage/facility_supplemental_status/overview` and add a new term (add values to every field).
 - [ ] After saving, visually verify the Status ID field is editable when viewing the term form.
 - [ ] Login as selina.cooper@va.gov (content admin), and visit the edit form for the term you created as an administrator.
 - [ ] Visually verify Status ID field is not editable.

### DIFF
--- a/docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
+++ b/docroot/modules/custom/va_gov_backend/src/EventSubscriber/EntityEventSubscriber.php
@@ -312,6 +312,9 @@ class EntityEventSubscriber implements EventSubscriberInterface {
     if (!$this->userPermsService->hasAdminRole(TRUE) && $bundle === 'health_care_service_taxonomy') {
       $form['field_health_service_api_id']['#disabled'] = TRUE;
     }
+    if (!$this->userPermsService->hasAdminRole(TRUE) && $bundle === 'facility_supplemental_status') {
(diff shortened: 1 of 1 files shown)


JayDarnell (approved):
Looks good to me @ethanteague

JayDarnell:
Passes QA and code review looks good.

Issue #9018, opened 2022-05-09
Implement MVP product tags for KB to increase findability


EWashb:
## Background ##

As a knowledge base user, I would like to see KB articles labeled in an obvious way to determine what is relevant for my tasks. 

## Assumptions ##

KB article tags empower the editor to help themselves troubleshoot and find the information they need to produce high-quality content
![image (4).png](https://images.zenhubusercontent.com/614b6cbe05afafb268be4c0b/ad72a9e0-091e-41cd-9054-684ec5d37b61)

## Acceptance Criteria ##

- [x] UX design and writing review before implementation

Revised ACs as of 5/17/22

-[ ] Review process and feasibility for linking the product tags
-[ ] determine if it is best practice to link only those that have a landing page to go to or if it is best to wait until all are create

msbtterswrth:
Approved design screenshot is below. First tag is a non-linked tag, second is linked. 
![image](https://user-images.githubusercontent.com/8375335/169098247-8af839f7-a194-48f0-81c5-7a95402be716.png)

EWashb:
We are going to put this on hold for Sprint 58 to determine the proper workflow. Some things to consider:

1. Should these be visible if they aren't taking to anywhere since the functionality appears to be a link
2. Should this go out with no links?
3. Should this go out with some linked and some not (since we don't have product landing pages for all tags)

rachel-kauff:
@EWashb I reckon we should kick this back into the icebox for now until we do a focused effort on KB stylin'

msbtterswrth:
Worked with cindy to refine the MVP design: 
![image](https://user-images.githubusercontent.com/8375335/180301680-c85dd7db-cb67-49d8-bf93-dd8e50c0dbca.png)

PR #9027, opened 2022-05-09
VACMS-9018: Add product tags to KB detail pages.

msbtterswrth:
## Description

Relates to #9018.

## Testing done
manual, cypress

## Screenshots
image

## QA steps

As cms user, 
1. Go to a KB page (`/help/vamc/how-to-edit-a-vamc-system-homepage`)
   - [ ] Validate that the product tags appear under the page introduction
   - [ ] Validate that the 'last updated' time appears above the page introduction
2. Then validate Acceptance Criteria from issue
   - [ ] UX design and writing review before implementation

### Definition of Done

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [x] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/docroot/themes/custom/vagovclaro/templates/node/node--documentation-page.html.twig
+++ b/docroot/themes/custom/vagovclaro/templates/node/node--documentation-page.html.twig
@@ -103,10 +103,13 @@
   {{ title_suffix }}
 
   <div{{ content_attributes.addClass('node__content') }}>
-    {{ content }}
-    <div class="last_updated">
-      Last updated: {{ node.changed.value|date("F j, Y") }} | {{ about }}
+    <div class="tags">
+      {{ content.field_products }}
     </div>
+    {{ content }}
+    {# <div class="last_updated">
+      Last updated: {{ node.changed.value|date("F j, Y") }} | {{ about }}
+    </div> #}
   </div>
 
 </article>
--- a/docroot/themes/custom/vagovclaro/assets/scss/pages/_kb-pages.scss
+++ b/docroot/themes/custom/vagovclaro/assets/scss/pages/_kb-pages.scss
@@ -152,3 +152,16 @@
     margin: 0;
   }
 }
+
+.field--name-field-products {
+  .field__item {
+    background: var(--va-gray-cool-light);
+    border-radius: 20px;
+    display: inline-block;
+    font-size: 12px !important;
+    font-weight: 700;
+    margin-right: var(--space-s) !important;
+    padding: var(--space-xs) var(--space-m);
+    text-transform: uppercase;
+  }
+}


docroot/themes/custom/vagovclaro/templates/node/node--documentation-page.html.twig
```diff
@@ -103,10 +103,13 @@
   {{ title_suffix }}
 
   <div{{ content_attributes.addClass('node__content') }}>
+    <div class="tags">
```
tjheffner, line 106: this and the commented out portion below can both be removed, i think. the `tags` class isn't actually being used as far as i can tell, because the blocks are set using layout builder.  This means they print under the `{{ content }}` variable anyway.

Screen Shot 2022-05-10 at 12 01 52 PM


if we want to continue to use layout builder for the KB pages that's fine (maybe worth a discussion?) but let's clean up the template then

tjheffner (changes requested):

docroot/themes/custom/vagovclaro/templates/node/node--documentation-page.html.twig
msbtterswrth: Whoops! good catch! I was going to go down the template path but forgot to take it out when I realized it was using layout builder. Let's bring layout builder usage up at out next scrum/coworking or something!

kevwalsh:
A long time ago (Q3 2021), we had a notion that the product tags might be clickable, and take the user to the product's landing page in the KB. That KB landing page can be specified in the product taxonomy term (eg, the same thing in the CMS that the tag uses). Did we consider that for this iteration?

msbtterswrth:
@kevwalsh I thought about it, but those pages are _pretty_ ugly at the moment and we thought it might be better to include that when we fix the display of the search a bit, which I think was our next focus for KB improvements.

kevwalsh:
not sure if we're talking about the same thing. It's not a search page. it's a curated landing page, the same one that is linked from the left nav. Like [VAMCs](https://prod.cms.va.gov/help/vamc) or [Vet Centers](https://prod.cms.va.gov/help/vet-centers).

msbtterswrth:
oh @kevwalsh i see what you mean. We could totally do it, or we can wait to add that, i think we _also_ talked about creating a new content type for kb landing pages and it might be better to wait for that, or not. Totally can go either way!

swirtSJW:
I'm stuck on validating this mystery 

> UX design and writing review before implementation

I dont see anything in the original issue to show the design and writing review.

msbtterswrth:
@swirtSJW that's for the review by @rachel-kauff and @cjterdi. I wonder if we should separate ACs for devs/ux?

swirtSJW:
> @swirtSJW that's for the review by @rachel-kauff and @cjterdi. I wonder if we should separate ACs for devs/ux?
Ideally I think it shouldn't matter.   Any one person should be able to pick up a PR and be able to run through the QA steps.   The design comps / words should be there and anyone should be able to validate that it does or does not match.   It should never be nebulous about what the criteria is.

msbtterswrth:
During coworking with @cjterdi we settled on the currently implemented design with the intent to revisit it when we re-theme the whole KB.

swirtSJW:
Rollout [conversation here](https://dsva.slack.com/archives/CT4GZBM8F/p1652895938904109)

EWashb:
Not sure if this is the proper place to note, but we will put this on hold for Sprint 58. I have made some notes in the original ticket for things to discuss before this is implemented. I will also update the ACs on the original ticket.

msbtterswrth:
I commented out the linked terms for now, until they're ready to be turned on in a future state per MVP convos with the team.

msbtterswrth:
Implemented design discussed during coworking. 
image

PR #9100, opened 2022-05-17
[DNM] Revert "Enable base next module without adding any settings."

timcosgrove:
Reverts department-of-veterans-affairs/va.gov-cms#9089


timcosgrove:
Not needed; see #9102

PR #9256, opened 2022-05-26
VACMS-9171 announcement block & alert updates

tjheffner:
## Description
Closes #9171. Had a design sync with @cjterdi and @rachel-kauff and we collaboratively updated some of the alert styles in concert with the new CMS announcement block type.

- Created new announcement block component in the design system w/ interactive storybook example for the different types (maybe should be split out to 4 standalone like the other alerts are)
- Updated existing theme templates for announcement blocks to use the DS
- Updated system status alerts to match new design direction

## Testing done
- New design system component in storybook
- Added a new announcement block to specific content types. Verified that all 4 types of announcement block render correctly
- Verified new alert styles render correctly

## Screenshots
Screen Shot 2022-05-26 at 1 39 17 PM
Screen Shot 2022-05-26 at 1 40 45 PM

## QA steps

- Visit the storybook link and see that there is a new component for the announcement block
- In storybook see that existing alerts now have a variant for no background (you may need to hit 'a' to bring up the controls panel)

- In Drupal, log in as an admin and visit `/block/add?destination=/admin/structure/block/block-content`.
- Add a cms announcement. Block description is the administrative label that will appear when it is placed in the block layout list, the title field is the title of the announcement.
- After creating the block, visit `/admin/structure/block` and place this new block in the Content Region (search for your label to find it quickly). Configure it to only be visible on certain content types (ex. VAMC Facility)

Screen Shot 2022-05-26 at 1 48 47 PM
- After placing the block, drag it to the top of the region to ensure it shows up where it's supposed to on the page. 

- Visit an existing node of the content type you restricted the block visibility to. See that hte block shows up on the page, and system status alerts appear below it with the updated styles.

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [x] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change?
- [ ] Yes, and it's written in issue ____ and queued for publication.
  - [ ] Merge and ping @ rachel-kauff so she's ready to publish after deployment
[...]

PR #9281, opened 2022-05-31
VACMS-9171 cms announcement block + alert updates

### LINKED ISSUE #9171: Update alert styles
## Background

As the CMS announcement block was introduced, it has caused inconsistent sizing with system status alerts that appear at full screen. The issue does not appear all the time (users won't necessarily have all of these alerts at any given time), but it could be possible. The VAMC Facility page > edit node page below shows an example of what it could be like with current state:
image

Figma:
https://www.figma.com/file/EO0oAAViKMpwGUwMYpAM9E/DS-Templates?node-id=29%3A145549

Desired appearance:
Screen Shot 2022-05-26 at 10 51 47 AM
Screen Shot 2022-05-26 at 10 51 59 AM
Screen Shot 2022-05-26 at 10 52 08 AM

## Acceptance Criteria
- [ ] CMS Announcement Block resized to match width of other alerts
- [ ] Background color removed from system status alerts, spacing adjusted
- [ ] Sitewide CMS design review for appropriate margins
- [ ] Implementation into Storybook

tjheffner:
## Description
Closes #9171. This PR replaces the PR open in #9256 which I dirtied the git history. Sorry for the noise. 

Had a design sync with @cjterdi and @rachel-kauff and we collaboratively updated some of the alert styles in concert with the new CMS announcement block type.

- Created new announcement block component in the design system w/ interactive storybook example for the different types (maybe should be split out to 4 standalone like the other alerts are)
- Updated existing theme templates for announcement blocks to use the DS
- Updated system status alerts to match new design direction
- Updated announcement block "announcement type" field to only the two options, per @rachel-kauff in the previous closed pr (dirty git history #9256)

## Testing done
- New design system component in storybook
- Added a new announcement block to specific content types. Verified that all 4 types of announcement block render correctly
- Verified new alert styles render correctly

## Screenshots
Screen Shot 2022-05-26 at 1 39 17 PM
Screen Shot 2022-05-26 at 1 40 45 PM

## QA steps

- Visit the storybook link and see that there is a new component for the announcement block
- In storybook see that existing alerts now have a variant for no background (you may need to hit 'a' to bring up the controls panel)

- In Drupal, log in as an admin and visit `/block/add?destination=/admin/structure/block/block-content`.
- Add a cms announcement. Block description is the administrative label that will appear when it is placed in the block layout list, the title field is the title of the announcement.
- After creating the block, visit `/admin/structure/block` and place this new block in the Content Region (search for your label to find it quickly). Configure it to only be visible on certain content types (ex. VAMC Facility)

Screen Shot 2022-05-26 at 1 48 47 PM
- After placing the block, drag it to the top of the region to ensure it shows up where it's supposed to on the page. 

- Visit an existing node of the content type you restricted the block visibility to. See that hte block shows up on the page, and system status alerts appear below it with the updated styles.

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [x] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`


rachel-kauff:
For some reason when I try to turn off the bg in storybook the component disappears from the preview box. Not sure if this is a problem with the component specifically, or a weird bug with Storybook in general.

Screen Shot 2022-06-01 at 8 50 15 AM

tjheffner:
> For some reason when I try to turn off the bg in storybook the component disappears from the preview box. Not sure if this is a problem with the component specifically, or a weird bug with Storybook in general.

It is a weird bug from how we have to render twig in Storybook. but good to know about! I will likely have to create separate variants for each case instead of a dynamic example

rachel-kauff:
@tjheffner I have some more suggestions for the UX writing for the field labels and help text for this block edit form, but want to check with you first to see if it's worth the effort. It's not bad, but could be more clear. Is that in scope for this ticket?

docroot/design-system/components/announcement-block/announcement-block.twig
```diff
@@ -0,0 +1,63 @@
...
+]
+%}
+
+{% set icon = '' %}
+
+{% if type is same as('information') %}
+  {% set icon = 'circle-info' %}
+{% elseif type is same as('warning') %}
+  {% set icon = 'warning' %}
+{% elseif type is same as('emergency') %}
```
msbtterswrth, line 45: Should we remove this statement if we're taking out 'emergency' as an option in the CMS?

docroot/design-system/components/announcement-block/announcement.scss
```diff
@@ -0,0 +1,68 @@
...
+    }
+  }
+
+  &.new {
+    background: var(--va-mint);
+
+    .title svg {
+      color: var(--va-green);
+    }
+  }
```
msbtterswrth, line 42: Should we also take these out if we're removing those options from the CMS?

msbtterswrth (changes requested):
Just a couple minor things, looks great!

docroot/design-system/components/announcement-block/announcement-block.twig
tjheffner: I left them in just in case we do add those options back in the future.... they render in SB only for now

docroot/design-system/components/announcement-block/announcement.scss
tjheffner: I left them in just in case we do add those options back in the future.... they render in SB only for now

msbtterswrth (approved):
Looks good! thanks for porting this over to DS :)

rachel-kauff:
Per our discussion about splitting the UI help text & label changes into a separate ticket, I'm gonna approve this one so it can get on with its life.

rachel-kauff (approved):

PR #9280, opened 2022-05-31
VACMS-9097 new facility notification

tjheffner:
## Description

Relates to #9097. First of four use cases for "dogfooding" the messages stack with our CMS Help Desk.

- Sends out a template message using the newly created CMS Help Desk notifications user to support@va-gov.atlassian.net with a message about any new facility created so that the appropriate follow-up steps may happen. 
- Coordinated with Stefanie on the desired wording of the email template.
- Created view modes to display email subject/body correctly.

## Testing done
Created new facilities of the 6 types locally, verified that an email was sent in mailhog ( https://va-gov-cms.ddev.site:8026/ )

## Screenshots
Email generated from a new facility being created:
Screen Shot 2022-05-31 at 2 05 15 PM

I'm not sure it's set up 100% correctly (or if it is mailhog changing something somewhere), because the `from:` and `to:` fields look a little off to me... but it's using the same setup as both @swirtSJW's example and the example in the module 🤔

## QA steps

What needs to be checked to prove this works?  
- create a new facility (any of the 6 facility types: VAMC Facility, VBA Facility, NCA Facility, Vet Center, Vet Center Mobile Vet Center, or Vet Center Outstation) and save it. exact section and menu things don't really matter for this initial step.
- as an admin user: check `/admin/content/message` and see that a new message was created for the facility creation. I am not sure how to check mailhog on tugboat, but if you are testing locally, visit mailhog at https://va-gov-cms.ddev.site:8026/ and see a corresponding email captured.
Screen Shot 2022-05-31 at 2 11 27 PM

What needs to be checked to prove it didn't break any related things? 
- No emails should be sent for any other actions at this time.

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [x] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

Issue #9097, opened 2022-05-17
Proof of concept: CMS notifications through helpdesk


EWashb:
## Background ##

This notifications POC will include VAMC flagged data such as: 

- [x] New facility added
- [x] Facility closed
- [x] Name change
- [x] Official legislative name vs. common name

## Acceptance Criteria ##
- [ ] 4 targeted PRs
- [x] Work with helpdesk to determine email address
- [x] Create test accounts
- [x] Create template and draft email

kevwalsh:
Removing the sprint because epics shouldn't have sprints.

EWashb:
I intended it to be a task and not an epic. Not sure why I did that. Apologies for the confusion.

EWashb:
@swirtSJW @tjheffner is this ready to be closed?

tjheffner:
No

swirtSJW:
https://github.com/department-of-veterans-affairs/va.gov-cms/pull/9412  which is the undo of the revert (with some fixes) is in.

swirtSJW:
I just merged this https://github.com/department-of-veterans-affairs/va.gov-cms/pull/9349  
so the issue can be closed.
Nice work @tjheffner

PR #9300, opened 2022-06-02
VACMS-9097 facility deleted notification

tjheffner:
## Description
Relates to #9097, companion to #9280. This contains the last of 4 use cases from the ticket.

Sends an email to the help desk when facilities are flagged as removed from the facility API. Collaborated with Stefanie on the email template.

## Testing done
- Created a new facility (any type, a newly created facility won't be in the facility API) and then ran the following command:

```sh 
$ ddev drush va_gov_migrate:flag-missing-facilities
[success] Flagged 2 facilities as removed from Facility API.
```

Checked mailhog locally @ https://va-gov-cms.ddev.site:8026/ to see that the notification was triggered.

Checked that the node in the email was correctly flagged as "removed from source" in Drupal flags. 

## Screenshots
mailhog:
Screen Shot 2022-06-02 at 2 47 00 PM
For some reason, this template renders the site url as "default" despite using the same url token from the referenced entity as the other templates. My thinking is this is due to this particular case being triggered by a drush command versus the entity event subscriber, but I haven't dug in to confirm yet.... the node url is correct below in the message log itself.

drupal message log (bottom of `/admin/content/message`. `/admin/content/messages` is also a valid route with a slightly similar view :thinking:):
Screen Shot 2022-06-02 at 2 52 22 PM

revision from migration command that flags the node:
Screen Shot 2022-06-02 at 2 54 21 PM

## QA steps

- The places in the testing steps above. Mailhog, messages, and the node revision.
- This email template should only trigger for flaggable facilities from the migration command

As an admin user,
- create a new facility with a bogus facility id (VBA facility doesn't have many required fields). you can flag it as a new facility  if you want (no email should be sent for that action in this branch).
- run this drush command to flag missing facilities: `ddev drush va_gov_migrate:flag-missing-facilities`. you should see at least one facility flagged (your new bogus facility)
- Check mailhog to see the email
- Check `/admin/content/message/` (you may need to scroll to the bottom)
- Check the node itself that a revision was created & it was flagged correctly (this functionality shouldn't have changed in this PR)

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [x] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

### DIFF
--- a/config/sync/message.template.va_facility_removed_from_source.yml
+++ b/config/sync/message.template.va_facility_removed_from_source.yml
@@ -1,0 +1,19 @@
+uuid: d8673740-2510-4e2e-9aef-f166239d0246
+langcode: en
+status: true
+dependencies:
+  config:
+    - filter.format.rich_text
+template: va_facility_removed_from_source
+label: 'VA Facility removed from source'
+description: 'Informs the CMS Help Desk when a facility has been removed upstream.'
+text:
+  -
+    value: "<p>Hello CMS Help Desk!</p>\r\n\r\n<p>An existing [message:field_target_entity:entity:content-type] named &lt;a href=\\\"[message:field_target_entity:entity:url]\\\"&gt;[message:field_target_entity:entity:title]&lt;/a&gt;&nbsp;was flagged as removed from the upstream Facility API on [message:field_target_entity:entity:changed].</p>\r\n\r\n<p><br />\r\nPlease follow up with the editor of the above facility and ask them to make sure they have completed all steps listed here:&nbsp;<a href=\"https://prod.cms.va.gov/help/vamc/about-locations-content-for-vamcs/how-to-archive-a-closed-facility\" target=\"_blank\">https://prod.cms.va.gov/help/vamc/about-locations-content-for-vamcs/how-to-archive-a-closed-facility</a><br />\r\n<br />\r\nThank you!</p>\r\n"
+    format: rich_text
+settings:
+  'token options':
+    clear: false
+    'token replace': true
+  purge_override: false
+  purge_methods: {  }
--- a/config/sync/core.entity_form_display.message.va_facility_removed_from_source.default.yml
+++ b/config/sync/core.entity_form_display.message.va_facility_removed_from_source.default.yml
@@ -1,0 +1,49 @@
+uuid: 2a697274-f316-440f-ab1d-8e2e62776c25
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.message.va_facility_removed_from_source.field_target_entity
+    - field.field.message.va_facility_removed_from_source.field_target_node_title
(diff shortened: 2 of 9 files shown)


tjheffner:
> In ConfigImportCommands.php line 359:
>                                                                                
>   The import failed due to the following reasons:                              
>   Deleted and replaced configuration entity "core.entity_form_display.message.va_facility_removed_from_source.default"

config/sync/message.template.va_facility_removed_from_source.yml
```diff
@@ -0,0 +1,19 @@
...
+status: true
+dependencies:
+  config:
+    - filter.format.rich_text
+template: va_facility_removed_from_source
+label: 'VA Facility removed from source'
+description: 'Informs the CMS Help Desk when a facility has been removed upstream.'
+text:
+  -
+    value: "<p>Hello CMS Help Desk!</p>\r\n\r\n<p>An existing [message:field_target_entity:entity:content-type] named &lt;a href=\\\"[message:field_target_entity:entity:url]\\\"&gt;[message:field_target_entity:entity:title]&lt;/a&gt;&nbsp;was flagged as removed from the upstream Facility API on [message:field_target_entity:entity:changed].</p>\r\n\r\n<p><br />\r\nPlease follow up with the editor of the above facility and ask them to make sure they have completed all steps listed here:&nbsp;<a href=\"https://prod.cms.va.gov/help/vamc/about-locations-content-for-vamcs/how-to-archive-a-closed-facility\" target=\"_blank\">https://prod.cms.va.gov/help/vamc/about-locations-content-for-vamcs/how-to-archive-a-closed-facility</a><br />\r\n<br />\r\nThank you!</p>\r\n"
```
swirtSJW, line 12: I think this message, may be misaligned with that article.   Either that or the article needs work.   
I think it should be pointing here since these are the steps helpDesk needs to kick off. 
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/new?assignees=&labels=Change+request&template=runbook-facility-closed.md&title=Facility+closed%3A+%3Cinsert_name%3E

swirtSJW (commented):
This looks good with the possible exception of the message.

swirtSJW (commented):
File changes look good.

However I could not get this to test on tugboat.
- I created a new facility "Bogus facility" and published it
- Then I used the terminal and ran `drush va_gov_migrate:flag-missing-facilities`
image
No removed email was created.
image

However, I think this is related to the removal script not functioning properly because the new facility was not flagged as removed
image

I think tugboat can not access the facility API.


Trying to test this a different way.

swirtSJW (approved):
Pulled locally.  Checks out good.

Issue #9203, opened 2022-05-23
Load test Authenticated GET requests to various content types


ElijahLynn:
In #8080 we got some baseline metrics for anonymous GET requests to the Benefits Detail Page which was around 217/second. We now need to test authenticated GET requests to various content types as that will be reflective of real world traffic when the new Next.js build system evolves to allowing for public requests. 

## AC
- [ ] Determine the average RPS (requests per second) for authenticated GET requests for a node of the Benefits Detail Page content type. This should be with a warmed cache. 

## Tasks
- [x] Get Umami demo working via simplytest.me (quick hello-world example, not sustained, avoid DDoS)
- [x] Write a test file for to login based on the Umami login code on local DEV
- [x] Document initial setup and how to run load test
- [x] Loadtest the Benefits Detail Page on staging.cms.va.gov
- [x] Get SOCKS passthrough working (https://github.com/tag1consulting/goose/issues/490) with example test file
- [x] Post results to this thread, or maybe Confluence

_Originally posted by @ElijahLynn in https://github.com/department-of-veterans-affairs/va.gov-cms/issues/8050#issuecomment-1134929759_

> We discussed in planning now, that the public API requests will be authenticated so it is still worth testing authenticated GET requests. I finally got Goose Load Testing working with concurrent requests on Friday so we should still pull in a new story this sprint to get these metrics on the authN GET requests.

ElijahLynn:
Working on getting requests routed through our SOCKS proxy with Goose > https://github.com/tag1consulting/goose/issues/490

ElijahLynn:
`content_build_api` user: 
```
 === PER REQUEST METRICS ===                                                     
 ------------------------------------------------------------------------------  
 Name                     |        # reqs |        # fails |    req/s |  fail/s  
 ------------------------------------------------------------------------------  
 GET auth /disability/h.. |        21,354 |         0 (0%) |    68.01 |    0.00  
```

ElijahLynn:
I cannot upload HTML files but here is the HTML report zipped up from this command, and I took a full page screenshot too:

```shell
 cargo run --release -- --host https://staging.cms.va.gov --report-file report-file.html --running-metrics 1 --users 125 --hatch-rate 10 --run-time 3m --no-reset-metrics
```

[...]

PR #9302, opened 2022-06-03
VACMS-9203: Introduce loadtesting framework for authenticated requests with Goose

ElijahLynn:
## Description
See https://github.com/department-of-veterans-affairs/va.gov-cms/issues/9203

## QA steps

1. Do this
   - [ ] Validate that
2. Then
   - [ ] Validate that
   - [ ] a
   - [ ] b
   - [ ] c

### Definition of Done

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [ ] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


tests/loadtest/src/admin.rs
```diff
@@ -0,0 +1,26 @@
...
+pub async fn log_in(user: &mut GooseUser) -> TransactionResult {
+    // Use ADMIN_USERNAME= to set custom admin username.
+    let admin_username = match env::var("ADMIN_USERNAME") {
+        Ok(username) => username,
+        Err(_) => "admin".to_string(),
+    };
+    // Use ADMIN_PASSWORD= to set custom admin username.
+    let admin_password = match env::var("ADMIN_PASSWORD") {
+        Ok(password) => password,
+        Err(_) => "P@ssw0rd1234".to_string(),
```
cweagans, line 15: This seems like something we shouldn't include. Even if it's not a real password, probably better to just throw an error if the env vars aren't set.

tests/loadtest/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+        // .register_scenario(
+        //     scenario!("Anonymous English user")
+        //         .set_weight(40)?
+        //         // .set_wait_time(Duration::from_secs(0), Duration::from_secs(1))?
+        //         .register_transaction(
+        //             transaction!(benefits_detail_page)
+        //                 .set_name("anon /")
+        //                 .set_weight(2)?,
+        //         ),
+        // )
```
cweagans, line 25: Can you please remove the commented code?

tests/loadtest/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+                    transaction!(log_in)
+                        .set_on_start()
+                        .set_name("auth login"),
+                )
+                .register_transaction(
+                    transaction!(benefits_detail_page)
+                      .set_name("auth /disability/how-to-file-claim"),
+                ),
+        )
+        .set_default(GooseDefault::Host, "https://drupal-9.ddev.site/")?
```
cweagans, line 40: Not sure this should have a default. I think it'd be okay to throw an error if this isn't set by the user? The CMS doesn't run at `drupal-9.ddev.site` normally, so this won't work for anyone at our team, right?

cweagans (dismissed):
Approved with a couple of quick (?) changes. If they aren't quick, don't worry about them.

tests/loadtest/src/main.rs
ElijahLynn: Yeah, I do want to add it back to be able to just run anon vs auth scenario but haven't figured out how to do that yet. Will delete for now.

tests/loadtest/src/main.rs
ElijahLynn: Correct, won't work for anyone, it came from the default in the goose-eggs crate. I think we should change to va-gov-cms.ddev.site.

tests/loadtest/src/admin.rs
ElijahLynn: Changed to be the export commands needed. Not sure how else to do it right now, the Err() needs to be there (tested).

PR #9472, opened 2022-06-15
VACMS 0000: nate plumbing the dreamlands of kadath

ndouglas:
just debugging stuff


ndouglas:
There's some environment stuff going on, but the Uri error that I suspect is causing a lot of the test failures is occurring in periods where Tugboat seems to be correctly identifying itself:
Screen Shot 2022-06-15 at 2 08 09 PM
Screen Shot 2022-06-15 at 2 07 44 PM

PR #9595, opened 2022-06-27
VACMS-0000: Enable twig debugging for local development.

dsasser:
## Description

I found I had to setup theme debugging (it was not already setup), and in doing so, realized that there is a missing services file for the local environment.


swirtSJW (commented):
I want to bring @cweagans into this since I know he made some changes related to other environments thinking they are "local"   I am not sure if this would get picked up by environments other than our ddev locals.

My other concern is that some of this ought to be a local choice (per developer), yet making this part of the repo takes that choice away since there is not another level of gitignored file that functions downstream from this one.

cweagans:
Twig debug and reload seem like good defaults to include in local environments, but IMO we really shouldn't be doing too much with caches completely disabled.

swirtSJW:
Some of this duplicates what is already in place here
https://github.com/department-of-veterans-affairs/va.gov-cms/blob/main/docroot/sites/development.services.yml

and gets brought in, in my case, by my settings.local.php   like this line from the example here
https://github.com/department-of-veterans-affairs/va.gov-cms/blob/main/docroot/sites/example.settings.local.php#L41

dsasser:
@swirtSJW @cweagans Since my main goal was to be able to more easily enable twig debugging, how does this sound:
1) remove this local services file and instead...
2) Modify the existing development.services.yml to:
```
# Local development services.
#
# To activate this feature, follow the instructions at the top of the
# 'example.settings.local.php' file, which sits next to this file.
parameters:
  http.response.debug_cacheability_headers: true
  twig.config:
    debug: true
    auto_reload: true
services:
  cache.backend.null:
    class: Drupal\Core\Cache\NullBackendFactory
```

Do the above will then only affect those that have a local settings file which adds the existing development.services.yml config to the container_yaml settings array, rather than applying to any "local" environment.

swirtSJW:
@dsasser That sounds like a good solution to me.

swirtSJW (approved):
Good solution.

swirtSJW:
The phpunit failure is an odd one.  
image

We'll see if it goes away with a rebuild.

swirtSJW:
image

PR #9598, opened 2022-06-27
VACMS-6528: archive related services/events when a facility is archived

### LINKED ISSUE #6528: Automate archiving of facility services when a facility is archived

JayDarnell:
## Description

Closes #6528

## Testing done

Manual / visual testing

## Screenshots

Screen Shot 2022-06-27 at 4 39 27 PM

Screen Shot 2022-06-27 at 4 51 38 PM

Screen Shot 2022-06-27 at 4 51 47 PM

Screen Shot 2022-06-27 at 4 52 05 PM

## QA steps

As an admin user
1. Load the Events view /admin/content/events and filter by: Facility title contains = **Rocky Mountain** (note the total number of items)
2. Load the Bulk edit view /admin/content/bulk and filter by: Title = **Rocky Mountain** and Content type = **VAMC Facility Health Service** (note the total number of items)
3. Load the Bulk edit view /admin/content/bulk and filter by: Title = **Rocky Mountain** and Content type = **VAMC Facility Non-clinical Service** (note the total number of items)
4. Load the following VAMC Facility Node page: /eastern-colorado-health-care/locations/rocky-mountain-regional-va-medical-center
5. Edit the Facility node, change the moderation state to Archived

- [ ] Verify the facility is unpublished
- [ ] Verify the status message shows a number of additional items that were archived
- [ ] Refresh the view pages listed above and verify the items are all archived
- [ ] Open one item of each type and ensure there is a revision log message stating that the item was automatically archived when the parent facility was archived.

### Definition of Done

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [ ] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [x] `⭐️ Facilities`
- [ ] `⭐️ User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
+++ b/docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
@@ -91,6 +91,45 @@ class FacilitiesSubscriber implements EventSubscriberInterface {
    */
   public function entityUpdate(EntityUpdateEvent $event): void {
     $entity = $event->getEntity();
+    $this->archiveRelatedFacilityContent($entity);
+  }
+
+  /**
+   * Clear status details when operating status is normal.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   Entity.
+   */
+  protected function clearNormalStatusDetails(EntityInterface $entity): void {
+    if ($entity instanceof NodeInterface) {
+      $facilitiesWithStatus = [
+        'health_care_local_facility',
+        'nca_facility',
+        'vba_facility',
+        'vet_center',
+        'vet_center_outstation',
+      ];
+      $bundle = $entity->bundle();
+      /** @var \Drupal\node\NodeInterface $entity */
+      if (in_array($entity->bundle(), $facilitiesWithStatus)
+      && ($entity->hasField('field_operating_status_facility'))
+      && ($entity->hasField('field_operating_status_more_info'))) {
+        $status = $entity->get('field_operating_status_facility')->value;
+        $details = $entity->get('field_operating_status_more_info')->value;
+        if ($status === 'normal' && !empty($details)) {
+          $entity->set('field_operating_status_more_info', '');
+        }
+      }
+    }
+  }
+
+  /**
+   * Archive related content when a facility has been archived.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   Entity.
+   */
+  protected function archiveRelatedFacilityContent(EntityInterface $entity): void {
     if ($entity instanceof NodeInterface && $entity->bundle() === 'health_care_local_facility') {
       // When a facility is archived auto archive related services and events.
       if (isset($entity->original) && ($entity->original instanceof NodeInterface)) {
@@ -150,35 +189,6 @@ class FacilitiesSubscriber implements EventSubscriberInterface {
           $this->messenger->addStatus($message);
         }
 
-      }
-    }
-  }
-
-  /**
-   * Clear status details when operating status is normal.
-   *
-   * @param \Drupal\Core\Entity\EntityInterface $entity
-   *   Entity.
-   */
-  protected function clearNormalStatusDetails(EntityInterface $entity): void {
-    if ($entity instanceof NodeInterface) {
-      $facilitiesWithStatus = [
-        'health_care_local_facility',
-        'nca_facility',
-        'vba_facility',
-        'vet_center',
-        'vet_center_outstation',
-      ];
-      $bundle = $entity->bundle();
-      /** @var \Drupal\node\NodeInterface $entity */
-      if (in_array($entity->bundle(), $facilitiesWithStatus)
-      && ($entity->hasField('field_operating_status_facility'))
-      && ($entity->hasField('field_operating_status_more_info'))) {
-        $status = $entity->get('field_operating_status_facility')->value;
-        $details = $entity->get('field_operating_status_more_info')->value;
-        if ($status === 'normal' && !empty($details)) {
-          $entity->set('field_operating_status_more_info', '');
-        }
       }
     }
   }
--- a/docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
+++ b/docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
@@ -91,7 +91,7 @@ class FacilitiesSubscriber implements EventSubscriberInterface {
    */
   public function entityUpdate(EntityUpdateEvent $event): void {
     $entity = $event->getEntity();
-    $this->archiveRelatedFacilityContent($entity);
+    $this->archiveRelatedHealthFacilityContent($entity);
   }
 
   /**
@@ -129,15 +129,10 @@ class FacilitiesSubscriber implements EventSubscriberInterface {
    * @param \Drupal\Core\Entity\EntityInterface $entity
    *   Entity.
    */
-  protected function archiveRelatedFacilityContent(EntityInterface $entity): void {
+  protected function archiveRelatedHealthFacilityContent(EntityInterface $entity): void {
     if ($entity instanceof NodeInterface && $entity->bundle() === 'health_care_local_facility') {
       // When a facility is archived auto archive related services and events.
-      if (isset($entity->original) && ($entity->original instanceof NodeInterface)) {
-        $currentState = $entity->get('moderation_state')->value;
-        $oldState = $entity->original->get('moderation_state')->value;
-        if ($oldState === 'archived' || $currentState !== 'archived') {
-          return;
-        }
+      if (isset($entity->original) && ($entity->original instanceof NodeInterface) && $this->isBeingArchived($entity)) {
         $facilityID = $entity->id();
         $relatedTypes = [
           'event',
@@ -150,7 +145,7 @@ class FacilitiesSubscriber implements EventSubscriberInterface {
         $query = $nodeStorage->getQuery();
         $query->condition('type', $relatedTypes, 'IN')
           ->condition('field_facility_location', $facilityID)
-          ->condition('moderation_state', 'published');
+          ->condition('moderation_state', 'archived', '!=');
         $nids = $query->execute();
 
         if (count($nids)) {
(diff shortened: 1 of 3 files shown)


docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
```diff
@@ -33,6 +83,76 @@ public function entityPresave(EntityPresaveEvent $event): void {
...
 
+  /**
+   * Entity update Event call.
+   *
+   * @param \Drupal\core_event_dispatcher\Event\Entity\EntityUpdateEvent $event
+   *   The event.
+   */
+  public function entityUpdate(EntityUpdateEvent $event): void {
+    $entity = $event->getEntity();
+    if ($entity instanceof NodeInterface && $entity->bundle() === 'health_care_local_facility') {
```
swirtSJW, line 94: Since this is specific to health care facilities,  should this go into va_gov_vamc instead?

I can see a case to be made for needing similiar removal for vet centers but I think that will be using different fields and may involve related outstations, mobile and caps... so I think the logic would be significantly different.

docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
```diff
@@ -33,6 +83,76 @@ public function entityPresave(EntityPresaveEvent $event): void {
...
+        }
+        $facilityID = $entity->id();
+        $relatedTypes = [
+          'event',
+          'health_care_local_health_service',
+          'vha_facility_nonclinical_service',
+        ];
+
+        // Find all related published services and events.
+        $query = \Drupal::entityQuery('node');
```
swirtSJW, line 110: Yep gotta agree with Stan here.  Pass this in as argument defined in the services yml

docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
```diff
@@ -33,6 +83,76 @@ public function entityPresave(EntityPresaveEvent $event): void {
...
+              // Set revision author to current user.
+              $node->setRevisionUserId($this->currentUser->id());
+              $node->setChangedTime(time());
+              $node->isDefaultRevision(TRUE);
+              $node->setRevisionCreationTime(time());
+              // Set revision log message.
+              $node->setRevisionLogMessage('Automatically archived when parent facility was archived.');
+              $node->save();
+              // Increment updated count.
+              $updated[$node->bundle()] += 1;
```
swirtSJW, line 139: minor: 
```suggestion
              $updated[$node->bundle()]++;
```

swirtSJW (changes requested):
Some changes needed.   We can kick around the idea of which module it should live in if you have other ideas.

swirtSJW:
This PR should also update the facility removed runbook template to remove the related manual steps.

docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
ethanteague: Not to pile on here, but all of this logic in entityUpdate (where ever this subscriber hook ultimately lives) should probably be called from a discrete function, e.g.: 
```
public function entityUpdate(EntityUpdateEvent $event): void {
  $entity = $event->getEntity();
  someDiscreetFunctionThatDescribesWhatYouAreDoing($entity);
}

Comment (JayDarnell) [reply docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php:94]: @ethanteague what is the main motivation behind moving the logic into a separate function call?

File: docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php [hunk 4]
```diff
@@ -62,4 +123,74 @@ protected function clearNormalStatusDetails(EntityInterface $entity): void {
...
+          'health_care_local_health_service',
+          'vha_facility_nonclinical_service',
+        ];
+
+        // Find all related published services and events.
+        $nodeStorage = $this->entityTypeManager->getStorage('node');
+        $query = $nodeStorage->getQuery();
+        $query->condition('type', $relatedTypes, 'IN')
+          ->condition('field_facility_location', $facilityID)
+          ->condition('moderation_state', 'published');
```
swirtSJW, line 153: I think we want this to be anything but 'archived'   If we just limit it to published, we will be creating orphans for anything that is in draft.

docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
```diff
@@ -62,4 +123,74 @@ protected function clearNormalStatusDetails(EntityInterface $entity): void {
...
+        if (count($nids)) {
+          $nodes = $nodeStorage->loadMultiple($nids);
+          $updated = [
+            'event' => 0,
+            'health_care_local_health_service' => 0,
+            'vha_facility_nonclinical_service' => 0,
+          ];
+          foreach ($nodes as $node) {
+            if ($node instanceof NodeInterface) {
+              // Make this change a new revision.
```
swirtSJW, line 165: minor:  If the functions being called are all well named like these are, the comments aren't adding much value. 
example: `// Make this change a new revision.`   is not much improvement over `->setNewRevision(TRUE);`

Almost all of these node operations are well named and don't benefit from the comments.

docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
```diff
@@ -62,4 +123,74 @@ protected function clearNormalStatusDetails(EntityInterface $entity): void {
...
+              $node->setRevisionCreationTime(time());
+              // Set revision log message.
+              $node->setRevisionLogMessage('Automatically archived when parent facility was archived.');
+              $node->save();
+              // Increment updated count.
+              $updated[$node->bundle()]++;
+            }
+          }
+
+          // Status message so editor knows the extent of the changes.
```
swirtSJW, line 183: This is a good comment because it explains why. and is not repeating what the code tells.

docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
```diff
@@ -62,4 +123,74 @@ protected function clearNormalStatusDetails(EntityInterface $entity): void {
...
+   */
+  protected function archiveRelatedFacilityContent(EntityInterface $entity): void {
+    if ($entity instanceof NodeInterface && $entity->bundle() === 'health_care_local_facility') {
+      // When a facility is archived auto archive related services and events.
+      if (isset($entity->original) && ($entity->original instanceof NodeInterface)) {
+        $currentState = $entity->get('moderation_state')->value;
+        $oldState = $entity->original->get('moderation_state')->value;
+        if ($oldState === 'archived' || $currentState !== 'archived') {
+          return;
+        }
```
swirtSJW, line 140: This is a good hunk to move into its own method
something like ->isBeingArchived($entity);
Then your "if" higher up becomes
if (isset($entity->original) && ($entity->original instanceof NodeInterface) && $this->isBeingArchived($entity)) {

docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
```diff
@@ -33,6 +83,17 @@ public function entityPresave(EntityPresaveEvent $event): void {
...
 
+  /**
+   * Entity update Event call.
+   *
+   * @param \Drupal\core_event_dispatcher\Event\Entity\EntityUpdateEvent $event
+   *   The event.
+   */
+  public function entityUpdate(EntityUpdateEvent $event): void {
+    $entity = $event->getEntity();
+    $this->archiveRelatedFacilityContent($entity);
```
swirtSJW, line 94: Since this method is specific to health care facilities, it would be good to make this more transparent.
```suggestion
    $this->archiveRelatedHealthFacilityContent($entity);
```

docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
```diff
@@ -62,4 +123,74 @@ protected function clearNormalStatusDetails(EntityInterface $entity): void {
     }
   }
 
+  /**
+   * Archive related content when a facility has been archived.
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   Entity.
+   */
+  protected function archiveRelatedFacilityContent(EntityInterface $entity): void {
```
swirtSJW, line 132:
```suggestion
  protected function archiveRelatedHealthFacilityContent(EntityInterface $entity): void {
```
If in the future we make this more broad, we can make the name more generic, but for now it is pretty specific.

docroot/modules/custom/va_gov_facilities/src/EventSubscriber/FacilitiesSubscriber.php
```diff
@@ -62,4 +123,74 @@ protected function clearNormalStatusDetails(EntityInterface $entity): void {
...
+              $node->set('moderation_state', 'archived');
+              $node->setUnpublished();
+              // Set revision author to current user.
+              $node->setRevisionUserId($this->currentUser->id());
+              $node->setChangedTime(time());
+              $node->isDefaultRevision(TRUE);
+              $node->setRevisionCreationTime(time());
+              // Set revision log message.
+              $node->setRevisionLogMessage('Automatically archived when parent facility was archived.');
+              $node->save();
```
swirtSJW, line 177: This is all likely to be re-used when we do vet centers, so it could be moved to method  like
archiveNode($node);

swirtSJW (commented):
Some suggestions to make this a little clearer and re-useable when we tackle this for vetcenters.

kevwalsh:
Now that i see it in practice, I worry that auto-archiving events may be too opinionated. Events are not really "child" content in the same way that facility services are.

To me, that step is really about reducing broken links, but i think we should worry about that in a different way, which Sitewide CMS team is thinking about in the broader archiving workflow (alerting editors to possible broken links that can happen if they archive a node). And If a redirect is in place for this facility URL, is it really so bad that an event (whether future or past) stays published with a reference to a closed facility?

swirtSJW:
> is it really so bad that an event (whether future or past) stays published with a reference to a closed facility?

It would annoy me if I were looking at an event and then discovered it only to get redirected to the facility that may not be the actual location.    Creates a recipe for a disgruntled veteran.

JayDarnell:
It is important to note that prior to this update our manual steps included archiving any events for the facility in question so this automatic archiving still accomplishes that same step, just without the manual work.

swirtSJW (approved):
This tests out great
image

image


Also checked recent log messages and no bad notifications or errors found.

Nice work @JayDarnell

swirtSJW:
image

Issue #9872, opened 2022-07-20
Update Drupal to 9.3.19


timcosgrove:
## Description
Security advisory: https://www.drupal.org/sa-core-2022-014 among others.

### CMS Team
Please check the team(s) that will do this work.

- [ ] `Program`
- [x] `Platform CMS Team`
- [ ] `Sitewide Crew`
- [ ] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public Websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

PR #9877, opened 2022-07-21
Vacms 9872 cleanup

timcosgrove:
## Description

Related to #9872 and also to https://github.com/department-of-veterans-affairs/va.gov-cms/pull/9595 which was previously merged. 

#9595 modified `docroot/sites/development.services.yml` to enable Twig debugging. However, this file is owned by Drupal core, and so every time Drupal core is updated, these changes would be removed by the drupal/core-composer-scaffold mechanism. These files came up as modified when I updated core to 9.3.19 today.

We want to allow people the option to enable or disable both the core-provided development services and also their own at will, without forcing that on them. We also want that enablement to be .gitignored. 

### Select Team for PR review

- [x] `Platform CMS Team`
- [ ] `Sitewide program`
- [x] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`


docroot/sites/default/default.settings.php
```diff
@@ -490,6 +490,29 @@
  */
 # $settings['file_public_path'] = 'sites/default/files';
 
+/**
```
timcosgrove, line 493: This change came in between Drupal 9.3.17 and 9.3.19.

docroot/sites/development.services.yml
```diff
@@ -4,9 +4,6 @@
 # 'example.settings.local.php' file, which sits next to this file.
 parameters:
   http.response.debug_cacheability_headers: true
-  twig.config:
```
timcosgrove, line 7: This is not in core, so drupal/core-composer-scaffold will remove it on every core update.

docroot/sites/default/settings/settings.local.php
```diff
@@ -44,6 +44,3 @@
 ];
 
 $settings['cms_datadog_api_key'] = getenv('CMS_DATADOG_API_KEY');
-
```
timcosgrove, line 47: Moved this to example.settings.personal.php.

.gitignore
```diff
@@ -59,6 +59,8 @@ node_modules/
 /docroot/sites/*/local.settings.php
 /docroot/sites/*/settings.local.php
 /docroot/sites/all/drush/drushrc.php
+/docroot/sites/default/settings/settings.personal.php
```
tjheffner, line 62: i guess i am unclear why need need `settings.personal.php` when settings.local.php is gitignored already two lines above. a recent change (i believe from @cweagans) has us using `sites/default/settings.local.php` (not inside the `settings/` directory for whatever reason

READMES/local.md
```diff
@@ -16,6 +16,12 @@ This occurs because Drupal checks and hardens file permissions under the setting
 
 ```$settings['skip_permissions_hardening'] = TRUE;```
 
+## Local development settings
+Common development settings shared for all local development are at `docroot/sites/default/settings/settings.local.php`. This is committed to the repo and shared for all engineers.
+
+Engineers can customize their local environment by creating a file at `docroot/sites/default/settings/settings.personal.php`. This file will be ignored by git. An example is provided at `docroot/sites/default/settings/example.settings.personal.php`. Drupal core also provides some common local development settings at `docroot/sites/default/example.settings.local.php`.
```
tjheffner, line 22: if you look at `example.settings.local.php` the directions are different from this for setting up local customizations. i think we should either consolidate around the one (imo, `settings.local.php`, which is used already) or update everything to `settings.personal.php`

docroot/sites/default/settings.php
```diff
@@ -185,6 +185,8 @@
   __DIR__ . '/settings/settings.deploy.active.php',
   // Local overrides
   __DIR__ . '/settings.local.php',
+  // Personal overrides for local development purposes.
+  __DIR__ . '/settings/settings.personal.php',
```
tjheffner, line 189: this is duplicating the line above, imo

docroot/sites/development.services.yml
dsasser: Good call on that, I completely forgot this was a Drupal owned file.

docroot/sites/example.vagov-dev.services.yml
```diff
@@ -0,0 +1,8 @@
+# Local development services.
```
dsasser, line 1: Personally, I prefer a filename format similar to the environment it will be used in. Something like `local.services.yml`.

Additionally, as devs may want to add/change development settings beyond the committed services settings, a git ignored `local-override.services.yml` in the same location would allow a dev to create such file without fear of it getting committed.

docroot/sites/default/settings/example.settings.personal.php
```diff
@@ -0,0 +1,27 @@
+<?php
```
dsasser, line 1: I prefer the naming convention: `example.settings.local.php`.

docroot/sites/default/settings/settings.local.php
timcosgrove: Still don't have a place for this.

timcosgrove: I think this doesn't need to be committed at all.

.gitignore
timcosgrove: Removed

docroot/sites/default/settings/example.settings.personal.php
timcosgrove: That one is provided by core, so we can't add to it. I removed this file entirely.

dsasser (approved):

PR #9878, opened 2022-07-21
VACMS-9514: Removes site_alert module.

### LINKED ISSUE #9514: Remove site alert module from codebase
After the site alert module has been uninstalled, we should remove it from composer and any references to site alert config/functionality should be updated as appropriate for the new module. Much of the information is likely to be relevant still, maybe a slightly different admin path or two for actually adding or managing alerts.

ndouglas:
Closes #9514.

Mostly documentation and Composer changes, no real technical QA needed.


composer.lock
```diff
@@ -27519,5 +27467,5 @@
     "platform-overrides": {
         "php": "7.4"
     },
-    "plugin-api-version": "2.3.0"
+    "plugin-api-version": "2.2.0"
```
swirtSJW, line 27470: Should this reversion be happening?  Seems unexpected.

swirtSJW (commented):
Changes look good except for maybe that sus change to the API version.

I also confirmed that this module has no hook uninstall so removal is good to go.

composer.lock
ndouglas: That's just metadata indicating which version of Composer last modified the lock file.  It's harmless for these to flip back and forth, at least at this time.  Whoever or whatever (looking at GHA) last ran Composer had 2.3.x installed, whereas I was on 2.2.9.  Probably not a bad idea for me to update, though, so I did 🤷🏻‍♂️

swirtSJW (approved):
This is good to go and can be merged right after conflict is resolved 
image

Issue #10249, opened 2022-08-15
Sprint 65 Quality Assurance incidental work


ndouglas:
### Table
| Date | Time | Duration | Description | Link(s) | Issue(s) or Resolution |
| -- | -- | -- | -- | -- | -- |
| 08/16 | 08:30 AM | 30m | Dependabot security alerts | [Dependabot Security Alerts](https://github.com/department-of-veterans-affairs/va.gov-cms/security/dependabot) | Updated graphql explorer package dependencies |
| 08/16 | 01:00 PM | 30m | Noticed Drupal's behind by a couple security updates [Drupal Security Alert](https://www.drupal.org/project/drupal/releases/9.3.21) | #10290 |
| 08/18 | 09:30 AM | 30m | content-build-gql tests failing | [Slack Thread](https://dsva.slack.com/archives/C03LFSPGV16/p1660830089022969) | |
| 08/22 | 09:30 AM | 30m | staging post-deploy tests failed | [Slack Thread](https://dsva.slack.com/archives/CJT90C0UT/p1661176152109289) | |
| 08/22 | 10:00 AM | 15m | module update failed | [PR](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/10362) | removed conflicting patch |
| 08/22 | 04:45 PM | 15m | dev was rightly confused by lack of documentation | [Slack Thread](https://dsva.slack.com/archives/CT4GZBM8F/p1661200688076879) | [Created documentation](https://github.com/department-of-veterans-affairs/va.gov-cms/commit/8b6024f146b5772de523c08376860e273319220d) |
| 08/23 | 06:30 AM | 90m | PRs failing | [Pull Request](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/10385) | [Pull Request to solve issue](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/10393).
| 08/23 | 09:45 AM | 15m | DDEV injecting cruft | [Pull Request comment](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/10279#pullrequestreview-1082138500) | [Pull Request to solve issue](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/10396). |
| 08/25 | 11:00 AM | 15m | module update failed (again) | [PR](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/10362) | removed conflicting patch (again) |

PR #10393, opened 2022-08-23
VACMS-10249: Directly require phpspec/prophecy.

ndouglas:
## Description

Relates to #10249.

`phpspec/prophecy` is being dropped as a dependency by `phpunit/phpunit` in 8.5.29.  This is causing PRs to fail.

See #10385.

If tests pass, this is good to merge.


swirtSJW (approved):
Good catch Nate

PR #10396, opened 2022-08-23
VACMS 10249: Add .ddev/providers/platform.yaml to .gitignore.

ndouglas:
## Description

Relates to #10249.

This file is continually recreated by at least some versions of DDEV.  I don't know why.  I know @edmund-dunn and I, at least, are affected by this.

I'm adding it to the .gitignore so that it doesn't pop up unexpectedly and annoy us.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,3 +92,6 @@ datadog-setup.php
 
 # Temporary .deb packages
 *.deb
+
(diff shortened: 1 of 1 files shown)


ndouglas:
Closed in favor of #10404

PR #9461, opened 2022-06-15
VACMS-8588: Convert all remaining lando references to ddev (2)

ElijahLynn:
I prematurely merged https://github.com/department-of-veterans-affairs/va.gov-cms/pull/9401 and had to revert it in https://github.com/department-of-veterans-affairs/va.gov-cms/pull/9459. So this is a revert of that revert and essentially is the same as the original #9401.

PR #10397, opened 2022-08-23
VACMS 0000: Work on making local (dev-specific) settings and container services files more consistent and intuitive 😬 

ndouglas:
**TL;DR**: Settings and container service overrides are a little messy and inconsistent right now.  I'd like to try and make this a little less confusing.  Possibly.

The original scope of this PR was just trying to include a local dev container service file, but `.gitignore` and the docs did not agree on where this file should be located, and even if they did, the file would not be included anyway.  Which meant you should add some custom PHP to include that file (not ideal for non-Drupal developers).  

While the appropriate place to add that code is documented and works (`docroot/sites/default/settings.local.php`), I think it's confusingly similar to `docroot/sites/default/settings/settings.local.php`.  It's also included in `settings.php`, and so an inadvertently-committed file could sneak through and have bad effects on prod.

It was difficult to determine what files actually were included, git-ignored, etc, so I wrote a script to do it for me.

## Current Status on this PR branch

```
$ ddev php ./settings_simulator.php local

Settings files that were actually included:
Array
(
    [0] => /var/www/html/docroot/sites/default/settings.php
    [1] => /var/www/html/docroot/sites/default/settings/settings.flysystem.php
    [2] => /var/www/html/docroot/sites/default/settings/settings.ddev.php
    [3] => /var/www/html/docroot/sites/default/settings/settings.personal.php
    [4] => /var/www/html/docroot/sites/default/settings/settings.fast_404.php
)


Possible local override settings files:
                                                   File Path          Exists  Is Git-Ignored     Is Included
                                    sites/local.settings.php           FALSE           FALSE           FALSE
                                    sites/settings.local.php           FALSE           FALSE           FALSE
                            sites/default/local.settings.php           FALSE           FALSE           FALSE
                            sites/default/settings.local.php           FALSE           FALSE           FALSE
                   sites/default/settings/local.settings.php           FALSE           FALSE           FALSE
                   sites/default/settings/settings.local.php           FALSE           FALSE           FALSE
                    sites/default/settings/settings.ddev.php            TRUE           FALSE            TRUE
                sites/default/settings/settings.personal.php           FALSE            TRUE            TRUE


Container YAML files that were actually included:
Array
(
    [0] => /var/www/html/docroot/sites/default/../default/services.yml
    [1] => /var/www/html/docroot/sites/default/services/services.personal.yml
    [2] => /var/www/html/docroot/sites/default/services/services.monolog.yml
    [3] => /var/www/html/docroot/sites/default/services/services.ddev.yml
)


Possible local override container YAML files:
                                                   File Path          Exists  Is Git-Ignored     Is Included
                                    sites/local.services.yml           FALSE           FALSE           FALSE
                                    sites/services.local.yml           FALSE           FALSE           FALSE
                            sites/default/local.services.yml           FALSE           FALSE           FALSE
                            sites/default/services.local.yml           FALSE           FALSE           FALSE
                   sites/default/services/local.services.yml           FALSE           FALSE           FALSE
                   sites/default/services/services.local.yml           FALSE           FALSE           FALSE
                    sites/default/services/services.ddev.yml            TRUE           FALSE            TRUE
                sites/default/services/services.personal.yml           FALSE            TRUE            TRUE

```

## Current Status on `main`

```
$ ddev php ./settings_simulator.php local


Settings files that were actually included:
Array
(
    [0] => /var/www/html/docroot/sites/default/settings.php
    [1] => /var/www/html/docroot/sites/default/settings/settings.local.php
    [2] => /var/www/html/docroot/sites/default/settings/settings.fast_404.php
    [3] => /var/www/html/docroot/sites/default/settings.local.php
)


Possible local override settings files:
                                                   File Path          Exists  Is Git-Ignored     Is Included
                                    sites/local.settings.php           FALSE           FALSE           FALSE
                                    sites/settings.local.php           FALSE           FALSE           FALSE
                            sites/default/local.settings.php           FALSE            TRUE           FALSE
                            sites/default/settings.local.php           FALSE            TRUE            TRUE
                   sites/default/settings/local.settings.php           FALSE           FALSE           FALSE
                   sites/default/settings/settings.local.php            TRUE           FALSE            TRUE


Container YAML files that were actually included:
Array
(
    [0] => /var/www/html/docroot/sites/default/../default/services.yml
    [1] => /var/www/html/docroot/sites/default/services/services.monolog.yml
    [2] => /var/www/html/docroot/sites/default/services/services.local.yml
)


Possible local override container YAML files:
                                                   File Path          Exists  Is Git-Ignored     Is Included
                                    sites/local.services.yml           FALSE            TRUE           FALSE
                                    sites/services.local.yml           FALSE           FALSE           FALSE
                            sites/default/local.services.yml           FALSE           FALSE           FALSE
                            sites/default/services.local.yml           FALSE           FALSE           FALSE
                   sites/default/services/local.services.yml           FALSE           FALSE           FALSE
                   sites/default/services/services.local.yml           FALSE           FALSE            TRUE
```

## To Do
I haven't updated any documentation yet, since there's currently no agreement about how any of this should function.


dsasser:
I think this is actually possible with the existing settings:

Local settings is git ignored: `/docroot/sites/*/local.settings.php`
Local services yml is ignored: `/docroot/sites/local.services.yml`

ndouglas:
Okay, this is breaking my brain, so I'm gonna go through it slowly.

## Local Settings

> Local settings is git ignored: `/docroot/sites/*/local.settings.php` 

Indeed, as of three years ago:

Screen Shot 2022-08-23 at 5 02 12 PM

But `settings.php` includes `/docroot/sites/settings.local.php`:

Screen Shot 2022-08-23 at 4 58 06 PM

And I can't find any reference to `local.settings.php` anywhere.  Maybe I'm missing it?

The documentation says:

>Engineers can customize their individual local environment by creating a file at `docroot/sites/default/settings.local.php`. This file will be ignored by git. Drupal core provides some common local development settings at `docroot/sites/example.settings.local.php`.

But this is wrong (because AFAICT `docroot/sites/default/settings.local.php` is not included anywhere).

I think it's also confusing, because `docroot/sites/default/settings/settings.local.php` is a shared file, and it's easy to miss one of those path segments.  I've been missing them ever since I started trying to figure out what is going on here.  I have a headache and I want to cry a little.

**Conclusion**: I think I should update the documentation to refer to `docroot/sites/default/local.settings.php`.  This might hopefully distinguish it a little bit more from `docroot/sites/default/settings/settings.local.php`, maybe?  Thoughts?

## Local Services

>Local services yml is ignored: `/docroot/sites/local.services.yml`

Yes, indeed, as of last month:

Screen Shot 2022-08-23 at 4 56 52 PM

But that's an incorrect path.  It should be, I think, `/docroot/sites/default/local.services.yml`.

That was added in this PR: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/9877/files

Which adds the following documentation:

>Drupal core provides some optional development services at `docroot/sites/default/development.services.yml`. Custom services can be added at `docroot/sites/default/local.services.yml`. This file will be ignored by git. An example is provided at `docroot/sites/default/example.local.services.yml`.

This is correct, and matches my expectation, but doesn't match the .gitignore.

**Conclusion**: I should fix the .gitignore to point at the right location.  Also, I should change my file location to point to that same location, because the location this PR currently points to is wrong (it points to `docroot/sites/default/services/services.local.yaml`; the model of `settings/settings.local.php` would indicate that file would be shared by all devs and committed to the repository).

But I still think this is a good change to make, since AFAICT the container file isn't included anywhere.

So I'll make the following changes, if they make sense to you:

1. Update documentation to refer to `docroot/sites/default/local.settings.php` as the .gitignore'd, purely local file.
2. Update `.gitignore` to ignore `docroot/sites/default/development.services.yml`.
3. Update `docroot/sites/default/settings/settings.local.php` to add the container YAML `docroot/sites/default/local.services.yml` if it exists.
4. Remove the bit in `docroot/sites/default/settings.php` that adds `docroot/sites/default/services/services.local.yml` if it exists, because it's redundant.
5. Add/expand/update documentation to explain all of this a little bit better, because holy crap.

Does that make sense?  Does that sound right?  Am I going crazy?

swirtSJW:
Related discussion https://dsva.slack.com/archives/D03MGPMNVNV/p1656362572437689
and I swear Tim and Daniel recently has some issues around this  still looking for it.

swirtSJW:
Here it is https://dsva.slack.com/archives/C03LFSPGV16/p1658365413893619

ndouglas:
So I wrote this script:

```php
<?php

// phpcs:ignoreFile

/**
 * @file
 * This is the VA.gov-CMS Settings Simulator™.
 *
 * Its purpose is to preview how settings files are read and processed for each
 * environment and validate the resulting configuration against an existing
 * baseline.
 * 
 * Run it with `php settings_simulator.php [ENV]`, where:
 * - ENV is an environment name normally provided by the CMS_ENVIRONMENT_TYPE 
 *   environment variable (e.g. "local", "tugboat", "prod").
 * 
 * WARNING: Note that this script may have unintended consequences when run
 * locally!
 * 
 * It will create files in a variety of locations to see which are and are not
 * included, which are and are not git-ignored, etc.  It will normally delete
 * them, but it's possible that it will fail to do so.  If these files are not
 * git-ignored, they will need to be manually deleted.
 */

// Some initial configuration.
const SITES_PATH = __DIR__ . '/docroot/sites';

// Retrieve arguments.
$env_type = $argv[1] ?? getenv('CMS_ENVIRONMENT_TYPE');
putenv("CMS_ENVIRONMENT_TYPE={$env_type}");

// Populate some variables to simulate a satisfying Drupal experience.
// This will vary from environment to environment.
$app_root = '/var/www/html/docroot';
// This should remain constant across environments.
$site_path = 'sites/default';

// Let's create some settings files in a variety of locations and see which
// are included, which are git-ignored, etc.
$local_settings_files = [
  'sites/local.settings.php',
  'sites/settings.local.php',
  'sites/default/local.settings.php',
  'sites/default/settings.local.php',
  'sites/default/settings/local.settings.php',
  // Skipped because we've established this should be shared local-dev config.
  // 'sites/default/settings/settings.local.php',
];
$local_settings_files_to_remove = [];
$local_settings_data = [];
foreach ($local_settings_files as $local_settings_file) {
  // All roads lead to Rome; all settings_files begin with $app_root.
  $full_path = $app_root . '/' . $local_settings_file;
  $exists = file_exists($full_path);
  if (!$exists) {
    $local_settings_files_to_remove[] = $full_path;
    file_put_contents($full_path, '<?php ');
  }
  // If the following does not return NULL, this file is git-ignored.
  $is_gitignored = shell_exec("git check-ignore {$full_path}") != NULL;

  $local_settings_data[] = [
    'path' => $local_settings_file,
    'exists' => $exists,
    'is_gitignored' => $is_gitignored,
  ];
}

// Let's do the same with some container YAML files.
$local_container_yamls = [
  'sites/local.services.yml',
  'sites/services.local.yml',
  'sites/default/local.services.yml',
  'sites/default/services.local.yml',
  'sites/default/services/local.services.yml',
  'sites/default/services/services.local.yml',
];
$local_container_yamls_to_remove = [];
$local_container_yaml_data = [];
foreach ($local_container_yamls as $local_container_yaml) {
  $full_path = $app_root . '/' . $local_container_yaml;
  $exists = file_exists($full_path);
  if (!$exists) {
    $local_container_yamls_to_remove[] = $full_path;
    file_put_contents($full_path, '# Empty file.');
  }
  // If the following does not return NULL, this file is git-ignored.
  $is_gitignored = shell_exec("git check-ignore {$full_path}") != NULL;

  $local_container_yaml_data[] = [
    'path' => $local_container_yaml,
    'exists' => $exists,
    'is_gitignored' => $is_gitignored,
  ];
}

// First, include the existing settings.php file.
require_once SITES_PATH . '/default/settings.php';

// Remove files we created above.
foreach ($local_settings_files_to_remove as $full_path) {
  unlink($full_path);
}
foreach ($local_container_yamls_to_remove as $full_path) {
  unlink($full_path);
}

// Now, which settings files did we include?
$included_settings_files = get_included_files();
// Remove first entry (which will be this very file).
$included_settings_files = array_slice($included_settings_files, 1);

echo "\n\nSettings files that were actually included:\n";

print_r($included_settings_files);

$format_string = "%60s %15s %15s %15s\n";

echo "\n\nPossible local override settings files:\n";

printf($format_string, "File Path", "Exists", "Is Git-Ignored", "Is Included");

function format_bool($bool) {
  return !!$bool ? 'TRUE' : 'FALSE';
}

foreach ($local_settings_data as $data) {
  $is_included = array_filter($included_settings_files, function ($included_path) use ($data) {
    return strpos($included_path, $data['path']) !== FALSE;
  });
  printf($format_string, $data['path'], format_bool($data['exists']), format_bool($data['is_gitignored']), format_bool(count($is_included)));
}

// Which container YAML files will we use?
$container_yamls = $settings['container_yamls'];

echo "\n\nContainer YAML files that were actually included:\n";

print_r($container_yamls);

echo "\n\nPossible local override settings files:\n";

printf($format_string, "File Path", "Exists", "Is Git-Ignored", "Is Included");

foreach ($local_container_yaml_data as $data) {
  $is_included = array_filter($container_yamls, function ($included_path) use ($data) {
    return strpos($included_path, $data['path']) !== FALSE;
  });
  printf($format_string, $data['path'], format_bool($data['exists']), format_bool($data['is_gitignored']), format_bool(count($is_included)));
}
```

This should automate the process of figuring out what is actually happening in the repo with these files.  On my machine, in DDEV, on this PR branch, it executes without errors and prints the following:

```bash
$ ddev php ./settings_simulator.php local


Settings files that were actually included:
Array
(
    [0] => /var/www/html/docroot/sites/default/settings.php
    [1] => /var/www/html/docroot/sites/default/settings/settings.local.php
    [2] => /var/www/html/docroot/sites/default/settings/settings.fast_404.php
    [3] => /var/www/html/docroot/sites/default/settings.local.php
)


Possible local override settings files:
                                                   File Path          Exists  Is Git-Ignored     Is Included
                                    sites/local.settings.php           FALSE           FALSE           FALSE
                                    sites/settings.local.php           FALSE           FALSE           FALSE
                            sites/default/local.settings.php           FALSE            TRUE           FALSE
                            sites/default/settings.local.php           FALSE            TRUE            TRUE
                   sites/default/settings/local.settings.php           FALSE           FALSE           FALSE


Container YAML files that were actually included:
Array
(
    [0] => /var/www/html/docroot/sites/default/../default/services.yml
    [1] => /var/www/html/docroot/sites/default/services/services.monolog.yml
    [2] => /var/www/html/docroot/sites/default/services/services.local.yml
    [3] => /var/www/html/docroot/sites/default/services/services.local.yml
)


Possible local override settings files:
                                                   File Path          Exists  Is Git-Ignored     Is Included
                                    sites/local.services.yml           FALSE            TRUE           FALSE
                                    sites/services.local.yml           FALSE           FALSE           FALSE
                            sites/default/local.services.yml           FALSE           FALSE           FALSE
                            sites/default/services.local.yml           FALSE           FALSE           FALSE
                   sites/default/services/local.services.yml           FALSE           FALSE           FALSE
                   sites/default/services/services.local.yml           FALSE           FALSE            TRUE
```

In short, it lists the settings files and container yamls used by a specified environment.  Then it loops through some conjectured local settings file names and container YAML paths and prints whether it existed _prior_ to the script running, whether that path is git-ignored, and whether it ends up being included by `settings.php` as it stands (again, I ran this script against this PR branch).

My opinion, echoing what I wrote (possibly incomprehensibly) in my previous comment, is that this output should look like this:

```
$ ddev php ./settings_simulator.php local


Settings files that were actually included:
Array
(
    [0] => /var/www/html/docroot/sites/default/settings.php
    [1] => /var/www/html/docroot/sites/default/settings/settings.local.php
    [2] => /var/www/html/docroot/sites/default/settings/settings.fast_404.php
    [3] => /var/www/html/docroot/sites/default/local.settings.php
)


Possible local override settings files:
                                                   File Path          Exists  Is Git-Ignored     Is Included
                                    sites/local.settings.php           FALSE           FALSE           FALSE
                                    sites/settings.local.php           FALSE           FALSE           FALSE
                            sites/default/local.settings.php           FALSE            TRUE            TRUE
                            sites/default/settings.local.php           FALSE            FALSE           FALSE
                   sites/default/settings/local.settings.php           FALSE           FALSE           FALSE


Container YAML files that were actually included:
Array
(
    [0] => /var/www/html/docroot/sites/default/../default/services.yml
    [1] => /var/www/html/docroot/sites/default/services/services.monolog.yml
    [2] => /var/www/html/docroot/sites/default/services/services.local.yml
    [3] => /var/www/html/docroot/sites/default/services/services.local.yml
)


Possible local override settings files:
                                                   File Path          Exists  Is Git-Ignored     Is Included
                                    sites/local.services.yml           FALSE            TRUE           TRUE
                                    sites/services.local.yml           FALSE           FALSE           FALSE
                            sites/default/local.services.yml           FALSE           FALSE           FALSE
                            sites/default/services.local.yml           FALSE           FALSE           FALSE
                   sites/default/services/local.services.yml           FALSE           FALSE           FALSE
                   sites/default/services/services.local.yml           FALSE           FALSE            TRUE
```

In other words, 

1) git-ignore and include `sites/default/local.settings.php` if it exists.
2) git-ignore and include `sites/default/local.services.yml` if it exists.

Does that make sense?  Is that acceptable?

(Then, obv, update documentation to match)

.gitignore
```diff
@@ -56,11 +56,9 @@ node_modules/
 /docroot/sites/*/private
 /docroot/sites/*/drushrc.php
 /docroot/sites/*/settings.php
+/docroot/sites/*/local.services.yml
```
ndouglas, line 59: Local (dev-specific) services overrides.

.gitignore
```diff
@@ -56,11 +56,9 @@ node_modules/
...
 /docroot/sites/*/local.settings.php
-/docroot/sites/*/settings.local.php
```
ndouglas, line 60: I prefer `docroot/sites/default/local.settings.php` over `docroot/sites/default/settings.local.php` because the latter looks very similar to `docroot/sites/default/settings/settings.local.php`, which is confusing.  I think we tend to pay more attention to basenames than full paths, so having a different basename for this file seems better to me.

.gitignore
```diff
@@ -56,11 +56,9 @@ node_modules/
...
 /docroot/sites/all/drush/drushrc.php
-/docroot/sites/local.services.yml
```
ndouglas, line 62: This didn't match documentation.

docroot/sites/default/settings.php
```diff
@@ -183,8 +183,6 @@
   __DIR__ . '/settings/settings.fast_404.php',
   // Ansible moves this file into place during deploy, so if it is present we are in deploy mode.
   __DIR__ . '/settings/settings.deploy.active.php',
-  // Local overrides
-  __DIR__ . '/settings.local.php',
```
ndouglas, line 187: I thought it best to include local dev-specific settings overrides at the very end of `docroot/sites/default/settings/settings.local.php` to guarantee that 1) these settings are actual overrides, and 2) so that an inadvertently-committed file doesn't override prod settings.

docroot/sites/default/settings/settings.local.php
```diff
@@ -44,3 +44,15 @@
...
+$local_settings_path = __DIR__ . '/../local.settings.php';
+if (file_exists($local_settings_path)) {
+  include_once $local_settings_path;
+}
+
+// Local overrides for services container.
+$local_services_path = __DIR__ . '/../local.services.yml';
+if (file_exists($local_services_path)) {
+  $settings['container_yamls'][] = realpath($local_services_path);
+}
```
ndouglas, line 58: Include local overrides only from the shared local config, to protect against inadvertently-committed files.

settings_simulator.php
```diff
@@ -0,0 +1,150 @@
+<?php
```
ndouglas, line 1: Epic script.  Feeling cute, might delete later, idk.  Or I might turn it into a GHA, since it doesn't require bootstrapping Drupal.

.gitignore
dsasser: Agreed.

docroot/sites/default/settings/settings.local.php
dsasser: Maybe it is just me, but it is confusing to me to have both:
settings.local.php
local.settings.php

Just looking at the filename, it isn't clear that local.settings.php is the final, git ignored override. Still, I wouldn't block the PR on that alone; I can make sense of whatever we come up with.

settings_simulator.php
dsasser, line 1: Should this be in the /scripts folder?

dsasser: EPIC script BTW

ndouglas: Yeah, that's a better place for it.

docroot/sites/default/settings/settings.local.php
ndouglas: Yeah, I'm not married to this either.

I was thinking it might be good to do `settings.ddev.php` for shared config and `settings.local.php` for actual local config?

But I also don't like "local" anyway... for reasons.  I think it means something else.

Ugh, naming things.

swirtSJW: Yah we didn't use to have this confusion until https://github.com/department-of-veterans-affairs/va.gov-cms/pull/9461 was merged in.  After that, this having two files with the same name has been a source of confusion several times over.  The change was unrelated to the issue of "Removing all Lando references"

edmund-dunn: I think the use of `settings.ddev.php` might eliminate some confusion. I see anything with local in it and I think it is for local, on my machine, development. Ihope this makes sense.

dsasser: What about 'settings.personal.php' ?

edmund-dunn: That would work for me too.

ndouglas: In `sites/default/` or `sites/default/settings/`?

dsasser: > In sites/default/ or sites/default/settings/?

I don't have a strong preference, but would prefer sites/default/settings/ for consistency.

ndouglas: yeah, if we do `settings.personal.php` I like that consistency too.

.gitignore
```diff
@@ -1,3 +1,8 @@
+# Your very own services.yml
+/docroot/sites/default/services.personal.yml
+# Your very own settings.php
+/docroot/sites/default/settings.personal.php
+
```
ndouglas, line 5: Moved up to top to serve as a form of documentation.  I'll add more elsewhere.

settings_simulator.php
ndouglas: Moved it 🙂

docroot/sites/default/settings/settings.local.php
ndouglas: `ddev` in the streets, `personal` in the sheets.

.ddev/config.yaml
```diff
@@ -41,7 +41,7 @@ web_environment:
 - BLACKFIRE_CLIENT_TOKEN=~
 - BLACKFIRE_SERVER_ID=~
 - BLACKFIRE_SERVER_TOKEN=~
-- CMS_ENVIRONMENT_TYPE=local
+- CMS_ENVIRONMENT_TYPE=ddev
```
ndouglas, line 44: This is so local environments look for e.g. `settings.ddev.php`.

.gitignore
```diff
@@ -1,3 +1,8 @@
+# Your very own services.yml
+/docroot/sites/default/services/services.personal.yml
+# Your very own settings.php
+/docroot/sites/default/settings/settings.personal.php
+
```
ndouglas, line 5: Moved up top to catch the eye of onboarding devs.

docroot/sites/default/services/services.ddev.yml
```diff
@@ -0,0 +1,2 @@
+# This file is created as a placeholder for any service modifications or 
+# additions we want to create within DDEV environments.
```
ndouglas, line 2: Just throwing that out there.

dsasser (dismissed):

ndouglas:
I need to update the `config_split` on this too.

ndouglas:
Eh. I might revisit this later but it's been dead for a while. Gonna close it.

PR #10488, opened 2022-08-27
Revert "Bump drupal/smart_date from 3.5.1 to 3.6.0"

ndouglas:
Reverts department-of-veterans-affairs/va.gov-cms#10362


ndouglas:
Merging because this update apparently broke things.

swirtSJW:
3.6.1 is out now.   Maybe it resolves this..... or we need an upstream issue.

ndouglas:
I have an issue open to debug and write tests to prevent this in the future: #10490 

All of the patches for 3.6.0 looked really innocuous.  Perhaps we're depending on something that's broken in 3.5.1.

Issue #9905, opened 2022-07-26
Isolate and rework PHPUnit performance tests.


ndouglas:
## Description
We have a few PHPUnit performance tests, but they aren't written very well; they just execute some operation and fail if that operation takes longer than a certain amount of time.

This is not great because outlying figures cause test failures that can irritate and distract the developer.

A more statistically defensible test would be to run multiple iterations (say, 10), remove the shortest and longest times, and do a very simple analysis on the remaining times -- mean/median, etc.  It doesn't need to be complex or terribly long; it's just nice to avoid spurious failures like this:

![Screen Shot 2022-07-26 at 9 32 09 AM](https://user-images.githubusercontent.com/1318579/181018609-6d44151a-24a4-459e-828c-4b29e5ef929b.png)

Another nice thing to do would be to keep these tests in a separate suite.  This would help to minimize the effects of this change on the duration of the CI test runs.

**EDIT**: Tim [suggested](https://dsva.slack.com/archives/CT4GZBM8F/p1666201071140519) that these be a metric with a monitor.  He may be a Xoomer, but he's spitting straight facts here.

## Acceptance Criteria
- [ ] PHPUnit performance tests have been modified as described above.
- [ ] PHPUnit performance tests are moved to a separate suite that is executed concurrently in CI.

### CMS Team
Please check the team(s) that will do this work.

- [ ] `Program`
- [ ] `Platform CMS Team`
- [ ] `Sitewide Crew`
- [x] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public Websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

ndouglas:
The increases I've been doing are just Band-Aid™ brand bandages.

![Screen Shot 2022-09-02 at 7 29 28 AM](https://user-images.githubusercontent.com/1318579/188141166-ccde34cf-7d90-40cc-9ad2-6b21d8c05998.png)

ndouglas:
![2reqtg](https://user-images.githubusercontent.com/1318579/188141250-55f05a2e-0c57-44f7-9ec4-f2675f511029.png)

PR #10522, opened 2022-08-30
VACMS-9905: Temporarily increase CreateMediaTest threshold.

ndouglas:
## Description

Relates to #9905.

The CreateMediaTest threshold was set when the tests ran serially.  It frequently fails now that tests are run concurrently.

I have an issue dedicated to remediating this in the longterm (see #9905), but in the short-term, these spurious test failures are really annoying and do not serve any purpose.

This PR increases the time limit from 2 (which is frequently exceeded, in my experience) to 4, which is very rarely exceeded and may indicate the existence of an actual problem.

If tests pass, this should be good to go.


swirtSJW (approved):
Good call.  Thanks for being proactive on this.

Issue #10466, opened 2022-08-25
Broken link checking for next-build pages in content-build


timcosgrove:
## Description
Content-build parses all HTML it knows about, gathers any internal links and files it finds in there, and checks those URLs against what it is build, gathering those links into a broken link report and in some cases halting the build.

Next-build's content is structured different, and so the parsing to identify these links will possibly be different. We need to test this content as well and ensure we are finding broken links in it.

First thing will be to correct content-build's local broken link checking.

## Acceptance Criteria
- [ ] Broken links within content created with next-build are reported in the broken links report correctly
- [ ] content-build pages that link to next-build URLs that are missing are identified as broken

### CMS Team
Please check the team(s) that will do this work.

- [ ] `Program`
- [x] `Platform CMS Team`
- [ ] `Sitewide Crew`
- [ ] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public Websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

### Implementation tasks
- [x] Fix local broken link checking for content-build (run with --nosymlink to download assets.)
- [x] Baseline report of broken links on a local content-build
- [x] Baseline report of broken links while running next-build on a local content-build
- [ ] Point at local drupal instance, introduce a purposefully broken link and see it in the report
- [ ] Point at local drupal instance, introduce a purposefully broken link in a next-build rendered page
- [ ] Ensure the time-to-run for the next-build steps are appropriately logged and reported
- [ ] ???
- [ ] profit

timcosgrove:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/VAgov-CMS-team-5c0e7b864b5806bc2bfc2087/issues/department-of-veterans-affairs/va.gov-cms/10466?planning-poker) with ZenHub @tjheffner @amponce

tjheffner:
ok - after pulling fresh drupal data i am seeing these baselines:

19 broken links with a standard content-build
33 broken links with a next-drupal enabled content-build, minus the 19 that were there, 14 new broken links. all appear to be from our demo pages.

Next I deleted the demo page and re-ran both types of builds for new baselines. 19 Broken links in the default content-build, 26 in the next-build-enabled content build.

[...]

PR #10630, opened 2022-09-06
VACMS-10466 increase jsonapi response max size for next

tjheffner:
## Description

Relates to #10466. 
Companion PR to https://github.com/department-of-veterans-affairs/content-build/pull/1295

## Testing done
in next-build, see that `yarn export` returns ~1600 nodes instead of ~950 without this override.

## Screenshots
n/a, sorry

## QA steps
in the next-build repo, point .env at your local ddev instance like so:
```
in .env.local: 
NEXT_PUBLIC_DRUPAL_BASE_URL=http://va-gov-cms.ddev.site
```

run `yarn export`. see that ~950 nodes are built.

now apply this service override in your local cms repo, clear cache

run `yarn export` in the next-build repo again, see that 1600 nodes are built. hooray!


timcosgrove (approved):

ndouglas:
Does this work by overriding the param that's normally populated from JSON:API's `SIZE_MAX` constant?

tjheffner:
@ndouglas yeah. it's kind of silly. jsonapi hardcodes to 50; jsonapi_extras lets you overwrite for specific endpoints in theory, but next_jsonapi sets that limit to 1000 (which happens to supersede any of the jsonapi_extras overrides). so we have to override it again, at least until pagination is respected 

we're overriding the value that comes from here: `docroot/modules/contrib/next/modules/next_jsonapi/next_jsonapi.services.yml`, used here: `docroot/modules/contrib/next/modules/next_jsonapi/src/Controller/EntityResource.php`, which in turn somehow gets read from the javascript package over here: https://github.com/chapter-three/next-drupal/blob/main/packages/next-drupal/src/client.ts#L798

it was a "fun" nesting doll to unpack

this slack thread has more context: https://dsva.slack.com/archives/C03MN1Y2X0V/p1662078509068309?thread_ts=1662066211.999429&cid=C03MN1Y2X0V

Issue #10298, opened 2022-08-16
Create Report of Modernized and Unmodernized URLs


EWashb:
## Description

Stakeholders request a report that pulls modernized versus unmodernized experience URLs. Bonus points if this report can be automated. 

[Slack thread](https://dsva.slack.com/archives/CT4GZBM8F/p1660675986641169) for request

This report should:

- Outright eliminate subdomains
- Compare anything remaining (www.va.gov/*) to see if it's from the CMS
- Provide a list of those URLS that are (www.va.gov/*) that are not matched to anything in the CMS to Mikki Northuis/OCTO
 for review (she will then cross-walk that smaller list with her view of VA.gov to see what's in the modernized experience)

The goal is to be able to provide this report and determine what URLs are modernized or not modernized. The modernized URLs mean that they are subject to the current platform scans for 508.

## Acceptance Criteria
- [x] A report of all unmodernized URLs
- [x] Report includes unique pageviews
- [x] Tech feasibility is determined for automation of this report
- [x] PM provides to PO

### CMS Team
Please check the team(s) that will do this work.

- [ ] `Program`
- [ ] `Platform CMS Team`
- [ ] `Sitewide Crew`
- [x] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public Websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

EWashb:
@ndouglas this is a special request by our stakeholders. Any idea on the lift of something like this? The automated part isn't required, but would be helpful when these requests come again.

ndouglas:
If I understand correctly, the approach from the technical side would be something like this:

- grab the [`sitemap`](https://www.va.gov/sitemap.xml)
- loop through it and parse each URL
  - if the path (e.g. `/abilene-vet-center`) corresponds to a piece of CMS content (e.g. https://prod.cms.va.gov/abilene-vet-center ), then put it in Column A (CMS)
  - otherwise (e.g. `/search`), put it in Column B (non-CMS)
- output Column B as a bullet list

Doing that in a PHP script is straightforward.  Making a page in Drupal that calculates this and then caches the result (because it _will_ take a while to calculate) is straightforward as well, but it should definitely be a separate issue and will have its own challenges.

Some concerns, though:

[...]

PR #10642, opened 2022-09-07
VACMS-10298: Creates script to create report of modernized and unmodernized pages.

ndouglas:
## Description

Relates to #10298.

No testing is needed; this will likely involve further development but this script isn't executed normally and doesn't need to be executed on production, so it's not a big deal.


ndouglas:
mind looking at this, @edmund-dunn ?  This should be ~zero risk.

edmund-dunn (approved):
Nicely done sir!!

PR #7981, opened 2022-02-11
VACMS-INTEGRATION recurring events

swirtSJW:
## Description
This is the PR that will track the integration branch into main.

Relates to #_issueid_.

## QA steps

As user _uid_ with _user_role_
1. Do this
2. Then that
3. Then validate Acceptance Criteria from issue
- [ ] This
- [ ] That
- [ ] The other thing

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide CMS Team`
  - [x] `⭐️ Content ops`
  - [ ] `⭐️ CMS Experience`
  - [ ] `⭐️ Offices`
  - [ ] `⭐️ Product Support`

### Is this PR blocked by another PR?

- [ ] `DO NOT MERGE`

### Does this PR need review from a Product Owner

- [ ] `Needs PO review`

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change?
- [ ] Yes, and it's written in issue ____ and queued for publication.
  - [ ] Merge and ping @ rachel-kauff so she's ready to publish after deployment
- [ ] Yes, but it hasn't yet been written
  - [ ] Don't merge yet -- ping @ rachel-kauff to prompt her to write and queue content
- [ ] No announcement is needed for this code change.
  - [ ] Merge & carry on unburdened by announcements

PR #10669, opened 2022-09-08
VACMS-10324: Hide all-day checkbox in recurring events

### LINKED ISSUE #10324: "All Day" event checkbox is off screen rather than hidden when creating a VAMC event
## Describe the defect
When editor creates a new event, there is a checkbox "All Day" associated with the date and time area that is off screen/hard to find, but should actually be hidden until All Day events are supported.

## To Reproduce
1. Go to any event in the VHA section of the CMS
2. Click on edit, then zoom out, making the text smaller until you see a little checkbox appear out to the left.

## AC / Expected behavior

- [ ] All Day events field should be hidden
- [ ] Add PR link to #3250 so how to unhide is clear when the time comes.

## Screenshots

Screen Shot 2022-08-25 at 3 24 11 PM

## Labels
- [ ] CMS subsystem (green)

### CMS Team

- [ ] `Program`
- [ ] `Platform CMS Team`
- [ ] `Sitewide Crew`
- [ ] `⭐️ Sitewide CMS`
- [x] `⭐️ Public Websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

chri5tia:
## Description

~Closes va.gov-cms issue 10324.~  <- Has no magic
Closes #10324.   <-has magic properties

## Testing done
On existing event (https://va-gov-cms.ddev.site/node/48108/edit), using the Chrome browser inspect tool:
1. Extend the length of the "All Day" label to make it more visible and expose it.
2. See that the `.hidden` property being applied to `.allday-label` is overridden. 
3. Find the `display:block` override, deselected it to see if removing the override fixes the issue.
4. Test was successful

(See screenshots below.)

## Screenshots
Screen Shot 2022-09-08 at 1 46 57 PM

Screen Shot 2022-09-08 at 1 47 02 PM

Screen Shot 2022-09-08 at 2 00 31 PM

Screen Shot 2022-09-08 at 2 00 36 PM

Notes: 

- This adds to work done in [PR #7981](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/7981)
- div sandwich:
Screen Shot 2022-09-08 at 5 23 52 PM

## QA steps
On https://cms-o4w3pg1yv2dgakucmkw9flhpss6wydd3.ci.cms.va.gov/node/48108/edit, in Chrome inspect or similar tool:
1. Extend the length of the "All Day" label with at least 300 characters (cut and paste some lorem ipsum or the label over and over) to make it more visible and expose it.
2. See that the `.hidden` property being applied to `.allday-label` is *not* overridden. 
3. See that there is not a `display:block !important` `.allday-label` property.
4. See that you still can't see the All Day label or checkbox, despite its absurdly long length.
Screen Shot 2022-09-08 at 5 31 26 PM

### Definition of Done

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [ ] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/docroot/themes/custom/vagovclaro/assets/scss/components/_recurring_events.scss
+++ b/docroot/themes/custom/vagovclaro/assets/scss/components/_recurring_events.scss
@@ -113,16 +113,6 @@
     margin: 0 5px 0 0;
   }
 
-  .allday-label {
-    clear: both;
-    /* display: block !important; // Don't allow display block override. */
-    float: none;
-    font-weight: 400;
-    left: -1000px;
-    margin: 10px 0 0;
-    position: absolute;
-  }
-
   .smartdate--time-inline {
     .form-datetime-wrapper {
       h4.form-item__label {
--- a/docroot/themes/custom/vagovclaro/assets/scss/components/_recurring_events.scss
+++ b/docroot/themes/custom/vagovclaro/assets/scss/components/_recurring_events.scss
@@ -115,7 +115,7 @@
 
   .allday-label {
     clear: both;
-    display: block !important;
+    /* display: block !important; // Don't allow display block override. */
     float: none;
     font-weight: 400;
     left: -1000px;


docroot/themes/custom/vagovclaro/assets/scss/components/_recurring_events.scss
```diff
@@ -115,7 +115,7 @@
 
   .allday-label {
     clear: both;
-    display: block !important;
+    /* display: block !important; // Don't allow display block override. */
```
dsasser, line 118: I would recommend removing this line, rather than commenting it out, unless there was a compelling reason to comment it?

docroot/themes/custom/vagovclaro/assets/scss/components/_recurring_events.scss
```diff
@@ -115,7 +115,7 @@
...
     float: none;
     font-weight: 400;
     left: -1000px;
```
dsasser, line 121: It looks like, instead of 'hiding' the element, the original code was moving the element off-screen. Did you determine that display:none was a better choice over absolutely positioning the element off screen? If so, removing the absolute positioning might make sense.

Alternatively, you could position the element further off screen be increasing the left attribute beyond any reasonable screen size (-10000px, for instance).

dsasser (commented):
Some questions/comments added.

docroot/themes/custom/vagovclaro/assets/scss/components/_recurring_events.scss
chri5tia: I decided to unblock the already existing .hidden property so that when this feature does work, I didn't make too many changes to how it already was previously. Whoever added the .hidden property can just unhide it and the placement will be as it was. I didn't want to mess with the absolute values or any of that since we don't know what it should be in the event that the checkbox becomes hidden again. On *my* screen, using different resolutions, the label was always right under or behind the "make recurring" and not off screen at all. I see no point in making it off screen when display:none makes it invisible and it will be visible again once that .hidden is removed from it.

dsasser: > Note: On my screen, using different resolutions, the label was always right under or behind the "make recurring" checkbox and not off screen at all. I see no value in making it off screen when display:none (already applied) makes it invisible and it can easily be visible again once that .hidden is removed from it.

Note that the issue describes _zooming out_ to see the issue. Different screen resolutions won't necessarily cause the issue to be visible.

> I unblocked the already existing .hidden property so that when this feature does work, there won't be too many changes to how it already was previously

But "how it already was previously" is incorrectly moving the element via absolute positioning. I recommend removing all absolute positioning to avoid this confusion later.

In summary, the changes I'm requesting are:
- Either - 
1) Remove the commented code. It is not a coding best practice, and won't be necessary to restore the visibility of the element in the future, because as you state, the .hidden class is the sole trigger for this going forward.
2) Remove the left and position properties. They won't be needed when we display this element in the future due to moving away from absolute positioning.
- OR - 
3) Consider removing the entire code block starting at line 116. It provides zero value while the element is hidden, and will only likely confuse things when we want to display the checkbox in the future, as element location may not need enhancement beyond what the smart_date module provides out of the box. Further, this element is receiving styling from a [different part of the theme](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/main/docroot/themes/custom/vagovclaro/assets/scss/components/_fields.scss#:~:text=.field%2D%2Dtype%2Dsmartdate%20%7B), so removing the entire code block here will reduce code duplication.

swirtSJW:
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/10324#issuecomment-1241417322

chri5tia:
The allday checkbox is added via javascript. 
At some point after that, a class "hidden" (`.hidden {display:none}`) was applied and then later a `display:none` as an inline style. Then more CSS was added which overrode the display:none as `display:block !important` but also set some parameters to push the block off screen. I remove that latter bit of CSS so that the `.hidden` class will work. 
When efforts are applied to get that allday checkbox working again, it can happen so that the `.hidden` class is removed.
What I did here is remove some technical debt. 

Front-end work is needed to actually fully remove the option, but that end, we may want to have that work applied to fixing it in the templates (the logic that can't apply no having a start or end date). That's for the future to determine. This is read to merge.

dsasser (approved):

swirtSJW (approved):
Performed a full creaet save, edit round.  All worked appropriately.
image

Issue #10617, opened 2022-09-02
Investigate best practices for using external GitHub Actions.


ndouglas:
## Description
We had a rude awakening today about GitHub Actions, but this was in addition to some natural pre-existing concerns about GitHub Actions.

We should research and implement better practices.

Thoughts:
- pinning GHA references to specific commits
- forking actions to VA repositories for internal use or replacing them with inline code wherever possible

## Acceptance Criteria
- [x] Best practices for using GitHub Actions safely and securely have been investigated.
- [x] Followup issue(s) opened to document and implement these practices.
  - [ ] #10947
  - [ ] #10973
  - [ ] #10974

ndouglas:
This might involve some changes to `content-build` as well.

See https://dsva.slack.com/archives/CU1E4CX9U/p1662129173118709

ndouglas:
Copypasta from Slack:

>I'm playing with a concept for improving this process in va.gov-cms.  I did some work on GitHub Actions recently, and now I have another ticket that is "review best practices for GitHub Actions".  There are substantial security considerations for using GHAs, and TBH I don't know if anyone has reviewed this stuff prior to or while working on our GHAs.  (I know I haven't.)  So I'm playing with the idea of having context-sensitive comments posted to the PR thread based on the code touched.  Sort of like codeowners, but more "what should be checked based on what you touched."

>For instance, if you touch any GHA workflow, did you check that:
>- the actions are up-to-date?
>- the actions are pinned to a commit, instead of a tag or not pinned at all?
>- no security vulnerabilities have been reported against the version?
>- the action doesn't inject unsanitized user input into trusted contexts?

>Etc etc etc.  I'm still in the very early stages of thinking about this, but given the burden of maintaining and updating codeowners and seeing that the subject-matter-expert load is distributed fairly, etc, it might be better if we can document in specific areas what needs to be checked and how.

>EDIT: Of course, these instructions would only be posted if they're relevant to the specific files touched by the PR.

ndouglas:
This document is great: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions

As mentioned in the previous commit, I'd like to post specific information in a comment on the PR thread corresponding to the areas of the codebase that are touched.  I think I'll open a PR and test this out.

PR #10715, opened 2022-09-13
VACMS-10619: Refactor GitHub Actions.

### LINKED ISSUE #10619: Refactor GitHub Actions
## Description
These share a lot of steps, variables (e.g. PHP version), etc.  Plus they're presumably consuming a lot of resources doing so.  Let's refactor our GHAs so they DRY.

## Acceptance Criteria
- [ ] GitHub Actions are well-factored.

ndouglas:
## Description

Closes #10619.

This accomplished less than I would like; there're just some things that GHA should probably support but don't yet.

That said, I think this does some good things.

I'll probably tweak a few more things.

## Notes
After merge, I need to modify the branch protection rules to require:
- Check Fields
- Composer Validate
- CodeQL Analyze
- ESLint
- PHP_CodeSniffer
- PHPStan
- PHPUnit
- StyleLint

and not require:
- runner / check-cer-fields
- runner / check-revision-log-fields
- runner / composer-validate
- runner / eslint
- runner / phpcs
- runner / phpstan
- runner / phpunit
- runner / stylelint


.github/actions/post-checkout/action.yml
```diff
@@ -0,0 +1,33 @@
+name: "VA.gov CMS Post-Checkout"
```
ndouglas, line 1: This is a Composite Action that collects some common steps we do with GHA post-checkout.

Doing this unfortunately does not reduce the number of times these steps are run, but it does reduce the number of places we have to change PHP versions, etc.

.github/actions/post-checkout/action.yml
```diff
@@ -0,0 +1,33 @@
...
+        php-version: '7.4'
+        extensions: dom, curl, libxml, mbstring, zip, pdo, mysql, pdo_mysql, bcmath, gd, exif, iconv
+        tools: composer:v2
+
+    - name: Install PHP packages via Composer
+      run: composer install --no-progress --prefer-dist --optimize-autoloader
+      shell: bash
+
+    - name: Setup ReviewDog
+      uses: reviewdog/action-setup@v1
```
ndouglas, line 33: We use ReviewDog in two places already, and hopefully will use it in a few more.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,141 @@
+name: Continuous Integration
```
ndouglas, line 1: I moved the respective jobs from separate files to a single combined file.

My reasoning is this:

1.  Highlight (by removing clutter) the files that _don't_ concern CI workflows, e.g. Dependabot auto-merge, S3 backup retention, etc.
2.  Centralize the CI jobs so we can change all of the things at once when we need to.  For instance, some of our jobs used `checkout@v1`, some `checkout@v2`, and some `checkout@v3`.  I have some upcoming stuff planned for pinning GHA action versions at commits (see #10617) and it's simply nicer to be able to make these changes in one file.
3. There was a lot of redundant information; we'd have a workflow named `phpstan`, and in it we'd name it `PHPStan`, then we'd have a job called `phpstan`, and in that we'd again say it's named `PHPStan`, and then have a step called `Run PHPStan`.  Some of this redundancy is inescapable, but this change peels off at least a couple of layers.
Screen Shot 2022-09-13 at 10 45 31 AM

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,141 @@
...
+  # Check whether certain fields exist in the CMS configuration.
+  check-fields:
+    name: Check Fields
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Check Corresponding Entity Reference Fields
+        run: ./tests/scripts/check-cer.sh
+      - name: Check Revision Log fields
+        run: ./tests/scripts/check-revision-logs.sh
```
ndouglas, line 14: Some of these actions don't require use of the post-checkout action because they only operate on the codebase.  I should probably indicate these with a comment.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,141 @@
...
+
+  # Validate that the `composer.lock` hash is up-to-date.
+  composer-validate:
+    name: Composer Validate
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/post-checkout
```
ndouglas: Not sure this line is necessary; I'll remove it and see what happens.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,141 @@
...
+          DIRECTORIES: 'docroot/modules/custom docroot/themes/custom tests'
+
+  # Analyze the codebase for programming errors with PHPStan.
+  phpstan:
+    name: PHPStan
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/post-checkout
+      - name: Run PHPStan and ReviewDog
```
ndouglas, line 82: I also added ReviewDog back in to the PHPStan action.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     
       # Checkout repo to make package allow list available
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
ndouglas, line 24: Not strictly related to the PR, but noticed this was outdated.

.github/workflows/pull-request-labels.yml
```diff
@@ -12,7 +12,7 @@ jobs:
   pr-team-labels:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
```
ndouglas, line 15: Not strictly related to the PR, but outdated.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,137 @@
...
+  codeql-analyze:
+    name: CodeQL Analyze
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: github/codeql-action/init@v2
+    - uses: github/codeql-action/analyze@v2
+
+  # Check style of ES/JS files with ESLint and ReviewDog.
+  eslint:
```
ndouglas, line 34: I also changed the naming scheme.

Previously, some of these had readable names like "CodeQL" and some had... uh, less good names, like `runner / phpcs`.  `runner` does not indicate much to most developers; if anything, it probably just seems like useless information.  Of course we know a runner's involved, if the thing is running.

So my goal here was to give more attractive and meaningful names.  Unfortunately, this does mean that I'll need to update the branch protection rules again 😑

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,137 @@
...
+    name: PHPUnit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/post-checkout
+      - name: Run PHPUnit (Unit Tests only)
+        run: bin/phpunit --group unit --exclude-group disabled tests/phpunit
+
+  # Check styles of modules and themes with Stylelint.
+  stylelint:
```
ndouglas, line 116: There were two workflows, `stylelint-modules` and `stylelint-themes`, but both were called `stylelint` and thus both were responsible for the same status check.  Whichever workflow executed last would set the status check, so this might have been masking errors.

.github/workflows/continuous_integration.yml
ndouglas: Aside from being a bit of an eyesore, the highly separate structure could mask issues.  

As noted elsewhere, there were two workflows, `stylelint-modules` and `stylelint-themes`, that appeared distinct from a filesystem view, appeared distinct when looking at the top few lines of YAML, but were named identically in the only place that mattered -- the job's name -- and thus they overlapped and potentially useful information was discarded.

swirtSJW (approved):
As much as I can follow it, this seems great.  Thanks for all the explanation in comments.

PR #10732, opened 2022-09-15
VACMS-0000: Improve/fix/remove set-tugboat-tests-pending action

ndouglas:
Dependabot PRs are currently showing failing tasks because they have read-only access, and setting the tasks to pending requires read-write access.  

This PR completely rewrites the action to try to make everyone happy.


ndouglas:
PR failing because of a deadlock, not because of anything in the PR 😩

ndouglas:
Merging because it kinda has to be merged to test it 😩   Won't affect prod, etc of course.

PR #10734, opened 2022-09-15
Update tasks-periodic.yml

ndouglas:
Fixes a pesky URL that started redirecting (but not invisibly to `curl`).


swirtSJW (approved):
Awesome.  Thank you.

PR #10762, opened 2022-09-20
VACMS-10691: Add validator to prevent linking to media pages.

### LINKED ISSUE #10691: Add filter to prevent linking to media view pages.
## Description
If #10651 goes in, there is a chance that an editor might copy and paste the path to the media page.  Linkit will not prevent this (although it will not encourage it either).  Therefore, we should detect this and alert the editor to the error.

## Acceptance Criteria
- [x] Editors cannot link to media view pages within content.

ndouglas:
## Description

Closes #10691.

## Testing done
Added unit test.  I'll do a manual test and write steps once it's deployed.

## QA steps

1. Do this
   - [ ] Validate that
2. Then
   - [ ] Validate that
   - [ ] a
   - [ ] b
   - [ ] c

### DIFF
--- a/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/PreventMediaViewLinks.php
+++ b/docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/PreventMediaViewLinks.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Drupal\va_gov_backend\Plugin\Validation\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * Checks that the text does not contain any relative links to media pages.
+ *
+ * @Constraint(
+ *   id = "PreventMediaViewLinks",
+ *   label = @Translation("Prevent Media View Links", context = "Validation"),
+ *   type = { "text_long" }
+ * )
+ */
+class PreventMediaViewLinks extends Constraint {
+
+  /**
+   * The error message for rich text fields.
+   *
+   * @var string
+   * @see \Drupal\va_gov_backend\Plugin\Validation\ConstraintPreventMediaViewLinksValidator
+   */
+  public $richTextMessage = '":link" uses a URL ( :url ) that\'s only available on the VA network. Update the link to a valid public-facing page.';
+
+}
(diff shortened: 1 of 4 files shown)


docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/PreventMediaViewLinks.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+class PreventMediaViewLinks extends Constraint {
+
+  /**
+   * The error message for rich text fields.
+   *
+   * @var string
+   * @see \Drupal\va_gov_backend\Plugin\Validation\ConstraintPreventMediaViewLinksValidator
+   */
+  public $richTextMessage = '":link" uses a URL ( :url ) that\'s only available on the VA network. Update the link to a valid public-facing page.';
```
swirtSJW, line 24: I think this needs some UX love.  I don't think our editors will know what to do with this.

swirtSJW (commented):
I love everything about this, with the exception of the message to the editor.  I have a low degree of confidence that they will know what to do with this message as it is.

> ":link" uses a URL ( :url ) that\'s only available on the VA network. Update the link to a valid public-facing page.

I am not sure who we have left that can tackle this kind of wording to make it more editor friendly.  @EWashb  any ideas?

docroot/modules/custom/va_gov_backend/src/Plugin/Validation/Constraint/PreventMediaViewLinks.php
ndouglas: Yeah, that's actually directly from the one I copypasta'ed the code from.  I figured you'd have complaints 😁

EWashb:
@swirtSJW we still have @valerierunde until 9/23, but I'll be around to wordsmith after that as well. I'm sure Val can whip something up that makes life a little easier. If not, I'll get with you tomorrow @ndouglas to provide some user-friendly text.

EWashb:
@ndouglas is the text blocking you here? I can give you something for it if so

ndouglas:
@EWashb Yeah, this is waiting on the error message.  I feel like it could be "This link points to a media page.  Don't link directly to media pages", but I'm a virtual unknown in the user interface industry 🤷🏻‍♂️ .

EWashb:
"Internal media pages should not be linked. Link to public-facing pages only."

How's this @ndouglas?

ndouglas:
@EWashb We need to also include the visible text and the problematic URL/path so they can find the link in CKEditor, I think.

":link" ( :url ) is a link to an internal media page. Link to public-facing pages only."

Maybe?

EWashb:
That looks good to me @ndouglas

edmund-dunn (approved):
I was excoriated as expected by the error message.

Issue #9799, opened 2022-07-15
Make broken link report from content build available to the CMS


cweagans:
## Description

We should get the broken link data to the CMS to support effort to surface metrics to Datadog + surface a list of broken links to editors. Step 1 is getting the data into the CMS somehow (it could be as simple as writing the JSON file to disk somewhere).

## Acceptance Criteria
- [ ] Broken link report JSON file is retrieved when a content release completes.

## Tasks
- CMS
  - [x] Make new module or use an existing module (e.g. va_gov_api, va_gov_backend, va_gov_build_trigger)
  - [x] Create an event subscriber that fires on content release completing
  - [x] Store the contents of the file to state
  - [ ] Write test

### CMS Team
Please check the team(s) that will do this work.

- [ ] `Program`
- [x] `Platform CMS Team`
- [ ] `Sitewide Crew`
- [ ] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public Websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

cweagans:
@ElijahLynn if you end up taking this on, we're going for dead-simple implementation. My thought was a controller to POST the data to and then shove the entire object into the Drupal KV store. We don't need a separate db table or anything fancy like that.

ElijahLynn:
K, I'm working on this now. I am going to use the Drupal State API https://www.drupal.org/node/1787278 to accept a JSON payload at the route of `/content-release/broken-link-report` and store it in the `key_value` table (w/State API) under a key format of `content_release.broken_link_report.action_run_id`. That is the working plan though, open to naming changes.

timcosgrove:
Broken links json is available publicly: https://vetsgov-website-builds-s3-upload.s3-us-gov-west-1.amazonaws.com/broken-link-reports/vagovprod-broken-links.json

Updated AC and tasks to reflect that instead of POSTing from content release to CMS, we can just pull within the CMS upon content release completing.

PR #10783, opened 2022-09-21
VACMS-9799 Ingest broken link report from content-build

timcosgrove:
## Description

Relates to #9799 

## Testing done
Verified in Tugboat instance. The state value can be checked with `drush sget content_release.broken_links `. 

Have also verified that failure to ingest does not interfere with normal content release operations.

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [x] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`


docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksSubscriber.php
```diff
@@ -0,0 +1,109 @@
...
+    if ($this->settings->get('broken_link_report_import_enabled')) {
+      if ($event->getNewReleaseState() === ReleaseStateManager::STATE_READY) {
+        if (in_array($event->getOldReleaseState(), [
+          ReleaseStateManager::STATE_COMPLETE,
+          ReleaseStateManager::STATE_ERROR,
+        ])) {
+          try {
+            $report_location = $this->settings->get('broken_link_report_location');
+            if (preg_match('/^http/', $report_location)) {
+              $response = $this->httpClient->request('GET', $this->settings->get('broken_link_report_location'));
```
swirtSJW, line 87: I am guessing the http check for the location is for environments other than prod? and the else clause is for prod where it has direct access?

Might be good to clarify the two different clauses with comments to the cases are more apparent.

Edit: now that I see the settings specifications I understand the reason, however I should not have to look to a different file to be able to understand the purpose of the two clauses.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksSubscriber.php
```diff
@@ -0,0 +1,109 @@
...
+            else {
+              $file_path = $report_location;
+              $file_stream = fopen($file_path, 'r');
+              $contents = fread($file_stream, filesize($file_path));
+              fclose($file_stream);
+              $this->state->set('content_release.broken_links', $contents);
+            }
+          }
+          catch (RequestException $e) {
+            throw new \Exception('Unable to retrieve broken links report.');
```
swirtSJW, line 101: Do we want this to be more informative, like surfacing the status code returned or the location that failed to be found?

docroot/sites/default/settings/settings.tugboat.php
```diff
@@ -77,3 +77,7 @@
   // Disable sso form.
   $config['simplesamlphp_auth.settings']['activate'] = FALSE;
 }
+
+// Settings supporting broken link report import.
+$settings['broken_link_report_import_enabled'] = TRUE;
+$settings['broken_link_report_location'] = '/var/lib/tugboat/docroot/vendor/va-gov/content-build/logs/vagovdev-broken-links.json';
```
swirtSJW, line 83: Nice!

docroot/sites/default/settings.php
```diff
@@ -176,6 +176,13 @@
 $settings['slack_webhook_url'] = getenv('CMS_VAGOV_SLACK_WEBHOOK_URL') ?: FALSE;
 $config['slack.settings']['slack_webhook_url'] = $settings['slack_webhook_url'];
 
+// Settings supporting broken link report import. Off by default.
+$settings['broken_link_report_import_enabled'] = FALSE;
+// Default prod location.
+// @todo should probably be managed via env var.
```
swirtSJW, line 182: I think it makes sense to remove the todo.  I think in other environments this could be overridden by reading an env variable , but by default for other lower environments including sandbox,  it makes sense to be able to use the prod location since dev and staging FE are built from data that is NOT from the cms equivalents.
```suggestion
```

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksSubscriber.php
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  /**
+   * Retrieve the broken links report when a content release finishes.
+   *
+   * @param \Drupal\va_gov_build_trigger\Event\ReleaseStateTransitionEvent $event
+   *   The release state transition event.
+   */
+  public function retrieveBrokenLinks(ReleaseStateTransitionEvent $event) {
+    if ($this->settings->get('broken_link_report_import_enabled')) {
```
swirtSJW, line 78: I like that you are basing this on a setting.  It seems like this should be available to a UI  to be able to disable or enable.  In case the data goes wonky and we need to shut it down either to silence the errors or prevent additional broken link processing and notifications (though notifications should have its own shunt).

swirtSJW (commented):
This is looking really good.  A couple of questions and comments ...

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksSubscriber.php
```diff
@@ -0,0 +1,109 @@
...
+              if ($response->getStatusCode() === 200) {
+                $this->state->set('content_release.broken_links', $response->getBody()->getContents());
+              }
+            }
+            else {
+              $file_path = $report_location;
+              $file_stream = fopen($file_path, 'r');
+              $contents = fread($file_stream, filesize($file_path));
+              fclose($file_stream);
+              $this->state->set('content_release.broken_links', $contents);
```
ndouglas, line 97: I think it'd be worth splitting this out into a separate class that can be subclassed per environment.  That'd improve testability, too.  That doesn't necessarily have to happen now, but I'd appreciate a followup ticket if nothing else.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksSubscriber.php
ndouglas: Yeah, a kill switch in State would be nice.  We should probably encourage that for all functionality that differs on BRD, given the time and PITA of the full deployment pipeline and off-cycle deployments.

timcosgrove: We should have a feature flag functionality to use for something like this. In the absence of that, I can put it into State. I don't think a separate form is warranted for this.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksSubscriber.php
timcosgrove: I can add comments to this effect, but it is pretty much, if the file location begins with http, it is networked, so use an http client to retrieve it. If it does not, assume that it is local and try to retrieve it from the local file system.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksSubscriber.php
```diff
@@ -0,0 +1,109 @@
...
+        if (in_array($event->getOldReleaseState(), [
+          ReleaseStateManager::STATE_COMPLETE,
+          ReleaseStateManager::STATE_ERROR,
+        ])) {
+          try {
+            $report_location = $this->settings->get('broken_link_report_location');
+            if (preg_match('/^http/', $report_location)) {
+              $response = $this->httpClient->request('GET', $this->settings->get('broken_link_report_location'));
+              if ($response->getStatusCode() === 200) {
+                $this->state->set('content_release.broken_links', $response->getBody()->getContents());
```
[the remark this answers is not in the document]
timcosgrove: Ideally yes. This will probably need to be a separate ticket or else push the ticket to another sprint, since currently the "JSON" file for broken links has comments written to it, which means it's not a JSON file.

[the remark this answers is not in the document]
ndouglas: ("Nah." is a complete sentence.)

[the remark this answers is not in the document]
ndouglas: Ooh, okay.

docroot/sites/default/settings.php
timcosgrove: My intention is "This should be *able* to be managed by env var." I can clarify that.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksSubscriber.php
ndouglas: I actually suspect middleware is causing a persistent infuriating issue with test reliability right now, and I gotta figure out a sustainable approach.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksIngestion.php
```diff
@@ -0,0 +1,126 @@
...
+use Drupal\va_gov_build_trigger\Event\ReleaseStateTransitionEvent;
+use Drupal\va_gov_build_trigger\Service\ReleaseStateManager;
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Exception\RequestException;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Re-queues a content release when an error condition happens.
+ */
+class ContentReleaseBrokenLinksIngestion implements EventSubscriberInterface {
```
timcosgrove, line 17: Name changed.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksIngestion.php
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  /**
+   * Retrieve the broken links report when a content release finishes.
+   *
+   * @param \Drupal\va_gov_build_trigger\Event\ReleaseStateTransitionEvent $event
+   *   The release state transition event.
+   */
+  public function retrieveBrokenLinks(ReleaseStateTransitionEvent $event) {
+    if ($this->settings->get('broken_link_report_import_enabled')) {
```
timcosgrove, line 78: Have not handled this yet.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksIngestion.php
```diff
@@ -0,0 +1,126 @@
...
+      if ($event->getNewReleaseState() === ReleaseStateManager::STATE_READY) {
+        if (in_array($event->getOldReleaseState(), [
+          ReleaseStateManager::STATE_COMPLETE,
+          ReleaseStateManager::STATE_ERROR,
+        ])) {
+          $report_location = $this->settings->get('broken_link_report_location');
+          // If the report location starts with http, use an http client to
+          // retrieve it. Otherwise, assume it is a local file and attempt to
+          // read it.
```
timcosgrove, line 87: Changes:
* This is now two try/catch blocks, so that we can log different error messages.
* No longer `throw`ing in the catch block, since that still results in an uncaught exception.
* Using logger for all messaging, including errors.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksIngestion.php
```diff
@@ -0,0 +1,126 @@
...
+          if (preg_match('/^http/', $report_location)) {
+            try {
+              $response = $this->httpClient->request('GET', $this->settings->get('broken_link_report_location'));
+              if ($response->getStatusCode() === 200) {
+                $this->state->set('content_release.broken_links', $response->getBody()->getContents());
```
timcosgrove, line 92: Have not separated this out yet. Will be follow-up ticket, I think.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksIngestion.php
```diff
@@ -0,0 +1,126 @@
...
+          ReleaseStateManager::STATE_ERROR,
+        ])) {
+          $report_location = $this->settings->get('broken_link_report_location');
+          // If the report location starts with http, use an http client to
+          // retrieve it. Otherwise, assume it is a local file and attempt to
+          // read it.
+          if (preg_match('/^http/', $report_location)) {
+            try {
+              $response = $this->httpClient->request('GET', $this->settings->get('broken_link_report_location'));
+              if ($response->getStatusCode() === 200) {
```
ndouglas, line 91: I was trying to remember if Drupal's HTTPClient follows redirects transparently and how the status codes are propagated.  I'm not finding anything in the Google, so I suspect it's fine.  A recent issue caused by `curl` not automatically following redirects makes me a tad gunshy though.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksIngestion.php
ndouglas: Would an uncaught exception pose a problem?  It might be appropriate in this case -- Sentry should post a comment in #cms-notifications, I'd get tagged, and would know, oh, something's up with the thing.  And I don't mind that.  Better that than silent failure, but that might be better with #10813.

But that's a matter of taste and comfort.

If this code's split out, we can reconsider that then, too.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksIngestion.php
ndouglas: I guess if the end URL is `https` it's not really a likely scenario anyway.

ndouglas (commented):
Looks good to me.  Bring [the followup issues] out to us, that we may _know_ them.

docroot/modules/custom/va_gov_links/src/EventSubscriber/ContentReleaseBrokenLinksIngestion.php
timcosgrove: It's throwing an error in the CMS, so I would expect Sentry to report that. If it doesn't, we should probably amend that.

What changes is, the queue run will continue; the failure won't halt subsequent code. 

I agree we don't want it to fail silently.

ndouglas (approved):

docroot/sites/default/settings.php
```diff
@@ -176,6 +176,12 @@
 $settings['slack_webhook_url'] = getenv('CMS_VAGOV_SLACK_WEBHOOK_URL') ?: FALSE;
 $config['slack.settings']['slack_webhook_url'] = $settings['slack_webhook_url'];
 
+// Settings supporting broken link report import. Off by default.
+$settings['broken_link_report_import_enabled'] = FALSE;
+// Default prod location, overrideable by env var.
+$settings['broken_link_report_location'] = getenv('CONTENT_RELEASE_BROKEN_LINK_REPORT') ?: 'https://vetssgov-website-builds-s3-upload.s3-us-gov-west-1.amazonaws.com/broken-link-reports/vagovprod-broken-links.json';
+
+
```
swirtSJW, line 184:
```suggestion
```

swirtSJW (approved):
Thanks for your work on this Tim.

Tested in tugboat and approved.
image


Unfortunately this can not be tested on BRD environments, so will be unable to test before it hits prod, but the risk is quite low.

PR #10955, opened 2022-09-28
VACMS-10947:  Contextual advice action

### LINKED ISSUE #10947: Investigate a contextual advice action to create context-specific comments on PRs.
## Description
This is a followup from #10617.

Whenever we edit certain files, it might be good to have some advice provided to reviewers (and the original submitter) about what things to check on the PR.

This is an investigatory ticket to see if such a thing can be implemented quickly and appealingly.

## Acceptance Criteria
- [x] I've created a GitHub Action that yammers on your PR based on which files you've edited.
- [ ] It has been reviewed and adopted or never spoken of again.

ndouglas:
## Description

Closes #10947.

Basically, a GHA diffs the PR HEAD and the base branch (not necessarily main), and looks for Markdown files in a specific directory.  It chooses the most specific one for that file.  All of those markdown files are then added as a PR comment (replacing any prior similar comments).

An example can be seen here: https://github.com/department-of-veterans-affairs/va.gov-cms-test/pull/1087

The resulting comment is way more verbose than I'd like.  Perhaps the final version would be a simple checklist of 4-5 items max, each being a brief sentence, along with links to deeper subject-specific documentation that lives elsewhere in the repository.  But I was thinking about both while writing these...

And I also don't think this idea is necessarily applicable throughout the codebase.  Maybe it is, maybe it ain't.  But for certain things like GitHub Actions, where we can pwn ourselves in unexpected ways, and where the guidelines and best practices will remain more-or-less static over time.

Mostly posting this as an RfC, get some feedback, gather thoughts.  Flame suit on, etc.


swirtSJW:
This is interesting... but I only pretend to understand what is going on here other than the documentation.  What am I looking to see happen... or not happen.

ndouglas:
As mentioned, the [PR comment](https://github.com/department-of-veterans-affairs/va.gov-cms-test/pull/1087#issuecomment-1261406746) is too verbose, but the basic idea is just to post a comment of "code review advice" on any PR touching certain files.

If I touch GitHub workflow files, then I get a comment saying something like:

>- [ ] Workflows do not directly inject user input
>- [ ] All third-party actions are pinned to commits, not hashes
>- [ ] ...

The idea is to think of the general QA steps that need to be performed for a section of code once, then have those automatically injected so that reviewers have some additional ideas of what to check.

swirtSJW (approved):

PR #10361, opened 2022-08-22
VACMS-9357 Shows facility hours in view display

omahane:
## Description

Relates to #9357
Only shows the hours on the view display.

## Testing done
Manually

## QA steps

What needs to be checked to prove this works?  
What needs to be checked to prove it didn't break any related things?  
What variations of circumstances (users, actions, values) need to be checked?  

As user _uid_ with _user_role_
1. Do this
   - [ ] Validate that
2. Then
   - [ ] Validate that
3. Then validate Acceptance Criteria from issue
   - [ ] a
   - [ ] b
   - [ ] c

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [ ] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

### Is this PR blocked by another PR?

- [ ] `DO NOT MERGE`

### Does this PR need review from a Product Owner

- [ ] `Needs PO review`

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change?
- [ ] Yes, and it's written in issue ____ and queued for publication.
  - [ ] Merge and ping the UX writer so they are ready to publish after deployment
- [ ] Yes, but it hasn't yet been written
  - [ ] Don't merge yet -- ping the UX writer to write and queue content
- [ ] No announcement is needed for this code change.
  - [ ] Merge & carry on unburdened by announcements

PR #11041, opened 2022-10-03
VACMS-0000: Fix function call causing WSOD editing CLPs on Prod

ndouglas:
See [this Slack thread](https://dsva.slack.com/archives/CJT90C0UT/p1664822940503299).

**TL;DR**: This function was renamed [here](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/10361/files#diff-567d531330304beb9522b26b7f7b660b4e481f6dcb381e1280b13b21786b1f50L305) but the change didn't make it to the function call [here](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/8f172aca74a2812a0a5a3b880858d22c454e3666/docroot/modules/custom/va_gov_backend/va_gov_backend.module#L276).


swirtSJW:
Closing this as a duplicate of https://github.com/department-of-veterans-affairs/va.gov-cms/pull/11046   The fix is exactly the same, but the other has more info for testing and the issue.

PR #11054, opened 2022-10-04
VACMS-0000: Remove distracting test output.

ndouglas:
## Description

This bit of spam can be misinterpreted as an error message and distract from actual error messages.

This should be safe to merge if tests pass.


edmund-dunn (approved):
ok

PR #11199, opened 2022-10-12
VACMS-10974: Pins GitHub Actions at commit SHAs instead of tags/branches.

### LINKED ISSUE #10974: Audit our use of GitHub Actions to ensure it's in line with best practices.
## Description
Follow up to #10617.

Actually ensuring conformance should be straightforward; just pin our commits, etc, in line with findings:

- Pin all GitHub Actions at a specific commit.  A tag is not sufficient, as it does not protect us against the third party losing control over their account.  A commit SHA should protect us against any anticipated injection.

- Review the source code of the action at the commit you are pinning.  This should be performed again whenever the action is updated, for whatever reason.

- Be careful with user-provided input, especially in scripts.  For example, do not read PR titles, comment bodies, or even the names of changed files in such a way that they might be executed as commands.

## Acceptance Criteria
- [ ] GitHub Actions have been audited and altered here applicable to conform to best practices.

ndouglas:
## Description

Closes #10974.  I did this fairly carefully, but possible I made a mistake copypastaing a SHA somewhere.  If so I might need to hotfix `main` after this is merged if I messed up one of the `pull_request_target` action SHAs.

### DIFF
--- a/.github/actions/post-checkout/action.yml
+++ b/.github/actions/post-checkout/action.yml
@@ -11,7 +11,8 @@ runs:
       shell: bash
 
     - name: Use the Composer cache, if possible
-      uses: actions/cache@v3
+      # https://github.com/actions/cache/releases/tag/v3.0.10
+      uses: actions/cache@56461b9eb0f8438fd15c7a9968e3c9ebb18ceff1
       with:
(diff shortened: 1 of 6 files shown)


edmund-dunn (approved):
Looks good!

Issue #11216, opened 2022-10-13
Some Operating Status pages are outputting Linky URLs


timcosgrove:
## Description
A rich text field on Operating Status pages is using the 'value' of its field rather than the 'processed' value.

## Example
https://www.va.gov/north-florida-health-care/operating-status/
* Health Chat App: https://www.va.gov/admin/content/linky/58

## Mitigation
Fix the GraphQL query

swirtSJW:
I believe this is closed as a result of the merged PRs ^^.  Closing.

PR #11220, opened 2022-10-14
DNM: Quick pass at helping with carriage return issues.

timcosgrove:
## Description
#11216 

Not to be merged yet; PR is for collab and discussion.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -374,7 +374,8 @@
                 "Internal Claro tweaks": "patches/claro-css-tweaks.patch",
                 "3099026 - Claro's preprocessing of field multiple value form's table header cell removes potential changes by others": "https://www.drupal.org/files/issues/2020-02-20/3099026-9.patch",
                 "2767243 - Create a theme suggestion for taxonomy terms by view mode": "https://www.drupal.org/files/issues/core-theme-suggestion-for-taxonomy-view-modes-2767243-14.patch",
-                "2775665 - MenuLinkContent updateLink function generates a PHP Warning for override-able keys that are not present in the loaded entity": "https://www.drupal.org/files/issues/2021-09-17/updateLink-2775665-14.patch"
+                "2775665 - MenuLinkContent updateLink function generates a PHP Warning for override-able keys that are not present in the loaded entity": "https://www.drupal.org/files/issues/2021-09-17/updateLink-2775665-14.patch",
+                "2975602 - Html::serialize() escapes backslash-r to &#13;": "https://www.drupal.org/files/issues/2018-05-27/2975602-2-html_serialize.patch"
             },
             "drupal/danse_content_moderation": {
                 "3309657 - Danse 2.2.7 compatibility": "https://www.drupal.org/files/issues/2022-09-14/3309657-danse-2-2-7-compatibility.patch"
--- a/docroot/modules/custom/va_gov_links/src/CorrectCarriageReturn.php
+++ b/docroot/modules/custom/va_gov_links/src/CorrectCarriageReturn.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\va_gov_links;
+
+use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
+use Drupal\Core\Entity\EntityInterface;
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -374,7 +374,8 @@
                 "Internal Claro tweaks": "patches/claro-css-tweaks.patch",
                 "3099026 - Claro's preprocessing of field multiple value form's table header cell removes potential changes by others": "https://www.drupal.org/files/issues/2020-02-20/3099026-9.patch",
                 "2767243 - Create a theme suggestion for taxonomy terms by view mode": "https://www.drupal.org/files/issues/core-theme-suggestion-for-taxonomy-view-modes-2767243-14.patch",
-                "2775665 - MenuLinkContent updateLink function generates a PHP Warning for override-able keys that are not present in the loaded entity": "https://www.drupal.org/files/issues/2021-09-17/updateLink-2775665-14.patch"
+                "2775665 - MenuLinkContent updateLink function generates a PHP Warning for override-able keys that are not present in the loaded entity": "https://www.drupal.org/files/issues/2021-09-17/updateLink-2775665-14.patch",
+                "2975602 - Html::serialize() escapes backslash-r to &#13;": "https://www.drupal.org/files/issues/2018-05-27/2975602-2-html_serialize.patch"
```
timcosgrove, line 378: https://www.drupal.org/project/drupal/issues/2975602

docroot/modules/custom/va_gov_links/src/CorrectCarriageReturn.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
```
timcosgrove, line 2: This is really basic and not intended for long life. All it does is find `&#13;` in the fields that are found and substitutes `\r`.

It applies to the same fields as the LinkyReplacerEntityOperations preSave does.

scripts/content/VACMS-11220-correct-carriage-returns.php
```diff
@@ -0,0 +1,133 @@
...
+      $node->setNewRevision(TRUE);
+      $node->setRevisionUserId(1317);
+      $node->setChangedTime($time);
+      $node->setRevisionCreationTime($time);
+      $node->setRevisionLogMessage('Saved to fix carriage returns in wysiwyg.');
+      $node->setSyncing(TRUE);
+      $node->save();
+    }
+    else {
+      log_message("https://prod.cms.va.gov/node/{$nid} may have a forward draft; latest revision {$latest_revision} is different than {$node_storage->getLatestRevisionId($nid)}. Skipping.");
```
swirtSJW, line 68: Do we have to skip.  We have the latest revision, we have the latest revision, can't we just save that too?  As long as it is saved after the default revision is saved, it should maintain the revision order. (it is a pattern we have done before)

scripts/content/VACMS-11220-correct-carriage-returns.php
```diff
@@ -0,0 +1,133 @@
...
+ */
+function process_chunk(int $chunk_id, array $chunk) {
+  $entity_type_manager = \Drupal::entityTypeManager();
+  $node_storage = $entity_type_manager->getStorage('node');
+  /** @var \Drupal\node\NodeInterface[] $nodes */
+  $nodes = $node_storage->loadMultiple($chunk);
+  $count = count($nodes);
+  log_message("Loaded {$count} nodes as chunk {$chunk_id}");
+  foreach ($nodes as $nid => $node) {
+    // Do not do this processing if there is a forward draft.
```
swirtSJW, line 53: We have some nice helper functions in here  that keep us from repeating this kind of save code.

/scripts/content/script-library.php

Things like take out some of the risk/variation.
get_node_at_latest_revision()
save_node_revision()

scripts/content/VACMS-11220-correct-carriage-returns.php
swirtSJW: Here is a pattern that updates both the default and the forward revision https://github.com/department-of-veterans-affairs/va.gov-cms/blob/f2ef3f0a73431917f8ea113d8a954c6b29e4c2ef/scripts/content/VACMS-9714-vamc-detail-page-crisis-to-988.php#L81-L88

timcosgrove:
Uneeded.

PR #11224, opened 2022-10-14
VACMS-11216: patch HTML utilities to properly handle carriage returns.

timcosgrove:
## Description

#11216

See https://drupal.org/node/2975602


ryguyk (approved):
Looks good conceptually, as I have context of what we're trying to fix (i.e. `/r` vs. `&#13;`), though I have not tested this personally.

PR #11398, opened 2022-10-31
VACMS-11280: Patch content-build to add timestamps in content-build-gql test.

### LINKED ISSUE #11280: Patch `content-build` as part of `content-build-gql` test to add timestamp logging.
## Description
Followup to #10391.  To assist in work on #11279 by identifying which (if any) tests might be causing issues with other tests running concurrently.

The `content-build-gql` test should patch `content-build` to log details to a file so that we can debug failures and other issues after the fact.

## Acceptance Criteria
- [x] `content-build-gql` test patches `content-build` as described above.
- [x] No other operations are adversely affected on any environment.

ndouglas:
Closes #11280.

This should fail-safe.  For instance, if content-build is updated and the patch starts to fail, the content-build-gql test should still pass.  I'll have to fix the patch at some point, but overall aggravation should be minimized.

This should only affect the `content-build-gql` test and not normal content builds, natch.


swirtSJW (approved):
Clever solution.

swirtSJW:
image

PR #11409, opened 2022-11-01
VACMS-9923: fix display for Lovell sub-sections in user section block

### LINKED ISSUE #9923: CMS: Lovell sub-sections show in user section block
## Describe the defect
After logging in, a user is taken taken to their user page which displays a user-section-block.  Since the subsections under Lovell were added
image

The block displays the subsections separately
image

## To Reproduce
1. Login as admin (admins have access to all sections)
2. See the user section block.  Notice two entries for Lovell.

## AC / Expected behavior
Lovell and its subsections should only show under VHA not as their own sections unless an editor is only assined to those explicitly.

 ## Additional context
This is likley only affecting true admins and content admins who are assigned to VHA at the upper level, not on a VAMC system level.

JayDarnell:
## Description

Closes #9923 

## Testing done

Manual testing

## Screenshots

Logged in as an admin before:
Screen Shot 2022-11-01 at 2 49 40 PM

Logged in as an admin after the fix:
Screen Shot 2022-11-01 at 2 39 27 PM

Expanded:
Screen Shot 2022-11-01 at 2 40 01 PM

Logged in as a Lovell editor after the fix:
Screen Shot 2022-11-01 at 2 40 10 PM

## QA steps

As an admin user
- [ ] Log into the site and verify the user sections list doesn't display either Lovell subsection at the top level
- [ ] Expand the VHA item, scroll down to find Lovell and verify all 3 sections are properly displayed

As a Lovell editor (example: Katie.Yearley@va.gov
- [ ] Log into the site and verify the Lovell subsections are properly displayed under Lovell Federal health care

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [ ] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [x] `⭐️ Facilities`
- [ ] `⭐️ User support`

### DIFF
--- a/docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php
+++ b/docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php
@@ -146,6 +146,15 @@ class UserSections extends BlockBase implements ContainerFactoryPluginInterface 
           ];
           break;
 
+        case 7:
+          $links[$parent_path[6]][$parent_path[5]][$parent_path[4]][$parent_path[3]][$parent_path[2]][$parent_path[1]][$parent_path[0]]['items'][$term->tid] = [
+            '#type' => 'link',
+            '#weight' => $term->weight,
+            '#title' => $term->name,
+            '#url' => Url::fromRoute('entity.taxonomy_term.canonical', ['taxonomy_term' => $term->tid]),
+          ];
+          break;
+
         default:
           $section_link = Link::fromTextAndUrl($term->name, Url::fromRoute('entity.taxonomy_term.canonical', ['taxonomy_term' => $term->tid]))->toString();
           $expand_button = $sections[$term->tid]['hasChildren'] ? Markup::create('<button class="toggle" aria-label="Toggle ' . $term->name . ' section" aria-pressed="false" aria-expanded="false" aria-controls="section-' . $term->tid . '"></button>') : NULL;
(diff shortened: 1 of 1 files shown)


docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php
```diff
@@ -146,6 +146,15 @@ public function build() {
           ];
           break;
 
+        case 7:
```
swirtSJW, line 149: This looks good.  Please update the docblock to speak the new truth
https://github.com/department-of-veterans-affairs/va.gov-cms/blob/66841c3d1f332903d5661337238e0f32535bc4b9/docroot/modules/custom/va_gov_user/src/Plugin/Block/UserSections.php#L114

swirtSJW: Also if you understand the reason why the cases only account for odd numbers, that would be good to explain... because I don't get that part.

JayDarnell: Ah yes, I can add notes about the odd numbers to the doc block. I found that particularly confusing when I first looked at the code. It boils down to the fact we build the tree in such a way that each valid link has a sub array called items which contains the child links.

swirtSJW (approved):
Nice work Jay.  Test out good.
image

Lovell shows like it should
image

And there are no notices.
image

PR #11436, opened 2022-11-03
VACMS-0000: Patch to stop Raven from exploding...

ndouglas:
... and killing everyone inside.

## Description

Base preview builds are currently failing with the following:

```
TypeError: Argument 1 passed to Sentry\Dsn::createFromString() must be of the type string, null given, called in /var/lib/tugboat/docroot/modules/contrib/raven/src/Config/SecKitOverrides.php on line 49 in Sentry\Dsn::createFromString() (line 77 of /var/lib/tugboat/docroot/vendor/sentry/sentry/src/Dsn.php).
>  [warning] Drush command terminated abnormally.

In SiteProcess.php line 214:
                                                                               
  The command "/usr/local/bin/drush config:import --uri=default --root=/var/l  
  ib/tugboat/docroot" failed.                                                  
                                                                               
  Exit Code: 1(General error)                                                  
                                                                               
  Working directory:                                                           
                                                                               
  Output:                                                                      
  ================                                                             
  TypeError: Argument 1 passed to Sentry\Dsn::createFromString() must be of t  
  he type string, null given, called in /var/lib/tugboat/docroot/modules/cont  
  rib/raven/src/Config/SecKitOverrides.php on line 49 in Sentry\Dsn::createFr  
  omString() (line 77 of /var/lib/tugboat/docroot/vendor/sentry/sentry/src/Ds  
  n.php).                                                                      
...

2022-11-03T10:26:21.925Z - Command Failed (Tugboat Error 1064): Exit code (1)
```

This is because Sentry uses the following bit of code:

Screen Shot 2022-11-03 at 7 43 26 AM

and Raven uses:

Screen Shot 2022-11-03 at 7 44 32 AM

Raven is not configured on Tugboat, either through environment variables or config, and in fact would be disabled immediately by a config import, so the expression evaluates to NULL.

This leads to Sentry throwing a TypeError, which is not the same as an InvalidArgumentException so it blows up.

This patch catches TypeError, so it permits the rest of the operations to proceed normally.  It shouldn't have any negative effects.

I'm not sure why this is happening now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -463,6 +463,9 @@
             "drupal/prometheus_exporter": {
                 "3214628 - Add published/unpublished node statistics to NodeCount.": "https://www.drupal.org/files/issues/2021-06-29/prometheus_exporter-add_status_metrics_to_node_count-3214628-3.patch"
             },
+            "drupal/raven": {
+                "Sentry/Dsn throws a TypeError, blowing up everything.": "patches/raven-catch-typeerror.patch"
+            },
             "drupal/schemata": {
                 "3198409 - Plugin definitions are not cached": "https://www.drupal.org/files/issues/2021-02-14/3198409_2.patch",
                 "3190131 - Possibly unnecessary logging in SchemaFactory::create().": "https://www.drupal.org/files/issues/2020-12-28/schemata-remove-logging-statement-3190131-2.patch"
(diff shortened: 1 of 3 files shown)


ndouglas:
Screen Shot 2022-11-03 at 7 54 50 AM

😩

ndouglas:
I think I might be the only person online right now, so I'm going to go ahead and merge this so that things can get back to normal.  Not expecting any ill effects since I tested this approach on Tugboat already.

olivereri:
Not all eastern time-zone heroes wear capes. If you do though, that's cool, but a bit weird.

PR #11441, opened 2022-11-03
VACMS-10870: php testing

edmund-dunn:
## Description

Relates to . (or closes?)

## QA steps

1. Do this
   - [ ] Validate that
2. Then
   - [ ] Validate that
   - [ ] a
   - [ ] b
   - [ ] c

### Definition of Done

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [ ] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/.tugboat/config.yml
+++ b/.tugboat/config.yml
@@ -175,6 +175,7 @@ services:
         - ./bin/npm install
         - composer va:theme:compile
         - composer va:web:install
+        - drush cr
         - drush cim -y
         - drush deploy
 
--- a/.tugboat/config.yml
+++ b/.tugboat/config.yml
@@ -175,8 +175,6 @@ services:
         - ./bin/npm install
         - composer va:theme:compile
         - composer va:web:install
-        - drush cr
-        - drush cim -y
         - drush deploy
 
         # Setup background processing service.  This uses runit to keep process up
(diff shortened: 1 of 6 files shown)


.tugboat/config.yml
```diff
@@ -175,6 +175,7 @@ services:
         - ./bin/npm install
         - composer va:theme:compile
         - composer va:web:install
+        - drush cr
         - drush cim -y
```
swirtSJW, line 179: Why are these being added?  Doesn''t drush deploy already clear cache and do config import?  Does new drush change this pattern?  Doing a config import before updb is going to cause problems as that is not the pattern everyone assumes from drush deploy,
image

edmund-dunn: I had to add the `drush cim -y` to get my table audit to deploy because the `updatedb` was failing because the `views_event_dispatcher`module wasn't installed yet. I then needed to add the `cr` to `cim` to run.

swirtSJW: That failure on tugboat would also be a failure on staging and prod.  
Those kind of things have to be an done as two separate merges a day apart or handle it in a hook_update_n.   We can't just alter the deploy process on one environment to circumvent a problem that is actually in the code.

swirtSJW:
Please remember to follow our QA process of waiting for tests to pass and getting an approval from an appropriate reviewer.
image

Issue #11493, opened 2022-11-07
Create and save a node of every content type in Cypress/behavioral tests.


ndouglas:
## Description
I thought I'd created this before, but apparently did not 😞 

Our behavioral tests should create and save a node (and verify it was saved correctly) of every content type.

This is an inexpensive way of verifying a lot of the basic (read: embarrassing to break) functionality of the CMS.  And timely, given the upgrades to PHP 8.1 and Drupal 10 we have ahead of us.

## Acceptance Criteria
- [x] Followup issues have been opened to create and save a node of every content type:
  - [x] #12456
  - [x] #12457
  - [x] #12458
  - [x] #12459
  - [x] #12460
  - [x] #12461
  - [x] #12462
  - [x] #12463
  - [x] #12464
  - [x] #12465
  - [x] #12466
  - [x] #12467
  - [x] #12468
  - [x] #12469
  - [x] #12470
  - [x] #12471
  - [x] #12472
  - [x] #12473
  - [x] #12474
  - [x] #12475
  - [x] #12476
  - [x] #12477
  - [x] #12478
  - [x] #12479
  - [x] #12480
  - [x] #12481
  - [x] #12482
  - [x] #12483
  - [x] #12484
  - [x] #12485
  - [x] #12486
  - [x] #12487
  - [x] #12488
  - [x] #12489
  - [x] #12490
  - [x] #12491
  - [x] #12492
  - [x] #12493
  - [x] #12494
  - [x] #12495
  - [x] #12496
  - [x] #12497
  - [x] #12498
  - [x] #12499
- [x] ~The relevant code is created in the right place.~ (see #9277)
- [x] ~A test or documentation is created advising engineers to extend these tests when they add a new content type.~ (see #12611)
- [x] ~Accessibility test is run against node creation form.~ (see #12610)
- [x] ~Accessibility test is run against node view page.~ (see #12610)

ndouglas:
@edmund-dunn I couldn't find the issue, so I created a new one.  (I think I had this in mind with #9277, but maybe I didn't verbalize it.)  Anything to add?

ndouglas:
Changed estimate for this issue to 3 after revising the scope.

ndouglas:
Closed in favor of various followups.

PR #11486, opened 2022-11-07
VACMS-8671: Removes problematic behavioral test.

ndouglas:
## Description

Closes #8671.

I added a Cypress test for basic functionality and also a PHPUnit unit test to hit some of the combinations.

**EDIT**: And then factored out some logic from the block class into a separate class, and added another unit test class focusing on it.  AND I'LL DO IT AGAIN

Unfortunately, this block is a complicated class as far as mocking is concerned; it has multiple responsibilities, uses calls that conceal heavy reliance upon the Drupal service container, and the critical function we want to test is ~60 lines long.

As this is a PR that replaces a test, the impact should be minimal.  However, I did some _slight_ rewriting to support dependency injection rather than concealed `Drupal::<service>` calls, so the EWA block _does_ need some manual testing.

## QA steps
The QA steps have to wait for the frontend build to complete.  ~And right now, frontend builds aren't completing because `content-build` deleted a file 😩~ that issue is fixed, but now builds are slow/unreliable for... unknown reasons 🤔 

[Convenient link to build log](https://cms-xxmehkfeohdgqiwhxpwbeginhkgdld2s.ci.cms.va.gov/sites/default/files/build.txt)

omg it's finished building go go go

   - [x] Validate that a page with a live frontend presence has the appropriate EWA block and that the URL is linked appropriately.
   e.g. https://cms-xxmehkfeohdgqiwhxpwbeginhkgdld2s.ci.cms.va.gov/new-haven-vet-center 
Screen Shot 2022-11-07 at 2 40 04 PM

   - [x] Validate that a page without a live frontend presence has the appropriate EWA block and that the URL is not linked at all.
   e.g. https://cms-xxmehkfeohdgqiwhxpwbeginhkgdld2s.ci.cms.va.gov/tacoma-vet-center/lacey-outstation
Screen Shot 2022-11-07 at 2 42 34 PM

  - [x] Validate that a page that is not supposed to have a live frontend presence has the appropriate EWA block and that there is no URL displayed. 
  e.g. https://cms-xxmehkfeohdgqiwhxpwbeginhkgdld2s.ci.cms.va.gov/central-ohio-health-care/staff-profiles/amy-lovell
Screen Shot 2022-11-07 at 2 47 04 PM

  - [x] Validate that a page with a differing number of parents (3) shows the correct behavior
  e.g. https://cms-xxmehkfeohdgqiwhxpwbeginhkgdld2s.ci.cms.va.gov/outreach-and-events/events/51074
[...]

Issue #8671, opened 2022-04-07
Research and implement alternative approach for the mock HTTP client in tests.


ndouglas:
## Description
We currently use the `va_gov_backend_http_client` module to perform certain behavioral tests outside of production (see `@mock_va_gov_urls`, specifically sections dealing with the behavior of "VA.gov URL", "(pending)", etc in the EWA block).  This is loaded and unloaded in `tests/behat/Drupal/ContextBase.php`.

This is unfortunate because it involves installing and uninstalling a module at runtime during tests, which is slow and costly and rather inelegant.  It might break other tests operating at the same time, or return nonsensical results for other people using the installation at the same time.

Beyond that, though, these are very costly tests.  I can't easily determine just how costly because of #8670, but it creates and saves a node repeatedly (five times!) to check whether the URL is displayed correctly in each case.

The EWA URL display logic should be factored out into a more testable form and its logic verified with HTTP mocks.  I don't think this is a valid problem for a behavioral test.

## Acceptance Criteria
- [x] Tests don't enable and disable this module in order to run.
- [x] The module has been removed from the codebase.
- [x] Tests pass reliably.

### CMS Team
Please check the team(s) that will do this work.

- [ ] `Sitewide program`
- [ ] `Platform CMS Team`
- [ ] `Sitewide crew ` (leave Sitewide unchecked and check the specific team instead)
  - [ ] `⭐️ Content ops`
  - [ ] `⭐️ CMS experience`
  - [ ] `⭐️ Public websites`
  - [ ] `⭐️ Facilities`
  - [ ] `⭐️ User support`

PR #11527, opened 2022-11-09
VACMS-9905: Remove PHPUnit performance tests.

### LINKED ISSUE #9905: Isolate and rework PHPUnit performance tests.

ndouglas:
## Description

Closes #9905.

No offense, but these just aren't good tests and I don't believe they're redeemable.

I'll go more into detail on the files page.

My original intent, as reflected in #9905, was to run these tests in a separate suite, sequentially with the other suites rather than concurrently.  But honestly I think they're just bad tests and not terribly meaningful even if I add a `for` loop around them.

If we want performance tests, we should create a story for identifying areas that we want to measure and track over time, we should determine useful ways of measuring them, and then we should write tests that measure them and report the results up to DataDog (or wherever) for tracking.

Currently, these tests are worse than not having the tests at all.  They do not reliably indicate problems, but they semi-reliably cause them.


tests/phpunit/Performance/CreateNodeTest.php
```diff
@@ -1,61 +0,0 @@
-<?php
-
-namespace tests\phpunit\Performance;
-
-use Tests\Support\Classes\VaGovExistingSiteBase;
```
ndouglas, line 5: This test does not do anything useful that isn't done far better and with far greater sensitivity to context in the Cypress tests.  This is not useful.  All it does is cause spurious, random test failures that aggravate engineers and require them to rerun the test until the recalcitrant thing finally passes.

swirtSJW (commented):
I am not opposed to removing these and I think your assertions are right about some of the cypress tests giving better coverage.   What do you think about re-purposing these from "performance tests"  to performance **monitoring**.  Where they do the same actions, but just log the time to datadog or something like that.... and only for staging.  So that we can spot trends like  if it takes 45 seconds to login.... or a minute to edit a page.

ndouglas:
@swirtSJW I like the idea of being able to monitor performance trends, but I think it's better done within Cypress than PHPUnit.

My thinking is that we're concerned about two different performance scenarios: editor-facing and content-build-facing.  

I think Tim's adding some good stuff for measuring the content-build metric, but it'd probably be wise to add a content-build-gql elapsed time metric since we're already performing the test, if only so that we can exclude a class of performance issues if we're ever in a blind panic trying to figure out why content builds are taking three hours.  That's probably not that tough.

The editor-facing performance monitoring is best done with Cypress IMHO, but we run into the issue of interference from concurrent tests.  I'm _hoping_ that with #8671 being resolved by #11486 that we'll lose the long tail we've seen in these performance tests on some runs and the numbers will settle down into something approaching a normal distribution.  Admittedly, knowing that loading an edit page _may_ take 30 seconds in some pathological situations is still of some value 🤷🏻‍♂️ 

Anyway, the Cypress tests do involve a whole lot of logging in, loading various node forms, saving nodes, etc, so I think it'd be useful to inject code before and after to measure some of those figures and then post 'em to DataDog.

So, followups:

- #11629 - these accessibility tests don't count login time, just the time to open various pages.  Since each has a unique path, it's trivial to do some text transformation on that and turn it into a metric (e.g. `/node/add/page` becomes `cms_bench._node_add_page` or something).
- #11630 - login time can be measured multiple times over the Cypress test runs.  Because of how long the Cypress tests are, they run concurrently with other tests and also by themselves, so it gives a decent range of conditions.  There might be other steps we can measure -- I'm not thinking of any ATM.
- #11493 - I wanted to ensure behavioral test coverage of creating and saving a node of each content type.  We can't quite get "statistically significant" coverage of this, but the measurements should still be useful, I'd think.
- #11631 - Similar to #11493, but for Media entities.

That's not exactly 1:1 for these performance tests, but I think it'll give better coverage of what matters.  I'm happy to add more issues or extend the ones above if you can think of anything else applicable.  And I think these should be fairly easy to extend in the future.

swirtSJW (approved):
Thanks for tackling this and spinning up all the new tickets.

Issue #11011, opened 2022-09-30
Disable and remove `vagovadmin` theme.


ndouglas:
## Description
Right?  I don't think we use it.  Eh, I'll open a PR and find out.

## Acceptance Criteria
- [ ] The theme has been removed and we no longer have to update two copies of everything.

timcosgrove:
It's an active theme. What's up here?

ndouglas:
Where did you see that this is an active theme?

ndouglas:
Not sure what "active" means, actually.  Locally it appears to be installed and enabled, but not default, which is AFAICT what we want right now.

My understanding is that `vagovclaro` was supposed to replace `vagovadmin`, so I opened this issue when I noticed that it was still hanging around.  (Before I saw #8950)  I also opened #11529 to remove it, and learned some stuff along the way, like that not _everything_ is completely shifted from `vagovadmin` to `vagovclaro`.  I think.  So I think I might be closing #11529 and instead opening a series of smaller PRs to do cleanup and verifying each of those changes individually.

PR #11529, opened 2022-11-09
VACMS-11011: Remove Vagovadmin theme.

ndouglas:
## Description

Relates to #11011.  This is going to have a lengthy QA process, most likely.


docroot/themes/custom/vagovadmin/yarn.lock
```diff
@@ -2874,7 +2874,7 @@ isarray@1.0.0, isarray@~1.0.0:
 isarray@2.0.1:
   version "2.0.1"
   resolved "https://registry.yarnpkg.com/isarray/-/isarray-2.0.1.tgz#a37d94ed9cda2d59865c9f76fe596ee1f338741e"
-  integrity sha512-c2cu3UxbI+b6kR3fy0nRnAhodsvR9dx7U5+znCOzdj6IfP3upFURTr0Xl5BlQZNKZjEtxrmVyfSdeE3O57smoQ==
```
swirtSJW, line 2877: I think I am a little confused as to why this file it being altered and not just removed.

swirtSJW (commented):
I was kind of expecting to see all of themes/custom/vagovadmin get removed in this PR.   Does this have to be done in two steps?
Other than that the code changes make sense and look very thorough.

Have not poked around to make sure the site is looking right yet.

ndouglas:
>I was kind of expecting to see all of themes/custom/vagovadmin get removed in this PR. Does this have to be done in two steps?

Yep, needs two steps unfortunately.  

>Other than that the code changes make sense and look very thorough.

Yeah, I disabled `vagovadmin` first just to see what would happen, and a lot of things broke -- which was when I realized we had a whole lot of stuff that I think was never properly updated.  

I think I'll probably need to go through each of the blocks, scripts, files, etc, figure out where exactly it's used, then confirm that it's functioning as intended.  Maybe open a PR based off this one that removes `vagovadmin` from the filesystem and actually use that one for the manual testing.

That all sounds suspiciously like work, so I haven't done it yet 😉

docroot/themes/custom/vagovadmin/yarn.lock
ndouglas: Yeah, that's weird.  I wonder if I was using the wrong version of `yarn` or something.  Honestly can't remember exactly why I was tinkering with the `vagovadmin` installation anyway 🤔 

I might need to go through and sort of retrace my steps when I return to this, now that I have a better picture of what needs to be done.  Maybe redo it from scratch 🤔   Probably the wise thing to do would be to carry out the changes that didn't get fully implemented, test those on an individual basis, and then ideally zapping the old theme should be trivial.

Issue #11450, opened 2022-11-03
Fix php 8.1 deprecation notices


cmaeng:

cmaeng:
could be done by either Platform CMS or Sitewide CMS folks

ElijahLynn:
Upstream issue: https://github.com/drupal-graphql/graphql/issues/1297

ElijahLynn:
This is also causing tests to fail with:

```
PHP Fatal error:  During inheritance of IteratorAggregate: Uncaught Behat\Testwork\Call\Exception\CallErrorException: 8192: Return type of Drupal\graphql\Plugin\TypePluginManagerAggregator::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /var/www/cms/docroot/modules/contrib/graphql/src/Plugin/TypePluginManagerAggregator.php line 49 in /var/www/cms/docroot/vendor/behat/behat/src/Behat/Testwork/Call/Handler/RuntimeCallHandler.php:90
   Stack trace:
   #0 /var/www/cms/docroot/modules/contrib/graphql/src/Plugin/TypePluginManagerAggregator.php(5): Behat\Testwork\Call\Handler\RuntimeCallHandler->handleError(8192, 'Return type of ...', '/var/www/cms/do...', 49)
   #1 /var/www/cms/docroot/vendor/composer/ClassLoader.php(571): include('/var/www/cms/do...')
   #2 /var/www/cms/docroot/vendor/composer/ClassLoader.php(428): Composer\Autoload\includeFile('/var/www/cms/do...')
   #3 [internal function]: Composer\Autoload\ClassLoader->loadClass('Drupal\\graphql\\...')
   #4 /var/www/cms/docroot/core/lib/Drupal/Core/DependencyInjection/Compiler/TaggedHandlersPass.php(143): ReflectionMethod->__construct('Drupal\\graphql\\...', 'addTypeManager')
   #5 /var/www/cms/docroot/core/lib/Drupal/Core/DependencyInjection/Compiler/TaggedHandlersPass.php(116): Drupal\Core\DependencyInjection\Compiler\TaggedHandlersPass->processServiceCollectorPass(Array, 'graphql.type_ma...', Object(Drupal\Core\DependencyInjection\ContainerBuilder))
[...]

PR #11620, opened 2022-11-16
VACMS-11450: Hide deprecation notices on prod in prep for PHP 8.1.

timcosgrove:
## Description

#11450

There are some libraries that we will not be able to update in advance of moving to PHP 8.1. We will hide deprecation notices on Prod until everything is up to date for PHP 8.1.


swirtSJW:
Double checking to make sure I am not misunderstanding something.  Do we still need this PR now that https://github.com/department-of-veterans-affairs/va.gov-cms/pull/11621 has been merged?

timcosgrove:
not needed

PR #11621, opened 2022-11-16
VACMS-11450: hide deprecation notices everywhere

timcosgrove:
## Description

#11450 

Hides deprecation notices in all environments.


docroot/sites/default/settings.php
```diff
@@ -253,3 +253,6 @@
 if (file_exists($env_services_path)) {
   $settings['container_yamls'][] = $env_services_path;
 }
+
+// Hide deprecation warnings during transition to PHP 8.1.
+ini_set('error_reporting', E_ALL & ~E_DEPRECATED & ~E_USER_DEPRECATED);
```
swirtSJW, line 258: Doesn't this cause notices to show on prod? Or does another ini override it?

ndouglas: mebbe we should do something like 

```php
$error_reporting = (int) ini_get('error_reporting');
$error_reporting = $error_reporting & ~E_DEPRECATED & ~E_USER_DEPRECATED;
ini_set('error_reporting', $error_reporting);
```

(untested)

to ensure that we're not turning on anything new, but definitely turning off `E_DEPRECATED` and `E_USER_DEPRECATED`?

timcosgrove: Good call. I tested Nate's idea; works fine.

swirtSJW: Pretty slick solution.     
Follow up question... do we want this off on local?  Or do we want them surfaced on local.  This hides em everywhere.

timcosgrove: I think they should be off and surfaced by an engineer tasked with fixing them, unless we want to encourage speed of correcting the issues through annoying people. 

That said, until we update GraphQL 3.x with a PHP 8.1-compliant patch, that is where the vast bulk of the warnings sit, and people don't do a ton of GraphQL locally AFAIK. Still, I'm not sure I feel it's necessary for all people working locally to potentially be subjected to deprecation warnings.

ndouglas: yeah, IMHO open an issue to address since we have dedicated tools (i.e. PHPStan) for finding deprecations.

ndouglas: #11648

timcosgrove: Added local display; we can always turn it off if it does get annoying.

swirtSJW: @timcosgrove I agree with the case you and @ndouglas  just made.

timcosgrove: OK, if you approve, go ahead!

ndouglas (approved):

PR #11703, opened 2022-11-23
VACMS-11698: Last-touched-by-an-editor field not always reflecting updates

### LINKED ISSUE #11698: Last-touched-by-an-editor field not always reflecting updates
## Description
See e.g. this node: https://prod.cms.va.gov/san-diego-health-care/locations/oceanside-va-clinic

Screen Shot 2022-11-23 at 12 14 37 PM

We should probably see a last-updated date here of November 23 2022 8:34am.

## Acceptance Criteria
- [ ] Last-touched-by-an-editor field updates reliably when the node is updated by a human.

edmund-dunn:
## Description

Closes #11698.

## Testing done
Tested locally. 

## Screenshots
CleanShot 2022-11-23 at 11 54 58
CleanShot 2022-11-23 at 11 54 45

## QA steps

When a piece of content is updated/added the Last Saved by and Editor field and the latest change int he Recent changes block should be the same if saved by a human. Select a piece of content, scroll to the bottom, change moderation to Published, add a revision log message such as "testing" and click save. The same timestamp should be reflected at the top of the Recent Changes block and in the Last Saved by an Editor field. If you click on the edit tab again, both times in the form meta in the top right corner should be the same. 

### Definition of Done

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [x] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


docroot/modules/custom/va_gov_backend/src/EventSubscriber/LastEditorSaveEventSubscriber.php
```diff
@@ -73,9 +73,8 @@ public function formAlter(FormAlterEvent $event): void {
       $node = $form_object->getEntity();
 
       $form['field_last_saved_by_an_editor']['#access'] = FALSE;
-      $form['actions']['submit']['#submit'][] = [
-        $this, 'lastSavedByEditorSetTimestamp',
-      ];
+      array_unshift($form['actions']['submit']['#submit'],
+        [$this, 'lastSavedByEditorSetTimestamp']);
```
swirtSJW, line 77: I am a little confused on the format here.  Why are we adding $this to the array of the submit handler that should  just be a callback name.
What does passing it $this do?  I can't find that pattern anywhere.
```suggestion
      array_unshift($form['actions']['submit']['#submit'], 'lastSavedByEditorSetTimestamp');
```

ndouglas: That's [this pattern](https://www.php.net/manual/en/language.types.callable.php):

Screen Shot 2022-11-25 at 8 16 38 AM

ndouglas (dismissed):

ndouglas:
This looked good to me.  However the instance rebuilt so I'd prefer to check this again just so I can be sure I've dotted all my t's and crossed all my i's, etc.

ndouglas:
Had an issue because of #11725.

docroot/modules/custom/va_gov_backend/src/EventSubscriber/LastEditorSaveEventSubscriber.php
edmund-dunn: @swirtSJW this is how you add the callback inside event subscribers. Otherwise, that method, in this case `lastSavedByEditorSetTimestamp`, would have to go in the `.module` file.

swirtSJW: Oh thanks for explaining. I see why you did that.  That seems like a fair amount of overhead to pass that instantiated object around the steps of a form process, just to be able to call a  4 line function.   A less intensive way to get at it, without moving that function to the .module, would be to change it to a public static function then call it using its namespaced path
like this

array_unshift($form['actions']['submit']['#submit'], 'Drupal\va_gov_backend\EventSubscriber\LastEditorSaveEventSubscriber::lastSavedByEditorSetTimestamp');

Granted, this particular object is pretty small and not a lot of overhead, but I wouldn't want somebody finding and following this pattern, when they have an object that is more gigantic.

swirtSJW: I made that function public and used the static call I referenced above and it works fine.

ndouglas: I'd be surprised if there were meaningful overhead to the `$this` approach for an object of any size, since it should just be passing a reference around.

edmund-dunn: Just pushed this change.

swirtSJW: @ndouglas you are right it would be by reference so there should be no additional overhead IF that event subscriber instance still exists.  I am not certain when event subscriber instances get cleaned up.  Maybe they never do (spine shivers).  In my head I was not picturing it as existing past the first load of node edit, but formAlter does fire on the submit too, so maybe it does persist, or maybe a new instance does?.  

I think in terms of architecture though... calling a static method seems a safer implementation.  
**Example from a distopian future:**  Somebody adds another submit handler to this class, and both submit handlers pass in the same instance in the callback.  However the new callback since it is using public method rather than static, alters the object in some way that the original callback does not expect.  The one callback could alter the outcome of the other, because they both have the ability/risk to interact with the same instance of the object.  

By using only static calls, we prevent one from inadvertently interacting with the other and all is right with the world  :)

swirtSJW (approved):
Looks great @edmund-dunn  
Tests out good.
image

docroot/modules/custom/va_gov_backend/src/EventSubscriber/LastEditorSaveEventSubscriber.php
ndouglas: Now I feel like you're just arguing against OOP 😛  You're not gonna get much argument from me there.  

But in many cases (most?), we _will_ need to consult services from the service container, and in those cases I think it's better to perform calls on an instantiated object with the services DI'ed than from a static method or function that calls `\Drupal::whatever()`.  That's the sort of situation where I'm worried about setting a precedent.

Issue #5087, opened 2021-04-21
Refactor tests.yml and improve tests.


ndouglas:
## Description
The `tests.yml` used by Task is a mess.  

Under Yaml-Tasks, the task file was a list of simple 1-2 line tasks.  Since Task is not a 1:1 replacement, however, a lot of the functionality of Yaml-Tasks was reimplemented, and since there was nowhere else to do it, that functionality was reimplemented within the task file.  The internal shell used by Task is not a 1:1 replacement for Bash, which led to #5081.

There's a ton of repetition and the repeated code is not subject to any automated tests.  So I think it's worth refining this further.

As much as possible of the reusable functionality should be implemented in `task` itself.

## Possible Implementation
- Extract the bodies of the tasks from the tests.yml file into separate scripts, e.g. `task-va-web-build.sh`, `task-va-test-status-error.sh`.
- Refactor these scripts to use a library script that defines a few Bash functions to make the scripts a little cleaner and reduce repetition.
- Write a new test (and task) for common functionality so that we can be assured that:
  - passed tests are always marked as passed
  - failed tests are always marked as failed

## Acceptance Criteria
- [x] `tests.yml` is refactored in such a way that it is easier to read/lint/maintain
- [x] `tests.yml` has automated tests

ndouglas:
Closed in #12192.

PR #11748, opened 2022-11-30
VACMS-11639: Fix status-error processing.

### LINKED ISSUE #11639: Change `status-error` test to parse Drush output rather than count lines.
## Description
We can use `--format=json` and `jq` to parse the output and not get thrown off by incidental notices/deprecation messages/etc.

## Acceptance Criteria
- [ ] `status-error` test parses JSON rather than counting lines from the useless `--format=list`.

ndouglas:
## Description

Closes #11639.

This is unfortunately difficult to test interactively; this whole file is a mess and needs to be completely rewritten (see #5087).

### DIFF
--- a/tests.yml
+++ b/tests.yml
@@ -125,10 +125,10 @@ tasks:
           output="\$(mktemp -d)/output"
           mkfifo \$output
           (cat \$output&)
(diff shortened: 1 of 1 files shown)


laflannery:
@ndouglas I don't understand why these issues are being flagged either, those are not the correct foreground colors

edmund-dunn (approved):
This makes sense to me.

PR #11762, opened 2022-12-01
VACMS-11761: Delay accessibility tests to allow CSS transitions to complete.

### LINKED ISSUE #11761: Debug new (apparently spurious) Cypress accessibility test failures
## Description
We've been encountering [some apparently unrelated but reasonably reliable accessibility test failures](https://dsva.slack.com/archives/C52CL1PKQ/p1669834551532619) across various PRs, so I need to figure out why and fix the issue.

## Acceptance Criteria
- [ ] PRs are no longer failing accessibility tests for things they didn't change.

ndouglas:
## Description

Closes #11761.  We're using CSS transitions with a short-but-nonzero duration when we draw various links, including in the toolbar.  Unfortunately, during this time the accessibility tests can run, and we're not _always_ accessible.

This change simply delays a second after page load and script injection to allow these transitions to complete before actually checking the accessibility of the DOM.

If accessibility tests pass, this should be safe to merge.

### DIFF
--- a/tests/cypress/integration/accessibility/accessibility.spec.js
+++ b/tests/cypress/integration/accessibility/accessibility.spec.js
@@ -64,6 +64,7 @@ describe("Component accessibility test", () => {
     it(testName, () => {
       cy.visit(route);
       cy.injectAxe();
+      cy.wait(1000);
       cy.checkA11y(axeContext, axeRuntimeOptions, (violations) => {
         cy.accessibilityLog(violations);
         const violationData = violations.map((violation) => ({


laflannery:
@ndouglas this seems logical and makes sense to me.  I know my tools will not run until the page loads which would probably account for the discrepancy I noted yesterday (the fact that I couldn't replicate the errors we were seeing). I'm good with this

swirtSJW (approved):
Makes sense to me.

Issue #11082, opened 2022-10-05
Investigate gathering function line count metrics.


ndouglas:
## Description
The line count of a function is a good indicator as to its general quality.  Generally speaking, about five lines should be sufficient for a function to Do One Thing.  A long line count likely means that the function is doing more than one thing, which leads to difficulty in analyzing, difficulty in testing, side effects due to ordering, overly acquisitive treatment of dependencies, poor naming, etc.  This is common with hook implementations, where we're doing several different modifications to different node forms; thus a general concept (such as Campaign Landing Pages) can have code spread out over several different functions.

## Acceptance Criteria
- [x] Gathering function line count metrics has been investigated.
- [x] If practical, function line count metrics are reported via #11073.

ndouglas:
Closed in #11923.

Issue #11081, opened 2022-10-05
Investigate gathering metrics on cyclomatic complexity


ndouglas:
## Description
[Cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) is a useful software metric indicating how easy it is to reason about code, how easy it is to test it, etc.  We should record this.  It'd also be nice if we reported function names too so that we can identify the top function and target it for refactoring.

## Acceptance Criteria
- [x] Gathering cyclomatic complexity metrics from our custom code has been investigated.
- [x] If practical, a followup issue has been opened to address gathering and reporting via #11073.

ndouglas:
Closed in #11923.

Issue #11079, opened 2022-10-05
Record custom module `.module` line count metrics.


ndouglas:
## Description
This is a very simple metric for reporting the total line count (comments excluded, ideally, to avoid creating a perverse incentive) of our custom module `.module` files, individually and aggregated.  Code in `.module` files tends to be without unit tests and be either hook implementations (see #11075 and #10964 ) or common code shared by hook implementations.

This should be very simple once #11073 is completed.

## Acceptance Criteria
- [x] Custom module `.module` line counts, separately and aggregated, are reported to DataDog.

ndouglas:
Closed in #11923.

PR #11923, opened 2022-12-13
VACMS-11082: Investigate gathering line-count metrics.

### LINKED ISSUE #11075: Record hook implementations in custom modules as a metric.
## Description
This should be straightforward provided that the hook implementations are appropriately commented.  In that event, we can just count the comments that match `/Implements\s+hook_.*/` or something like that.

This depends on #11073.

## Acceptance Criteria
- [x] Verified that we can accurately create a metric for hook implementations.
- [ ] Metric for hook implementations is collected and reported upstream to DataDog.

ndouglas:
## Description

See #11082, #11081, #11079.  This is a big step in the right direction, even if I was kind of hoping for a deeper look.

**EDIT**: Also closes #11075 now.

## Testing done
This should be safe to merge, but I'm going to run it on CMS-Test to confirm the dumb thing actually works (more-or-less).

Mostly works, at least it appears I have no syntax errors, etc:

## CMS-Test Output
Screen Shot 2022-12-13 at 2 51 48 PM


swirtSJW (dismissed):
This is pretty amazing.  Nice work. 
At the same time it is a bit creepy, and makes me want to go move my closet skeletons to somewhere they will not be audited by this ;)

PR #12041, opened 2022-12-20
VACMS-11897: Remove reference to vertical-tabs.js

### LINKED ISSUE #11897: Discovery: Determine cause "failed to open stream" warning related to vagovclaro.info.yml
The following warning has been found and we need to determine the cause and solution. This was not related to the PHP 8.1 upgrade. The issue was reported in this [slack thread](https://dsva.slack.com/archives/CT4GZBM8F/p1670535435011379)

possibly could be the result of a subtheme issue:

### Acceptance Criteria
- [ ] Determine the cause of the file error
- [ ] Document possible solutions
- [ ] Create fix implementation ticket

### CMS Team

- [ ] `Program`
- [ ] `Platform CMS Team`
- [ ] `Sitewide Crew`
- [X] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public Websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

ndouglas:
## Description

Closes #11897.

The line references a file that exists... in the Claro supertheme.  Not the subtheme.  I believe this is causing errors, but also forcing the core Drupal behavior, which in at least some cases is desired.  So I think I need to override this in such a way that the core Drupal behavior is used without triggering errors.

**TL;DR**: Disable the Claro vertical tabs library rather than point it to a nonexistent file.

### Current Prod Behavior

Screen Shot 2022-12-20 at 1 04 11 PM

### Claro Behavior (AFAICT)

Screen Shot 2022-12-20 at 1 04 46 PM

So the goal is to get the Prod behavior, but without the errors caused by referring to a nonexistent file.

### PR Behavior

Screen Shot 2022-12-20 at 2 09 15 PM


docroot/themes/custom/vagovclaro/vagovclaro.info.yml
```diff
@@ -46,7 +46,7 @@ libraries-override:
       component:
         misc/vertical-tabs.css: false
     js:
-      misc/vertical-tabs.js: js/vertical-tabs.js
+      misc/vertical-tabs.js: false
```
timcosgrove, line 49: Re: 12041: I see this issue for extending Claro: https://www.drupal.org/project/claro/issues/3085894#comment-13555431

It has a different location for the Claro-provided vertical tabs js:
```
  core/drupal.vertical-tabs:
    css:
      component:
        /core/themes/stable/css/core/vertical-tabs.css: false
    js:
      misc/vertical-tabs.js: /core/themes/claro/js/vertical-tabs.js
```

Would that work?

ndouglas: I expect that would make the interface behave as in the second image.  We don't want that.  

If I just remove the line, the file resolves correctly, but that leads to it making the form behave incorrectly.

timcosgrove (approved):

PR #12109, opened 2022-12-23
VACMS-11834: Improve password_policy test.

ndouglas:
## Description

Closes #11834.

Rather than scripting the UI interactions, I just changed this to Drush config settings.

If this passes all tests, it should be safe to merge.


edmund-dunn (approved):
looks good to me!

PR #12131, opened 2022-12-28
Install `jq` and `nano` on Tugboat.

ndouglas:
`jq` is needed for one of the tests.  `nano` is just useful for people who don't use Vim.


swirtSJW (approved):
Nice addition and as a Nano fan... I applaud this.

PR #12210, opened 2023-01-11
VACMS-0000: Update loadtest dependencies.

ndouglas:
## Description

Resolves [this Dependabot issue](https://github.com/department-of-veterans-affairs/va.gov-cms/security/dependabot/82).

Also adds a reasonable default admin user.


edmund-dunn (approved):
Makes sense to me. Approved.

PR #12393, opened 2023-02-01
VACMS-10579: Adds content validator for links using absolute URLs as relative paths.

### LINKED ISSUE #10579: Content validator for /<scheme>://<url> URLs.
## Description
These pop up occasionally, or might pop up occasionally, per Swirt.

## Acceptance Criteria
- [ ] Malformed links like `/http://example.org/` or `/https://example.org/` are caught when inserted in content.

ndouglas:
## Description

Closes #10579.

## QA steps

As any user with the ability to edit content:
1. Edit a node, e.g. [VAMC System Health Service Whole health at VA Cincinnati health care](https://cms-pxuj8mm5h1fhqfidhiax67zm0v28bxf4.ci.cms.va.gov/node/24426/edit)
2. Insert a relative link containing a URL into a rich text field.  For instance: `<a href="/https://www.example.org/">Example Link!</a>`.
3. Attempt to save the document.  You should see:
  - [x] An error message at the top of the screen:
Screen Shot 2023-02-01 at 11 00 38 AM
  - [ ] validate error message link jumps to the error.

  - [x] An error message at the field:
Screen Shot 2023-02-01 at 11 01 11 AM

  - [x] An error message at the bottom of the field:
Screen Shot 2023-02-01 at 11 00 52 AM

4. Edit a media image item, e.g. [August2022.jpg](https://cms-pxuj8mm5h1fhqfidhiax67zm0v28bxf4.ci.cms.va.gov/media/image/16769)

5.  Insert a relative link containing a URL into a plain text field. For instance, `/https://www.example.org/`.

6. Attempt to save the media image item.  You should see:
  - [x] An error message at the top of the screen:
Screen Shot 2023-02-01 at 11 02 47 AM
  - [ ] validate error message link jumps to the error.

  - [x] An error message at the field:
Screen Shot 2023-02-01 at 11 02 51 AM

  - [x] An error message at the bottom of the field:
Screen Shot 2023-02-01 at 11 02 56 AM

### Definition of Done

### Select Team for PR review

- [x] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -202,7 +202,7 @@
         "symfony/process": "^4.0",
         "symfony/routing": "^4.0",
         "traviscarden/behat-table-comparison": "~0.2",
-        "va-gov/content-build": "^0.0.3186",
+        "va-gov/content-build": "^0.0.3187",
         "vlucas/phpdotenv": "^5.3",
         "webflo/drupal-finder": "^1.0.0",
         "webmozart/path-util": "^2.3",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "41a222d9c68a040b6cea7eb2fd50a703",
+    "content-hash": "dab23f2c0dbde8ea7d3b45def1dc2e5e",
     "packages": [
         {
             "name": "acquia/drupal-spec-tool",
@@ -6588,29 +6588,34 @@
         },
         {
             "name": "drupal/flood_control",
-            "version": "2.2.4",
+            "version": "2.3.0",
             "source": {
                 "type": "git",
                 "url": "https://git.drupalcode.org/project/flood_control.git",
-                "reference": "2.2.4"
+                "reference": "2.3.0"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://ftp.drupal.org/files/projects/flood_control-2.2.4.zip",
-                "reference": "2.2.4",
-                "shasum": "8dcebe57abdc9ca0ff11c42006436a7fa0694750"
+                "url": "https://ftp.drupal.org/files/projects/flood_control-2.3.0.zip",
+                "reference": "2.3.0",
+                "shasum": "194fc2134fd02c4f944656be8af2b05cd5b73d74"
             },
             "require": {
-                "drupal/core": "^8 || ^9 || ^10"
+                "drupal/core": "^8 || ^9"
             },
             "type": "drupal-module",
             "extra": {
                 "drupal": {
-                    "version": "2.2.4",
-                    "datestamp": "1663328704",
+                    "version": "2.3.0",
+                    "datestamp": "1674851365",
                     "security-coverage": {
                         "status": "covered",
                         "message": "Covered by Drupal's security advisory policy"
+                    }
+                },
+                "drush": {
+                    "services": {
+                        "drush.services.yml": "^9 || ^10"
                     }
                 }
             },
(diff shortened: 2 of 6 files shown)


tests/phpunit/Content/PreventAbsoluteUrlsAsPathsInLinksValidatorTest.php
```diff
@@ -0,0 +1,109 @@
...
+        'This contains a <a href="/https://www.example.org">path consisting of an absolute URL</a> inside a tag and should fail.',
+        'text_long',
+        'plain_text',
+      ],
+      [
+        TRUE,
+        '/http://paths-that-look-like-absolute-urls outside of <a href="https://www.example.org/">anchor tags</a> should not trigger the validator.',
+        'text_long',
+        'rich_text',
+      ],
```
swirtSJW, line 105: There are two other cases that we might want to catch here, though I could also be talked into them being in a separate validator

`<a href="//www.example.org/"`
`<a href="/www.example.org/"`
`<a href="www.example.org/"`

swirtSJW (commented):
This is fantastic @ndouglas.  Just one scope type question.

tests/phpunit/Content/PreventAbsoluteUrlsAsPathsInLinksValidatorTest.php
swirtSJW: I am guessing these ought to be separate as the directive text on how to fix would be different.

ndouglas: The first one is already taken care of by a separate validator (protocol-relative link validator, I think it's called)

#12395 handles the second case, but only for `va.gov` domains -- I think we discussed this briefly at some point and landed on VA domains to avoid being too greedy, but maybe I misremember.

The third scares me a bit since I can imagine a path segment that looks a whole lot like a domain name.  If it's restricted to VA.gov domains that's less concerning.  Thoughts?

swirtSJW:
I just added an additional QA step.
`validate error message link jumps to the error.`
image

swirtSJW:
I was thinking this was missing coverage for link fields but it looks like link fields have their own validator
image

Yours has better error message text ;)

swirtSJW:
Everything else checks out great with the exception of the jump link.   I thought it was just related to the wysiwyg because CKEeditor messes around with the markup, but it didn't work for the plain text field either.

ndouglas:
That jumping-to-the-right-place thing is nasty; IIRC validators have to have a bit of extra code added in to handle fields within paragraphs and fields on the base entity the same way.  I might've not gotten that right 🤔

ndouglas:
Weirdly, the link generated is `#edit-field-body-0-value`.  When clicked, the link does not work.  If entered directly in the address bar, it is rewritten to`#cke_edit-field-body-0-value` and works.

ndouglas:
It's [this Core bug again](https://www.drupal.org/project/drupal/issues/3229493), but I think something changed and we need to reïnvestigate the underlying behavior again.  Might be as simple as removing my patch entirely.

ndouglas:
Blocked by #12415.

swirtSJW:
Suggesting here too that we not block it on the core defect.   Reaching out to Laura for consensus.

swirtSJW (approved):
Tested and approved.  I don't think we need to block these on the core defect.

swirtSJW (dismissed):
Tested and approved.  I don't think we need to block these on the core defect.

ndouglas:
Because this and the other validation PR both altered `va_gov_backend.module`, I needed to resolve a trivial conflict.  I'm going to go ahead and merge this rather than asking for a re-review 🙂

Issue #12217, opened 2023-01-12
Spike - Update WYSIWYG with user behavior


Dottisea:
To get the work done for #10708, the WYSIWYG field for COVID-19 Status details needs to be updated, which seemed like it would be the most difficult technical piece of the work, due to the nature of CKEditor and the `<iframe>`.

omahane:
I started this work during this sprint. There's a lot of code in this PR: #12185 . I was running into a 502 bad gateway error when visiting the VAMC Facility node:add form as a content editor. I confirmed that it wasn't happening on main, so it seemed to be related to my code. @swirtSJW thought it might have something to do with the userPerms service (possibly). I'm still investigating.

smythe5:
@omahane can you provide an update on this ticket that I can use for sprint reporting?  I need to provide a brief statement as to why it is not done and/or what all is needed to get this complete.  Thank you.

omahane:
Bad gateway issue is resolved.

Issue #10708, opened 2022-09-12
CMS: Update how COVID status feature works on VAMC facilities


dsinla:
## Description

The current (Sep 2022) bulleted descriptive text for COVID status policies does not meet the operational reality at the facilities - it's not granular enough to give Veterans the variation of policies that are in place at each facility.

- The editorial interface should be updated to match the design from #10842
- Will also need to update the field that sends information to the COVID status
- May be intertangled with #11794

The content team is providing [guidance for editors](https://github.com/department-of-veterans-affairs/va.gov-team/issues/45774), which will need to be added to a new KB article that will be referenced from the COVID-19 status area of the editorial interface. 

## User Stories

As a Veteran, I want information about COVID policies that reflects the reality at the facility, in order to prepare for my visit.
As a VAMC editor, I want guidance on the best way to author content that accounts for the reality of the policies at my facility. 

## Acceptance Criteria
- [ ] ~The **COVID-19 status** descriptive text is rewritten and a link to the KB guidelines is provided.~ See Figma
- [ ] ~The **Guidelines** section is removed~  This section is not a section, it is content on the three taxonomy terms.  The solution is to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12111
- [x] **Operating status** descriptive text is updated - ~it should instruct editors to use the open field for COVID status (perhaps link to the KB again in this area)~  follow Figma
- [x] New field field_supplemental_status_more_i is added to health_care_local_facility content type
   - Cardinality 1.
   - rich text limited 
   - required
   - 500 character count limit excluding html
 - [x] field layouts labels and guidance text matches [FIGMA](https://www.figma.com/file/IYpeP2ZAmBTT8jqhdtJklu/Covid-%2F-Facility-Status?node-id=128%3A32180&t=GcTmUSaSHKT2y5BN-0)
 - [x] When a selection in field_supplemental_status is made/changed, the  term facility_supplemental_status->description immediately populates the node's field_supplemental_status_more_i with the same html and content.
 - [x] Once populated, field_supplemental_status_more_i can be changed as needed by the editor.
 - [x] cypress test fails if field_supplemental_status_more_i is not populated on change of field_supplemental_status

### CMS Team
Please check the team(s) that will do this work.

- [ ] `Program`
- [ ] `Platform CMS Team`
- [ ] `Sitewide Crew`
- [ ] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public Websites`
[...]

Issue #10805, opened 2022-09-22
[RESEARCH] Investigate bundle classes as an avenue to refactoring and improving code quality.


ndouglas:
## Description
See https://www.drupal.org/node/3191609.

Introduced in Drupal 9.3.x, bundle classes allow you to associate a specific PHP class with a content type, meaning that you can write clean object-oriented code that follows the Law of Demeter rather than incomprehensible spaghetti code.

>Encapsulating all the required logic for each bundle into its own subclass opens up many possibilities for making more clear, simple, maintainable, and testable code. Drupal itself has no opinion on how you should structure your code, but this change allows you to use whatever object oriented style/design/paradigm you prefer.

We should investigate adopting these classes and using this concept to aid improving the overall quality of our codebase.

## Acceptance Criteria
- [ ] A bundle class has been written for at least one content type.
- [ ] At least some code currently targeting that content type has been refactored to target the bundle class.
- [ ] Unit tests have been written to cover the above refactoring.
- [ ] If the approach satisfies QA goals, findings are documented in the codebase.
- [ ] Otherwise, findings are documented in this thread.

PR #12403, opened 2023-02-01
VACMS-12217: Update wysiwyg from user behavior

omahane:
## Description

Relates to #12217 
Relates to #10708

This PR demonstrates through manual testing and through Cypress testing (`tests/cypress/integration/behavioral/content_editing_vamc_facility.feature`) that the WYSIWYG is updated from user interaction in another field.

It also shows that the field values persist on save, are shown in node:view and are loaded into the WYSIWYG on page load.

Not included is complete UI fidelity to the [FIGMA mockup](https://www.figma.com/file/IYpeP2ZAmBTT8jqhdtJklu/Covid-%2F-Facility-Status?node-id=128%3A32180&t=GcTmUSaSHKT2y5BN-0), though some of the work to approximate the mockup has been done.

## Testing done
Manually
Cypress

## Screenshots
ezgif com-video-to-gif

### Create a facility
As an admin, create a VAMC Facility
- [x] Fill in any required fields
- [x] Validate that when choosing a COVID-19 Status that the WYSIWYG changes boilerplate copy, replacing the old copy
- [x] Validate that you can select all the copy and replace it with your own copy
- [x] Save the facility node
- [x] Validate that the node:view shows the data that was in the Details 
- [x] Click "Edit"
- [x] Validate that your own copy is showing in the COVID-19 Status Details WYSIWYG

### Edit a facility
As an editor, edit a VAMC Facility
- [x] Validate that on load, the COVID-19 Details field uses boilerplate copy, based on the default selection of the radio button
- [x] Validate that when choosing a different COVID-19 Status that the WYSIWYG changes boilerplate copy, replacing the old copy
- [x] Validate that you can select all the copy and replace it with your own copy
- [x] Save the facility node
- [x] Validate that the node:view shows the data that was in the Details 
- [x] Click "Edit"
- [x] Validate that your own copy is showing in the COVID-19 Status Details WYSIWYG

### Definition of Done

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [x] `Facilities`
- [ ] `User support`

- [x] `DO NOT MERGE`

- [ ] `Needs PO review`


omahane:
The Cypress tests show as failing due to `behavioral/content_release.feature`:
> 1) Content Release
       The content release page should show a pending default release initiated from the command line:
     AssertionError: Timed out retrying after 10000ms: Expected to find content: 'Preparing' but never did.
      at Context.eval (https://pr12403-yjq5zexhavmhassuj4kctou3kauqjue2.ci.cms.va.gov/__cypress/tests?p=tests/cypress/integration/behavioral/content_release.feature:73222:28)
      at Context.resolveAndRunStepDefinition (https://pr12403-yjq5zexhavmhassuj4kctou3kauqjue2.ci.cms.va.gov/__cypress/tests?p=tests/cypress/integration/behavioral/content_release.feature:27189:29)
      at Context.eval (https://pr12403-yjq5zexhavmhassuj4kctou3kauqjue2.ci.cms.va.gov/__cypress/tests?p=tests/cypress/integration/behavioral/content_release.feature:26510:35)

The Cypress test I added (`behavioral/content_editing_vamc_facility.feature`) passed.

tests/cypress/support/commands.js
```diff
@@ -94,6 +94,21 @@ Cypress.Commands.add("type_ckeditor", (element, content) => {
   });
 });
 
+Cypress.Commands.add("read_ckeditor", (element, content) => {
```
ndouglas, line 97: trying to figure out why I made `type_ckeditor` snake_cased instead of camelcased 🤔

tests/cypress/integration/behavioral/content_editing_vamc_facility.feature
```diff
@@ -0,0 +1,27 @@
...
+  As anyone involved in the project
+  I need to have certain functionality available
+
+  Scenario: Log in and edit a VAMC Facility as an editor
+    When I am logged in as a user with the roles "vamc_content_creator, content_publisher"
+    And my workbench access sections are set to "372"
+    When I am at "/alaska-health-care/locations/anchorage-va-medical-center"
+    And I click the edit tab
+    And I select the radio button with the value "1037"
+    And I wait "2" seconds
```
ndouglas, line 13: Are these lines still necessary with the 5 second delay in `read_ckeditor`?

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -18,14 +21,14 @@
...
-const USER_CMS_HELP_DESK_NOTIFICATIONS = 4050;
-
 /**
  * VA.gov VAMC Entity Event Subscriber.
  */
 class EntityEventSubscriber implements EventSubscriberInterface {
 
+  // The UID of the CMS Help Desk account subscribing to facility messages.
+  const USER_CMS_HELP_DESK_NOTIFICATIONS = 4050;
+
```
ndouglas, line 31: Nice!

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -230,10 +235,46 @@ public function addCovidStatusTermTextToSettings(array &$form): void {
...
+          . $term_storage->load($status)->getDescription();
+          $form['field_supplemental_status_more_i']['widget']['0'] = [
+            "#title" => $widget_title,
+            '#description' => $widget_description,
+            '#type' => 'text_format',
+            '#default_value' => $chosen_term_description,
+            '#format' => 'rich_text_limited',
+          ];
+        }
+      }
```
ndouglas, line 270: Could this be factored out into a separate method?  Or two?  Or three?

tests/cypress/support/commands.js
ndouglas, line 97: I'm confused about what this function is doing exactly.  

Above:

```
Then(`I should see {string} in ckeditor {string}`, (value, label) => {
  cy.read_ckeditor(label, value);
});
```

I feel like that's expecting that `cy.read_ckeditor()` will perform an assertion, since there's nothing explicitly making an assertion anywhere else.

But the only place `content` is used is in `win.CKEDITOR.instances[element].getData(content);` -- which doesn't look to me like it takes an argument:

Screen Shot 2023-02-07 at 8 28 48 AM

and it doesn't seem to me to perform any assertions.

I feel like `read_ckeditor` should retrieve the data and nothing else -- maybe we want to log it, maybe we want to check if it _contains_ a string, or if it _exactly matches_ a string or a regex, etc.

Then the step definition for `I should see {string} in ckeditor {string}` should contain the assertion itself.

omahane: > I feel like read_ckeditor should retrieve the data and nothing else -- maybe we want to log it, maybe we want to check if it contains a string

@ndouglas Yes, that's definitely what I was going for, checking to see if the WYSIWYG has certain text in it. It looks like I missed the mark, though. I'm not sure how to get that assertion in there. Also, I was having no luck with debugging a Cypress test as I might other JS, no matter the number "debugger"s I put in.

ndouglas: I think you can just do `return cy.wrap(win.CKEDITOR.instances[element].getData()` in `read_ckeditor`.  

Then in the step definition, you can do `cy.read_ckeditor(label).should('contain', value)`.  I _think_ that'll work.

Cypress is definitely... an experience to debug.  And in general.  Its very unusual and doesn't resemble anything else in the space AFAICT.  `console.log` doesn't go where you expect.  I didn't have much luck with the debugger either -- maybe there's something we're both missing.

I typically end up doing very caveman-debugging stuff.

I assume that at some point another tool will come along that'll replace Cypress and do so in an elegant, more JS-like kind of way.  It'll probably come just as we finally understand Cypress 😩

JayDarnell:
QA passes. Waiting on a code refactor before I do a code review.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -230,12 +232,79 @@ public function addCovidStatusTermTextToSettings(array &$form): void {
...
+   * @return bool
+   *   TRUE if COVID status is set but Details are empty.
+   *   Otherwise, FALSE.
+   */
+  private function isCovidStatusSetAndDetailsEmpty(NodeInterface $node, int $status) {
+    $statusIsEmpty = FALSE;
+    if (isset($node->get("field_supplemental_status")['0'])
+    && isset($node->get("field_supplemental_status")['0']->getValue()['target_id'])) {
+      if ($node->get("field_supplemental_status")['0']->getValue()['target_id'] == $status
+      && empty($node->get("field_supplemental_status_more_i")['0']->getValue())) {
```
ndouglas, line 272: **question**: How did you develop this?  Because this sort of thing with the fielditemvaluelists and the fielditemvalues and entityreferencevalues and so forth is a nightmare to me.  I tend to write unit tests just so I can more-or-less know in real time just what the heck is going on.  But this is a private method so it's excruciating to test in PHPUnit.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -230,12 +232,79 @@ public function addCovidStatusTermTextToSettings(array &$form): void {
...
+   *   The node to be interrogated.
+   * @param int $status
+   *   A single COVID status value.
+   *
+   * @return bool
+   *   TRUE if COVID status is set but Details are empty.
+   *   Otherwise, FALSE.
+   */
+  private function isCovidStatusSetAndDetailsEmpty(NodeInterface $node, int $status) {
+    $statusIsEmpty = FALSE;
```
ndouglas, line 268: **suggestion**: This seems like a misleading variable name, since it's only TRUE if the 'status' field is set to something.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -230,12 +232,79 @@ public function addCovidStatusTermTextToSettings(array &$form): void {
...
+   *
+   * @param array $form
+   *   The form.
+   * @param int $status
+   *   A single COVID status value.
+   * @param \Drupal\taxonomy\TermStorage $term_storage
+   *   The taxonomy term.
+   */
+  private function setCovidStatusDetails(array &$form, int $status, TermStorage $term_storage) {
+    $chosen_term_description = "";
```
ndouglas, line 290: **nitpick**: Setting a default value for this variable seems unnecessary.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -230,12 +232,79 @@ public function addCovidStatusTermTextToSettings(array &$form): void {
...
+    $widget_title = "";
+    $widget_description = "";
+    if (isset($form['field_supplemental_status_more_i']['widget']['0']['#title'])) {
+      $widget_title = $form['field_supplemental_status_more_i']['widget']['0']['#title'];
+    }
+    if (isset($form['field_supplemental_status_more_i']['widget']['0']['#description'])) {
+      $widget_description = $form['field_supplemental_status_more_i']['widget']['0']['#description']->__toString();
```
ndouglas, line 297: **thought**: I think this conceals a call to the Drupal string translation service, or some other service, IDK which.  i.e. somewhere in the called code there's a method that says `return \Drupal::stringTranslation->translate($this);` or something like that.  That makes adding unit tests a PITA.

That's not really an issue for right now, but another thing to think of for #10805.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -230,12 +232,79 @@ public function addCovidStatusTermTextToSettings(array &$form): void {
...
     }
-    $form['#attached']['library'][] = 'va_gov_vamc/set_covid_term_text';
-    $form['#attached']['drupalSettings']['vamcCovidStatusTermText'] = $terms_text;
+    $chosen_term_description = $term_storage->load($status)->getName()
+    . $term_storage->load($status)->getDescription();
```
ndouglas, line 300: **nitpick**: I believe will die if there's no corresponding term for `$status`, i.e. if `load($status)` returns NULL.  Perhaps it should because it indicates a problem with the data integrity of the site, but I'd prefer it be an explicit exception.

tests/cypress/integration/behavioral/content_editing_vamc_facility.feature
```diff
@@ -0,0 +1,37 @@
+@content_editing_vamc_facility
+Feature: CMS Users may effectively interact with the VAMC Facility form
+  In order to confirm that cms users have access to the necessary functionality
+  As anyone involved in the project
+  I need to have certain functionality available
+
+  Scenario: Log in and edit a VAMC Facility as an editor
+    When I am logged in as a user with the roles "vamc_content_creator, content_publisher"
+    And my workbench access sections are set to "372"
+    When I am at "/alaska-health-care/locations/anchorage-va-medical-center"
```
ndouglas, line 10: **thought**: Because this uses an existing node, this can cause problems.  For instance, if the test fails, the node is left locked, which makes it aggravating to develop the test.

It's preferable to create a new, clean node for each test.

This likely has to wait until #12463.

ndouglas:
Cypress test failures appear legit:

https://user-images.githubusercontent.com/1318579/217557508-6fbd4816-2772-4b6c-ba1e-10b1a477fe65.mp4

tests/cypress/integration/behavioral/content_editing_vamc_facility.feature
omahane: I changed this to create a new one as an admin with only the required fields. As as soon as someone (perhaps me or another developer) takes on #12463, we can refactor the node creation part out.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
omahane: I agree. That's a holdover from when it was being set inside a condition.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
omahane: I agree. I changed it.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -230,12 +232,79 @@ public function addCovidStatusTermTextToSettings(array &$form): void {
...
+   *   A single COVID status value.
+   *
+   * @return bool
+   *   TRUE if COVID status is set but Details are empty.
+   *   Otherwise, FALSE.
+   */
+  private function isCovidStatusSetAndDetailsEmpty(NodeInterface $node, int $status) {
+    $statusIsEmpty = FALSE;
+    if (isset($node->get("field_supplemental_status")['0'])
+    && isset($node->get("field_supplemental_status")['0']->getValue()['target_id'])) {
```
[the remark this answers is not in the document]
omahane: I changed this.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -230,12 +232,79 @@ public function addCovidStatusTermTextToSettings(array &$form): void {
...
+   * @param int $status
+   *   A single COVID status value.
+   *
+   * @return bool
+   *   TRUE if COVID status is set but Details are empty.
+   *   Otherwise, FALSE.
+   */
+  private function isCovidStatusSetAndDetailsEmpty(NodeInterface $node, int $status) {
+    $statusIsEmpty = FALSE;
+    if (isset($node->get("field_supplemental_status")['0'])
```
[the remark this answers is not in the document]
omahane: OK, using only single quotes now.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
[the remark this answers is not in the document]
ndouglas: I guess it's cast transparently? 🤔

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -230,12 +232,79 @@ public function addCovidStatusTermTextToSettings(array &$form): void {
...
+    && isset($node->get("field_supplemental_status")['0']->getValue()['target_id'])) {
+      if ($node->get("field_supplemental_status")['0']->getValue()['target_id'] == $status
+      && empty($node->get("field_supplemental_status_more_i")['0']->getValue())) {
+        $statusIsEmpty = TRUE;
+      }
+    }
+    return $statusIsEmpty;
```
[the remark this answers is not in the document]
omahane: Wow, _sounds_ like both useful and daunting for me, especially as blocker for this issue, which is one of our sprint goals. This is the first I'm hearing of bundle classes, so I'd need to do a little more research before I committed to this.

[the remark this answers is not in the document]
ndouglas: Yeah, totally **not blocking** this PR.  Just a thought of how we can improve the codebase in the future.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
ndouglas: They're not fun TBH.  I'm hoping to make it better.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
omahane: If it dies on that. It will do it here, first on line 239   
```
 $terms_text[$status]['name'] = $term_storage->load($status)->getName();
 $terms_text[$status]['description'] = $term_storage->load($status)->getDescription();

```
I am not sure exactly how to defend against that here. I'm seeing other places in the code where we go loading a term or terms without accounting for the null case, but that doesn't make it right, of course.

ndouglas: You can't defend against it, realistically, but you can do something like this (not tested):

```php
$term = $term_storage->load($status);
if (!$term) {
  throw new \Exception('Term not found.');
}
```

Otherwise the error will be something like "Can't call getName() on null" or something like that.  It'd still give the line number, so we'd probably figure it out pretty quickly.  🤷🏻 

It will probably never matter, and like you said, we definitely don't check return values very often.

So... eh, nitpicky.

ndouglas (commented):
Looks great to me 🙂   Thank you!

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -230,12 +232,78 @@ public function addCovidStatusTermTextToSettings(array &$form): void {
...
+   * @param int $status
+   *   A single COVID status value.
+   *
+   * @return bool
+   *   TRUE if COVID status is set but Details are empty.
+   *   Otherwise, FALSE.
+   */
+  private function isCovidStatusSetAndDetailsEmpty(NodeInterface $node, int $status) {
+    $statusButNoDetails = FALSE;
+    if (isset($node->get('field_supplemental_status')[0])
```
JayDarnell, line 269: I'm surprised by the [0] values in this section. The $node parameter passed into this function comes from `$node = $form_object->getEntity();` above. I don't understand the need for the [0] in each of these lines. `$node->get('field_supplemental_status')->value` seems like it should work to me instead of the `[0]->getValue()` but maybe this is some strange Drupal thing because we're in an EventSubscriber vs an old school .module form_alter. @swirtSJW any thoughts?

swirtSJW: I dunno.  It should work and should be the ideal.  DId you test that it didn't work?

JayDarnell:
@christianmburk QA steps pass. The only strange thing I noticed is that trying to use CMD+A on my mac to select all the text in the new field doesn't work. CMD+A will select all the text but when I release the keys the text is no longer selected. I've never seen this behaviour before. 

If I try to use CMD+A to select all the text in the VAMC system service description field on the VAMC System Health Service type it works as expected. Have you seen this behaviour?

omahane:
> 

It was the way the interpolated string was being transpiled from ES6 to JS. It was being prepended with an empty string. Who can say why? I put in a `<p>` to give CKEditor a bone to chew.
Screen Shot 2023-02-13 at 2 49 39 PM

config/sync/field.field.node.health_care_local_facility.field_supplemental_status_more_i.yml
```diff
@@ -0,0 +1,30 @@
...
+      - rich_text_limited
+  epp:
+    value: ''
+    on_update: 0
+id: node.health_care_local_facility.field_supplemental_status_more_i
+field_name: field_supplemental_status_more_i
+entity_type: node
+bundle: health_care_local_facility
+label: Details
+description: 'Add relevant details about how this status will impact Veterans and visitors. This section is pre-populated with suggested text based on the level you selected but is editable to in order to make changes that fit your unique situation.'
```
swirtSJW, line 24: **Could:**
```suggestion
description: 'Add relevant details about how this status will impact Veterans and visitors. This section is pre-populated with suggested text based on the level you selected, but is editable to in order to make changes that fit your unique situation.'
```

docroot/modules/custom/va_gov_vamc/css/set_covid_term_text.css
```diff
@@ -14,3 +14,5 @@
 #covid-safety-guidelines-status-text-target {
   display: inline-block;
 }
+
```
swirtSJW, line 17: Code style, should be only one trailing empty line.
```suggestion
```

docroot/modules/custom/va_gov_vamc/js/set_covid_term_text.js
```diff
@@ -23,27 +16,22 @@
...
-      window.addEventListener("DOMContentLoaded", textSetter);
-      document.getElementById("group-covid-19-safety-guidelines").addEventListener("click", textSetter);
+      var supplementalStatusChoices = document.querySelectorAll(".form-item--field-supplemental-status [id^='edit-field-supplemental-status-']");
+
+      supplementalStatusChoices.forEach(function (choice) {
+        document.getElementById(choice.id).addEventListener("click", wysiwygSetter);
+      });
     }
   };
 })(Drupal);
```
swirtSJW, line 37: Code style
```suggestion
})(Drupal);

```

swirtSJW (approved):
Outstanding work @omahane   I left only a couple of minor suggestions.  Do not feel you have to act on any of them.

config/sync/field.field.node.health_care_local_facility.field_supplemental_status_more_i.yml
omahane: You don't need that comma as the clause at the end is not an independent one. It has no subject.

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -18,14 +19,14 @@
 use Drupal\va_gov_workflow\Service\Flagger;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
-// The UID of the CMS Help Desk account subscribing to facility messages.
-const USER_CMS_HELP_DESK_NOTIFICATIONS = 4050;
```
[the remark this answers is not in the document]
omahane: Thank you for doing that.

omahane:
@ndouglas I successfully ran Cypress in a terminal in the PR environment:
Screen Shot 2023-02-15 at 6 19 00 AM

ndouglas:
Great news!  If there are any issues with reliability, etc I'll take care of 'em 🙂

docroot/modules/custom/va_gov_vamc/src/EventSubscriber/EntityEventSubscriber.php
ryguyk: I'm curious about this change. Am I understanding correctly that it's simply a matter of not wanting the constant defined outside the class when it is only needed inside the class?

ndouglas: Also that the constant "comes along" with the class/interface if you refer to it from another namespace.  Simpler `use` statement.  And for encapsulation purposes you can alter the visibility of a class constant, but not (I believe) a constant defined outside a class.  And I just like it because then you're still defining one thing inside a file, rather than two things, one of which might easily be missed if you need to copypasta or whatever.

docroot/modules/custom/va_gov_vamc/js/set_covid_term_text.es6.js
```diff
@@ -29,38 +14,35 @@
         statusId = element.value;
       }
     });
```
ryguyk, line 16: **CONSIDER**
It feels like there is a cleaner way to handle getting this `statusId` value:
```
document.querySelector('input[name="field_supplemental_status"]:checked')?.value;
```

**Better yet**, this function `wysiwygSetter` feels like it should be getting the event passed in as a parameter, which would have the clicked option as an attribute:
```
const wysiwygSetter = e => {
   const statusId = e.target.value;
}
```

docroot/modules/custom/va_gov_vamc/js/set_covid_term_text.es6.js
```diff
@@ -29,38 +14,35 @@
...
+    if (
+      document.querySelector(
+        "#cke_edit-field-supplemental-status-more-i-0-value iframe"
+      )
+    ) {
+      iframeDocument = document.querySelector(
+        "#cke_edit-field-supplemental-status-more-i-0-value iframe"
+      ).contentDocument;
+      if (iframeDocument.body) {
+        iframeDocument.body.innerHTML = `<div>${drupalSettings.vamcCovidStatusTermText[statusId].description}</div>`;
```
ryguyk, line 29: **SHOULD**
It seems possible that `statusId` could be undefined. It also seems possible, though less likely, that `vamcCovidStatusTermText` and even `drupalSettings` might be undefined. This might not actually result in any apparent error to the editor, as it'll likely just result in keeping the `innerHTML` property empty, but it'd throw an error in the background and it'd probably be a good idea to account for it.

This can likely be achieved with some optional chaining:
```
iframeDocument.body.innerHTML = `<div>${drupalSettings?.vamcCovidStatusTermText?.[statusId]?.description || ''}
```

ryguyk (changes requested):
I reviewed the JS work. I shared a few thoughts with some small requested changes.

docroot/modules/custom/va_gov_vamc/js/set_covid_term_text.es6.js
ryguyk: Note: With my other suggestions, it seems `statusId` should never be undefined. Still, I think this is good to consider.

omahane: Based on our conversation, as we don't have optional chaining and the mostly likely outcome is going to be some text the editors are going to replace, the checks don't seem worth it.

omahane:
Screen Shot 2023-03-01 at 9 25 41 AM

Tests passing and worked approved.

ElijahLynn:
From [Slack](https://dsva.slack.com/archives/CJT90C0UT/p1677695837594769), this merge is causing tests to fail on STAGING with: 

> Expected to include "Visitors are welcome"

image

PR #12434, opened 2023-02-03
VACMS-12433: Adds Lovell check for exact string to hide systems

omahane:
## Description

Closes #12433

## Testing done

Manually
Using Cypress

## Screenshots
Screen Shot 2023-02-03 at 12 53 37 PM

### As a Lovell Editor 
- [ ] [Log in](https://pr12434-xrwdp08mgsdy3psjepz6prnttvzedovt.ci.cms.va.gov/) as Katie.Yearley@va.gov
- [ ] [Create a "VAMC System Health Service"](https://pr12434-xrwdp08mgsdy3psjepz6prnttvzedovt.ci.cms.va.gov/node/add/regional_health_care_service_des)
- [ ] [Edit a "VAMC System Health Service"](https://pr12434-xrwdp08mgsdy3psjepz6prnttvzedovt.ci.cms.va.gov/node/39217/edit)

### As a Non-Lovell Editor
- [ ] [Log in](https://pr12434-xrwdp08mgsdy3psjepz6prnttvzedovt.ci.cms.va.gov/) as Samuel.Hudson@va.gov
- [ ] [Create a "VAMC System Health Service"](https://pr12434-xrwdp08mgsdy3psjepz6prnttvzedovt.ci.cms.va.gov/node/add/regional_health_care_service_des)
- [ ] [Edit a "VAMC System Health Service"](https://pr12434-xrwdp08mgsdy3psjepz6prnttvzedovt.ci.cms.va.gov/node/53057/edit)

### Definition of Done

- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [x] `Facilities`
- [ ] `User support`

### Lovell editor-facing announcement

Is an announcement needed to let editors know of this change?
- [ ] Yes, an email should be sent to the reporters of the error

Issue #12433, opened 2023-02-03
Lovell editors can not select a System when creating VAMC System Health Service


omahane:
## Describe the defect
When creating or editing a VAMC System Health Service, Lovell editors have no options in the VAMC System dropdown. Because of that, they can not create the service.

## AC / Expected behavior
An editor of a given system, including Lovell, should be able to create or edit a VAMC System Health Service, including choosing the appropriate VAMC System from the VAMC System dropdown.

### As a Lovell Editor
- Create a "VAMC System Health Service"
- Edit a "VAMC System Health Service"

### As a Non-Lovell Editor
- Create a "VAMC System Health Service"
- Edit a "VAMC System Health Service"

## Additional context
Add any other context about the problem here. Reach out to the Product Managers to determine if it should be escalated as critical (prevents users from accomplishing their work with no known workaround and needs to be addressed within 2 business days).

## Labels
(You can delete this section once it's complete)
- [x] Issue type (red) (defaults to "Defect")
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [x] CMS workstream (orange) (not needed for bug tickets)
- [ ] CMS-supported product (black)

### Team
Please check the team(s) that will do this work.

- [ ] `CMS Team`
- [ ] `Public Websites`
- [ ] `Facilities`
- [ ] `User support`

omahane:
From @ryguyk via Slack
> The existing (previous) code was hiding options that included "Lovell Federal health care". That worked before because the other options were `Lovell Federal VA health care` and `Lovell Federal TRICARE health care`, so those names did not include the federal text.

> Now that the new names are `Lovell Federal health care - VA` and `Lovell Federal Health Care - TRICARE`, these names include "Lovell Federal health care" and those names will always be hidden. The updated code changes things to be an exact check, which should be exactly what we want.

PR #12441, opened 2023-02-06
VACMS-12433: Removes Lovell Federal health care from Section dropdown (restored functionality)

omahane:
## Description

Relates to #12433 
I discovered a regression from [my last PR](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/12434). 

"Lovell Federal health care" should not be a selectable option in the Section field in the VAMC Health System Services or the VAMC Facility Health Services.

The Cypress tests should have caught it (had they been written correctly), but they have now been changed to check for what I thought they were checking for in the first place.

Lastly, I have added the checks for the correct Lovell names (which I had commented out in the name change code to get it to pass before the content change).

## Testing done
Manually
Cypress

## Screenshots
Screen Shot 2023-02-06 at 8 25 07 AM

## QA steps
- [x] [Log in](https://pr12441-hp55zz32ssvbgranqumqaryucr5hw74v.ci.cms.va.gov) as Katie.Yearley@va.gov
- [x] [Create a "VAMC System Health Service"](https://pr12441-hp55zz32ssvbgranqumqaryucr5hw74v.ci.cms.va.gov/node/add/regional_health_care_service_des)
- [x] [Edit a "VAMC System Health Service"](https://pr12441-hp55zz32ssvbgranqumqaryucr5hw74v.ci.cms.va.gov/node/39217/edit)
- [x] [Create a "VAMC Facility Health Service"](https://pr12441-hp55zz32ssvbgranqumqaryucr5hw74v.ci.cms.va.gov/node/add/regional_health_care_service_des)
- [x] [Edit a "VAMC Facility Health Service"](https://pr12441-hp55zz32ssvbgranqumqaryucr5hw74v.ci.cms.va.gov/node/39266/edit)
- [x] Validate that the user can create content in the section to which they have access
- [x] Validate that the user can edit content in the section to which they have access

### As a Non-Lovell Editor
- [x] [Log in](https://pr12441-hp55zz32ssvbgranqumqaryucr5hw74v.ci.cms.va.gov) as Samuel.Hudson@va.gov
- [x] [Create a "VAMC System Health Service"](https://pr12441-hp55zz32ssvbgranqumqaryucr5hw74v.ci.cms.va.gov/node/add/regional_health_care_service_des)
- [x] [Edit a "VAMC System Health Service"](https://pr12441-hp55zz32ssvbgranqumqaryucr5hw74v.ci.cms.va.gov/node/53057/edit)
- [x] [Create a "VAMC Facility Health Service"](https://pr12441-hp55zz32ssvbgranqumqaryucr5hw74v.ci.cms.va.gov/node/add/regional_health_care_service_des)
- [x] [Edit a "VAMC Facility Health Service"](https://pr12441-hp55zz32ssvbgranqumqaryucr5hw74v.ci.cms.va.gov/node/53059/edit)
- [x] Validate that the user can create content in the section to which they have access
- [x] Validate that the user can edit content in the section to which they have access

### Definition of Done

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [x] `Facilities`
- [ ] `User support`


ndouglas:
I was actually poking around in this area last week because of a test failure -- I couldn't actually figure out how the test was passing, but it passed _most_ of the time.  "Eh," I thought, "maybe it's because of the database version run for that particular test" or something like that.

How did you develop the test -- did you use `npm test:cypress:interactive`?

docroot/modules/custom/va_gov_lovell/js/limit_lovell.es6.js
```diff
@@ -74,8 +74,12 @@
...
+      // Multiple "-" prefix is used in Drupal menus.
+      // We can't get an exact match with it.
+      const optionsPattern = /^-+/;
       domElement.forEach((i) => {
-        if (i.text === textMatch) {
+        const trimmedText = i.text.replace(optionsPattern, "");
+        if (trimmedText === textMatch) {
           i.classList.add("hidden-option");
         }
       });
```
ryguyk, line 85: **CONSIDER**

While I understand this change and the regression that initiated it, I wonder if it might be more clear to be more explicit with what you're passing into the function rather than changing the function.

```
if (adminFieldOptions) {
	seekHide(adminFieldOptions, `---${lovellFederalText}`);
}
if (regionPageOptions) {
  seekHide(regionPageOptions, lovellFederalText);
}
 ```

That would keep the function more general use, and, in my opinion, easier to understand as well. The prefixed dashes (i.e. `"----"`) are very specific to only one context in which this function is called, so the function feels too tightly coupled to that one context.

If you needed to, eventually, you could perhaps alter the function to accept an array of possible matches. Right now, though, I'm not sure that is needed. 

I do think the proposed solution works, though, so I'm approving. This commentary is intended as a consideration.

**Review by ryguyk — APPROVED** [review]
Approving with one change for consideration.

File: docroot/modules/custom/va_gov_lovell/js/limit_lovell.es6.js (hunk 1 above)
Comment (omahane) [reply docroot/modules/custom/va_gov_lovell/js/limit_lovell.es6.js:85]: @ryguyk I agree about the coupling, and would prefer a more general usage. I am trying to consider that it's possible that items move around in the Drupal menu, causing the number of prefix hyphens to change. I am hesitant to hardcode the "---" for that reason.

**Comment by omahane** [comment]
> I was actually poking around in this area last week because of a test failure -- I couldn't actually figure out how the test was passing, but it passed _most_ of the time. "Eh," I thought, "maybe it's because of the database version run for that particular test" or something like that.
> 
> How did you develop the test -- did you use `npm test:cypress:interactive`?

@ndouglas Yes, I use the interact Cypress test so that it can just keep grinding as I make changes to the test. I am _pretty sure_ I know why it passed. One of the tests has to find some thing, that it has to be visible. 
```
an element with the selector "#edit-field-administration" should be visible
```
Depending on what's happening on the top of the page in the messages section, the menu may be pushed down far enough that the position absolute is set in the css, which is what I was getting as the error here:
```
This element <select#edit-field-administration.form-select.required.form-element.form-element--type-select> is not visible because its ancestor has position: fixed CSS property and it is overflowed by other elements. How about scrolling to the element with cy.scrollIntoView()?
```

I just added a scroll to the bottom, which should solve this.

File: docroot/modules/custom/va_gov_lovell/js/limit_lovell.es6.js (hunk 1 above)
Comment (ryguyk) [reply docroot/modules/custom/va_gov_lovell/js/limit_lovell.es6.js:85]: I hear that, definitely. In that light, I'd like to try to emphasize the general point I made above rather than the specific example. The general point is that I think this function concerns itself too much with outside knowledge/context. I think we should strive to write it in a way that does not concern itself with the specifics of one caller.

As you point out, the way that is suggested above might not work because of the potential for the text match to change. My preference is to try to address that while still adhering to the general principle described above. In short, I think that we would adhere to that principle if we end up with a function that does not have the `/^-+/` pattern hard-coded in its implementation. BUT, _we could pass it in_ :)

```
function seekHide(domElement, regexMatch) {
  domElement.forEach((i) => {
      if (regexMatch.test(i.text)) {
          i.classList.add("hidden-option");
      }
  });
}

if (adminFieldOptions) {
	seekHide(adminFieldOptions, new RegExp(`^-+${lovellFederalText}$`));
}
if (regionPageOptions) {
  seekHide(regionPageOptions, new RegExp(`^${lovellFederalText}$`));
}
```

JayDarnell:
QA steps have passed

Issue #12501, opened 2023-02-07
Make DataDog credentials available in Tugboat environment.


ndouglas:
## Description
I thought this was a thing already, but after poking around in Tugboat and the codebase I can't find any evidence of this.  A lot of support is there, just not the credentials.

This is needed for #12396, and a Good Thing™ for some other things we've casually mentioned in the past like performance monitoring, etc.

## Acceptance Criteria
- [x] DataDog credentials are surfaced in Tugboat PR review environments.

ndouglas:
Completed [here](https://dsva.slack.com/archives/CT4GZBM8F/p1676046035671639)!

PR #12502, opened 2023-02-07
VACMS-12396: Incorporates heading violation check into tugboat content-build.

### LINKED ISSUE #12396: Incorporate heading violation check into Tugboat content-build actions.
## Description

See #12127.

The script exists, but it's not executed regularly.  It should be executed as part of Tugboat content-build processes.

Laura said:

>a report from the day before would be fine.  We are just looking for something repeatable, predictable and historical.  So that we can easily see what impact our own changes have but more importantly when we start to make changes and get our editors involved, we want to have something  so we can say "1 month ago we had XX issues but since we did this, our editors improved and we now only have XX issues"

>The report you gave me the other day, I noticed I haven't been using the HTML as much as I thought.  I think it's because there are just so many I can't even deal with it right now - and because I specifically need to find very very specific violations so I unfortunately just need to look at things.  But I do think it will be useful in the future, but I don't think we need that much more data than what you already provided

Based on this, I think I'll focus on getting this working in Tugboat and creating a known path so that we can access the HTML report easily, but I won't bother with optimizing the HTML report for frequent and wide access or interfering with the regular content-build process.

## Acceptance Criteria
- [ ] The script:
  - [x] creates an HTML report with a fixed location on the server.
  - [x] creates a JSON file with various metrics about the build:
    - [x] total files
    - [x] total files with violations
    - [x] total violations
  - [ ] sends the metrics from the JSON file to DataDog.
  - [ ] provides a plain text summary with URLs, etc.

### Team

- [x] `CMS Team`
- [ ] `Public Websites`
- [ ] `Facilities`
- [ ] `User support`

ndouglas:
## Description

Closes #12396.

## Testing done

None other than iteratively developing and deploying the PR.

## QA Steps

Once automated tests are complete, and the content-build has had time to run:

- [x] view [this file](https://pr12502-bxkzipnsc0ickdebitlsr89bq04ywmdl.ci.cms.va.gov/sites/default/files/build.txt) and verify:
  - [x] The file actually exists.
  - [x] You see the text `==> Done` displayed, indicating that the script completed successfully.
- [x] view [this file](https://pr12502-bxkzipnsc0ickdebitlsr89bq04ywmdl.ci.cms.va.gov/heading_order_violations.html) and verify:
  - [x] The page actually exists.
  - [x] It contains an absolutely rudimentary list of pages (linking to VA.gov) and properly formatted HTML snippets after each item.


tugboat.yml
```diff
@@ -1,6 +1,3 @@
-# If any of these test names get updated they need to be updated here too
-# https://github.com/department-of-veterans-affairs/va.gov-cms/settings/branches
-
```
ndouglas, line 3: This was copypasta'ed from `tests.yml` but didn't need to be.

scripts/build-frontend.sh
```diff
@@ -82,11 +82,25 @@ echo "==> Build complete" >> ${logfile}
 drush va-gov:content-release:advance-state complete
 drush va-gov:content-release:advance-state ready
 
+# After this point, we are less concerned with errors; the build has completed.
+set +e
```
ndouglas, line 86: Problems in the previous versions of the script here were causing problems, as I'll explain.

scripts/build-frontend.sh
```diff
@@ -82,11 +82,25 @@ echo "==> Build complete" >> ${logfile}
...
+
 echo "==> Broken link report" >> ${logfile}
-cat ./vendor/va-gov/content-build/logs/vagovdev-broken-links.json >> ${logfile}
+broken_links_path="${reporoot}/docroot/vendor/va-gov/content-build/logs/vagovdev-broken-links.json"
+cat "${broken_links_path}" | jq >> ${logfile}
```
ndouglas, line 90: 1. This path was incorrect.
2. So there was never a file at that path.
3. So the `cat` command would encounter an error.
4. So the script would terminate early (because `set -e`).
5. So the line that calls it (in `scripts/queue_runner/queue_runner.sh`) would short-circuit:

```bash
[ -f "./docroot/sites/default/files/.buildrequest" ] && ./scripts/build-frontend.sh && rm ./docroot/sites/default/files/.buildrequest
```

6. So the `.buildrequest` file would not be removed.
7. So Tugboat content builds would loop forever (see #12505), at least in some cases.

scripts/build-frontend.sh
```diff
@@ -82,11 +82,25 @@ echo "==> Build complete" >> ${logfile}
...
+
+echo "==> List heading order violations" >> ${logfile}
+pushd ./web
+yarn list-heading-order-violations 2>&1 | grep -vE '^Processing file ' &>> ${logfile}
+cp -v heading_order_violations.html ${reporoot}/docroot/ &>> ${logfile}
+curl -X POST "https://api.ddog-gov.com/api/v1/series" \
+  -H "Content-Type: text/json" \
+  -H "DD-API-KEY: ${HA_HA_HA_NO_DATADOG_KEY_YET}" \
+  -d @- < heading_order_violations.json &>> ${logfile}
+popd
```
ndouglas, line 100: Unfortunately, we're blocked by #12501.  Not a big deal, though.

scripts/build-frontend.sh
```diff
@@ -82,11 +82,25 @@ echo "==> Build complete" >> ${logfile}
...
 
 # Make sure other builds can start.
-rm ${reporoot}/.buildlock
+rm -f ${reporoot}/.buildlock
```
ndouglas, line 103: Don't cause an error if the file is already removed.

edmund-dunn (approved):
Everything is where it should be. Nicely done sir!

PR #12521, opened 2023-02-08
VACMS-0000: Disables failure-on-watchdog-messages in ExistingSiteBase tests.

ndouglas:
Also reënables the TrimNodeWhitespaceTest test, since it's not inherently flawed and is just an innocent bystander in all this.

The query for watchdog messages isn't restricted to the running process, so they will fail any test that runs concurrently with _any_ failure of a concurrent process, e.g. `content-build-gql`.


ElijahLynn (approved):

timcosgrove (approved):
From my visual perspective, all is well.

ndouglas:
Failures are unrelated (Behat reporting deadlock and a JS exception outside of tested code that Cypress treats as an error by default).

PR #12613, opened 2023-02-15
VACMS-12612: Prevent Cypress failing tests on uncaught application exceptions.

### LINKED ISSUE #12612: Tell Cypress to ignore uncaught exceptions in application code.
## Description
The default behavior of Cypress is to fail a test when an exception in application code occurs.  This appeals to me as a way of rooting out bugs that we might not know about.

In practice, this has turned out to be a major pain; people's tests fail, they don't know why, they get anxious and upset, they consult the log/video, and... it was something unrelated.  And it passes on a re-run.  Everyone's time is wasted.

Also, in my experience, these errors that occur in the course of normal use are already reported and triaged via Sentry.  The behavioral tests are a subset of normal usage, and so are inferior and redundant.

Therefore, I should configure Cypress to ignore uncaught application exceptions for the good of all involved.

## Acceptance Criteria
- [ ] Cypress is configured to ignore uncaught application exceptions.

ndouglas:
## Description

Closes #12612.  Uncaught application exceptions should already be logged and reported upstream by Sentry.  Failing whenever an AJAX call fails randomly is frustrating and wastes time and resources.

## QA steps

See [here](https://docs.cypress.io/api/events/catalog-of-events#To-turn-off-all-uncaught-exception-handling).  This is basically a direct copypasta from upstream documentation.


edmund-dunn (approved):
oh thank you!!!!

PR #12649, opened 2023-02-16
VACMS-0000: Updates eslint-plugin-import for great committing success.

ndouglas:
See https://dsva.slack.com/archives/CT4GZBM8F/p1676552855223089


omahane (approved):
Works without issue.

PR #12771, opened 2023-02-27
VACMS-0000: Use substring of generated text in link-text field to avoid running afoul of character limit.

ndouglas:
Rarely, this test would break because there's a 40-character limit on this field, and Faker might generate 41 or 43 😩


edmund-dunn (approved):
nicely done!

swirtSJW (approved):

swirtSJW:
image

PR #12849, opened 2023-03-04
VACMS-0000: account for anchorage vamc name change.

timcosgrove:
## Description

An integration test that depends on the name of the Anchorage VAMC is failing due to a name change of that facility.

> CMS Migrator
> Flagged: Changed name
> Update of Facility API data by migration.
> The title of this facility changed from 'Anchorage VA Medical Center' to 'Colonel Mary Louise Rasmuson Campus of the > Alaska VA Healthcare System'.
> Published

test:
>   # Non-Lovell test
>  Then I select option "---VA Alaska health care" from dropdown "Section"
>  And I wait "2" seconds
>  Then I click the button with selector "#edit-group-health-service-and-facilit"
>  Then I select option "Anchorage VA Medical Center | VA Alaska health care" from dropdown with selector "#edit-field-facility-location"
>  Then I select option "Audiology and speech at VA Alaska health care" from dropdown with selector "#edit-field-regional-health-service"


ndouglas (dismissed):

swirtSJW:
This name change is an odd one and I have a feeling it might shift in the near future when VHA DM looks at it.   My suggestion is we convert this test to not use anything in Alaska.  Alaska is an outpost kind of place that may not fully follow the rules of naming.   I would switch to someplace like Boston or Lebannon PA, which hopefully is a little more established and less likely to change.

swirtSJW:
Failure is related to this name change fix.. which is odd
image

swirtSJW:
I just pushed a change to convert to Boston.

PR #12868, opened 2023-03-07
Revert "bump drupal/tablefield from 2.3.0 to 2.4.0"

swirtSJW:
Reverts department-of-veterans-affairs/va.gov-cms#12781

This is being reverted due to https://dsva.slack.com/archives/CDHBKAL9W/p1678134778777159    There were no db updates as part of the last release so this should be safe.

##QA and testing

- [x] Go to here.  https://web-6m9y4j2xgdtnqvfhwuoncclz1dpnabil.ci.cms.va.gov/education/survivor-dependent-benefits/
- [x] scroll to the first table and validate there are only two columns, not a weird counting column off to the right.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -12411,26 +12411,26 @@
         },
         {
             "name": "drupal/tablefield",
-            "version": "2.4.0",
+            "version": "2.3.0",
             "source": {
                 "type": "git",
                 "url": "https://git.drupalcode.org/project/tablefield.git",
-                "reference": "8.x-2.4"
+                "reference": "8.x-2.3"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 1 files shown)


swirtSJW:
It is possible that the first content build from this will still show the bad table.  It might need a save and publish in the CMS and then another content-build to see if it worked.

https://cms-6m9y4j2xgdtnqvfhwuoncclz1dpnabil.ci.cms.va.gov/education/survivor-dependent-benefits/

swirtSJW:
As I suspected, the new release is not enough to resolve this.
image


I am going to try the editing and re-release to see if it resolves the problem.

swirtSJW:
I edited the table  Had to remove the row by hand,  then published.  Waiting on a content-release.

swirtSJW:
image

PR #12992, opened 2023-03-16
VACMS-0000: Allow newer versions of Node 16 locally.

ndouglas:
Seeing issues attempting to download exactly v16.0.0 of Node.  Felt like it was worth at least opening a PR to check out the latest 16.x.x.

Screenshot 2023-03-16 at 2 39 02 PM

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -322,7 +322,7 @@
         },
         "mouf": {
             "nodejs": {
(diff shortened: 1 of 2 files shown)


edmund-dunn (approved):
fingers crossed

PR #13007, opened 2023-03-17
VACMS-0000: Sort permissions arrays in RolePermissionsTest.

ndouglas:
The sorting here appears to have become nondeterministic recently, which leads to dumb errors and people having to make unrelated changes to get PRs to pass.  As I did recently.  And then the test fails in the opposite way.


ndouglas:
hol up, might be developer error

PR #13016, opened 2023-03-20
VACMS-0000: Removes CodeQL CI test.

ndouglas:
This test is throwing errors because we have the default configuration switched on.  Therefore, we've apparently been running two identical tests for some time 🤷🏻‍♂️


edmund-dunn (approved):
Let's fix this!

PR #13055, opened 2023-03-23
VACMS-0000: Sort permissions arrays before comparison.

### LINKED ISSUE #13057: RolePermissionsTest subject matter is suddenly non-deterministic.
## Description
We have a PHPUnit functional test, `RolePermissionsTest`, which retrieves a list of permissions for a role from the database and compares it to a list of permissions maintained within the test.  At some point, at least two engineers noticed that occasionally some of the permissions checks would fail unexpectedly because the permissions were ordered slightly differently.

I looked at the code pathway for retrieving permissions, and although they're almost always sorted, there is no guarantee that they will be so.  As a result, I'm suggesting that the expected and actual arrays be sorted prior to comparison so that these comparisons be more accurate.

## Acceptance Criteria
- [ ] Expected and actual arrays are sorted alphabetically prior to comparison.

ndouglas:
Closes #13057.

### DIFF
--- a/tests/phpunit/Security/RolesPermissionsTest.php
+++ b/tests/phpunit/Security/RolesPermissionsTest.php
@@ -20,11 +20,13 @@ class RolesPermissionsTest extends VaGovExistingSiteBase {
    * @dataProvider expectedPerms
    */
   public function testSecurityRolesPermissions($roleMatch, $expectedPerms) {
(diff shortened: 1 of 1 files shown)


edmund-dunn (commented):
Clean!

edmund-dunn (approved):
Nice!

PR #13060, opened 2023-03-23
VACMS-13057: Sort permissions arrays, take 2

### LINKED ISSUE #13057: RolePermissionsTest subject matter is suddenly non-deterministic.

ndouglas:
Closes #13057.

I think my previous attempt was causing segmentation faults because the array was passed into the test by reference, then `sort()` performed an in-place sort.


tests/phpunit/Security/RolesPermissionsTest.php
```diff
@@ -30,11 +30,16 @@ public function testSecurityRolesPermissions($roleMatch, $expectedPerms) {
     else {
       $message = 'The ' . $roleMatch . ' role is missing from the system.';
     }
+
```
ndouglas, line 33:
```suggestion

```

tests/phpunit/Security/RolesPermissionsTest.php
```diff
@@ -30,11 +30,15 @@ public function testSecurityRolesPermissions($roleMatch, $expectedPerms) {
     else {
       $message = 'The ' . $roleMatch . ' role is missing from the system.';
     }
+    // Copy $expectedPerms to a new array, then sort it.
+    // Doing this without copying the array can cause segmentation faults.
+    $expected = $expectedPerms;
+    sort($expected);
+
```
ndouglas, line 37:
```suggestion

```

tests/phpunit/Security/RolesPermissionsTest.php
```diff
@@ -30,11 +30,14 @@ public function testSecurityRolesPermissions($roleMatch, $expectedPerms) {
...
+    $expected = $expectedPerms;
+    sort($expected);
+    // Sort $permissions for consistency.
+    sort($permissions);
 
-    // Test assertion.
-    $match = ($expectedPerms == $permissions);
-
-    $this->assertTrue($match, $message);
+    $this->assertEqual($permissions, $expected);
```
timcosgrove, line 40:
```suggestion
    $this->assertEquals($permissions, $expected);
```

timcosgrove (approved):

PR #13151, opened 2023-04-01
VACMS-13150 Remove path alias truncation in core.path

### LINKED ISSUE #13150: Path alias management for Lovell is foiled by truncation
## Description
Trying to investigate or resolve alias issues on prod is hampered by drupal core truncation of aliases in the alias management UI.

image

## Acceptance Criteria
- [ ] When I visit /admin/config/search/path?search=lovell-federal-health-care-tricare/health-services/returning-service-member-care I can see all the aliases in full, so that I can properly make decisions.
- [ ] Patch is contributed upstream https://www.drupal.org/project/drupal/issues/3351638
- [ ] patch is applied

swirtSJW:
## Description

Closes #13150 

## Screenshots
image

## QA steps

As an admin
1. Go to /admin/config/search/path?search=/lovell-federal-health-care-tricare/health-services/returning-service-member
   - [x] Validate that the alias data is not truncated and that you can see the difference in alias between the two results.
2. Go to https://www.drupal.org/project/drupal/issues/3351638 and leave a comment that you reviewed and tested it and that it works.


omahane (approved):
So much better!
Screen Shot 2023-04-03 at 12 51 07 PM

Issue #13141, opened 2023-03-30
Add workflow functionality to VA Benefit Taxonomy


wesrowe:
### User story

**AS AN** Editor
**I WANT** VA Benefit taxonomy terms to have workflow rules in place
**SO THAT** content life cycle is easier to manage.

### Engineering notes / background

**Not launch blocking for merging the Taxonomy to CMS.**

Facilities is working on adding workflows to taxonomies in [this ticket](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/13079). ~~Our work will likely precede theirs.~~ _Update: Facilities went first_

Can't leverage the existing node workflow, but can create a new workflow that mimics it, for this type of entity (taxonomy terms).

## Acceptance criteria

- [x] As a Content Editor 
  - [x] Benefit taxonomy terms follow existing MVP node workflow (draft, needs review, published, archived)
  - [x] Benefit terms require me to enter a revision log message with each save
- [x] I expect as a Content Admin or Drupal Admin to be able to use these workflow steps (Draft, Published, etc) and see Revisions.
- [x] [Any changes required to permissions for Official Name and API Name fields]

### Team
Please check the team(s) that will do this work.

- [ ] `CMS Team`
- [X] `Public Websites`
- [ ] `Facilities`
- [ ] `User support`

davidmpickett:
Retitling this to kill the BCM acronym. The benefit content model includes multiple Drupal entities. The specific entity this ticket will be addressing in the VA Benefit Taxonomy

wesrowe:
Per refinement today, PW will most likely be the first to implement workflow on a taxonomy. 

We also confirmed that this is _**not**_ launch blocking for the MVP. Rationale: the MVP won't lead to content being actually published until there is a FE actually presenting content to Veterans. So this workflow issue is a "launch-blocker" for release of a FE product built upon the content model.

wesrowe:
Note that it sounds like Facilities may beat PW to this after all: [Swirt's comment](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/13079#issuecomment-1509145580).

chri5tia:
Workflow functionality is available now and since revision logs are now _required_, I went ahead and added this. It should be tested as part of the PR review.

chri5tia:
[...]

Issue #13075, opened 2023-03-24
VA Benefits Taxonomy Build Ph4: Complete and merge


wesrowe:
### User story

**AS AN** Editor
**I WANT** benefit data to be structured in the CMS
**SO THAT** I can create content once and have it published everywhere with consistent messaging. (COPE – create once publish everywhere)

**AS AN**  Veteran 
**I WANT** VA benefits to be discussed in any product using the same consistent language 
**SO THAT** I feel reassured that the information is accurate and trustworthy.

### Engineering notes / background
Phase 4 build for the benefits taxonomy includes finishing up any tickets for technical pieces extracted through phase 1-3 that needed a closer look, writing automated tests, fine code review and pushing to production.

### Quality / testing notes
This ticket should include automated tests.
* No circular dependencies on taxonomy entity references
* Title can't be edited by anyone but an Admin
* API field lock behavior

## Acceptance criteria
Phase 4:
- [ ] Final CMS config
- [ ] Automated tests cover listed test cases and are passing
- [ ] PR review and functional testing
- [ ] PR concerns addressed
- [ ] Final PR signed off by Code Reviewers (Daniel, Tim, Steve) and merged
- [ ] Product sign off from: Wes, Dave P., Jordan, before merge
- [ ] PR merged
- [ ] Verified in production: Christia demo to Wes

### Team
Please check the team(s) that will do this work.

- [ ] `CMS Team`
- [X] `Public Websites`
- [ ] `Facilities`
- [ ] `User support`

chri5tia:
Note from phase 1: 

- [ ] Decide strategy for taxonomy name property label change: weight doing a hook form alter for this vs. base field overrides in config (phase 1 closed using hook form alter)

chri5tia:
- [ ] fix unrelated cypress test failure

wesrowe:
@chri5tia, noting here Steve's suggestion in another ticket for regarding moving uniqueness error messages to the fields they're addressing.

wesrowe:
@jilladams, Notes from sync Fri 4/28:
- The oustanding requirement that we need to resolve is the immutability of Official Name and API ID. @chri5tia and I have the impression that the workflow module would enable Content Editors to create new terms, including those 2 required fields. But our model, VA Services, doesn't seem to have resolved this piece of the term creation experience either.

I'm going to slack Steve W about it now.

jilladams:
From discussion with Wes this morning: 
[...]

Issue #13071, opened 2023-03-24
VA Benefits Taxonomy Build Ph 2: Add design elements and prettify per design


wesrowe:
### User story

**AS A** Content Admin
**I WANT** the VA Benefits Taxonomy UI to follow the design
**SO THAT** it is clear and easy for me to enter benefit content.

**AS A** PW developer
**I WANT** to iterate the UI to its designed state
**SO THAT** I and team members can review and identify any iterations that are needed to the product.

### Engineering notes / background

Source of truth for what should be implemented is the [BCM spec](https://docs.google.com/spreadsheets/d/1gP6x-QLFRrJlLWdCEE1COAZXU-kRRmFKSc0sIlc_tyM/edit?usp=share_link) (Google sheet; please request an Excel copy if you need one.)

Follow design created in #12383. 
**MOCKUP:** https://www.figma.com/file/1xUGIKRiuVmYMUoXYTr0BI/Benefit-Content-Model?node-id=6%3A4791&t=eniBhcrYY9Zy61ot-1

### Analytics considerations
n/a

### Quality / testing notes
- Basic QA on this pass
- Detailed code review

## Acceptance criteria

- [ ] UI now follows CMS designs from #12383 
  - [ ] help text added for all fields (per spec)
  - [ ] field/form widget added for all fields (per spec)
  - [ ] text input appropriate per text format field in spec (ie formatter)
  - [ ] inputs laid out per design, e.g. by fieldset
  - [ ] followed considerations in the BCM spec column R, "Notes for Drupal Engineers"
- [ ] CMS and Design roadblocks, changes, or questions uncovered during stage 1 are considered:
  - [ ] discussed in mini-sprints and documented in the ticket
  - [ ] new tickets cut per item discovered if they warrant more than a 30 minute conversation or need to involve more parties (e.g. we can’t make Drupal to do the thing we want)
- [ ] Field config and UI is ready for collab cycle review
- [ ] revision log messaging matches what editors experience on editing nodes (note: workflow moved to [new ticket](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/13141))
- [ ] reviewed, preferably in a group demo meeting, by:
  - [ ] LauraF (508) (Final sign off will be later)
  - [ ] Wes
  - [ ] Jordan (design)
  - [ ] Dave Pickett (content strategy)
- [ ] light code review by one or more of: Daniel, Tony, Swirt (to avoid late surprises with CMS Collab Flow)

[...]

Issue #13070, opened 2023-03-24
VA Benefits Taxonomy Build Ph 1: Implement structure of fields in new taxonomy


wesrowe:
### User story

**AS A** PW developer
**I WANT** to stand up the basic implementation
**SO THAT** I and team members can check for surprises in a review instance.

### Engineering notes / background

Source of truth for what should be implemented is the [VA Benefit Taxonomy spec](https://docs.google.com/spreadsheets/d/1gP6x-QLFRrJlLWdCEE1COAZXU-kRRmFKSc0sIlc_tyM/edit?usp=share_link) (Google sheet; please request an Excel copy if you need one.)

### Analytics considerations
n/a

### Quality / testing notes
No QA, this is a rough pass.

## Acceptance criteria

***The ACs below apply ~~only to rows 4-17~~ to all rows of the spec spreadsheet (the ones finalized in the meetings on 3/29 and 4/6)***

- [x] All fields implemented per structural details in spec (not design or content), specifically:
  - [x] field label
  - [x] machine name
  - [x] field type
  - [x] required
  - [x] cardinality
  - [x] permissions
- [ ] ~~Taxonomy terms implements existing editorial workflow similar to  node content (draft, in review, approved, published, archived)~~ _moved this AC to [Taxonomy workflow ticket](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/13141)_
- [ ] ~~Taxonomy term editing requires revision log message entry in order to save. (similar to node content)~~ _4/6: this AC is also part of the [Taxonomy workflow ticket](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/13141)_
- [ ] reviewed by Jill, Daniel, DaveP
- [ ] shared for high-level review (not detailed design & code review) by: Jordan, Tony, Swirt

NOT in scope:
- rename the term label 
- lock down the api field 
- enforce uniqueness
- enforce that terms can’t be linked to themselves

### Team
Please check the team(s) that will do this work.

- [ ] `CMS Team`
- [X] `Public Websites`
- [ ] `Facilities`
- [ ] `User support`

swirtSJW:
I just updated the ACs to reflect revisions and workflow.

wesrowe:
@swirtSJW, we just discussed internally and decided to punt the workflow ACs to a future ticket so that we can leverage Facilities' work on adding the workflow module to taxonomies. Hopefully that makes sense to you.

jilladams:
[...]

PR #13154, opened 2023-04-03
VACMS-13070: New VA Benefits Taxonomy

chri5tia:
## Description

Relates to #13070
#13071 
#13206 
#13231 
#13141 
#13075 

## Testing done
Manual functional testing.
Automated testing.

## Screenshots
n/a

## QA steps
1. As a user with uid 4186 with user role administrator:
   - [ ] Go to Go to https://cms-i7n92wfzswpe358hivrspihf2ntzshzp.ci.cms.va.gov/admin/structure/taxonomy/manage/va_benefits_taxonomy/add

2. Under **Benefit name**:
   - [x] Enter a test **Official Benefit name** shorter than 255 characters, such as `VA Benefit test 1`; is a required field
   - [x] Enter a random test **VA Benefit API ID** value; is a required field
   - [x] Enter test **Alternate benefit name(s)** with comma separated values; not a required field

3. Under **Benefit Relationships**:
   - [x] Enter a test **Related legislation** external URL, such as `https://va.gov` and Link text; not a required field
   - [x] Click **Add another link**
   - [x] Enter a test **Related legislation** internal URL, such as `Post-9/11 GI Bill monthly housing allowance (47411)` and link text
   - [ ] Search in the **Pre-requisite benefits** field for `VA Benefit Test 2` and add it; not a required field
   - [ ] Search in the **Co-considered benefits** field for `VA Benefit Test 3` and add it; not a required field

4. Under **Benefit Description**:
   - [x] Enter test **Teaser summary** summary plain text. The counter should go down and into the negative after 300 characters; not a required field
   - [x] Enter test **Brief summary** summary rich text limited text. The counter should go down and into the negative after 500 characters; not a required field
   - [x] Enter test **Longer summary** summary rich text limited text, uncounted characters; not a required field
  
5. Under **Benefit Eligibility**:
   - [x] Select one or more checkboxes for **Beneficiaries**; this is a required field
   - [ ] Enter test rich text limited text in **Eligibility summary**, uncounted characters; not a required field

6. Under **How to apply**:
   - [x] Search for a form in **Related VA Form** field, such as `About VA Form 22-8873`; not a required field
   - [x] Enter test rich text limited text in **Preparing to apply**, uncounted; not a required field
   - [x] Enter test rich text limited text in **How to get help with your application**, uncounted; not a required field

7. Under **After you apply summary**:
   - [x] Enter test rich text limited text in **After you apply summary**, uncounted characters; not a required field

8. Under **Editorial workflow**:
   - [x] Select `Published` under **Save as**
   - [x] Enter a test revision log message
   - [x] Click **Save**

9. Logout  
10. As user with uid 4433 with user role content_admin

Go to https://cms-i7n92wfzswpe358hivrspihf2ntzshzp.ci.cms.va.gov/admin/structure/taxonomy/manage/va_benefits_taxonomy/add
   - [x] Validate that you can not edit text in the **Official Benefit name** field or the **VA Benefit API ID** field.

Note: field configs spec to match:
https://docs.google.com/spreadsheets/d/1gP6x-QLFRrJlLWdCEE1COAZXU-kRRmFKSc0sIlc_tyM/edit#gid=0
Columns ABGHIJLMN

### Definition of Done

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/config/sync/core.extension.yml
+++ b/config/sync/core.extension.yml
@@ -234,6 +234,7 @@ module:
   va_gov_api: 0
   va_gov_backend: 0
   va_gov_banner: 0
+  va_gov_benefits: 0
   va_gov_build_trigger: 0
   va_gov_bulk: 0
   va_gov_clp: 0
--- a/config/sync/field.field.taxonomy_term.va_benefits_taxonomy.field_enforce_unique_id.yml
+++ b/config/sync/field.field.taxonomy_term.va_benefits_taxonomy.field_enforce_unique_id.yml
@@ -1,0 +1,46 @@
+uuid: 83c6a8dc-0736-4d68-9f86-bf3bb8713fd3
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.taxonomy_term.field_enforce_unique_id
+    - taxonomy.vocabulary.va_benefits_taxonomy
+  module:
+    - allow_only_one
+    - tmgmt_content
+third_party_settings:
+  allow_only_one:
+    field_eligibility_summary: 0
+    field_va_benefit_after_applying: 0
+    field_va_benefit_api_id: 1
+    field_va_benefit_app_form: 0
+    field_va_benefit_app_help: 0
+    field_va_benefit_app_preparation: 0
+    field_va_benefit_app_type: 0
+    field_va_benefit_beneficiaries: 0
+    field_va_benefit_brief_summary: 0
+    field_va_benefit_coconsidered: 0
+    field_va_benefit_legislation: 0
+    field_va_benefit_long_summary: 0
+    field_va_benefit_plain_name: 0
+    field_va_benefit_prerequisite: 0
+    field_va_benefit_teaser_summary: 0
+    title: 0
+    case_sensitive: '0'
+    limit_validation_to_published: 0
+  tmgmt_content:
+    excluded: false
+id: taxonomy_term.va_benefits_taxonomy.field_enforce_unique_id
+field_name: field_enforce_unique_id
+entity_type: taxonomy_term
+bundle: va_benefits_taxonomy
+label: 'Enforce Unique API ID'
+description: ''
+required: false
+translatable: false
+default_value:
+  -
+    value: 0
+default_value_callback: ''
+settings: {  }
+field_type: allow_only_one
--- a/config/sync/field.field.taxonomy_term.va_benefits_taxonomy.field_enforce_unique_value.yml
+++ b/config/sync/field.field.taxonomy_term.va_benefits_taxonomy.field_enforce_unique_value.yml
@@ -1,0 +1,46 @@
+uuid: 02400cd1-576c-43db-9c35-74b3e6c7a0ef
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.taxonomy_term.field_enforce_unique_value
+    - taxonomy.vocabulary.va_benefits_taxonomy
+  module:
+    - allow_only_one
+    - tmgmt_content
+third_party_settings:
+  allow_only_one:
+    field_eligibility_summary: 0
+    field_va_benefit_after_applying: 0
+    field_va_benefit_api_id: 0
+    field_va_benefit_app_form: 0
+    field_va_benefit_app_help: 0
+    field_va_benefit_app_preparation: 0
+    field_va_benefit_app_type: 0
+    field_va_benefit_beneficiaries: 0
+    field_va_benefit_brief_summary: 0
+    field_va_benefit_coconsidered: 0
+    field_va_benefit_legislation: 0
+    field_va_benefit_long_summary: 0
+    field_va_benefit_plain_name: 0
+    field_va_benefit_prerequisite: 0
+    field_va_benefit_teaser_summary: 0
+    title: 1
+    case_sensitive: '0'
+    limit_validation_to_published: 0
+  tmgmt_content:
+    excluded: false
+id: taxonomy_term.va_benefits_taxonomy.field_enforce_unique_value
+field_name: field_enforce_unique_value
+entity_type: taxonomy_term
+bundle: va_benefits_taxonomy
+label: 'Enforce Unique Benefit'
+description: ''
+required: false
+translatable: false
+default_value:
+  -
+    value: 0
+default_value_callback: ''
+settings: {  }
+field_type: allow_only_one
--- a/docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
+++ b/docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace Drupal\va_gov_benefits\EventSubscriber;
+
+use Drupal\core_event_dispatcher\Event\Form\FormAlterEvent;
+use Drupal\core_event_dispatcher\FormHookEvents;
+use Drupal\Core\Entity\ContentEntityForm;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\va_gov_user\Service\UserPermsService;
+use Drupal\Core\StringTranslation\StringTranslationTrait;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * VA.gov VA Benefits Entity Event Subscriber.
+ */
+class BenefitsSubscriber implements EventSubscriberInterface {
+  use StringTranslationTrait;
+
+  /**
+   * The User Perms Service.
+   *
+   * @var \Drupal\va_gov_user\Service\UserPermsService
+   */
+  protected $userPermsService;
+
+  /**
+   * Constructs the EventSubscriber object.
+   *
+   * @param \Drupal\va_gov_user\Service\UserPermsService $user_perms_service
+   *   The user perms service.
+   */
+  public function __construct(UserPermsService $user_perms_service) {
+    $this->userPermsService = $user_perms_service;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents(): array {
+    return [FormHookEvents::FORM_ALTER => 'formAlter'];
+  }
+
+  /**
+   * Form alter Event call.
+   *
+   * @param \Drupal\core_event_dispatcher\Event\Form\FormAlterEvent $event
+   *   The event.
+   */
+  public function formAlter(FormAlterEvent $event): void {
+    $form = &$event->getForm();
+    $form_state = $event->getFormState();
+    if (!empty($form['#form_id']) && $form['#form_id'] === 'taxonomy_term_va_benefits_taxonomy_form') {
+      $this->protectSelectFields($form, $form_state);
+      $this->renameTitleLabelAndDescription($form);
+    }
+  }
+
+  /**
+   * Lock down certain fields from non-admins.
+   *
+   * @param array $form
+   *   The form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The form state.
+   */
+  public function protectSelectFields(array &$form, FormStateInterface $form_state): void {
+    $form_object = $form_state->getFormObject();
+    if ($form_object instanceof ContentEntityForm) {
+      $bundle = $form_object->getEntity()->bundle();
+      if (!$this->userPermsService->hasAdminRole(TRUE) && $bundle === 'va_benefits_taxonomy') {
+        // Disable Official Benefit name field from non-editors.
+        $form['name']['#disabled'] = TRUE;
+        // Hide API ID field from non-editors.
+        $form['field_va_benefit_api_id']['#access'] = FALSE;
+        // Hide taxonomy Relations field from non-editors.
+        $form['relations']['#access'] = FALSE;
+      }
+    }
+  }
+
+  /**
+   * Renames VA Benefits title label and adds description.
+   *
+   * @param array $form
+   *   The form.
+   */
+  public function renameTitleLabelAndDescription(array &$form): void {
+    $form['name']['widget'][0]['value']['#title'] = 'Official Benefit name';
+    $form['name']['widget'][0]['value']['#description'] = $this->t('The full name of the benefit.');
+  }
+
+}
--- a/docroot/modules/custom/va_gov_benefits/va_gov_benefits.info.yml
+++ b/docroot/modules/custom/va_gov_benefits/va_gov_benefits.info.yml
@@ -1,0 +1,5 @@
+name: 'VA.gov Benefits'
+type: module
+description: 'Tools to improve user experience for VA Benefits.'
+core_version_requirement: ^9 || ^10
+package: 'Custom'
--- a/tests/cypress/integration/common/i_create_a_node.js
+++ b/tests/cypress/integration/common/i_create_a_node.js
@@ -54,7 +54,8 @@ const creators = {
     );
 
     // Hero banner
-    cy.contains("Hero banner").scrollIntoView().click({ force: true });
+    cy.contains("Hero banner").scrollIntoView();
+    cy.contains("Hero banner").click({ force: true });
     cy.contains("Hero banner")
       .parent()
       .then(($el) => {
@@ -77,7 +78,8 @@ const creators = {
     cy.contains("Hero banner").click({ force: true });
 
     // Why this matters
-    cy.contains("Why this matters").scrollIntoView().click({ force: true });
+    cy.contains("Why this matters").scrollIntoView();
+    cy.contains("Why this matters").click({ force: true });
     cy.contains("Why this matters")
       .parent()
       .findAllByLabelText("Introduction")
@@ -87,7 +89,8 @@ const creators = {
     cy.contains("Why this matters").click();
 
     // What you can do
-    cy.contains("What you can do").scrollIntoView().click({ force: true });
+    cy.contains("What you can do").scrollIntoView();
+    cy.contains("What you can do").click({ force: true });
     cy.contains("What you can do")
       .parent()
       .within(() => {
@@ -100,9 +103,12 @@ const creators = {
         cy.findAllByLabelText("Introduction").type(faker.lorem.sentence(), {
           force: true,
         });
-        cy.get("#edit-field-clp-what-you-can-do-promos-actions-ief-add")
-          .scrollIntoView()
-          .click({ force: true });
+        cy.get(
+          "#edit-field-clp-what-you-can-do-promos-actions-ief-add"
+        ).scrollIntoView();
+        cy.get("#edit-field-clp-what-you-can-do-promos-actions-ief-add").click({
+          force: true,
+        });
         cy.contains("Add media").click({ force: true });
       });
     cy.get('div[role="dialog"]').within(() => {
@@ -125,9 +131,10 @@ const creators = {
     cy.contains("What you can do")
       .parent()
       .within(() => {
-        cy.findAllByLabelText("URL")
-          .focus()
-          .type(faker.internet.url(), { force: true });
+        cy.findAllByLabelText("URL").focus();
+        cy.findAllByLabelText("URL").type(faker.internet.url(), {
+          force: true,
+        });
         cy.findAllByLabelText("Link text").type(faker.lorem.sentence(), {
           force: true,
         });
@@ -136,11 +143,13 @@ const creators = {
     cy.contains("What you can do").click();
 
     // VA Benefits
-    cy.contains("VA Benefits").scrollIntoView().click({ force: true });
-    cy.contains("VA Benefits")
+    cy.get("#edit-group-va-benefits").scrollIntoView();
+    cy.get("#edit-group-va-benefits").click({ force: true });
+    cy.get("#edit-group-va-benefits")
       .parent()
       .within(() => {
-        cy.contains("Related benefits").scrollIntoView().click({ force: true });
+        cy.contains("Related benefits").scrollIntoView();
+        cy.contains("Related benefits").click({ force: true });
         cy.contains("Select Benefit Hub(s)").click({ force: true });
       });
     cy.get("iframe.entity-browser-modal-iframe").should("exist");
@@ -157,7 +166,7 @@ const creators = {
         cy.get("#edit-submit").click({ force: true });
       });
     cy.get("iframe.entity-browser-modal-iframe").should("not.exist");
-    cy.contains("VA Benefits").click();
+    cy.get("#edit-group-va-benefits").click();
 
     return cy.wait(1000);
   },
--- a/tests/cypress/integration/common/i_create_a_taxonomy_term.js
+++ b/tests/cypress/integration/common/i_create_a_taxonomy_term.js
@@ -46,6 +46,24 @@ const creators = {
 
     return cy.wait(1000);
   },
+  va_benefits_taxonomy: () => {
+    cy.visit("admin/structure/taxonomy/manage/va_benefits_taxonomy/add");
+    cy.scrollTo("top");
+    cy.findAllByLabelText("Official Benefit name").type(
+      `[Test Data] ${faker.lorem.sentence()}`,
+      { force: true }
+    );
+    cy.get("#edit-field-va-benefit-api-id-0-value").type(
+      faker.datatype.number(),
+      { force: true }
+    );
+
+    cy.get("#edit-moderation-state-0-state").select("published", {
+      force: true,
+    });
+
+    return cy.wait(1000);
+  },
 };
 
 Given("I create a {string} taxonomy term", (vocabulary) => {
--- a/config/sync/core.entity_form_display.message.vamc_outdated_content.default.yml
+++ b/config/sync/core.entity_form_display.message.vamc_outdated_content.default.yml
@@ -1,0 +1,39 @@
+uuid: c4fa806c-d095-41c9-80bb-201cfe68ce98
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.message.vamc_outdated_content.field_subject
+    - message.template.vamc_outdated_content
+id: message.vamc_outdated_content.default
+targetEntityType: message
+bundle: vamc_outdated_content
+mode: default
+content:
+  created:
+    type: datetime_timestamp
+    weight: 10
+    region: content
+    settings: {  }
+    third_party_settings: {  }
+  field_subject:
+    type: string_textfield
+    weight: 11
+    region: content
+    settings:
+      size: 60
+      placeholder: ''
+    third_party_settings: {  }
+    field_name: field_subject
+  uid:
+    type: entity_reference_autocomplete
+    weight: 5
+    region: content
+    settings:
+      match_operator: CONTAINS
+      match_limit: 10
+      size: 60
+      placeholder: ''
+    third_party_settings: {  }
+    '#group': advanced
+hidden: {  }
--- a/config/sync/core.entity_form_display.message.vet_center_outdated_content.default.yml
+++ b/config/sync/core.entity_form_display.message.vet_center_outdated_content.default.yml
@@ -1,0 +1,39 @@
+uuid: c953cf36-d568-4916-9b56-cd91fc31fb00
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.message.vet_center_outdated_content.field_subject
+    - message.template.vet_center_outdated_content
+id: message.vet_center_outdated_content.default
+targetEntityType: message
+bundle: vet_center_outdated_content
+mode: default
+content:
+  created:
+    type: datetime_timestamp
+    weight: 10
+    region: content
+    settings: {  }
+    third_party_settings: {  }
+  field_subject:
+    type: string_textfield
+    weight: 11
+    region: content
+    settings:
+      size: 60
+      placeholder: ''
+    third_party_settings: {  }
+    field_name: field_subject
+  uid:
+    type: entity_reference_autocomplete
+    weight: 5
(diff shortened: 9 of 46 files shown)


chri5tia:
Not sure the PHP failure is: https://tugboat.vfs.va.gov/log/642b271a63b3e8775edf2bc3

swirtSJW:
Converted to draft so it does not get merged until complete.

config/sync/taxonomy.vocabulary.va_benefit_taxonomy.yml
```diff
@@ -0,0 +1,8 @@
+uuid: 81602658-7bb1-40fe-afcc-8466d062694b
+langcode: en
+status: true
+dependencies: {  }
+name: 'VA Benefits Taxonomy'
+vid: va_benefit_taxonomy
+description: 'Single source of truth for VA Benefits including official names, related legislation, related / pre-requisite benefits, applicat'
```
swirtSJW, line 7: If the description is being truncated due to length, we should re-word so it is not cutoff.  If not we should fix wording.

config/sync/syslog.settings.yml
```diff
@@ -0,0 +1,5 @@
+_core:
```
swirtSJW, line 1: Is this related to this issue?

config/sync/password_policy.password_policy.va_gov_cms.yml
```diff
@@ -5,12 +5,9 @@ dependencies: {  }
 id: va_gov_cms
 label: va_gov_cms
 password_reset: 0
-send_reset_email: false
```
swirtSJW, line 8: The changes in this file seem unrelated and out of place.

config/sync/migrate_plus.migration.va_node_health_care_local_facility.yml
```diff
@@ -158,6 +158,12 @@ process:
   field_office_hours:
     plugin: va_field_office_hours
     source: hours
+  field_facility_hours/value:
```
swirtSJW, line 161: These changes are very unexpected and should not be here.  It is reverting changes that were made last week.  I think there may have been a misshap with a rebase.

config/sync/core.extension.yml
```diff
@@ -185,6 +185,7 @@ module:
   post_api: 0
   preprocess_event_dispatcher: 0
   prometheus_exporter: 0
+  raven: 0
```
chri5tia, line 188: Trying to figure out why this is here. Should it be removed?

config/sync/migrate_plus.migration.va_node_health_care_local_facility.yml
chri5tia: Hey there, I didn't even realize you made these comments. I already pushed revised commits, excluding several unexpected and unrelated files. This should likely address a lot of this.

config/sync/password_policy.password_policy.va_gov_cms.yml
chri5tia: Oops, some of these slipped by me. Should be fixed.

chri5tia:
Locked down fields can be seen in the tugboat instance now
Screenshot 2023-04-06 at 8 12 07 PM
 if you log in as a content admin or non admin.

config/sync/taxonomy.vocabulary.va_benefit_taxonomy.yml
chri5tia: For sure. We are working on this, thanks for taking a look.

chri5tia:
Noticing that content that is automatically generated doesn't use rich limited text, so when a non-admin editor goes in, the fields are locked down due. ~~Need to enforce the wywiwyg type and/or figure out how plain text is being used.~~ This seems to be a bug with devel generate. Manually entered fields do not provide another option.

Screenshot 2023-04-07 at 5 54 57 AM

Screenshot 2023-04-07 at 6 06 18 AM

There is also a mysterious failed cypress test relating to campaign pages: https://cms-i7n92wfzswpe358hivrspihf2ntzshzp.ci.cms.va.gov/cypress/screenshots/actual/content_type/campaign_landing_page.feature/Content%20Type%20Campaign%20Landing%20Page%20--%20Log%20in%20and%20create%20a%20campaign_landing_page%20(failed).png

chri5tia:
@swirtSJW @ndouglas Watch at 45 seconds in on this cypress test as it jumps to the taxonomy menu and goes to the new benefits taxonomy instead of running through the items on the campaign landing page. It's weird and kind of hilarious.
https://cms-i7n92wfzswpe358hivrspihf2ntzshzp.ci.cms.va.gov/cypress/videos/content_type/campaign_landing_page.feature.mp4

ndouglas:
Yeah, it looks like it finds a link in the admin toolbar and clicks that instead of whatever is expected.

chri5tia:
@ndouglas 

```
    // VA Benefits
    cy.contains("VA Benefits").scrollIntoView().click({ force: true });
    cy.contains("VA Benefits")
      .parent()
      .within(() => {
        cy.contains("Related benefits").scrollIntoView().click({ force: true });
        cy.contains("Select Benefit Hub(s)").click({ force: true });
      });
```

Maybe we need to remove that second line.

docroot/modules/custom/va_gov_benefits/va_gov_benefits.module
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Implements hook_form_alter().
+ */
+function va_gov_benefits_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  if ($form_id == 'taxonomy_term_va_benefits_taxonomy_form') {
+    $form['name']['widget'][0]['value']['#title'] = 'Official Benefit name';
```
dsasser, line 15: Just a note that this should possible to do with base field overrides, which is config. It is a bit weird to expose the override config. Reading this may be useful https://www.previousnext.com.au/blog/overriding-base-field-labels-and-descriptions-of-drupal-entities-without-custom-code

chri5tia: Thanks for this! I assume this can work for taxonomy and not just content type nodes. But will it be clear and obvious to the next developer who needs to change this? I don't like hard-coding content, but doing it this way feels more hidden. I am willing to hear your thoughts on which is the best case scenario for this project and/or best practice.

docroot/modules/custom/va_gov_benefits/va_gov_benefits.module
```diff
@@ -0,0 +1,19 @@
...
+ * @file
+ * Contains va_gov_benefits.module.
+ */
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Implements hook_form_alter().
+ */
+function va_gov_benefits_form_alter(&$form, FormStateInterface $form_state, $form_id) {
```
swirtSJW, line 13: Is there a reason to have both  a form_alter in the event subscriber and a form_alter here?  If there is not a good reason, it would be helpful to keep them living together.

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
```diff
@@ -0,0 +1,152 @@
...
+   * Form alter Event call.
+   *
+   * @param \Drupal\core_event_dispatcher\Event\Form\FormAlterEvent $event
+   *   The event.
+   */
+  public function formAlter(FormAlterEvent $event): void {
+    $form = &$event->getForm();
+    $form_state = $event->getFormState();
+    $this->lockApiIdEditing($form, $form_state);
+    $this->lockBenefitName($form, $form_state);
```
swirtSJW, line 111: Consider: We are blindly calling these and then checking bundle in each method.   To avoid repeated code can we check the bundle or better yet, the form ID beforehand and then only call the function when needed?  form_alter is pretty generic and gets called a ton of times.

tests/cypress/integration/common/i_create_a_node.js
```diff
@@ -84,7 +84,7 @@ const creators = {
       .type(faker.lorem.sentence(), {
         force: true,
       });
-    cy.contains("Why this matters").click();
+    cy.contains("Why this matters").click({ force: true });
```
swirtSJW, line 87: I'm a little concerned about switching to blanket force: true.   It makes it harder to debug with screenshots and videos that cypress creates because it skips all these. 

> We will NOT perform these:
- Scroll the element into view
- Ensure it is visible
- Ensure it is not disabled
- Ensure it is not detached
- Ensure it is not readonly
- Ensure it is not animating
- Ensure it is not covered
- Fire the event at a descendent

Is there a specific issue we are trying to solve?

chri5tia: I included this to force local cypress test failures to pass, which seem to affect more than one developer. I can remove them. The error it gets in those places is that it can't find the center of the DOM in order to click it because a modal or some other element is in the way.

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
```diff
@@ -0,0 +1,152 @@
...
+    MessengerInterface $messenger,
+    TranslationInterface $string_translation,
+    UserPermsService $user_perms_service
+  ) {
+    $this->currentUser = $currentUser;
+    $this->entityTypeManager = $entityTypeManager;
+    $this->entityFieldManager = $entity_field_manager;
+    $this->messenger = $messenger;
+    $this->stringTranslation = $string_translation;
+    $this->userPermsService = $user_perms_service;
```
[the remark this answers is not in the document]
chri5tia: Thank you for taking a look at that. I assumed they weren't needed but broke things once they were removed. I will focus now on removing them properly instead of whether or not they should be removed.

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
chri5tia: I may need a recommendation for this. That sounds wise.

tests/cypress/integration/common/i_create_a_node.js
```diff
@@ -54,7 +54,8 @@ const creators = {
     );
 
     // Hero banner
-    cy.contains("Hero banner").scrollIntoView().click({ force: true });
+    cy.contains("Hero banner").scrollIntoView();
+    cy.contains("Hero banner").click({ force: true });
```
chri5tia, line 58: Updated chained actions to satisfy new chaining standards, and the linter:
https://docs.cypress.io/guides/core-concepts/retry-ability#Actions-should-be-at-the-end-of-chains-not-the-middle

tests/cypress/integration/common/i_create_a_node.js
```diff
@@ -136,11 +143,13 @@ const creators = {
     cy.contains("What you can do").click();
 
     // VA Benefits
-    cy.contains("VA Benefits").scrollIntoView().click({ force: true });
-    cy.contains("VA Benefits")
+    cy.get("#edit-group-va-benefits").scrollIntoView();
+    cy.get("#edit-group-va-benefits").click({ force: true });
+    cy.get("#edit-group-va-benefits")
```
chri5tia, line 148: Identified element instead of the verbiage, because cypress testing was finding it in the menu (Taxonomy > VA Benefits) and leaving the campaign landing page.

tests/cypress/integration/common/i_create_a_taxonomy_term.js
```diff
@@ -44,6 +44,24 @@ const creators = {
...
+  va_benefits_taxonomy: () => {
+    cy.visit("admin/structure/taxonomy/manage/va_benefits_taxonomy/add");
+    cy.scrollTo("top");
+    cy.findAllByLabelText("Official Benefit name").type(
+      `[Test Data] ${faker.lorem.sentence()}`,
+      { force: true }
+    );
+    cy.get("#edit-field-va-benefit-api-id-0-value").type(
+      faker.datatype.number(),
+      { force: true }
```
chri5tia, line 58: This is a fake number for now. We can use different criteria once there is an AC for how to restrict the entry.

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public function lockApiIdEditing(array &$form, FormStateInterface $form_state) {
+    $form_object = $form_state->getFormObject();
+    $bundle = NULL;
+    if ($form_object instanceof ContentEntityForm) {
+      $bundle = $form_object->getEntity()->bundle();
+    }
+    if (!$this->userPermsService->hasAdminRole(TRUE) && $bundle === 'va_benefits_taxonomy') {
+      $form['field_va_benefit_api_id']['#disabled'] = TRUE;
+    }
```
swirtSJW, line 78: Could: Reduce lines of code and make the logic a little easier to follow.   If the first if-clause never gets acted on, there is no reason to entertain considering the second.  

```suggestion
    if ($form_object instanceof ContentEntityForm) {
      $bundle = $form_object->getEntity()->bundle();
      if (!$this->userPermsService->hasAdminRole(TRUE) && $bundle === 'va_benefits_taxonomy') {
        $form['field_va_benefit_api_id']['#disabled'] = TRUE;
      }
    }
```

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public function lockBenefitName(array &$form, FormStateInterface $form_state) {
+    $form_object = $form_state->getFormObject();
+    $bundle = NULL;
+    if ($form_object instanceof ContentEntityForm) {
+      $bundle = $form_object->getEntity()->bundle();
+    }
+    if (!$this->userPermsService->hasAdminRole(TRUE) && $bundle === 'va_benefits_taxonomy') {
+      $form['name']['#disabled'] = TRUE;
+    }
```
swirtSJW, line 97: Could:  reduce lines of code and unnecesary checks.
```suggestion
    if ($form_object instanceof ContentEntityForm) {
      $bundle = $form_object->getEntity()->bundle();
      if (!$this->userPermsService->hasAdminRole(TRUE) && $bundle === 'va_benefits_taxonomy') {
        $form['name']['#disabled'] = TRUE;
      }
    }
```

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
```diff
@@ -0,0 +1,115 @@
...
+  public function renameTitleLabelAndDescription(array &$form) {
+    $form['name']['widget'][0]['value']['#title'] = 'Official Benefit name';
+    $form['name']['widget'][0]['value']['#description'] = $this->t('The full name of the benefit.');
+  }
+
+  /**
+   * Hides relations portion of the form.
+   */
+  public function hideRelations(array &$form) {
+    $form['relations']['#access'] = FALSE;
```
swirtSJW, line 112: Minor concern about blanket hiding this for everyone.  It ties our hands as admins a bit.  However if we need to make it a structured taxonomy with depth and order we can do it by drag and drop through the term list page, so losing this functionality is not a deal breaker.

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * Renames VA Benefits title label and adds description.
+   */
+  public function renameTitleLabelAndDescription(array &$form) {
+    $form['name']['widget'][0]['value']['#title'] = 'Official Benefit name';
+    $form['name']['widget'][0]['value']['#description'] = $this->t('The full name of the benefit.');
+  }
+
+  /**
+   * Hides relations portion of the form.
```
swirtSJW, line 109: Must: needs a param  defined.

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
```diff
@@ -0,0 +1,115 @@
...
+      $bundle = $form_object->getEntity()->bundle();
+    }
+    if (!$this->userPermsService->hasAdminRole(TRUE) && $bundle === 'va_benefits_taxonomy') {
+      $form['name']['#disabled'] = TRUE;
+    }
+  }
+
+  /**
+   * Renames VA Benefits title label and adds description.
+   */
```
swirtSJW, line 102: Must: Needs a param defined.

docroot/modules/custom/va_gov_benefits/va_gov_benefits.libraries.yml
swirtSJW, line 1: Do we need this file?

tests/cypress/integration/common/i_create_a_node.js
swirtSJW, line 148: Is this line needed?  It doesn't seem to be doing anything.

docroot/modules/custom/va_gov_benefits/va_gov_benefits.info.yml
```diff
@@ -0,0 +1,5 @@
+name: 'VA.gov Benefits'
+type: module
+description: 'Tools to improve user experience for VA Benefits.'
+core_version_requirement: ^8 || ^9
```
swirtSJW, line 4:
```suggestion
core_version_requirement:  ^9 || ^10
```

swirtSJW (changes requested):
Overall this looks great.  A couple "musts" and a couple "coulds".

swirtSJW:
Did design sign off on this?
image

swirtSJW:
Is this field type right?
image

I am trying to figure out how we would use a string of free form text.  It seems like there should be shorter sting field with multiple cardinality so then we have a list we can work with.

config/sync/field.field.taxonomy_term.va_benefits_taxonomy.field_enforce_unique_id.yml
```diff
@@ -0,0 +1,46 @@
...
+    field_va_benefit_brief_summary: 0
+    field_va_benefit_coconsidered: 0
+    field_va_benefit_legislation: 0
+    field_va_benefit_long_summary: 0
+    field_va_benefit_plain_name: 0
+    field_va_benefit_prerequisite: 0
+    field_va_benefit_teaser_summary: 0
+    title: 0
+    case_sensitive: '0'
+    limit_validation_to_published: 1
```
swirtSJW, line 30: Must:  This should be 0.   We don't want editors creating a second one and then not being able to publish.
image

config/sync/field.field.taxonomy_term.va_benefits_taxonomy.field_enforce_unique_value.yml
```diff
@@ -0,0 +1,46 @@
...
+    field_va_benefit_brief_summary: 0
+    field_va_benefit_coconsidered: 0
+    field_va_benefit_legislation: 0
+    field_va_benefit_long_summary: 0
+    field_va_benefit_plain_name: 0
+    field_va_benefit_prerequisite: 0
+    field_va_benefit_teaser_summary: 0
+    title: 1
+    case_sensitive: '0'
+    limit_validation_to_published: 1
```
swirtSJW, line 30: Must:  should be 0
image

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
chri5tia: I would defer to your opinion on this. We want to try to keep the taxonomy flat so it seemed like a good idea to hide unnecessary form components. At the same time, a different approach to making taxonomy flat for this taxonomy type would probably be a better move or hide it only for non-admins. 

Would you suggest that I fix this to hide from non-admins only?

chri5tia:
> Did design sign off on this? image

Verbiage, field types and content model is documented in the the [spec/content model](https://docs.google.com/spreadsheets/d/1gP6x-QLFRrJlLWdCEE1COAZXU-kRRmFKSc0sIlc_tyM/edit#gid=0).

@davidmpickett 
Instead of this help text:

> The "VA Benefit API ID" is used to match the benefit with other external data sources. This value must be exact, or else the desired data will not appear. Once created, this can never be changed (enforced by Drupal).

Can we consider:

"The "VA Benefit API ID" will be used to match the benefit with external data sources. This value must fit specific criteria, [TBD]. Once the API ID is used, is should not be changed. Temporarily, any value will work."

I wonder why we don't disable this field until we have an external data source consuming it and get guidance from those teams for the criteria.

chri5tia:
> Is this field type right? image
> 
> I am trying to figure out how we would use a string of freeform text. It seems like there should be a shorter string field with multiple cardinality so then we have a list we can work with as data.
> 
> If this is the correct field, the help text could be more helpful. Should they separate by commas or new lines?

This field provides information to the end-user only. When someone is looking at the page, they might understand the benefit better under a different name. It will not be referenced in other places.

@davidmpickett 

Instead of the help text

> Add one or more nicknames for the benefit.

Can we consider 

"Add one or more Veteran-facing nicknames for the benefit, separated by commas."

davidmpickett:
## VA Benefit API ID

### Help text
The "VA Benefit API ID" will be used to match the benefit with external data sources. This value must fit specific criteria [TBD]. Once the API ID is published, it should not be changed.

### How to handle governance of this field (for now)
> I wonder why we don't disable this field until we have an external data source consuming it and get guidance from those teams for the criteria.

Yes. I think that it for the period of initial population of content with the Content Team that this should be set to not required and disabled and/or hidden. Right now doing anything with this is just going to cause unnecessary problems. Technically a call for @wesrowe to make though.

We could probably use some more detailed user permission stories for the the VA Benefit Taxonomy similar to [work I am currently doing](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/13272) on Facilities this Sprint. Could be a place for us to collab next sprint?

## Alternate benefit names

### Help Text
> Can we consider
> 
> "Add one or more Veteran-facing nicknames for the benefit, separated by commas."

👍🏻

### Field type questions

@swirtSJW The parallel field on the Service Taxonomy is Common Conditions. A comma-separated list of alternate names for a benefit. 

We could potentially use a different field widget. The single-line text widget is what we use over there. It perhaps gets across the vibe of how much information should be here better.

Screenshot 2023-05-02 175940

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
chri5tia: I went with hiding only for non-admins and added a test. Let me know what you think.

chri5tia:
@davidmpickett I can hide the API field from editors on the back end for now.

chri5tia:
> We could potentially use a different field widget. The single-line text widget is what we use over there. It perhaps gets across the vibe of how much information should be here better.

@davidmpickett I personally prefer the better visibility of a textarea field since there could be more than one nickname at whatever verbose length.

tests/cypress/integration/common/i_create_a_taxonomy_term.js
```diff
@@ -44,6 +44,24 @@ const creators = {
...
+    );
+
+    cy.get("#edit-moderation-state-0-state").select("published", {
+      force: true,
+    });
+
```
chri5tia, line 64: This test makes sure the form still works in general with respect to fields that have custom code vs. the other test, which is for editor experience. We can change the type of faker data when we further restrict the value rules.

dsasser (dismissed):

docroot/modules/custom/va_gov_benefits/va_gov_benefits.info.yml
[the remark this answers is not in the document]
chri5tia: Oops, that should have been updated after Steve pointed it out.

docroot/modules/custom/va_gov_benefits/va_gov_benefits.module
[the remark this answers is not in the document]
chri5tia: Oops, good find.

swirtSJW:
> @swirtSJW The parallel field on the Service Taxonomy is Common Conditions. A comma-separated list of alternate names for a benefit.
> 
> We could potentially use a different field widget. The single-line text widget is what we use over there. It perhaps gets across the vibe of how much information should be here better.

I think we made a mistake on that one making it blobby, which is why I suggest making this not blobby.  We can't make connections or support chat bot or other consumers as well when this is blobby and they are expected to parse their own blobs.

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+   * Locks down Official Benefit name for non-admins.
+   *
+   * @param array $form
+   *   The form.
+   * @param \Drupal\Core\Form\FormStateInterface $form_state
+   *   The form state.
+   */
+  public function lockBenefitName(array &$form, FormStateInterface $form_state) {
```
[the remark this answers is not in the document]
chri5tia: I want to hold on this because everything to do with the API is a bit up in the air. By having it separate, it'll be more straight forward to update later, at which point I should consolidate the code. That's my take on it, but I welcome thoughts.

swirtSJW:
> @swirtSJW It's worth noting that there are other fields that are blobby in this iteration, until we talk to the chatbot team. I like the idea of separating the nicknames for the next phase. We are going to be doing some work to separate components of the eligibility requirements, etc. as well.

Yes, those don't have the hair on the back of my neck standing up the way this one does.  We are taking something that we know is going to have multiple values and are forcing it into a single value field.   We are coming up with a data model that we are going to be feeding to several APIs.   Any one of them might want to have Veterans searching for things that are in common language and not the official term.  I can think of no reason to make that more complicated and have to do a string contains kind of query instead of string equals kind of query.

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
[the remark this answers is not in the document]
chri5tia: Scratch that. Never know when "later" will be.

swirtSJW:
@davidmpickett and @chri5tia 
VBA benefit API id description seems problematic for me because is it trying to serve both admins (the only people who can set it) and Term editors who can not do anything with it.   We are also trying to explain all the rules and where it will be used.  These kind of things raise anxiety.  I suggest keeping it more generic.

Something like
The "VA Benefit API ID" is the unique identifier for this benefit.  Once it is established, it can not be changed.

wesrowe:
> > @swirtSJW It's worth noting that there are other fields that are blobby in this iteration, until we talk to the chatbot team. I like the idea of separating the nicknames for the next phase. We are going to be doing some work to separate components of the eligibility requirements, etc. as well.
> 
> Yes, those don't have the hair on the back of my neck standing up the way this one does. We are taking something that we know is going to have multiple values and are forcing it into a single value field. We are coming up with a data model that we are going to be feeding to several APIs. Any one of them might want to have Veterans searching for things that are in common language and not the official term. I can think of no reason to make that more complicated and have to do a string contains kind of query instead of string equals kind of query.

@swirtSJW, the decision to create a comma-separated field was discussed at some length in the MVP finalization meeting with Dave and Danielle. There were several drivers, including that no one knows how a field like this will be used by the Content team or consuming products. So the intent is to use this as a place to jot notes to drive post-MVP iteration. I don't think anyone intends for products to consume this field as-is. We should leave it as a single, comma-separated field as designed, as that's what stakeholders who are closest to the content are expecting and asked for.

docroot/modules/custom/va_gov_benefits/src/EventSubscriber/BenefitsSubscriber.php
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+   * Constructs the EventSubscriber object.
+   *
+   * @param \Drupal\va_gov_user\Service\UserPermsService $user_perms_service
+   *   The user perms service.
+   */
+  public function __construct(
+    UserPermsService $user_perms_service
+  ) {
```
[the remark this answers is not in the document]
chri5tia: Can't declare a return type for constructor. There is a php bug report for it, as others agree that it seems important to be able to declare `void` here; they marked it as "not a bug":
https://bugs.php.net/bug.php?id=75263

Trying to do so returns an error. 

```
  public function __construct(UserPermsService $user_perms_service): void {
    $this->userPermsService = $user_perms_service;
  }
```

```
Fatal error: Method Drupal\va_gov_benefits\EventSubscriber\BenefitsSubscriber::__construct() cannot declare a return type
```

Is there a way that you have dealt with this?

dsasser (dismissed):

jilladams:
Touched base with Christia in 1:1. I enabled automerge, since: Daniel approved before the last push. This way, when he re-approves, this will merge. @swirtSJW happy to talk through any concerns you have re: follow up tickets that we should create potentially.

swirtSJW (approved):
Works as intended.  Nice work Christia.

I included a QA check to make sure that unique applied to title.  It worked great.
image

swirtSJW:
image

PR #13524, opened 2023-04-27
Revert "Bump trim, @storybook/addon-essentials and @storybook/html in /docroot/design-system"

ndouglas:
Reverts department-of-veterans-affairs/va.gov-cms#13493.

Seems to have broken the Storybook build process.


omahane (approved):
Revert away.

PR #13562, opened 2023-05-02
VACMS-13431 Restore Lovell menu

### LINKED ISSUE #13431: Lovell menu missing in CMS
## Describe the defect
When visiting any lovell page in the CMS there should be a sidebar menu visible on the left.

We have narrowed it down to being missing for at least a month and not  more than 4 months.

[Slack details](https://dsva.slack.com/archives/C0FQSS30V/p1682086518104549)

## To Reproduce
1. In the CMS go to '/lovell-federal-health-care-tricare', or '/lovell-federal-health-care-va' , or '/lovell-federal-health-care'
2. You should see the system menu on the left.

## AC / Expected behavior
- [x] I expect the lovell menu to appear on all lovell pages 
- [x] I expect there is a test that covers this (cypress test)

## Additional context
Add any other context about the problem here. Reach out to the Product Managers to determine if it should be escalated as critical (prevents users from accomplishing their work with no known workaround and needs to be addressed within 2 business days).

## Labels
- [ ] CMS subsystem (green)

### Team

- [ ] `CMS Team`
- [ ] `Public Websites`
- [ ] `Facilities`
- [ ] `User support`

swirtSJW:
## Description

Closes #13431

## QA steps

As an anon user
1. go here https://pr13562-xku3f71uyf1vcgobwdstsgk2glihmthc.ci.cms.va.gov/lovell-federal-health-care
   - [x] Validate that sidebar menu shows
1. go here https://pr13562-xku3f71uyf1vcgobwdstsgk2glihmthc.ci.cms.va.gov/lovell-federal-health-care-tricare
   - [x] Validate that sidebar menu shows
1. go here https://pr13562-xku3f71uyf1vcgobwdstsgk2glihmthc.ci.cms.va.gov/lovell-federal-health-care-va
   - [x] Validate that sidebar menu shows
1. go here https://pr13562-xku3f71uyf1vcgobwdstsgk2glihmthc.ci.cms.va.gov/lovell-federal-health-care-va/work-with-us
   - [x] validate that sidebar menu shows.

### DIFF
--- a/docroot/modules/custom/va_gov_lovell/src/LovellMenuLinkTreeManipulators.php
+++ b/docroot/modules/custom/va_gov_lovell/src/LovellMenuLinkTreeManipulators.php
@@ -117,8 +117,7 @@ class LovellMenuLinkTreeManipulators {
    */
   protected function setMenuEntities(array $tree) {
     if (empty($this->menuEntities)) {
-      $ids = [];
-      $this->getIdsFromTree($ids, $tree);
+      $ids = $this->getIdsFromTree($tree);
       if (!empty($ids) && is_array($ids)) {
         $this->menuEntities = $this->entityTypeManager->getStorage('menu_link_content')->loadMultiple($ids);
       }
@@ -128,21 +127,22 @@ class LovellMenuLinkTreeManipulators {
   /**
    * Traverse a menu tree and extract all menu entity ids.
    *
-   * @param array $ids
-   *   An array of menu ids accumulated by reference.
    * @param \Drupal\Core\Menu\MenuLinkTreeElement[] $tree
    *   An array of MenuLinkTreeElements.
    *
-   * @return void
+   * @return array
+   *   An array of menu ids accumulated from all recursion of tree.
    */
-  protected function getIdsFromTree(array &$ids, array $tree): void {
+  protected function getIdsFromTree(array $tree): array {
+    $ids = [];
     foreach ($tree as $branch) {
       $menu_item = $this->accessProtected($branch->link, 'entity');
       $ids[] = $menu_item->id();
       if (!empty($branch->subtree)) {
-        $this->getIdsFromTree($ids, $branch->subtree);
+        $ids = array_merge($ids, $this->getIdsFromTree($branch->subtree));
       }
     }
+    return $ids;
   }
 
   /**
--- a/docroot/modules/custom/va_gov_lovell/src/LovellMenuLinkTreeManipulators.php
+++ b/docroot/modules/custom/va_gov_lovell/src/LovellMenuLinkTreeManipulators.php
@@ -117,11 +117,30 @@ class LovellMenuLinkTreeManipulators {
    */
   protected function setMenuEntities(array $tree) {
     if (empty($this->menuEntities)) {
-      $element = reset($tree);
-      $link = $element->link;
-      $ids = $this->accessProtected($link, 'entityIdsToLoad');
+      $ids = [];
+      $this->getIdsFromTree($ids, $tree);
       if (!empty($ids) && is_array($ids)) {
-        $this->menuEntities = $this->entityTypeManager->getStorage('menu_link_content')->loadMultiple(array_values($ids));
+        $this->menuEntities = $this->entityTypeManager->getStorage('menu_link_content')->loadMultiple($ids);
+      }
+    }
+  }
+
+  /**
+   * Traverse a menu tree and extract all menu entity ids.
+   *
+   * @param array $ids
+   *   An array of menu ids accumulated by reference.
+   * @param \Drupal\Core\Menu\MenuLinkTreeElement[] $tree
+   *   An array of MenuLinkTreeElements.
+   *
+   * @return void
+   */
+  protected function getIdsFromTree(array &$ids, array $tree): void {
+    foreach ($tree as $branch) {
+      $menu_item = $this->accessProtected($branch->link, 'entity');
+      $ids[] = $menu_item->id();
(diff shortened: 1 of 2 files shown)


docroot/modules/custom/va_gov_lovell/src/LovellMenuLinkTreeManipulators.php
```diff
@@ -117,11 +117,30 @@ public function reduceLovellMenu(array $tree, $lovell_type, $firstrun = TRUE) {
    */
   protected function setMenuEntities(array $tree) {
     if (empty($this->menuEntities)) {
-      $element = reset($tree);
-      $link = $element->link;
-      $ids = $this->accessProtected($link, 'entityIdsToLoad');
```
swirtSJW, line 122: When Translation was enabled, 'entityIdsToLoad' went from having 73 menu ids in it, to being empty.  :(

docroot/modules/custom/va_gov_lovell/src/LovellMenuLinkTreeManipulators.php
```diff
@@ -117,11 +117,30 @@ public function reduceLovellMenu(array $tree, $lovell_type, $firstrun = TRUE) {
...
+      $ids = [];
+      $this->getIdsFromTree($ids, $tree);
```
omahane, line 121:
```suggestion
      $ids = $this->getIdsFromTree($tree);
```

docroot/modules/custom/va_gov_lovell/src/LovellMenuLinkTreeManipulators.php
```diff
@@ -117,11 +117,30 @@ public function reduceLovellMenu(array $tree, $lovell_type, $firstrun = TRUE) {
...
+        $this->menuEntities = $this->entityTypeManager->getStorage('menu_link_content')->loadMultiple($ids);
+      }
+    }
+  }
+
+  /**
+   * Traverse a menu tree and extract all menu entity ids.
+   *
+   * @param array $ids
+   *   An array of menu ids accumulated by reference.
```
omahane, line 132:
```suggestion
```

docroot/modules/custom/va_gov_lovell/src/LovellMenuLinkTreeManipulators.php
```diff
@@ -117,11 +117,30 @@ public function reduceLovellMenu(array $tree, $lovell_type, $firstrun = TRUE) {
...
+   * @param \Drupal\Core\Menu\MenuLinkTreeElement[] $tree
+   *   An array of MenuLinkTreeElements.
+   *
+   * @return void
```
omahane, line 136:
```suggestion
   * @return array
   *   An array of menu ids.
```

docroot/modules/custom/va_gov_lovell/src/LovellMenuLinkTreeManipulators.php
```diff
@@ -117,11 +117,30 @@ public function reduceLovellMenu(array $tree, $lovell_type, $firstrun = TRUE) {
...
+   */
+  protected function getIdsFromTree(array &$ids, array $tree): void {
```
omahane, line 138:
```suggestion
  protected function getIdsFromTree(array $tree): array {
```

docroot/modules/custom/va_gov_lovell/src/LovellMenuLinkTreeManipulators.php
```diff
@@ -117,11 +117,30 @@ public function reduceLovellMenu(array $tree, $lovell_type, $firstrun = TRUE) {
...
+    foreach ($tree as $branch) {
+      $menu_item = $this->accessProtected($branch->link, 'entity');
+      $ids[] = $menu_item->id();
+      if (!empty($branch->subtree)) {
+        $this->getIdsFromTree($ids, $branch->subtree);
       }
     }
```
omahane, line 145:
```suggestion
    $ids = [];
    foreach ($tree as $branch) {
      $menu_item = $this->accessProtected($branch->link, 'entity');
      $ids[] = $menu_item->id();
      if (!empty($branch->subtree)) {
        $leaf = $this->getIdsFromTree($branch->subtree);
            foreach ($leaf as $leaflet) {
                array_push($ids, $leaflet);
            }
        $this->getIdsFromTree($branch->subtree);
      }
    }
    return $ids;
```

omahane (commented):
@swirtSJW This works great. It's good to have our old menu back. 

I was wondering if we can make this function not have any side effects by returning the array instead of manipulating it by reference. But I am not sure about the performance hit I am might be introducing this way.

docroot/modules/custom/va_gov_lovell/src/LovellMenuLinkTreeManipulators.php
swirtSJW: This seems like a recursive anti-pattern.  It is doing partial recursion into the leaflet  but not just simple recursion on what is present.  Then on 148 it is not catching the  array returned by getIdsFromTree().  so we are losing picking up the recursed values.   This is a lot of work to make this a pure function.  But then we are using array_push() which itself is altering the array by reference. 

Performance wise if we were always catching the returned array, it would work, but now we have instances of the array in multiple places (because arrays are not referenced the way an object is) where as with doing it by reference, we only ever have one array. So it is a bit more performant.

omahane (approved):
Looks great and still works. Thanks for indulging my pure function predilection.

swirtSJW:
image

PR #13569, opened 2023-05-02
VACMS-0000: Fixes status-error test comment quoting.

ndouglas:
Quoting was broken, giving an unhelpful comment when this test failed.


ndouglas:
I altered the script live _only_ to include the classes and severities of issues that are normally excluded so that we could get an idea of what this looks like when it fires.

It's not pretty, but it's at least readable and quoted correctly 🤷🏻‍♂️  Unfortunately, anything deeper is going to require non-trivial work, e.g. to parse CSV (which in limited testing appears to be formatted improperly) or TSV or JSON and create a good Markdown table from it.

ndouglas:
This looks like it failed because of the intentional failure I introduced to show the comment above.

edmund-dunn (approved):
Nicely done sir!

Issue #13780, opened 2023-05-22
Demo Feedback: Conflicting item on job check out page


productmike:
## User story / Problem Statement
When creating translation jobs, if the user adds a piece of content for translation that has already been added to another translation job, they will see a message like the following:

[...]

PR #13894, opened 2023-05-30
Set up UX languages changes for TMGMT.

timcosgrove:
## Description

#13780 

## Testing done
Same as QA steps.

## QA steps

1. Log in as admin.
2. Visit https://pr13894-swyq1xt9wttom9yleleqcke7jeuka3rv.ci.cms.va.gov/admin/tmgmt/sources.
3. Select any two items, select 'Spanish' as the target language, and hit 'Request translation'.
4. On the next page, hit 'Submit to provider'. You'll be taken back to the Sources page.
  - [x] The items you selected in 3 should have blue hourglasses in the Spanish column on the right.
5. Select *one* of the two items from steps 3/4. as well as 1 or more additional 'fresh' items. Select 'Spanish' as the target language and hit 'Request translation'.
  - [x] On the next page, at the top, you should see a message 'The translation for the following item is already in progress, so it will not be added to this job: `already translated item`'
  - [x] The message should be formatted with the CMS-standard 'warning' style. 
6. Hit 'Submit to provider'. You'll be taken back to Sources page.
  - [x]  The warning message at top should include '1 item is already in translation and was not included in the job.'.
  - [x] Note: you will also see the message from the previous step. This is known and proving extremely difficult to prevent and a low impact non-blocking bug.
7. Repeat the process, but select *2 or more* items that have the blue hour glass on the right, in addition to fresh items. Hit 'Submit to provider'.
  - [x] The warning messages should include 'The translation for the following `number` items is already in progress, so they will not be added to this job: `items`' 
  - [x] After submission, the warning message should include "The translation for the following `number` items is already in progress, so they will not be added to this job: `items`'
8. Repeat the process, but this time select *only* items that are already in progress.
  - [x] When you hit 'Submit to provider', you should receive a form error: All items in this job are already being translated. This job cannot be submitted.

### Select Team for PR review

- [x] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -507,7 +507,7 @@
                 "3272128 - Calling a static method, or accessing a static property directly on a trait is deprecated": "https://www.drupal.org/files/issues/2022-04-27/deprecated-static-function-call-in-trait-3272128.patch"
             },
             "drupal/tmgmt": {
-                "3316264 - Empty lines on job overview": "https://www.drupal.org/files/issues/2022-10-19/tmgmt_empty_job_lines.patch"
+                "UX language updates for TMGMT": "patches/tmgmt_language_changes.patch"
             },
             "drupal/viewfield": {
                 "3252356 - Add option to display default view on node edit when always use default value selected": "https://www.drupal.org/files/issues/2023-02-07/viewfield-option-display-default-view-on-node-edit-3252356-5.patch"
--- a/docroot/modules/custom/va_gov_multilingual/modules/va_gov_multilingual_tmgmt/src/EventSubscriber/FormEventSubscriber.php
+++ b/docroot/modules/custom/va_gov_multilingual/modules/va_gov_multilingual_tmgmt/src/EventSubscriber/FormEventSubscriber.php
@@ -57,24 +57,26 @@ class FormEventSubscriber implements EventSubscriberInterface {
       /** @var Drupal\tmgmt\Form\JobForm $form_object */
       $form_object = $event->getFormState()->getFormObject();
       $job = $form_object->getEntity();
-      $conflicting_items_by_item = $job->getConflictingItems();
-      $form = &$event->getForm();
-      $form['message'] = [];
-      if (count($conflicting_items_by_item)) {
-        // Make a list of links to the existing items causing the conflicts.
-        $conflicts = [];
-        $num_of_existing_items = 0;
-        foreach ($conflicting_items_by_item as $conflicting_items) {
-          $num_of_existing_items += count($conflicting_items);
-          foreach (JobItem::loadMultiple($conflicting_items) as $id => $conflicting_item) {
-            $conflicts[] = Link::createFromRoute($conflicting_item->label(), 'entity.tmgmt_job_item.canonical', ['tmgmt_job_item' => $id])->toString();
+      if (!$job->isContinuous()) {
+        $conflicting_items_by_item = $job->getConflictingItems();
+        $form = &$event->getForm();
+        $form['message'] = [];
+        if (count($conflicting_items_by_item)) {
+          // Make a list of links to the existing items causing the conflicts.
+          $conflicts = [];
+          $num_of_existing_items = 0;
+          foreach ($conflicting_items_by_item as $conflicting_items) {
+            $num_of_existing_items += count($conflicting_items);
+            foreach (JobItem::loadMultiple($conflicting_items) as $id => $conflicting_item) {
+              $conflicts[] = Link::createFromRoute($conflicting_item->label(), 'entity.tmgmt_job_item.canonical', ['tmgmt_job_item' => $id])->toString();
+            }
           }
+          // Make the links usable in formatPlural().
+          $conflict_list = Markup::create(implode(', ', $conflicts));
+          $message = $this->translationManager
+            ->formatPlural($num_of_existing_items, $this::CONFLICTING_ITEMS_MESSAGE_SINGLE, $this::CONFLICTING_ITEMS_MESSAGE_PLURAL, ['@items' => $conflict_list]);
+          $this->messenger->addWarning($message);
         }
-        // Make the links usable in formatPlural().
-        $conflict_list = Markup::create(implode(', ', $conflicts));
-        $message = $this->translationManager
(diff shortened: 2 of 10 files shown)


composer.json
```diff
@@ -507,7 +507,7 @@
                 "3272128 - Calling a static method, or accessing a static property directly on a trait is deprecated": "https://www.drupal.org/files/issues/2022-04-27/deprecated-static-function-call-in-trait-3272128.patch"
             },
             "drupal/tmgmt": {
-                "3316264 - Empty lines on job overview": "https://www.drupal.org/files/issues/2022-10-19/tmgmt_empty_job_lines.patch"
```
timcosgrove, line 510: Version bump, patch was incorporated.

docroot/modules/custom/va_gov_multilingual/modules/va_gov_multilingual_tmgmt/src/EventSubscriber/FormEventSubscriber.php
```diff
@@ -0,0 +1,93 @@
+<?php
```
timcosgrove, line 1: This whole thing is to replace code in TMGMT that looks for items that are already in translation and removes them from the translation job you're working on: https://git.drupalcode.org/project/tmgmt/-/blob/8.x-1.x/src/Form/JobForm.php#L274

We do this for two reasons:
* To change the language to be less confusing
* To make the duplicate item messages a proper message.

Possible this could just be in the patch.

docroot/modules/custom/va_gov_multilingual/va_gov_multilingual.info.yml
```diff
@@ -0,0 +1,5 @@
+name: VA.gov multilingual
```
timcosgrove, line 1: Nothing to see here at the moment but I expect it will be used eventually.

patches/tmgmt_language_changes.patch
```diff
@@ -0,0 +1,33 @@
+diff --git a/src/Form/JobForm.php b/src/Form/JobForm.php
+index 8c9396ee..eb6d1beb 100644
+--- a/src/Form/JobForm.php
++++ b/src/Form/JobForm.php
+@@ -640,7 +640,7 @@ class JobForm extends TmgmtFormBase {
+       if ($existing_items_ids = $form_state->get('existing_item_ids')) {
+         $storage->delete($storage->loadMultiple($existing_items_ids));
+         $num_of_items = count($existing_items_ids);
+-        $this->messenger()->addWarning(\Drupal::translation()->formatPlural($num_of_items, '1 conflicting item has been dropped.', '@count conflicting items have been dropped.'));
++        $this->messenger()->addWarning(\Drupal::translation()->formatPlural($num_of_items, '1 item is already in translation and was not included in the job.', '@count items are already in translation and were not included in the job.'));
```
timcosgrove, line 10: Language replacements. There's no other way to change this text that I'm aware of.

docroot/modules/custom/va_gov_multilingual/modules/va_gov_multilingual_tmgmt/src/EventSubscriber/FormEventSubscriber.php
```diff
@@ -0,0 +1,93 @@
...
+   */
+  public function formAlter(FormAlterEvent $event) {
+    if ($event->getFormId() === 'tmgmt_job_edit_form') {
+      /** @var Drupal\tmgmt\Form\JobForm $form_object */
+      $form_object = $event->getFormState()->getFormObject();
+      $job = $form_object->getEntity();
+      if (!$job->isContinuous()) {
+        $conflicting_items_by_item = $job->getConflictingItems();
+        $form = &$event->getForm();
+        $form['message'] = [];
```
timcosgrove, line 63: This is TMGMT's custom location for the message, which we're replacing. I suspect they did this for exactly the reasons I describe above.

docroot/modules/custom/va_gov_multilingual/modules/va_gov_multilingual_tmgmt/src/EventSubscriber/FormEventSubscriber.php
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+   * Form alters for va_gov_home.
+   *
+   * @param \Drupal\core_event_dispatcher\Event\Form\FormAlterEvent $event
+   *   The form event.
+   */
+  public function formAlter(FormAlterEvent $event) {
+    if ($event->getFormId() === 'tmgmt_job_edit_form') {
+      /** @var Drupal\tmgmt\Form\JobForm $form_object */
```
timcosgrove, line 57: What we are doing here is replacing TMGMT's version of the warning around duplicate items being dropped from a new translation job. They use alarming language, weird non-standard formatting, and place it incorrectly for a warning message.

Everything we're doing in the form alter is to send the information to a standard warning message. This itself presents a problem. Ideally we would only want this message to show when the user is looking at the job form. However, since we do this in formAlter, and since formAlter is invoked prlor to form build, which is called during the submission process, we cannot prevent the message from showing on the success page, after form submission is complete. 

I tried doing this in a `#pre_render` callback, with some success in terms of when a message would show. However, `#pre_render` operates on a render array, which means we don't have access to the form, form_state, or the TMGMT job objects. We need the job object to construct the actual message of the text.

ndouglas: Out of curiosity, why not make what you've written above a comment? It's fantastic context.

docroot/modules/custom/va_gov_multilingual/modules/va_gov_multilingual_tmgmt/src/EventSubscriber/FormEventSubscriber.php
ndouglas: looking at that `form()` method like
dust off this old gem

ndouglas (dismissed):
All steps passed QA process 🎉

docroot/modules/custom/va_gov_multilingual/modules/va_gov_multilingual_tmgmt/src/EventSubscriber/FormEventSubscriber.php
timcosgrove: Because you'll have to re-review.

Issue #13948, opened 2023-05-31
Capture events/metrics about general CMS product delivery performance.


ndouglas:
## Description
There are some metrics that we can and should be capturing for CMS product delivery performance, but that I don't believe we are at present.

We should ensure that these metrics are being recorded and reported upstream to Datadog.

These involve changes to the BRD CD pipeline, so each of these might be nontrivial and need to be split off into a separate issue.

## Events
- **A**: PR merged to `main` -> commit, timestamp
- **B**: staging post-deploy tests started -> commit, timestamp
- **C**: staging post-deploy tests succeeded -> commit, timestamp
- **D**: staging post-deploy tests failed -> commit, timestamp
- **E**: release created -> list of commits, timestamp?

## Metrics
- Deployment Lead Time - **E - A**
- Delivery Lead Time - **C - A**
- Change Failure Rate - **D / B** or **( B - C ) / B**
- Mean Time to Restore - difference of the time between a change failure and the next subsequent success? because of sporadic commit schedule this might be expressed better as number of subsequent failures.
- Deployment Frequency - this is kinda boring because we do 1 per day, and I'm not comfortable escalating that for Drupal in the near future. But maybe 👉🏻 👈🏻 🥺 what if we kissed by the continuous deployment banner?

## Acceptance Criteria
- [ ] Above events/metrics are reported to Datadog

productmike:
@ndouglas @olivereri I'm good to move forward with this in a hypothesis manner (e.g. we think these metrics will best represent a good 1st slice of measurements based on what we know now). Who all has access to datadog and/or how can we best socialize these measurements in an ongoing manner (once we feel comfortable they are accurate and not especially negative). Though we didn't have a chance to refine together (pretty much only story pointing left), moving into STRETCH for Nate to tear into when he's back next week.

ndouglas:
I was intending to use events for A, B, C, D, E, then use those events to calculate metrics. I don't think that's going to work well; it would seem to require that I set a tag value for the event to the commit SHA, which would then cause the custom metrics billing to scale according to the number of commits we make! This would be very wasteful financially.

After some thinking, I decided to just use the commit timestamp as stored in the Git history as the start point and compute dates relative to that for each subsequent step.

I don't think this actually changes anything, but wanted to note this for future reference. If we create similar issues in the future, we should beware of this cost scaling complication.

productmike:
[...]

PR #14089, opened 2023-06-15
VACMS-13948: Record continuous delivery metrics for CMS.

ndouglas:
## Description

See #13948. 

This just adds an app that is executed in the CD pipeline, so this shouldn't require much QA.


ndouglas:
Failure caused by a 502 :(

olivereri (approved):

PR #14135, opened 2023-06-22
VACMS-0000: A small change to the CD metrics script.

ndouglas:
This script alters the Time-to-Restore calculation algorithm.

Previously, we assumed we had been in a failure condition if and only if the previous commit's combined status was `"failure"`.

However, this breaks under cases like the following:

Screenshot 2023-06-22 at 11 26 03 AM

Here, we enter failure mode at commit `9ab040a`. The following commit (`055b532`) only changes documentation, and therefore no tests are run. Then `7d51500` runs tests and they succeed.

The correct failure period was `9ab040a..7d51500`, but this failure period is never reported.

This PR updates this logic to loop through previous commits until a non-pending commit is found.

I'm slightly concerned that there might be a pathological state where GitHub indicates _all_ commits have a status of "pending". As a result, I've added some logic to escape the loop after 25 "pending" statuses. This shouldn't happen normally, but I have in the past created large queues of commits changing only documentation.

### DIFF
--- a/scripts/cd_metrics/time_to_restore.lib.js
+++ b/scripts/cd_metrics/time_to_restore.lib.js
@@ -120,7 +120,7 @@ export async function calculateAccruedTimeInFailure(sha) {
     repo,
     previousSha
   );
-  if (previousStatus !== "failure") {
+  if (previousStatus === "success") {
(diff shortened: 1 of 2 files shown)


edmund-dunn (approved):
Methinks this is good to go!

PR #14157, opened 2023-06-26
VACMS-13845: Creates `va_gov_environment` module and Environment Discovery service.

### LINKED ISSUE #13845: Create new `EnvironmentDiscovery` service.
## Description

The `EnvironmentDiscovery` service does not indicate anything about environmental _behavior_, but merely vends the environment identifier, so that we can decouple the functional environment (and thus behavior) from the actual environment.

This environment identifiers are constant strings defined in `EnvironmentDiscoveryInterface` so that code elsewhere can easily benefit from this functionality. Their values do not matter. Currently, code that varies behavior by environment is not easily testable, and is mostly done with ad-hoc environment detection in Drupal hook implementations.

It is tempting to make these "environments" plugins rather than just string constants, but I believe this will lead to coupling behavior directly to the environment (e.g. adding the logic to the plugin class). That's the antipattern that inspired this issue.

This should exist in a new module, `va_gov_environment`, so that we do not continue to clutter the `va_gov_backend` module.

## Acceptance Criteria

- [x] A `va_gov_environment` module has been created.
- [x] An `EnvironmentDiscoveryInterface` interface has been created.
- [x] An `EnvironmentDiscovery` class has been created implementing that interface.
- [x] A `va_gov_environment.discovery` service is created, instantiating this service.
- [x] Functional tests exist, enabled only on appropriate environments, confirming correct service behavior.
- [x] Unit tests exist, confirming correct behavior.

ndouglas:
## Description

Closes #13845.

## Testing done
Added unit and functional tests. Will test on CMS-TEST to verify correct function there.

Ran on CMS-TEST staging:

```
$ drush eval 'print \Drupal::service("va_gov_environment.discovery")->getRawEnvironment();'
staging
$ drush eval 'print \Drupal::service("va_gov_environment.discovery")->getEnvironment();'
staging
$ drush eval 'print \Drupal::service("va_gov_environment.discovery")->isCmsTest() ? "TRUE" : "FALSE";'
TRUE
```


docroot/sites/default/settings.php
```diff
@@ -138,6 +138,15 @@
 
 $env_type = getenv('CMS_ENVIRONMENT_TYPE') ?: 'ci';
 
+/**
+ * Environment discovery service settings, vended by the Environment Discovery
+ * service (`va_gov.environment_discovery`).
+ */
+$settings['va_gov_environment'] = [
+  'environment_raw' => $env_type,
```
ndouglas, line 146: Mostly for introspection. `environment` will be set in the environment-specific settings files.

docroot/sites/default/settings/settings.tugboat.php
```diff
@@ -15,11 +17,6 @@
 // Use the TUGBOAT_REPO_ID to generate a hash salt for Tugboat sites.
 $settings['hash_salt'] = hash('sha256', getenv('TUGBOAT_REPO_ID'));
 
-/**
- * GLOBAL SETTINGS.PHP OVERRIDES
- *
- * Put regular settings.php overrides below this section.
- */
```
ndouglas, line 22: Meaningless cruft.

edmund-dunn (approved):
Nicely done, sir! This will be super helpful going forward.

PR #14160, opened 2023-06-26
VACMS-13847: Create ContentReleaseStrategy plugin system.

### LINKED ISSUE #13847: Create new `ContentReleaseStrategy` plugin system.
## Description

The `ContentReleaseStrategyPluginManager` manages a small collection of `ContentReleaseStrategyPlugin` classes. Each class represents a strategy for initiating a content release and encapsulates the specific logic and dependencies of that strategy.

These would be, at present, something like:
- Filesystem (i.e. check out `content-build` locally and execute it, targeting the local Drupal server)
- GitHub (i.e. issue a repository dispatch event on a specific workflow)

We could extend these with new PoCs for AP, different scenarios for AP, etc. I'm also inclined to add some strategies that only mock content releases, e.g. for testing how the system responds to dispatch errors, filesystem errors, etc. The manager itself has no additional methods because I want to be able to mock it easily. It should be a dumb plugin manager.

## Acceptance Criteria

- [x] A `ContentReleaseStrategyPluginInterface` interface has been created, specifying the following methods:
  - [x] `triggerContentRelease(): void`
- [x] A `ContentReleaseStrategyPluginBase` base class has been created.
  - [x] `triggerContentRelease()` is abstract
- [x] A `ContentReleaseStrategyPluginManagerInterface` interface has been created, specifying the following methods:
  - [x] `getStrategy(string $id): ContentReleaseStrategyPluginInterface`
  - [x] `triggerContentRelease(string $id): void`
- [x] A `GitHubRepositoryDispatch` strategy plugin exists encapsulating whatever logic is currently used in whatever class this is done.
  - [x] Unit tests exist covering all functionality of this class.
- [x] A `LocalFilesystem` strategy plugin exists encapsulating whatever logic is currently used to do these builds on Tugboat.
  - [x] Unit tests exist covering all functionality of this class.
- [x] Followup issues have been opened for additional strategy plugins that should be created.

ndouglas:
## Description

Closes #13847.

This _does_ affect the current content release system somewhat. I'll note specifics in the affected files.


tests/phpunit/va_gov_environment/DiscoveryTest.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace tests\phpunit\Environment;
+namespace tests\phpunit\va_gov_environment;
```
ndouglas, line 3: Moved to make matching up tests and code less irritating and subjective.

phpunit.xml
```diff
@@ -25,8 +25,8 @@
  should be no spaces between the hyphens).
 -->
   <php>
-    <!-- Set error reporting to E_ALL. -->
-    <ini name="error_reporting" value="32767"/>
+    <!-- Set error reporting to E_ALL & ~E_DEPRECATED & ~E_USER_DEPRECATED. -->
+    <ini name="error_reporting" value="8191"/>
```
ndouglas, line 29: This was incredibly annoying, since Guzzle has an unresolved deprecation and is stuck at its current version.

docroot/modules/custom/va_gov_content_release/va_gov_content_release.info.yml
```diff
@@ -0,0 +1,8 @@
+name: 'VA.gov Content Release'
+type: module
+description: 'Internal content release state management and tools'
+core_version_requirement: ^9 || ^10
+package: 'Custom'
+dependencies:
+  - va_gov_consumers
+  - va_gov_environment
```
ndouglas, line 8: Added a new module because "build trigger" doesn't really match how we converse about this process. Also, makes it easier to see what is and is not changing at any point in this process.

phpunit.xml
edmund-dunn: That will be changing soon, I hope.

docroot/modules/custom/va_gov_consumers/src/GitHub/GitHubClient.php
ndouglas, line 1: I made some changes here that affect the current content release process. The only changes should be that the response codes of the requests are checked and that errors and exceptions are rethrown with a wrapping exception.

PR #7105, opened 2021-11-30
VACMS-7113: Forbids PHPUnit to reuse connections.

ndouglas:
## Description

Relates to #7023.  Closes #7113.  This patches the `weitzman/drupal-test-traits` package to forbid it from reusing connections, in the hopes that this might keep it from basically sorta SlowLoris'ing our own machines during PHPUnit tests.

## QA steps

As user _uid_ with _user_role_
1. Do this
1. Then that
1. Then validate Acceptance Criteria from issue
- [ ] This
- [ ] That
- [ ] The other thing

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [ ] `Platform CMS team`
- [ ] `Sitewide CMS Team`

### Is this PR blocked by another PR?

- [ ] `DO NOT MERGE`

### Does this PR need review from a Product Owner

- [ ] `Needs PO review`

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change?
- [ ] Yes, and it's written in issue ____ and queued for publication.
  - [ ] Merge and ping @ rachel-kauff so she's ready to publish after deployment
- [ ] Yes, but it hasn't yet been written
  - [ ] Don't merge yet -- ping @ rachel-kauff to prompt her to write and queue content
- [ ] No announcement is needed for this code change.
  - [ ] Merge & carry on unburdened by announcements

PR #14204, opened 2023-06-28
VACMS-0000: Allow connection reuse in ExistingSiteBase tests.

ndouglas:
## Description

See #7105. The KeepAliveTimeout is no longer 1805 seconds after https://github.com/department-of-veterans-affairs/terraform-aws-vsp-internal-alb/pull/9, so this patch should no longer be needed.


edmund-dunn (approved):
Good deal!

PR #14212, opened 2023-06-29
VACMS-13849: Create form resolver service.

### LINKED ISSUE #13849: Create new `ContentReleaseFormResolver` service.
## Description

The `ContentReleaseFormResolver` service encapsulates the logic to translate the perceived environment, as retrieved from `EnvironmentDiscovery` (see #13845), into an appropriate content release page form. 

This logic is currently currently very tightly coupled into the rest of the environment system. This means that we have to override the environment just to look at a different version of the content-release form, which is really aggravating. We have a functional test that looks at this form, and it does so by overriding the environment in the Settings service, then injecting that back into the Drupal container and rebuilding the router. This is ugly.

Instead, this can be a very simple service that translates the environment identifier, retrieved from the injected `EnvironmentDiscovery` service, into a specific form class. This would just be hard-coded so that we can benefit from static analysis.

## Acceptance Criteria

- [x] A `ContentReleaseFormResolverInterface` interface exists, with the following methods:
  - [x] `getForm(): BuildTriggerForm`
- [x] A `ContentReleaseFormResolver` class exists, implementing the above interface.

ndouglas:
## Description

Closes #13849.


tests/phpunit/va_gov_content_release/unit/Strategy/Resolver/ResolverTest.php
ndouglas, line 1: Misnamed this the other day.

edmund-dunn (approved):
Good to go!

PR #14228, opened 2023-06-30
VACMS-0000: Updates NPM packages to fix Dependabot alerts.

ndouglas:
## QA Steps

- [x] See if Tugboat log has any unexpected error messages
- [x] Check design system and see if it looks okay


edmund-dunn (approved):
Everything works and built correctly.

PR #14229, opened 2023-06-30
VACMS-0000: Enable syntax highlighting for *.module and *.install files

ndouglas:
See [here](https://github.com/github-linguist/linguist/issues/1792).

Won't work immediately, [apparently](https://github.com/github-linguist/linguist/issues/1792#issuecomment-286418120).

@swirtSJW Any other extensions we need to add here? 🤔

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.module linguist-language=PHP
+*.install linguist-language=PHP


swirtSJW (approved):
Nice idea @ndouglas.   

The only other one that came to mind was .feature for cypress, but I just checked and it already seems to be highlighted correctly.

PR #14259, opened 2023-07-05
VACMS-0000: Disable sitewide_alert alerts on Tugboat.

ndouglas:
## Description

`sitewide_alert` alerts added to prod can propagate through the lower environments, but that's not ideal. The information they contain is generally not relevant. If it is relevant, we already know about it and are trying to fix the underlying issue. Either way, the alert can interfere with browser tests that expect certain elements to be visible. 

This can be somewhat difficult to predict, since some elements vary in length, and one that is a few characters too wide can be partially obscured by the JSD widget and thus not visible for test purposes.

This PR should disable all `sitewide_alert` alerts so that our tests run against a more predictable configuration.

Screenshot 2023-07-05 at 8 55 09 AM

Screenshot 2023-07-05 at 8 56 13 AM

### DIFF
--- a/.tugboat/config.yml
+++ b/.tugboat/config.yml
@@ -189,6 +189,9 @@ services:
         # https://www.drush.org/latest/deploycommand/ (updatedb, cache:rebuild, config:import, deploy:hook)
         - drush deploy
 
+        # Disable sitewide alerts so as not to interfere with testing.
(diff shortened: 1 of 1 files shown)


edmund-dunn (approved):
Good call! Nicely done!

PR #14314, opened 2023-07-11
VACMS-13854: Create Response Strategy plugin system.

### LINKED ISSUE #13854: Create `ContentReleaseEntityEventResponseStrategy` plugin system.
## Description

The `ContentReleaseEntityEventResponseStrategyPluginManager` class manages entity event response strategy plugins. Each plugin encapsulates the logic for responding to entity events. This is so that we can easily test the logic for how prod should respond to node updates, or compare multiple algorithms for that logic for equivalence.

These should be few in number and actually fairly simple:
- One that returns TRUE, e.g. always invokes a content release in response to any change.
- One that returns FALSE, e.g. never invokes a content release in response to any change. This is the default, since non-prod environments do not behave like this.
- One that uses the logic as currently implemented in `Drupal\va_gov_build_trigger\EventSubscriber\EntityEventSubscriber`. This is the strategy that would be used on production, and should be considerably simpler because most of the logic is implemented in bundle classes.

## Acceptance Criteria

- [x] A `ContentReleaseEntityEventResponseStrategyPluginInterface` interface has been created, specifying the following methods:
  - [x] `shouldTriggerContentRelease(VaNodeInterface $node): bool`
- [x] A `ContentReleaseEntityEventResponseStrategyPluginBase` base class has been created.
  - [x] `shouldTriggerContentRelease(VaNodeInterface $node)` is abstract
- [x] A `ContentReleaseEntityEventResponseStrategyPluginManagerInterface` interface has been created, specifying the following methods:
  - [x] `getStrategy(string $id): ContentReleaseEntityEventResponseStrategyPluginInterface`
  - [x] `shouldTriggerContentRelease(VaNodeInterface $node): bool`
- [x] Unit tests exist for the `ContentReleaseEntityEventResponseStrategyPluginManager` class.
- [x] A strategy plugin exists that uses the logic currently used on production.
  - [ ] Unit tests exist covering all functionality of this class.
- [x] Other strategy plugins exist for other use cases, e.g. tests and normal behavior.
- [x] Followup issues have been opened for additional strategy plugins that should be created.

ndouglas:
## Description

Closes #13854.

This (again) exists independently alongside the content release system within the CMS, and should not affect it yet.

## QA Steps
- [ ] Automated tests pass.

### DIFF
--- a/docroot/modules/custom/va_gov_content_release/src/Annotation/ContentReleaseStrategy.php
+++ b/docroot/modules/custom/va_gov_content_release/src/Annotation/ContentReleaseStrategy.php
@@ -23,7 +23,7 @@ class ContentReleaseStrategy extends Plugin {
   public $id;
 
   /**
-   * The human-readable name of the environemnt.
+   * The human-readable name of the strategy.
    *
    * @var \Drupal\Core\Annotation\Translation
    */
--- a/docroot/modules/custom/va_gov_content_release/src/Annotation/EntityEventStrategy.php
+++ b/docroot/modules/custom/va_gov_content_release/src/Annotation/EntityEventStrategy.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Drupal\va_gov_content_release\Annotation;
+
+use Drupal\Component\Annotation\Plugin;
+
+/**
+ * Defines the annotation object for entity event strategy plugins.
+ *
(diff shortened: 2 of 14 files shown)


edmund-dunn (approved):
More clean code!! Excellent!

Issue #14395, opened 2023-07-17
NPM JS build tooling seems broken.


ndouglas:
## Description
I don't tend to use this much, but just found that `npm run build` seems to be broken:

```
$ ddev npm run build

> va-gov-cms@1.0.0 build
> npm run build:js


> va-gov-cms@1.0.0 build:js
> cross-env BABEL_ENV=legacy node ./scripts/js/babel-es6-build.js

node:internal/modules/cjs/loader:988
  throw err;
  ^

Error: Cannot find module 'babel-core'
Require stack:
- /var/www/html/scripts/js/compile.js
- /var/www/html/scripts/js/changeOrAdded.js
- /var/www/html/scripts/js/babel-es6-build.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:985:15)
    at Function.Module._load (node:internal/modules/cjs/loader:833:27)
    at Module.require (node:internal/modules/cjs/loader:1057:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (/var/www/html/scripts/js/compile.js:3:15)
    at Module._compile (node:internal/modules/cjs/loader:1155:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1209:10)
    at Module.load (node:internal/modules/cjs/loader:1033:32)
    at Function.Module._load (node:internal/modules/cjs/loader:868:12)
    at Module.require (node:internal/modules/cjs/loader:1057:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/var/www/html/scripts/js/compile.js',
    '/var/www/html/scripts/js/changeOrAdded.js',
    '/var/www/html/scripts/js/babel-es6-build.js'
  ]
}
Failed to run npm run build: exit status 1
```

This is because we replaced `babel-core` with `@babel/core` at some point. After making that change, though, the output of any generated files is quite a bit different from the committed versions:

![Screenshot 2023-07-17 at 4 20 09 PM](https://github.com/department-of-veterans-affairs/va.gov-cms/assets/1318579/ebdc7f23-3653-4b49-b40f-346ab5774c5b)

And those updated versions don't match the expected formatting, throwing red squigglies all over the place.

## Acceptance Criteria
- [ ] `npm run build` command works without errors
- [ ] Updated files pass our own linting :facepalm:
- [ ] This command gets run during ci checks so we can verify it works

PR #14402, opened 2023-07-18
VACMS-14386: Create new content release forms.

### LINKED ISSUE #14386: Create new form classes for content release UI.
## Description

We currently have the following forms in `va_gov_build_trigger`:
- `BrdBuildTriggerForm`
- `BuildTriggerForm` (base class)
- `LocalBuildTriggerForm`
- `TugboatBuildTriggerForm`

These are tightly coupled to the environment concept and cannot really be tested in an environment-independent way.

## Tasks
- Create new versions of the forms, using current language and appearance, but environment-independent
- Redirect `va_gov_content_release` form route subscriber to point to these forms
- Redirect `va_gov_content_release` routes to point to these forms
- Find a way of injecting stub services for the test routes
- Add behavioral tests for testing these forms on every environment

## Acceptance Criteria
- [x] New forms exist
- [x] Old forms continue to exist
- [x] Existing functional and unit tests continue to pass
- [x] Behavioral tests have been added for new classes
- [x] `va_gov_content_release` and its services only refer to the new forms

ndouglas:
## Description

Closes #14386.

This should not yet affect the current content release architecture.

Briefly, this adds new routes and forms to handle those routes:

- `/admin/content/deploy/git` -- Submit a request to perform content release with a local `content-build` Git repository. This allows selection of the `default` branch or specifying another branch or PR (depending on current access to the GitHub API).
- `/admin/content/deploy/simple` -- Submit a request to perform content release "simply". This does not permit selecting a branch or performing other operations.

These are not intended for normal use (for instance, it doesn't make any sense to do a local, git-based content release on prod, and it's harmless but unusual to make a "simple" content release on Tugboat). Rather, these exist so that we can perform automated behavioral tests on the forms regardless of environment.

To this end, I've added a new Cypress step definition, `I stub form submission for the current page`, which can be executed on a form page. This will intercept the form submission and set an `is_under_test` form value to `"true"`. If a form submission method then checks the `is_under_test` value of the form state, it can modify its behavior to avoid costly outcomes.

There is a third route:

- `/admin/content/deploy/new` -- this is a replacement for the form currently found at `/admin/content/deploy`. It should look and behave the exact same way on all environments. But it does have a different backing form class and engages different machinery.

## To Do
- [x] ~I need to ensure I can't initiate prod content releases from Tugboat and local environments. That's not necessarily the concern of this issue, but it is becoming relevant.~ Dispatching a build from the "Simple" (BRD) form does not affect how the build is actually processed, so this is fine. I'm digging deeper into this in #14460.
- [x] ~I want to investigate switching backend services based on route. This would involve some additional complexity and need functional tests verifying correct behavior, but this might be useful for testing the content block. Or perhaps the content block should have separate tests altogether.~ I think the form submission handler should be a separate service.
- [x] ~Git branch selection is broken on Tugboat. It's broken on `main` too, but I want to cover this well with tests. Part of this is likely an infra issue with Git permissions, because a nobody user is trying to check git branches on Tugboat.~ Resolved in #14424.
- [x] ~PR branch selection is broken on DDEV because of the lack of a GitHub token (at least by default). I should add an error message or something and behavioral tests to account for this. This needs a test too.~ Will address in #14460.

## Testing done

Added some tests.

## QA steps

- [x] Verify that the [frontend](https://web-gcm0updj2ixyqscaoqvfm1lqr9rrdjec.ci.cms.va.gov/) was actually built.
- [x] Visit the [Release content](https://cms-gcm0updj2ixyqscaoqvfm1lqr9rrdjec.ci.cms.va.gov/admin/content/deploy) page.
- [x] Verify that the form works normally (optionally, perform a content release).
- [x] Visit the [new form](https://cms-gcm0updj2ixyqscaoqvfm1lqr9rrdjec.ci.cms.va.gov/admin/content/deploy/new).
- [x] Verify that it looks the same as the old form and works the same.
- [x] Visit the [git form](https://cms-gcm0updj2ixyqscaoqvfm1lqr9rrdjec.ci.cms.va.gov/admin/content/deploy/git).
- [x] Verify that it looks the same as the old form and works the same.
- [x] Visit the [simple form](https://cms-gcm0updj2ixyqscaoqvfm1lqr9rrdjec.ci.cms.va.gov/admin/content/deploy/simple).
- [x] Verify that it looks the same as the current CMS content release form.
- [x] Click the "Release content" button and verify that this form, unlike some other, inferior forms, actually requires that you confirm you understand the implications of your actions.
- [x] Click the "I understand" checkbox, click the "Release content" button, and verify that yes, this is a working confirmation.


scripts/js/compile.js
ndouglas, line 1: Related to #14395.

package.json
```diff
@@ -73,6 +73,7 @@
     "pixelmatch": "^5.3.0",
     "pngjs": "^6.0.0",
     "prettier": "^2.8.4",
+    "qs": "^6.11.2",
```
ndouglas, line 76: For query string parsing in Cypress.

edmund-dunn (approved):
good work!

Issue #14405, opened 2023-07-18
Fix issues with JSON API responses


timcosgrove:
## Acceptance Criteria
- [ ] Testable_Outcome_X
- [ ] Testable_Outcome_Y
- [ ] Testable_Outcome_Z
- [ ] Requires design review

### Team
Please check the team(s) that will do this work.

- [ ] `CMS Team`
- [ ] `Public Websites`
- [ ] `Facilities`
- [ ] `User support`

timcosgrove:
Tanner

ndouglas:
@timcosgrove 

I left a comment in #14451 but I'd like to rap about adding automated test coverage (within CMS repo) for any API schema requirements we have to minimize the threat of disruption to your work, especially considering that the connection between the changes in the PR and its effects on the JSON:API response is non-obvious to me.

tjheffner:
There were two unrelated issues with JSON:API:
1. it seems config related to translations may have removed certain base fields from JSON:API responses. The path alias for a given entity seemed to be the main one impacted in our case. See: https://www.drupal.org/project/drupal/issues/3274419 

We have some [mock json values captured in next-build](https://github.com/department-of-veterans-affairs/next-build/blob/main/src/mocks/newsStory.mock.json) from the prior effort, and those responses seemed to match on everything else but the `path` object. (There are some new values in the response related to translation source & status)

2. There was a feature in the `consumer_image_styles` module that we were relying on, it provided derived image dimensions for different image styles as part of the JSON:API response. During AP/next-build's pause, this feature was removed from the module because it actually created a HTTP request for each image style requested for each image in order to correctly derive the dimensions of the image styles.

Images are rendering again, but the only width/height values that are passed come from the initial image upload, so they are not always in the correct aspect ratio for the stated image style. I am not entirely sure we want to be making all of those requests under-the-hood either for AP. Maybe we do store all images, original and derived, on the S3 bucket and it doesn't matter as much, because the image will exist already and the dimensions will be known so we can include it. Further investigation needed. 

additional context:
https://www.drupal.org/project/consumer_image_styles/issues/3069385 [feature we initially depended on]
https://www.drupal.org/project/consumer_image_styles/issues/3252023 [issue that reverts that feature due to causing many http requests]
https://www.drupal.org/project/consumer_image_styles/issues/3301224 [this is the one we added a patch from, it fixes a portion that the previous issue missed. the missing piece was causing validation errors in our response.]



[...]

PR #14451, opened 2023-07-19
VACMS-14405 fix jsonapi responses

tjheffner:
## Description

Relates to #14405. This PR relates to next-build functionality -- the CMS's JSONAPI endpoint wasn't returning all the expected data without these two patches included. 

I also needed to remove the config for news story node's path value... something apparently related to base field override (possibly translations?) seems to mysteriously drop some fields in the json response. The BaseFieldOverride patch fixes that behavior, but the config was committed prior. Removing that config shows the path variable in the response as expected. Other content types may need their `core.base_field_override.node.NODE_TYPE.path.yml` files removed in the future for JSONAPI as well.

## Testing done
running locally, hitting a jsonapi endpoint for a news story node should include a `path` object underneath the `attributes` key:
https://va-gov-cms.ddev.site/jsonapi/node/news_story/1dbd5b1e-0fb2-43fb-8581-a2f9edfe3c83

## Screenshots
Screenshot 2023-07-13 at 1 36 34 PM

## QA steps
- What needs to be checked to prove this works?  
Visit the same endpoint without the patches from this PR applied and see that the path variable is missing.

You could also try running the next-build repo as-is. Story listing pages will fail because `entity.path.alias` is not found in the data for any of the listed news story nodes. There are other cleanup tasks pushed to that repo here: https://github.com/department-of-veterans-affairs/next-build/pull/126

- What needs to be checked to prove it didn't break any related things?  
- [ ] Do News Stories still translate as expected?
- [ ] Validate that the path of a newly created news story matches that of a newly created news story on staging.

### Definition of Done

### Select Team for PR review

- [x] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`
- [x] `Accelerated Publishing`


ndouglas:
If y'all're gonna need specific things from the API, I'd like to have test coverage of those requirements so we don't inadvertently create obstacles for you, especially with things that depart from the JSON:API spec, default Drupal behavior, etc.

Could you add PHPUnit tests to perform a JSON:API request and verify that the raw response conforms to the schema that you need? 

If this is something you're not comfortable with, I can open up a followup issue and start creating tests and utility traits for this sort of thing.

Just wanna make sure we're holding up our end of the bargain 🙂

tjheffner:
@ndouglas that is a great callout and something we will definitely want. I don't mind taking a crack at the PHPUnit tests, but do you mind if I include them in the next followup (that has updates for image styles?)

These two patches are the bare minimum right now for next-build to successfully run locally, and I'd like to get the AP team unblocked on that

ndouglas:
@tjheffner Yeah, that's fine 🙂

tjheffner:
Turns out there will not need to be a follow-up for image styles at this time... they're there, just moved slightly (and the missing height/width data is currently intentional, due to the "bug" the patch is fixing). I'll include a pass at schema tests in this PR.

swirtSJW:
I added a QA step to validate the path of  news story since that override is being removed.

ndouglas (approved):

ndouglas:
Story nodes aren't currently translatable. If and when they become translatable, I assume this might become an issue, but at present I'm not inclined to set up translation for the content type just to see if this breaks it 🙂 

Created a story node, nothing weird happened 🤷🏻‍♂️  Path alias unaffected.

tjheffner:
i murdered the git history again attempting to rebase.... going to close and reopen a clean one off the updated main. ugh. sorry for the noise.

PR #14557, opened 2023-07-28
VACMS-0000: Make Cypress test more reliable.

ndouglas:
This adds a Cypress command to run Drush and forcibly unlock node 1010 before the relevant test in `content_editing.feature` just in case the previous test failed unexpectedly.

## QA Steps

- [ ] Automated tests pass.


tonytaylor (approved):
Stupid 1010 - who needs it?  We do!
Looks good though.

Issue #14565, opened 2023-07-28
Error prevents deleting a facility


swirtSJW:
## Describe the defect
When deleting a facility, which should be avoided, there is an error that prevents it.

> Drupal\va_gov_content_types\Exception\NoOriginalExistsException: No original version exists for this node. in Drupal\va_gov_content_types\Entity\VaNode->getOriginal() (line 42 of /var/www/cms/docroot/modules/custom/va_gov_content_types/src/Traits/GetOriginalTrait.php). [dd.trace_id=9223372036854775807 dd.span_id=8118747153603467487]

## To Reproduce
1. Go to '/node/3819/delete'
2. Click on  the delete button
5. The deletion does not occur.
6. See recent log messages for the cause.

![image](https://github.com/department-of-veterans-affairs/va.gov-cms/assets/5752113/dc9adb5f-2b7e-4dbc-822f-e52b4baae9ec)

## AC / Expected behavior
The facility node should delete successfully.

## Additional context
[This line](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/main/docroot/modules/custom/va_gov_content_types/src/Traits/GetOriginalTrait.php#L47) is making an assumption that $node->original is there, when in many cases (node creation, deletion) it is not there at all.

jilladams:
@swirtSJW can you add a comment here to clarify status - we have 2 closed PRs / nothing merged but ticket is closed. Not sure what that means, for reportings' sake.

jilladams:
@swirtSJW @ndouglas : But now confused again. 

1. This ticket had a PR: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/14567 -- closed in favor of a different PR
2. The different PR: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/14569, merged
3. There is also a ticket: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14576 - closed in favor of a different ticket
4. The different ticket: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14580
5. THis ticket is also closed but I'm not clear on whether: 
    a. It's closed as a dupe of #14580 also, OR
    b. It's closed bc it was resolved by https://github.com/department-of-veterans-affairs/va.gov-cms/pull/14569, meaning this bug should not repro in Prod any longer

[...]

PR #14567, opened 2023-07-29
VACMS-14565 bugfix make facilities deleteable

swirtSJW:
## Description

Relates to #14565

## QA steps

As an admin
1. Go to a [Abilene Vet center facility](https://pr14567-obclexe1yea8ftut4va5uquxmd47gobi.ci.cms.va.gov/abilene-vet-center) and delete it.
   - [ ] Validate that it deletes.
2. Go to the migration and run it as an update migration (check the "update" check box)
image  

   - [ ] validate that the vet center you deleted, was recreated by [going here](https://pr14567-obclexe1yea8ftut4va5uquxmd47gobi.ci.cms.va.gov/abilene-vet-center) 
   - [ ] Open a new tab for [dblog](https://pr14567-obclexe1yea8ftut4va5uquxmd47gobi.ci.cms.va.gov/admin/reports/dblog)  validate there are no related errors/warnings in watchdog
3. Edit a facility node and give it status and publish it.
   - [ ] validate in dblog that a content release should have been triggered.
4. Edit the same node but do not change its status. and publish and save the revision
   - [ ] validate in dblog that a content release should **not** have been triggered
5. Edit the same node and change the status, save and publish it.
   - [ ] Validate in dblog that a content release should have been triggered.

- [ ] `Needs PO review`

### DIFF
--- a/docroot/modules/custom/va_gov_content_types/src/Interfaces/GetOriginalInterface.php
+++ b/docroot/modules/custom/va_gov_content_types/src/Interfaces/GetOriginalInterface.php
@@ -35,10 +35,10 @@ interface GetOriginalInterface {
    * @param string $field_name
    *   The machine name of the field to get.
    *
-   * @return \Drupal\Core\Field\FieldItemListInterface
-   *   The value of the field.
+   * @return \Drupal\Core\Field\FieldItemListInterface|null
+   *   The fieldItemList of the field if it exists. NULL otherwise.
    */
-  public function getOriginalField(string $field_name): FieldItemListInterface;
+  public function getOriginalField(string $field_name): ?FieldItemListInterface;
 
   /**
    * Checks if the value of the field on the node changed.
--- a/docroot/modules/custom/va_gov_content_types/src/Interfaces/GetOriginalInterface.php
+++ b/docroot/modules/custom/va_gov_content_types/src/Interfaces/GetOriginalInterface.php
@@ -35,10 +35,13 @@ interface GetOriginalInterface {
    * @param string $field_name
    *   The machine name of the field to get.
    *
-   * @return \Drupal\Core\Field\FieldItemListInterface|null
+   * @return \Drupal\Core\Field\FieldItemListInterface
    *   The fieldItemList of the field if it exists. NULL otherwise.
+   *
+   * @throws \Drupal\va_gov_content_types\Exception\NoOriginalExistsException
+   *   Thrown when the node has no original version.
    */
-  public function getOriginalField(string $field_name): ?FieldItemListInterface;
+  public function getOriginalField(string $field_name): FieldItemListInterface;
 
   /**
    * Checks if the value of the field on the node changed.
--- a/docroot/modules/custom/va_gov_content_types/src/Traits/GetOriginalTrait.php
+++ b/docroot/modules/custom/va_gov_content_types/src/Traits/GetOriginalTrait.php
@@ -31,6 +31,8 @@ trait GetOriginalTrait {
   /**
    * Retrieve the original version of this node.
    *
+   * To fail gracefully, precede this with a separate check for ->hasOriginal().
+   *
    * @return \Drupal\va_gov_content_types\Entity\VaNodeInterface
    *   The original version of this node.
    *
@@ -54,12 +56,18 @@ trait GetOriginalTrait {
    * @param string $fieldName
    *   The machine name of the field to get.
    *
-   * @return \Drupal\Core\Field\FieldItemListInterface
-   *   The value of the field.
+   * @return \Drupal\Core\Field\FieldItemListInterface|null
+   *   The fieldItemList of the field if it exists. NULL otherwise.
    */
-  public function getOriginalField(string $fieldName): FieldItemListInterface {
-    $original = $this->getOriginal();
-    return $original->get($fieldName);
+  public function getOriginalField(string $fieldName): ?FieldItemListInterface {
+    $field = NULL;
+    if ($this->hasOriginal()) {
+      $original = $this->getOriginal();
+      if ($original->hasField($fieldName)) {
+        $field = $original->get($fieldName);
+      }
+    }
+    return $field;
   }
 
   /**
@@ -74,6 +82,11 @@ trait GetOriginalTrait {
   public function didChangeField(string $fieldName): bool {
     $originalField = $this->getOriginalField($fieldName);
     $currentField = $this->get($fieldName);
+    if (!$originalField) {
+      // Consider a new save (no original exists) to be a change of value.
+      return TRUE;
+    }
+
     return !$currentField->equals($originalField);
   }
 
--- a/docroot/modules/custom/va_gov_content_types/src/Traits/GetOriginalTrait.php
+++ b/docroot/modules/custom/va_gov_content_types/src/Traits/GetOriginalTrait.php
@@ -56,18 +56,21 @@ trait GetOriginalTrait {
    * @param string $fieldName
    *   The machine name of the field to get.
    *
-   * @return \Drupal\Core\Field\FieldItemListInterface|null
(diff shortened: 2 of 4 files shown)


docroot/modules/custom/va_gov_content_types/src/Traits/GetOriginalTrait.php
```diff
@@ -74,6 +82,11 @@ public function getOriginalField(string $fieldName): FieldItemListInterface {
   public function didChangeField(string $fieldName): bool {
     $originalField = $this->getOriginalField($fieldName);
     $currentField = $this->get($fieldName);
+    if (!$originalField) {
+      // Consider a new save (no original exists) to not be a change of value.
+      return FALSE;
```
swirtSJW, line 87: I am torn on this a bit.  First commit, I was of the mind that going from no value to a value should be considered a change.  Second commit (Nate's patch) I aligned it with it not being a change.   They both seem equally misleading.

swirtSJW:
The status report after the migration runs is a lie
image

It creates a new vet center for any you deleted, it just does not consider them as new because they already have an entry for the uinque id in the migrate map table.

swirtSJW:
Blarg

```
There was 1 error:

1) tests\phpunit\va_gov_content_types\unit\Traits\GetOriginalTraitTest::testGetOriginalField
Error: Call to a member function getValue() on null

/home/runner/work/va.gov-cms/va.gov-cms/tests/phpunit/va_gov_content_types/unit/Traits/GetOriginalTraitTest.php:123
```

and 
```
There was 1 failure:

1) tests\phpunit\va_gov_content_types\unit\Traits\GetOriginalTraitTest::testDidChangeField
Expectation failed for method name is "get" when invoked 1 time(s).
Method was expected to be called 1 times, actually called 0 times.

/home/runner/work/va.gov-cms/va.gov-cms/docroot/vendor/phpunit/phpunit/src/Framework/MockObject/Matcher.php:237
```

tests/phpunit/va_gov_content_types/functional/Traits/GetOriginalTraitTest.php
```diff
@@ -73,17 +73,6 @@ public function testGetOriginalField() {
...
 
-  /**
-   * Test that the getOriginalField method throws an exception appropriately.
-   */
-  public function testGetOriginalFieldException() {
-    $node = $this->createNode([
-      'bundle' => 'page',
-    ]);
-    $this->expectException(NoOriginalExistsException::class);
-    $node->getOriginalField('title');
```
swirtSJW, line 84: This check for an exception is testing a method that throws no exception.  
->getOriginalField() calls ->getOriginal() which does throw an exception if called without being defensive.  But ->getOriginalField() is defensive about its call to ->getOriginal().

swirtSJW:
@ndouglas 
The phpunit errors are related to calls to 'get()' on this test
image

But before I dig into figure out why, I am confused how this tests 'getOriginalField()' when it never calls that method.  It seems like this is more testing inner drupal workings rather than our additional method.  I am sure I am missing something, but I don't see it.  :(

ndouglas:
>But before I dig into figure out why, I am confused how this tests 'getOriginalField()' when it never calls that method. It seems like this is more testing inner drupal workings rather than our additional method. I am sure I am missing something, but I don't see it. :(

Nope, that's a typo on my part. That should have been

```php
$this->assertEquals($node->getOriginalField('title')->value, $original->get('title')->value);
```

docroot/modules/custom/va_gov_content_types/src/Interfaces/GetOriginalInterface.php
```diff
@@ -35,10 +35,10 @@ public function getOriginal(): VaNodeInterface;
    * @param string $field_name
    *   The machine name of the field to get.
    *
-   * @return \Drupal\Core\Field\FieldItemListInterface
-   *   The value of the field.
+   * @return \Drupal\Core\Field\FieldItemListInterface|null
+   *   The fieldItemList of the field if it exists. NULL otherwise.
```
ndouglas, line 39: I strongly disagree with this change.

I feel that calling `getOriginalField($fieldName)` fairly explicitly lays out some preconditions -- that an original version of the node exists, and that the field exists -- and the method should communicate which (if either) precondition is not met. That's why it propagates exceptions.

Changing this to optionally return a NULL makes it harder to debug. It also discards the information about which precondition was not met.

ndouglas: "value" -> "fieldItemList" is A++++ though.

docroot/modules/custom/va_gov_content_types/src/Traits/GetOriginalTrait.php
ndouglas: I'm not sure that it makes sense to call `didChangeField()` on something without an original version. I think it should probably throw an exception.

tests/phpunit/va_gov_content_types/functional/Traits/GetOriginalTraitTest.php
ndouglas: I disagree. I think it's perfectly reasonable to check for exception propagation.

As noted above, I believe that `getOriginalField()` embeds two assumptions about the callee. I don't think it's an appropriate intermediary for handling the exception from `getOriginal()`. If I were a real man, I would've written `getFieldValue()` to wrap `get()` and thrown an exception from that too 😂

docroot/modules/custom/va_gov_content_types/src/Interfaces/GetOriginalInterface.php
swirtSJW: I'm listening to you.  I am trying to get onboard, I am largely troubled by this though because Drupal's internal magic property of $node->fieldname fails gracefully(returns NULL if we request something that does not exist, as does $node->get('fieldname').  Other than the ability to test more rigidly, what do we gain by making our method that mimics get() fail more strict than Drupal's internal function that we are replicating?  (hah, I just saw that you joked about that farther down)  For developer sanity, I think we should try to replicate Drupal's approach than create our more uptight doppleganger.

docroot/modules/custom/va_gov_content_types/src/Traits/GetOriginalTrait.php
swirtSJW: I can get behind that idea but it moves the defensiveness required higher up the chain we need to make less assumptions.

tests/phpunit/va_gov_content_types/functional/Traits/GetOriginalTraitTest.php
swirtSJW: It does??
You are right and that just blew my arguement to make it fail gracefully out of the water.   Apparently it only fails gracefully if you access the property directly
image

docroot/modules/custom/va_gov_content_types/src/Interfaces/GetOriginalInterface.php
swirtSJW: I was talking jibbereish ^^ and our comment farther down made me realize you are right.

tests/phpunit/va_gov_content_types/functional/Traits/GetOriginalTraitTest.php
swirtSJW: Adding this check back in because now getOriginalField throws its own exceptions.

swirtSJW:
@ndouglas I have adjusted these based on your wise advice.  It looks like the unit test is still angry.  I will look at it closer in the morning after I've had some sleep.

ndouglas:
Screenshot 2023-07-31 at 7 51 49 AM

I hope you're getting more sleep that this screenshot implies ❤️

swirtSJW:
Closing this in preference of #14569

PR #14569, opened 2023-07-31
VACMS-14565: Make content release operations on nodes without original versions safer.

ndouglas:
A somewhat different approach to @swirtSJW's in #14567, though incorporating some of his changes.

This makes the change in a different place and adds some tests that I think will provide more confidence that these errors won't cause problems during normal editing operations 😩 .

## QA steps (Copied from #14567)

As an admin
1. Go to a [Abilene Vet center facility](https://pr14569-rbbisnso6sf3muduwsd8ozkb4orvvv85.ci.cms.va.gov/abilene-vet-center) and delete it.
   - [ ] Validate that it deletes.
2. Go to the migration and run it as an update migration (check the "update" check box)
image  

   - [ ] validate that the vet center you deleted, was recreated by [going here](https://pr14569-rbbisnso6sf3muduwsd8ozkb4orvvv85.ci.cms.va.gov/abilene-vet-center) 
   - [ ] Open a new tab for [dblog](https://pr14569-rbbisnso6sf3muduwsd8ozkb4orvvv85.ci.cms.va.gov/admin/reports/dblog)  validate there are no related errors/warnings in watchdog
3. Edit a facility node and give it status and publish it.
   - [ ] validate in dblog that a content release should have been triggered.
4. Edit the same node but do not change its status. and publish and save the revision
   - [ ] validate in dblog that a content release should **not** have been triggered
5. Edit the same node and change the status, save and publish it.
   - [ ] Validate in dblog that a content release should have been triggered.

6. Edit the same node, make a change to status but save as draft
    - [ ] validate that no release trigger took place.
7. Edit edit the same node and change to publish
   - [ ] validate that a release trigger took place.


docroot/modules/custom/va_gov_content_types/src/Traits/GetOriginalTrait.php
```diff
@@ -56,6 +56,11 @@ public function getOriginal(): VaNodeInterface {
    *
    * @return \Drupal\Core\Field\FieldItemListInterface
    *   The value of the field.
```
swirtSJW, line 58:
```suggestion
   *   The fieldItemList of the field.
```

swirtSJW (commented):
I like this approach.  A couple minor adds.

ndouglas:
I'm looking closer at that PHPUnit failure, but it doesn't appear to be related.

docroot/modules/custom/va_gov_content_types/src/Interfaces/GetOriginalInterface.php
```diff
@@ -36,7 +36,12 @@ public function getOriginal(): VaNodeInterface;
    *   The machine name of the field to get.
    *
    * @return \Drupal\Core\Field\FieldItemListInterface
-   *   The value of the field.
+   *   The fieldItemList of the field.
+   *
+   * @throws \Drupal\va_gov_content_types\Exception\NoOriginalExistsException
+   *   Thrown when the node has no original version.
+   * @throws \InvalidArgumentException
+   *   Thrown when the field does not exist.
```
ndouglas, line 44:
```suggestion
```

ndouglas:
The PHPUnit test failure didn't repeat this time. I'm reasonably confident saying it wasn't related because it was complaining that the local checkout of `content-build` wasn't a git repository 😕  I'm going to open an issue for that separately. It sounds like some of the steps of the content release might have been taking place, and that shouldn't be the case.

swirtSJW (commented):
All this looks good.  When tests pass, I will run through the QA steps.

ndouglas:
@swirtSJW Tests are all passed 🙂

swirtSJW:
I am seeing a failure on the following steps.
1) make a change to status but save as draft
2) edit the same node and change to publish
- [ ] validate that a release trigger took place.

swirtSJW:
I validated this is working on Main.  
image

swirtSJW:
> @swirtSJW Is the desired behavior there for `didChangeOperatingStatus()` (or something it calls) to compare the operating status of the current, published revision, to the previous published revision and return TRUE if there's a difference?

@ndouglas  I think the ideal way to handle it IFF it was published, original was draft and Current is published, then load the default revision to see if it changed.   We did that for data pushes to lighthouse.   However, I think the original code on this punted and just said, if it is going from draft to published and was published, triggere a release because we don't know what the previous state was because it is too far back in the history (more than one back in the history).

swirtSJW:
Probably didChangeField should have the clause that gets default revision to compare as original if  

- original status was draft AND
- current status is published AND
- was previously published

Or  maybe getOriginal should do that. ??

ndouglas:
I've added this clause here:

Screenshot 2023-07-31 at 2 02 09 PM

That did (as it should) cause a test to fail, but after changing the test it passes.

That's not ideal, but I'm concerned about the code on prod causing issues for editors.

I'd prefer to devote another ticket to modifying/extending this to compare to default revisions, if you're okay with that. I'm fine with making it a high priority ticket.

ndouglas:
I opened #14575 to change this to behave more intelligently in the future.

swirtSJW (approved):
As an admin
1. Go to a [Abilene Vet center facility](https://pr14569-rbbisnso6sf3muduwsd8ozkb4orvvv85.ci.cms.va.gov/abilene-vet-center) and delete it.
   - [x] Validate that it deletes.
2. Go to the migration and run it as an update migration (check the "update" check box)
image  

   - [x] validate that the vet center you deleted, was recreated by [going here](https://pr14569-rbbisnso6sf3muduwsd8ozkb4orvvv85.ci.cms.va.gov/abilene-vet-center) 
   - [x] Open a new tab for [dblog](https://pr14569-rbbisnso6sf3muduwsd8ozkb4orvvv85.ci.cms.va.gov/admin/reports/dblog)  validate there are no related errors/warnings in watchdog
3. Edit a facility node and change its status and publish it.
   - [x] validate in dblog that a content release should have been triggered.
4. Edit the same node but do not change its status. and publish and save the revision
   - [x] validate in dblog that a content release should **not** have been triggered
5. Edit the same node and change the status, save and publish it.
   - [x] Validate in dblog that a content release should have been triggered.

6. Edit the same node, make a change to status but save as draft
    - [x] validate that no release trigger took place.
7. Edit edit the same node and change to publish
   - [x] validate that a release trigger took place.

This all checks out great. 
image


I did surface [one existing issue](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14580) that pre-existed this.  Deleting a facility does not trigger a release. 
image

It is not a critical problem since admins are the only one who can delete a facility so it would go out with normally scheduled content releases.  


Incredible work Nate.

PR #14042, opened 2023-06-09
VACMS-13143: Magichead with the entity_reference_hierarchy module

chri5tia:
## Description

Relates to #13143

Just a draft. Still trying to figure out how to use this module/show the functionality I was looking for in it.

## QA steps

What needs to be checked to prove this works?  
What needs to be checked to prove it didn't break any related things?  
What variations of circumstances (users, actions, values) need to be checked?  

As user _uid_ with _user_role_
1. Do this
   - [ ] Validate that
2. Then
   - [ ] Validate that
3. Then validate Acceptance Criteria from issue
   - [ ] a
   - [ ] b
   - [ ] c

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

### Is this PR blocked by another PR?

- [ ] `DO NOT MERGE`

### Does this PR need review from a Product Owner

- [ ] `Needs PO review`

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change?
- [ ] Yes, and it's written in issue ____ and queued for publication.
  - [ ] Merge and ping the UX writer so they are ready to publish after deployment
- [ ] Yes, but it hasn't yet been written
  - [ ] Don't merge yet -- ping the UX writer to write and queue content
- [ ] No announcement is needed for this code change.
  - [ ] Merge & carry on unburdened by announcements

Issue #12826, opened 2023-03-02
CMS mirror is using tip of main instead of released tag


swirtSJW:
## Describe the defect
The CMS mirror in tugboat that provides data to all the lower environment FE content-builds is using the database from prod, but is building with where the tip of main is at the time it builds (2-3AM).  This means that FE tests are running against CMS code that is not actually on prod yet.  This creates a disconnect in our testing pipeline where it is possible that FE tests pass for fields that have not made it to main yet, and since the daily FE build happens an hour before the cms build, the result can break the build and send people scrambling to resolve it.

## AC / Expected behavior
- [ ] The tugboat mirror uses the latest released tag, not the latest commit when it is rebuilt each night.

## Additional context
Add any other context about the problem here. Reach out to the Product Managers to determine if it should be escalated as critical (prevents users from accomplishing their work with no known workaround and needs to be addressed within 2 business days).

## Labels
(You can delete this section once it's complete)
- [x] Issue type (red) (defaults to "Defect")
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [x] CMS workstream (orange) (not needed for bug tickets)
- [ ] CMS-supported product (black)

### Team
Please check the team(s) that will do this work.

- [ ] `CMS Team`
- [ ] `Public Websites`
- [ ] `Facilities`
- [ ] `User support`

productmike:
@olivereri @ndouglas I'd like to get @teeshe into the action in the next sprint. Do y'all think this would be a good starter ticket?

ndouglas:
@olivereri knows more about this than I do, and he might talk me down, but this ticket scares me. 

I think there would be a lot of work involved in this and I have a clear path forward on very little of it. I would guess that this needs to be an epic and @teeshe and @olivereri and I would need to do some discovery, discussion about how to proceed, etc.

Or Eric might come back and say "nah, we just set this environment variable/tweak a script and this'll happen automatically." 🤷🏻‍♂️

productmike:
@ndouglas That's enough to set aside and back into Needs Refined status -- even if just for a quick connection to validate your perspective or otherwise :) We'll set @teeshe ahead on #14075!

PR #14657, opened 2023-08-04
Revert "VACMS-13143: Magichead Model-3"

ndouglas:
Reverts department-of-veterans-affairs/va.gov-cms#14042


edmund-dunn (approved):
Git r done!

ndouglas:
This is undeployable because the commit made it onto `main` and was part of the Tugboat base preview build this morning. See [this issue](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/12826).

PR #14879, opened 2023-08-21
VACMS-14878: Make Cypress tests a little more reliable.

### LINKED ISSUE #14878: Repeated facilities API Cypress test failures
Think this is my fault.

## Acceptance Criteria

- [ ] Cypress tests can run with fewer failures.

ndouglas:
Closes #14878.

- Increases Cypress' responseTimeout to 60s.
- Refactors some long tests that were the source of some repeated failures into smaller, more finely-grained tests.

As seen here: https://pr14868-lkkjjomadlkbvxbrl75wu9ldmyi322v0.ci.cms.va.gov/cypress/videos/facilities/facilities_api.feature.mp4

At timestamp 2:30, an attempt is made to load `/node/175/edit`. Whether because of hooks or form complexity or other factors or just server load, the load cannot complete within 30 seconds. 

Based on limited testing, node 175 seems to be a frequent offender.

So this PR attempts to redress this by increasing the timeout for individual page loads and refactors the Facilities API test from a test with a chain of testable conditions into a list of tests, each with a shorter list of steps.

## QA Steps

- [x] Automated tests pass.


tests/cypress/integration/features/facilities/facilities_api.feature
```diff
@@ -4,87 +4,26 @@ Feature: CMS Users may not unintentionally change information in fields populate
...
+  # Content Admin
+  Scenario: Log in and edit a <contentType>: <title> (<nid>) as <roleTitle>
+    Given I am logged in as a user with the roles "<roles>"
+    And my workbench access sections are set to "<sections>"
+    When I unlock node <nid>
+    And I am at "/node/<nid>/edit"
+    Then an element with the selector "<shouldExistSelector>" should exist
+    And an element with the selector '<shouldNotExistSelector>' should not exist
+
+  Examples:
```
ndouglas, line 16: The examples are incredibly hard to read, but they transpose the information in the above test steps and comments into a table that is then used and logged directly.

swirtSJW: If I make it wide enough, I can read it.  ;)

swirtSJW (commented):
@ndouglas I get how adding the timeout makes it a little more reliable.   How does the change to the table make it more reliable?

ndouglas:
@swirtSJW 

Well, shifting the data into tables is mostly for the sake of DRY, since we're doing the same set of operations a dozen times. We might look at using different selectors or test-id attributes or something similar in the future to make the table narrower, or just abstracting some of that into a separate step definition.

The general refactor should help with reliability in a couple of different ways by splitting long tests with multiple Given-When-Then sequences into shorter tests, each with a single Given-When-Then sequence. Of course, a test with fewer risky operations is less likely to encounter failure than a test with four times as many of the same risky operation. Since tests are retried, this has a compounding effect with making retries less likely to fail.

The table restricts the sections assigned to the user to those necessary to edit the node in question, rather than a superset of those necessary. Given that the test is verifying (among other things) the accuracy of permissions and section access, this seems like a reasonable change to me.

It also improves reporting by exposing the node, content type, nid, etc in the test name. Rather than needing to trace through the Tugboat log and watch the video to see where the failure actually occurred, you can just see "oh, we had repeated failure to load the Cheyenne Vet Center in under 30 seconds" or "this error is thrown by content admins on Vet Center Outstations, but nowhere else."

Since each test is performed with a different user, this gives increased traceability for watchdog errors, etc.

And this change should make it easier to add new test cases, should that become useful. 

BBL gotta figure out why prod's dying.

ndouglas:
@swirtSJW I had a few minutes while waiting for a meeting so I wrote a custom step definition ("the Facility Locator API ID field should not be editable") and threw it under `facilities/`. That makes the table two columns narrower and abstracts away some of the implementation details.

swirtSJW:
Here is a great reason why we need this :) https://dsva.slack.com/archives/C0FQSS30V/p1692985619239349

swirtSJW (approved):
Nice work.  I think this makes it cleaner and hopefully more reliable.

PR #14988, opened 2023-08-28
VACMS-14983: Remove old CodeQL workflow.

ndouglas:
No longer necessary now and might be causing complaints.


edmund-dunn (approved):
Nice!!

Issue #15092, opened 2023-09-05
Cross team intake request - CMS tugboat server infra


tjheffner:
## Overview
This is a template to be used by the Platform CMS, Accelerated Publishing, Public Websites, Facilities, Sitewide CMS, and CMS User Support teams in order to:
- help determine which team ought to do the work
- help any discussions at the PO or PM levels around scope and timeline
- help capture the needed information to make good decisions on prioritization by the intake team

Please be as detailed as possible knowing that any recommended path forward may change. This ticket will help give a baseline for discussions.

Please provide:

### Description of the problem / issue / enhancement
* please include all relevant context known
 
When next-build sends requests to the content-build tugboat endpoint (https://tugboat.vfs.va.gov/6189a9af690c68dad4877ea5 / https://content-build-medc0xjkxm4jmpzxl3tfbcs7qcddsivh.ci.cms.va.gov/ ) we occasionally see `502 Bad Gateway` errors returned. It appears to be caused by next-build making too many concurrent requests to the endpoint.

I'd like AP DevOps and CMS DevOps to work together on this issue to update the server infrastructure to allow for more requests (or larger sized requests, or whatever the issue ultimately turns out to be)

Related ticket: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15138

## Description of known dependencies
check nginx logs in the php container on that tugboat
update nginx + php-fpm configuration if needed. light googling shows `fastcgi` options may be one lever to pull

In next-build, there is this section in `next.config.js` that limits the # of requests to 1 at a time. this greatly slows down speed when generating static pages.
```
  experimental: {
    // This is experimental but can
    // be enabled to allow parallel threads
    // with nextjs automatic static generation
    workerThreads: false,
    cpus: 1,
  },
```

Ideally, this piece of configuration is able to be removed after the server infrastructure is updated to allow the # of requests next-build may send.

## Rough estimate of urgency and impact
* urgency as it relates to timeframe needed and why
* impact as it relates to # of people affected, their roles, and severity (i.e. can't perform their job duties)

Urgency: relatively urgent. Needed before AP team can go to production.
Impacts: everyone on the AP team, consumers of pages built by next-build

## Acceptance Criteria
- [ ] The Product Manager for the recommended intake team is pinged through Slack about the issue
- [ ] The Product Manager has discussed with the respective Product Owner to determine prioritization
- [ ] Any decisions made on prioritization or implementation are communicated back to the originating team

[...]

PR #15341, opened 2023-09-21
VACMS-15092: Debug Tugboat 502s impacting next-build

ndouglas:
See #15092.

Not normally executed, no risk.


ndouglas:
Per [this comment](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15092#issuecomment-1732069166), the priority of this issue has dropped. Debugging the issue further will require altering logging settings on the Tugboat server, which I'm unwilling to do at this time.

This should be committed so it can be used to debug further issues or if the priority of this issue is increased.

Issue #15242, opened 2023-09-13
BE: Ensure required revision log field is visible on promo block forms


wesrowe:
### Engineering notes / background

In #15187 it was found that the promo block UI (used for publishing full-width banners?) requires a revision log, but the revision log field is hidden – thus preventing saving. [Example promo block in Prod](https://prod.cms.va.gov/block/6).

<details><summary>Screenshots</summary>

### Error Message:
![image](https://github.com/department-of-veterans-affairs/va.gov-cms/assets/108767736/12a08e94-52f3-4213-9871-1626df43c1e8)

### Editorial Workflow does not populate:
![image](https://github.com/department-of-veterans-affairs/va.gov-cms/assets/108767736/e2c5a8c8-73c4-467f-8d70-ec4aefc64dec)
</details>

## Acceptance criteria

>Consider
> - Design / Accessibility reviews
> - Collab cycle requirements
> - Device sizes (mobile first)
> - Documentation updates / Change management
< - [Content model documentation](https://prod.cms.va.gov/admin/structure/cm_document)
> - Testing notes

- [x] Revision log field shows and is required in Promo blocks, [example](https://prod.cms.va.gov/block/6)
- [ ] Checkbox to optionally include a Revision log should be hidden / removed
- [x] Spelunked codebase to look for any similar block types that might need to be fixed to have a required Revision log

chri5tia:
https://github.com/department-of-veterans-affairs/va.gov-cms/pull/15243 addresses the first AC and is ready to merge. I need to put some more thought or planning into the other two ACs. We need to decide whether or not to hold on the PR or push it. It would be nice if this didn't have to wait a day while I look into the other ACs, which require enough testing to delay this.

jilladams:
Push it!

chri5tia:
Done. I suggest moving the last two ACs to a new ticket and assigning it to me. Call it a 5.

jilladams:
@chri5tia counter-proposal: 
* Your current PR closes #15187 
* When merged, you check off the 1st AC here
* We keep this ticket for the remaining 2 ACs
Less ticketing busy work.

chri5tia:
I am noticing that the CMS announcement block doesn't have a revision log message and seems broken, but shouldn't be affected by my change.
https://prod.cms.va.gov/block/add/cms_announcement

The third AC is done also.

chri5tia:
[...]

Issue #15187, opened 2023-09-11
Publishing custom block results in Error Message - required hidden fields


TroyA6:
## Describe the defect
Saving changes to page: https://prod.cms.va.gov/block/6 results in the following message: 'Please fill out this field', but there is no empty field displayed -- Perhaps the Editorial Workflow / Revision Log field is hidden, as it does not seem to populate correctly.
 
## To Reproduce
1. Go to 'https://prod.cms.va.gov/block/6'
2. Click on 'Publish' in the status section
3. See error

## AC / Expected behavior
When saving changes in any of the three statuses, Editor should be required to complete a Revision Log message, and the changes should be saved successfully.

### Error Message:
![image](https://github.com/department-of-veterans-affairs/va.gov-cms/assets/108767736/12a08e94-52f3-4213-9871-1626df43c1e8)

### Editorial Workflow does not populate:
![image](https://github.com/department-of-veterans-affairs/va.gov-cms/assets/108767736/e2c5a8c8-73c4-467f-8d70-ec4aefc64dec)

 ## Additional context
This is preventing a the page update, which may be time sensitive.

## Labels
(You can delete this section once it's complete)
- [x] Issue type (red) (defaults to "Defect")
- [ ] CMS subsystem (green)
- [ ] CMS practice area (blue)
- [x] CMS workstream (orange) (not needed for bug tickets)
- [ ] CMS-supported product (black)

### Team
Please check the team(s) that will do this work.

- [ ] `CMS Team`
- [ ] `Public Websites`
- [x] `Facilities`
- [ ] `User support`
- [ ] `Accelerated Publishing`

TroyA6:
Associated JIRA Help Ticket [VAHELP-6355](https://va-gov.atlassian.net/jira/servicedesk/projects/VAHELP/queues/custom/22/VAHELP-6355)

ndouglas:
I think this might've been broken by #14842.

If I visit [this preview environment](https://pr14753-kcqw2tduyeeldmgxulxphjyhxf0ymatu.ci.cms.va.gov/block/6), last touched August 16:

![Screenshot 2023-09-11 at 12 58 59 PM](https://github.com/department-of-veterans-affairs/va.gov-cms/assets/1318579/ecae2b39-3f8e-4bcb-85f3-a355a040b446)

then the revision log message field appears as normal:

[...]

PR #15355, opened 2023-09-22
VACMS-15242: Make Revision log message requirement more inclusive

chri5tia:
## Description

Relates to #15242 
#15187 

~~Revises the workflow module to hide the revision checkbox on promo block types as well as the other custom block types, without unintentionally hiding the required revision log message field on that type.~~

Updates the promo block architecture to match the other custom block types by adding a field set for section settings and moving revision_log, section and workflow state into it.

## Testing done
local functional

## Screenshots
Screenshot 2023-09-27 at 1 20 01 PM

## QA steps

As a user with any role, check both promo and non-promo blocks that a) revision log message works, is required and the checkbox is hidden.

### Definition of Done

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
+++ b/docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
@@ -277,9 +277,6 @@ class EntityEventSubscriber implements EventSubscriberInterface {
       return;
     }
 
-    // Disable the checkbox functionality.
-    unset($form['revision_log']['#states']);
-
     // Hide the checkbox that lets you opt into making a revision.
     // Promo blocks have a different form structure.
     if ($entity->bundle() === 'promo') {
--- a/docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
+++ b/docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
@@ -265,26 +265,37 @@ class EntityEventSubscriber implements EventSubscriberInterface {
    * @param string $form_id
    *   The form id.
    */
-  public function requireRevisionMessage(array &$form, FormStateInterface &$form_state, $form_id) {
-    // Check if the block-content form is of type promo and exit (#15187)
+  public function requireRevisionMessage(array &$form, FormStateInterface &$form_state, $form_id): void {
+    // See what kind of form we are dealing with and do some safety checks.
     $formObject = $form_state->getFormObject();
-    if ($formObject instanceof EntityFormInterface) {
-      $entity = $formObject->getEntity();
-      if ($entity instanceof BlockContentInterface) {
-        if ($entity->bundle() === 'promo') {
-          return;
-        }
-      }
+    if (!$formObject instanceof EntityFormInterface) {
+      return;
     }
 
+    $entity = $formObject->getEntity();
+    if (!$entity instanceof BlockContentInterface) {
+      return;
+    }
+
+    // Disable the checkbox functionality.
+    unset($form['revision_log']['#states']);
+
+    // Hide the checkbox that lets you opt into making a revision.
+    // Promo blocks have a different form structure.
+    if ($entity->bundle() === 'promo') {
+      $form['revision']['#access'] = FALSE;
+    }
+    else {
+      $form["revision_information"]["#attributes"]['class'][] = 'visually-hidden';
+      $this->bypassRevisionLogValidationOnIef($form, $form_state);
+    }
+
+    // Make revision log required.
     $form['revision_log']['#required'] = TRUE;
     $form['revision_log']['widget']['#required'] = TRUE;
     $form['revision_log']['widget'][0]['#required'] = TRUE;
     $form['revision_log']['widget'][0]['value']['#required'] = TRUE;
     $form['#validate'][] = '_va_gov_workflow_validate_required_revision_message';
-    // Hide the checkbox that lets you opt into making a revision.
-    $form["revision_information"]["#attributes"]['class'][] = 'visually-hidden';
-    $this->bypassRevisionLogValidationOnIef($form, $form_state);
   }
 
   /**
--- a/docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
+++ b/docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
@@ -2,7 +2,6 @@
 
 namespace Drupal\va_gov_workflow\EventSubscriber;
 
-use Drupal\block_content\BlockContentInterface;
 use Drupal\Core\Entity\ContentEntityForm;
 use Drupal\Core\Entity\EntityFormInterface;
 use Drupal\Core\Entity\EntityInterface;
@@ -266,27 +265,22 @@ class EntityEventSubscriber implements EventSubscriberInterface {
    *   The form id.
    */
   public function requireRevisionMessage(array &$form, FormStateInterface &$form_state, $form_id): void {
-    // See what kind of form we are dealing with and do some safety checks.
+    // Hide the checkbox that lets user opt into making a revision.
     $formObject = $form_state->getFormObject();
-    if (!$formObject instanceof EntityFormInterface) {
-      return;
-    }
-    $entity = $formObject->getEntity();
-    if (!$entity instanceof BlockContentInterface) {
-      return;
-    }
-
-    // Hide the checkbox that lets you opt into making a revision.
-    // Promo blocks have a different form structure.
-    if (in_array($entity->bundle(), ['promo', 'cms_announcement'])) {
-      $form["revision_information"]["#access"] = TRUE;
-      unset($form['revision_log']['#states']);
-      $form["revision"]["#access"] = FALSE;
-      $form["revision_log"]["#access"] = TRUE;
-    }
-    else {
-      $form["revision_information"]["#attributes"]['class'][] = 'visually-hidden';
-      $this->bypassRevisionLogValidationOnIef($form, $form_state);
+    if ($formObject instanceof EntityFormInterface) {
+      $entity = $formObject->getEntity();
+      if ($entity->bundle() === 'promo') {
+        // Hide revision checkbox on Promo blocks.
+        $form["revision_information"]["#access"] = TRUE;
+        unset($form['revision_log']['#states']);
+        $form["revision"]["#access"] = FALSE;
+        $form["revision_log"]["#access"] = TRUE;
+      }
+      else {
+        // Hide revision checkbox on all other custom blocks.
+        $form["revision_information"]["#attributes"]['class'][] = 'visually-hidden';
+        $this->bypassRevisionLogValidationOnIef($form, $form_state);
+      }
     }
 
     // Make revision log required.
--- a/docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
+++ b/docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
@@ -264,31 +264,15 @@ class EntityEventSubscriber implements EventSubscriberInterface {
    * @param string $form_id
    *   The form id.
    */
-  public function requireRevisionMessage(array &$form, FormStateInterface &$form_state, $form_id): void {
-    // Hide the checkbox that lets user opt into making a revision.
-    $formObject = $form_state->getFormObject();
-    if ($formObject instanceof EntityFormInterface) {
-      $entity = $formObject->getEntity();
-      if ($entity->bundle() === 'promo') {
-        // Hide revision checkbox on Promo blocks.
-        $form["revision_information"]["#access"] = TRUE;
-        unset($form['revision_log']['#states']);
-        $form["revision"]["#access"] = FALSE;
-        $form["revision_log"]["#access"] = TRUE;
-      }
-      else {
-        // Hide revision checkbox on all other custom blocks.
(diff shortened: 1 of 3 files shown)


chri5tia:
@swirtSJW We'll have to rebuild this when the tugboat issues are resolved (I see that nothing is being deployed/environment isn't being built) but wondering if you can pull down the branch and test it and/or take a look at the code revision and give me any feedback or preferences. Thanks!

docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -265,26 +265,37 @@ public function entityDelete(EntityDeleteEvent $event): void {
...
+      return;
+    }
+
+    $entity = $formObject->getEntity();
+    if (!$entity instanceof BlockContentInterface) {
+      return;
     }
 
+    // Disable the checkbox functionality.
+    unset($form['revision_log']['#states']);
```
swirtSJW, line 281: This seems potentially overreaching to obliterate all states.  Is there way to be more targeted about which state is getting removed?  


Also this seems like it might be an upstream bug.  If this checkbox is set, 
image  
Users should  not be presented with the option to opt out of creating a new revision.

If it is not an upstream issue, it makes sense (to me anyway) that the logic we employ ought to rely on that setting across all block bundles.

chri5tia:
Cypress failure is the media function, which is unrelated. Still trying to figure out what to do about that.

https://cms-7blfsvkcnwkjikm4zu3iee5ko5eheqhn.ci.cms.va.gov/cypress/screenshots/actual/platform/media.feature/

Screenshot 2023-09-22 at 10 04 32 AM

.tugboat/config.yml
```diff
@@ -177,6 +177,7 @@ services:
         # Currently, resetting to the remote branch/PR SHA since [Task](https://github.com/go-task/task) runs tests status against this commit.
         # Tugboat currently merges the branch into master, and we don't need that because we have a GitHub branch rule
         # setup to ensure branch is up-to-date.
+        - git rm --cached -r . && git reset --hard
```
chri5tia, line 180: I'll remove this before merging.

docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
chri5tia: Thank you.

chri5tia:
I noticed that the `cms_announcemet` block type was also affected by the defect reported in #15187 

Engineering notes to incorporate the fix:

`dpm($form)` shows the revision log structure for this bundle as:

```
  "revision_log" => array:9 [▼
    "#type" => "container"
    "#parents" => array:1 [▶]
    "#attributes" => array:1 [▼
      "class" => array:3 [▼
        0 => "field--type-string-long"
        1 => "field--name-revision-log"
        2 => "field--widget-string-textarea"
      ]
    ]
    "widget" => array:13 [▶]
    "#access" => true
    "#weight" => 25
    "#cache" => array:3 [▶]
    "#group" => "revision_information"
    "#states" => array:1 [▼
      "visible" => array:1 [▼
        ":input[name="revision"]" => array:1 [▼
          "checked" => true
        ]
      ]
    ]
  ]
```

`alert`, `cta_with_link` block, `benefit_promo`, `news_promo` work with the following previously-implemented alter:

```
$form["revision_information"]["#attributes"]['class'][] = 'visually-hidden';
```

docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
chri5tia: I seem to only be able to get this to work on promo and cms_announcement if I unset the revision_log state (checkbox), because changing access to false makes the log disappear, even if I tell it access true. But I have moved unsetting the state to just these two blocks.

I think this is because of how we add the revision block on the config sync as part of the Editorial workflow group when it isn't actually. I'm not sure though.

docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -265,26 +265,36 @@ public function entityDelete(EntityDeleteEvent $event): void {
...
+    if (!$formObject instanceof EntityFormInterface) {
+      return;
+    }
+    $entity = $formObject->getEntity();
+    if (!$entity instanceof BlockContentInterface) {
+      return;
     }
 
+    // Hide the checkbox that lets you opt into making a revision.
+    // Promo blocks have a different form structure.
```
swirtSJW, line 280: Minor:  Your comment and your code are in disagreement.  Your comment says promo blocks are different, while your code says promo and cms_announcement are different.

I suggest altering your comment to say some blocks are different.... or better yet, explain what about them is different.

swirtSJW: Also cms announcement block is probably not worth addressing since it is a dead block
https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15419

tests/cypress/support/main_content_blocks.js
```diff
@@ -51,5 +49,5 @@ Cypress.Commands.add("addMainContentBlockWithFile", (type) => {
     `[Test Data] ${faker.lorem.sentence()}`,
     { force: true }
   );
-  cy.get("form.node-form").find("input#edit-submit").click();
+  // cy.get("form.node-form").find("input#edit-submit").click();
```
swirtSJW, line 52: If we are commenting this out, can we comment on why it is being commented out and what it might take to comment it back in?

docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -265,26 +265,36 @@ public function entityDelete(EntityDeleteEvent $event): void {
...
-        if ($entity->bundle() === 'promo') {
-          return;
-        }
-      }
+    if (!$formObject instanceof EntityFormInterface) {
+      return;
+    }
+    $entity = $formObject->getEntity();
+    if (!$entity instanceof BlockContentInterface) {
+      return;
```
swirtSJW, line 276: Critical:  This method 'requireRevisionMessage()' Is being called on to operate on both blocks and nodes.  It is called in alterBlockContentForm() and alterNodeForm()   but with this one check we are bailing out of the function if it is something other than a block.  So it is no longer operating on nodes.   I am pretty sure this is a breaking change.

swirtSJW (changes requested):
I think this has one major breaking change that has to be remedied.

tests/cypress/support/main_content_blocks.js
chri5tia: My understanding was that saving the form during this test is not desirable. @ndouglas, can you confirm? Test fails because it is looking for form elements after it saves, which is odd, but I don't want to remove critical functionality of the test if I misunderstood feedback from earlier.

docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
chri5tia: Removing this all together.

tests/cypress/support/main_content_blocks.js
swirtSJW: If we should not be saving during the test, why leave the code commented out?  The commented out code just becomes cruft.  Usually we only comment out code if we are leaving breadcrumbs to come back to later.

docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -266,27 +265,22 @@ public function entityDelete(EntityDeleteEvent $event): void {
...
-      $form["revision"]["#access"] = FALSE;
-      $form["revision_log"]["#access"] = TRUE;
-    }
-    else {
-      $form["revision_information"]["#attributes"]['class'][] = 'visually-hidden';
-      $this->bypassRevisionLogValidationOnIef($form, $form_state);
+    if ($formObject instanceof EntityFormInterface) {
+      $entity = $formObject->getEntity();
+      if ($entity->bundle() === 'promo') {
+        // Hide revision checkbox on Promo blocks.
```
swirtSJW, line 273: It would still be good to have some breadcrumb left here in a comment that describes why promos are different from all other blocks.  I've been on this PR for a while now and it is still not clear to me why promos are fundamentally different than our other blocks.  If present-me can't connect the dots, there is no way future-me is going to .   We need to be kind to our future selves.   ;)

swirtSJW: Rather than create exceptions, would it be better for promo blocks to be made to be just like other blocks?

tests/cypress/support/main_content_blocks.js
```diff
@@ -51,5 +49,4 @@ Cypress.Commands.add("addMainContentBlockWithFile", (type) => {
     `[Test Data] ${faker.lorem.sentence()}`,
     { force: true }
   );
-  cy.get("form.node-form").find("input#edit-submit").click();
```
swirtSJW, line 54: Nice.  This makes good sense.  The block test should not be responsible for saving the node.

docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -265,26 +264,31 @@ public function entityDelete(EntityDeleteEvent $event): void {
...
+  public function requireRevisionMessage(array &$form, FormStateInterface &$form_state, $form_id): void {
+    // Hide the checkbox that lets user opt into making a revision.
     $formObject = $form_state->getFormObject();
     if ($formObject instanceof EntityFormInterface) {
       $entity = $formObject->getEntity();
-      if ($entity instanceof BlockContentInterface) {
-        if ($entity->bundle() === 'promo') {
-          return;
-        }
+      if ($entity->bundle() === 'promo') {
```
swirtSJW, line 272: Minor:  You had good instincts before to make this check specific to blocks named promo.   We don't have a node named promo or a menu named promo but it is a simple machine name that could easily appear so it makes sense to keep your original check.
```suggestion
      if (($entity instanceof BlockContentInterface) && ($entity->bundle() === 'promo')) {
```

docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
```diff
@@ -2,7 +2,6 @@
 
 namespace Drupal\va_gov_workflow\EventSubscriber;
 
-use Drupal\block_content\BlockContentInterface;
 use Drupal\Core\Entity\ContentEntityForm;
```
swirtSJW, line 5:
```suggestion
use Drupal\block_content\BlockContentInterface;
use Drupal\Core\Entity\ContentEntityForm;
```

See the change I suggested below that will require this to be added.

swirtSJW (commented):
This looks much better, functionally.  It is no longer a breaking change.   A couple of minor suggestions.

tests/cypress/support/main_content_blocks.js
swirtSJW: Your new commit resolves this.

swirtSJW:
> I noticed that the `cms_announcement` block type was also affected by the defect reported in #15187
> 

CMS announcement is a dead block type that should be removed by CMS Team https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15419

docroot/modules/custom/va_gov_workflow/src/EventSubscriber/EntityEventSubscriber.php
chri5tia: I am checking to see if there is just a field group in all but the promo block that could easily be put in.

chri5tia:
@swirtSJW Creating a fieldset and moving the revision_log in config to match the other blocks fixed the issue without the need to change what you had in the event subscriber originally. *epic facepalm*

Thanks for working through this ticket with me; I learned a lot!

swirtSJW (changes requested):
Code wise this looks good.
I am seeing a difference in the order of the fields on promo VS other blocks.   The order on the promo seems a little scrambled.

image

I think we want those to be consistent, and the alert seems to make more sense the order of 

- section
- moderation state
- revision log message

swirtSJW:
The QA steps were a little loose, so tightening them up a bit makes it easier on reviewers

1. As  an admin [edit a promo block](https://pr15355-7blfsvkcnwkjikm4zu3iee5ko5eheqhn.ci.cms.va.gov/block/263?destination=/admin/structure/block/block-content) 
   - [ ] Validate that there is no checkbox option to not create a revision.
   - [ ] Validate that revision log message is required.
   - [ ] Validate the field order is
      - section
      - workflow
      - revision log message
2. Visit another block type ([alert block](https://pr15355-7blfsvkcnwkjikm4zu3iee5ko5eheqhn.ci.cms.va.gov/block/256?destination=/admin/structure/block/block-content)) to make sure it is unaffected 
   - [ ] Validate that there is no checkbox option to not create a revision.
   - [ ] Validate that revision log message is required.
   - [ ] Validate the field order is
      - section
      - workflow
      - revision log message

swirtSJW (approved):
This tests out.  Thank you.

PR #15375, opened 2023-09-25
setup advancedqueue_runner

### LINKED ISSUE #8108: Move responsibility for kicking off content build to the CMS
## Background

We've spent a lot of time figuring out how to deduplication build requests, coordinate between manual builds (when banners or status nodes are published) and scheduled builds, etc. Because this coordination is happening across two different systems, it's extremely complicated, fragile, and error prone.

Let's move the triggering logic out of the content build workflow into the CMS so that Drupal is entirely in control of when content build is triggered.

### Acceptance Criteria

- [ ] Add a cron job that runs every minute
- [x] Ensure builds are still requested at the top of the hour
- [x] Ensure builds are requested when an editor creates/updates banner or status nodes
- [x] Make sure we're saving timestamps somewhere for:
    * the last time the build was requested in the CMS, 
    * the last time the build request was processed (GHA API call was fired),
    *  and the last ~time the govdelivery callback notified us that the build was complete~ timestamp passed to us in the govDelivery callback.

### Tasks

- [ ] Debounce build requests so that we're not clogging up the GHA queue
- [ ] If there's a build request when the every minute tasks run, fire the GHA API request

In the future, we could look at:
* Run the builds more frequently (potentially back to back)
* Ignore manual build requests during business hours in favor of frequent scheduled/automatic builds and only allow manual builds to go through after hours
* Push events to Datadog when builds are requested + when build requests are processed

### CMS Team

- [x] `Platform CMS Team`

### LINKED ISSUE #8849: Set up advancedqueue_runner on all envs
https://www.drupal.org/project/advancedqueue_runner

After #8108, we're leaning more heavily on advancedqueue. It would be awesome to not have a once per minute cron job and instead have a persistent background worker that we can farm out tasks to.

### CMS Team

- [x] `Platform CMS Team`
- [ ] `Sitewide program`
- [ ] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

ariperez:
closes #8849

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
         "drupal/admin_feedback": "^2.2",
         "drupal/admin_toolbar": "^3.3",
         "drupal/advancedqueue": "^1.0@RC",
+        "drupal/advancedqueue_runner": "^2.0@alpha",
         "drupal/allow_only_one": "1.0.2",
         "drupal/allowed_formats": "^2.0",
         "drupal/animated_gif": "^2.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "f5b00d1a3689d04ec1d97aa12c844755",
+    "content-hash": "1b1d4d2d037cfab1189b5a244aade82b",
     "packages": [
         {
             "name": "alchemy/zippy",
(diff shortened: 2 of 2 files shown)


ariperez:
@ndouglas , I'm confirming these are the correct directories....
image


the process kept failing silently until I kept the above directories, and modified the following line in the local package:
https://git.drupalcode.org/project/advancedqueue_runner/-/blob/2.x/src/Scripts/jobs.php?ref_type=heads#L17

becomes

$autoloader = require $_SERVER['PWD'] . '/vendor/autoload.php';

ariperez:
@ndouglas , I'm confirming these are the correct directories....
image


the process kept failing silently until I kept the above directories, and modified the following line in the local package:
https://git.drupalcode.org/project/advancedqueue_runner/-/blob/2.x/src/Scripts/jobs.php?ref_type=heads#L17

becomes

$autoloader = require $_SERVER['PWD'] . '/vendor/autoload.php';

ariperez:
@ndouglas , also, do you have a quick way to test this with a job that's prebuilt?  having a bit of trouble trying to kick a test one off that increments the queue

ndouglas:
@ariperez Yeah, we currently have `vendor` located within `docroot`. That's not best practices, and almost no one does it that way, but the decision was made before I was here and I honestly don't want to fight with it at this point.

We can use `composer-patches` to patch the local version of that file. See e.g. [here](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/main/patches/3380997-fix-textfield_counter-respect-count-only-mode.patch) for an example.

Screenshot 2023-09-25 at 9 49 22 AM

>also, do you have a quick way to test this with a job that's prebuilt? having a bit of trouble trying to kick a test one off that increments the queue

Hmm, nope. What issue are you encountering? 

I'll poke around with this a bit, if you don't mind.

**Note**: You're referencing #8108 but I think you want #8849.

ndouglas:
(converted to draft since we're still evaluating)

I'm... eh... not crazy about this module so far. It seems really hard to inspect and debug -- the silent failure you mentioned and the fact that it's apparently set up correctly but isn't logging any errors to either the Drupal watchdog log or PHP logs. I can't find any output anywhere 🤔 

I'm not enthusiastic about risking a working-but-suboptimal system for this module -- I kinda dread something going wrong and us having to debug it live on prod. And a few months from now, the suboptimal system will be gone anyway.

But that's fine, because the ticket as revised was for discovery anyway 🙂 

I would like to close this PR and consider this a possibility that we checked out but did not find anything worth pursuing, unless you disagree.

ariperez:
@ndouglas , I won't lie that this module did stress me out lol. 

I concur that the right decision is probably not to implement atm.  It's not battle hardened pro code yet.  

At least I learned alot about the project, excited for the next ticket!

PR #15464, opened 2023-09-28
VACMS-000 Remove Pittsburgh VAMC as default value on VAMC operating status node

davidmpickett:
## Description

Currently when a user goes to [create a new VAMC System Operating Status](https://prod.cms.va.gov/node/add/vamc_operating_status_and_alerts), the VAMC system field defaults to VA Pittsburgh. This is a legacy configuration from VAMC pilot. This should default to none. 

## Screenshots

Screenshot 2023-09-28 172921

### DIFF
--- a/config/sync/field.field.node.vamc_operating_status_and_alerts.field_office.yml
+++ b/config/sync/field.field.node.vamc_operating_status_and_alerts.field_office.yml
@@ -16,9 +16,7 @@ label: 'VAMC system'
 description: 'To do: hide this field or disable it after node creation. '
 required: true
 translatable: false
-default_value:
-  -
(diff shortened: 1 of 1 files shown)


davidmpickett:
Verified on tugboat that the default is correctly showing up
Screenshot 2023-09-28 175020

swirtSJW (approved):
Nicely done Mr Pickett.

swirtSJW:
This is a good fix.  We don't have an issue against this but it is important to note that no editor has the option to add this page
image

These would only be created by an admin or content-admin when we stand up a new system.  These were all originally created by the VAMC upgrade team.

@davidmpickett first config contribution should be celebrated though. :man_dancing: 

Approved but won't merge until monday.

PR #15530, opened 2023-10-03
VACMS-15529 Remove duplicative field menu item.

### LINKED ISSUE #15529: Remove Field list link from one of two spots in the menu
## Describe the defect
There is a "Fields" link in the reports menu that points to the same place as the content model fields list.
image

This is a remnant because we use a View for that list and the upstream module can not manage that View once the module is installed.
## To Reproduce
1. Pull down the reports menu.
2. See the two menu items that both take you to the same place (see screenshot)

## AC / Expected behavior
- [ ] The "Fields" menu item not longer exists
- [ ] The content model fields menu item still exists.

swirtSJW:
## Description

closes #15529

## QA steps

As an admin (they are the only users who can see this)
1. [login to the site](https://pr15530-aajkwumdnteo2u1jgy3dgwvtkg5nv9hx.ci.cms.va.gov/)
2. Pull down the reports menu.
   - [x] Validate that the `Reports > Fields` no longer exists as a menu item
   - [x] Validate that `Reports > Content model > Content model fields` does exist
   
image

### DIFF
--- a/config/sync/views.view.content_model_fields.yml
+++ b/config/sync/views.view.content_model_fields.yml
@@ -2,8 +2,6 @@ uuid: 9cce4e26-7273-40be-99c0-9fb19a6823c3
 langcode: en
 status: true
 dependencies:
-  config:
-    - system.menu.admin
   module:
     - config_views
     - content_model_documentation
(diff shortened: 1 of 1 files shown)


omahane (approved):
Checked the PR environment against prod. The change is good to go.

swirtSJW:
image

Issue #14215, opened 2023-06-29
Push mental health phone number field to lighthouse with the status push.


kmariepat-cityfriends:
Note that we cannot merge and perform the push outlined in the below AC until Lighthouse is ready to accept.  PRs for this should be against the integration-mental-health-phone branch.

- [x] I expect that the mental health phone number field is pushed to lighthouse with the status push
   - [x] If the mental_health_contact_phone is populated, send that.
   - [x] If it is not populated send  field_mental_health_phone  (this field will be slated for removal after full compliance with populating mental_health_contact_phone has been achieved.
- [ ] ~I expect the content model documentation is updated to show that it's a push and not a pull~ Moved to https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15576 because it can't be documented until it lands on prod.

jilladams:
VHA DM needs to buy into options outlined in #. Then we will work with Lighthouse to define the shape of the data push.

mmiddaugh:
This is blocked pending Changes to existing field field_mental_health_phone on VAMC facilities, change management and final signoff by OMHSP

PR #15688, opened 2023-10-16
VACMS-14215 VAMC mental health phone data push

swirtSJW:
## Description

Relates to #14215

## QA steps

As `QA Content Publisher` 
1. Test push with no mental health phone Edit this https://pr15688-9qhcwttn1a2ppwra05saolrc3vvoczq4.ci.cms.va.gov/node/2181/edit
2. Change the status and publish the change
   - [x] Go to  https://pr15688-9qhcwttn1a2ppwra05saolrc3vvoczq4.ci.cms.va.gov/admin/config/post-api/queue and validate that the push for this facility has no mental health number entry
3. Test push with only API sourced mental health phone: edit this https://pr15688-9qhcwttn1a2ppwra05saolrc3vvoczq4.ci.cms.va.gov/node/2124/edit
4. change nothing and just publish the save.
   - [x] Go to  https://pr15688-9qhcwttn1a2ppwra05saolrc3vvoczq4.ci.cms.va.gov/admin/config/post-api/queue and validate that the push for this facility has an entry in the core payload for  "mental health phone"
5. Test push with mental health contact (editor sourced) mental health phone: edit the node and for Mental Health Contact phone number enter `555-444-1111`
6. Publish and save the node.
   - [x] Go to  https://pr15688-9qhcwttn1a2ppwra05saolrc3vvoczq4.ci.cms.va.gov/admin/config/post-api/queue and validate that the push for this facility has an entry in the core payload for  "mental health phone" that matches the `555-444-1111` number you entered.

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/docroot/modules/custom/va_gov_post_api/src/Service/PostFacilityStatus.php
+++ b/docroot/modules/custom/va_gov_post_api/src/Service/PostFacilityStatus.php
@@ -172,6 +172,9 @@ class PostFacilityStatus extends PostFacilityBase implements PostServiceInterfac
           'additional_info' => (strtoupper($this->statusToPush) != 'NORMAL') ? $this->additionalInfoToPush : NULL,
         ],
       ];
+      if ($this->getFacilityMentalHealthPhone()) {
+        $payload['core']['mental_health_phone'] = $this->getFacilityMentalHealthPhone();
+      }
 
       $this->addSupplementalStatus($payload);
       $this->getRelatedSystemInfo($payload);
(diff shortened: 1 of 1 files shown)


swirtSJW:
No phone
image


API sourced phone
image


Editor sourced phone

swirtSJW:
image

Pushed and received without error.   The phone number is not processed to anything but at least is not causing a push failure.

docroot/modules/custom/va_gov_post_api/src/Service/PostFacilityStatus.php
```diff
@@ -400,4 +403,37 @@ protected function shouldPushSystem(NodeInterface $entity) {
...
+  /**
+   * Retrieves a field and returns the value if there is one or empty string.
+   *
+   * @param string $field_name
+   *   The field name to get.
+   *
+   * @return string
+   *   The field value if it exists, empty string otherwise.
+   */
+  protected function getFieldSafe(string $field_name): string {
```
omahane, line 432: This is very cool.

Not that it's urgent or even necessary, but we have ternary operators around the code that could be refactored with this. This would be cleaner, plus, it would consistently send an empty string. Not sure if we want that, though.

omahane (approved):
I had a bit of a time with the QA steps because the user QA Content Publisher didn't have rights to edit all the nodes, but I made do. 

The code worked as expected.

swirtSJW:
> I had a bit of a time with the QA steps because the user QA Content Publisher didn't have rights to edit all the nodes, but I made do.

Oyy :(  Sorry yes I think we need a new QA "person" that has both VAMC roles, because this is impossible to test with with either of out existing QA VAMC roles because one can't publish and the other can't edit.

docroot/modules/custom/va_gov_post_api/src/Service/PostFacilityStatus.php
swirtSJW: Yah I agree.   I think some of the ternaries are fine as they are but this one had to both see if the field exists AND if it had a value.  It is also only for strings so it has to be used cautiously.

PR #15780, opened 2023-10-23
VACMS-15779: Set `rel="noopener"` in preview links.

### LINKED ISSUE #15779: Set `rel="noopener"` in links that target `_blank`.
## Description

See [this thread](https://dsva.slack.com/archives/CDHBKAL9W/p1697059698606299). Editors are reporting that they encounter 404s when they click preview links.

This seems to be reproducible for me in CAG, but not in SOCKS.

After playing with it, I believe it's because of an interaction between the `Upgrade-Insecure-Requests` header is set to 1, and our `target="_blank"` links by default refer to the security context of the referring page. I don't know why this causes problems in CAG and not locally, but it might be because of a result of security policies or something else beyond my understanding.

Adding the `rel="noopener"`, so that the security context of the referring tab is not reused by the new tab, seems to fix this for me. Doing that is best practices anyway.

## Acceptance Criteria

- [ ] Preview buttons have an additional `rel="noopener"` attribute.

ndouglas:
Closes #15779.

Screenshot 2023-10-24 at 10 17 11 AM

## QA Steps

- [ ] Find a node view somewhere on the site that has a preview link
- [ ] Verify that the link contains a `rel="noopener"` attribute.

### DIFF
--- a/docroot/modules/custom/va_gov_backend/va_gov_backend.module
+++ b/docroot/modules/custom/va_gov_backend/va_gov_backend.module
@@ -1011,7 +1011,7 @@ function va_gov_backend_preprocess_page(&$variables) {
         $host = \Drupal::request()->getHost();
         $preview_form = new PreviewForm();
         $url = $preview_form->getEnvironment($host, $nid);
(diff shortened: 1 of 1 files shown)


edmund-dunn (approved):
Simply brilliant! Well done!

Issue #13529, opened 2023-04-27
Update Events CMS feature toggle to expose Events checkbox to a specific set of usernames


kmariepat-cityfriends:
## User story
As a product owner, I want to ship an update to a small group of editors to verify it is easy to use, before rolling out change management to all Events editors.

## Description
Expand the CMS Feature Toggle binary on/off for National Events calendar checkbox to allow the feature to be visible to a subset of users.

List of emails is available in DSVA slack: https://dsva.slack.com/archives/C52CL1PKQ/p1697148973755669?thread_ts=1697148950.813469&cid=C52CL1PKQ

##### Acceptance Criteria 
- [x] Events CMS feature toggle, when enabled, will only expose the feature to specific Editors listed, and Drupal Admins

jilladams:
@dsasser @chri5tia this ticket is regarding this earlier [thread in Slack](https://dsva.slack.com/archives/C52CL1PKQ/p1696610411984349?thread_ts=1696434452.043579&cid=C52CL1PKQ) , around the idea that the Events-related feature toggle could be expanded to enable display to a subset of users. 

Dave will be having some conversations in the next couple of days that will help us know if this is required. If so, it's likely to be that we expand to show to a set of explicit users by username/email. If that's required, we will need to inject into sprint, and it will slot in after the Events related sprint goal. (It will deprioritize the Notifications framework discovery). 

Please let me know if you have questions / concerns. A stab at points would also be much appreciated, based on the assumption we'll be enabling for a list of ~3-10 usernames.

dsasser:
> @dsasser @chri5tia this ticket is regarding this earlier [thread in Slack](https://dsva.slack.com/archives/C52CL1PKQ/p1696610411984349?thread_ts=1696434452.043579&cid=C52CL1PKQ) , around the idea that the Events-related feature toggle could be expanded to enable display to a subset of users.
> 
> Dave will be having some conversations in the next couple of days that will help us know if this is required. If so, it's likely to be that we expand to show to a set of explicit users by username/email. If that's required, we will need to inject into sprint, and it will slot in after the Events related sprint goal. (It will deprioritize the Notifications framework discovery).
> 
> Please let me know if you have questions / concerns. A stab at points would also be much appreciated, based on the assumption we'll be enabling for a list of ~3-10 usernames.

[...]

PR #15964, opened 2023-11-01
Add new Editors to beta per VHA DM

jilladams:
## Description

Relates to #13529

Somewhere wires got crossed between Facilities PO discussion with VHA DM, list of users that was provided to Public Websites, and list of Editors that VHA DM notified to try the new feature. This diff represents users who were notified by VHA DM, but not in our list, so do not currently have access to the feature. Adding them so they do and can test.


dsasser (approved):

jilladams:
QQ: once this merges, will it have to go through CMS nightly build to take effect in Prod? Guessing so but want to confirm.

dsasser:
> QQ: once this merges, will it have to go through CMS nightly build to take effect in Prod? Guessing so but want to confirm.

Not the nightly build, but the daily CMS deploy, so it will go live at ~12pm tomorrow 11/2/23.

Issue #13834, opened 2023-05-25
Minimize PR overhead for non-code changes.


ndouglas:
## Description

As someone who hates that we don't have all of the branch protection rules enabled on `main`,
I should prevent Tugboat deploys and GHA workflow runs on documentation-only PRs
So that people don't look like Walter Donovan in _Indiana Jones and the Last Crusade_ when he drank from the fake Grail while waiting for completely unnecessary test runs to complete.

## Acceptance Criteria
- [ ] Tugboat deploy doesn't happen on documentation-only changes
- [ ] GHA workflows don't happen on documentation-only changes.

ndouglas:
This could turn out to be a major PITA/time sink.

JunTaoLuo:
FYI, `paths-ignore` should be very helpful here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths.

BerniXiongA6:
@JunTaoLuo how many points of this ticket would you say were completed so far? We can carry over ticket if needed (I can calculate the points manually on my planning calculator). cc: @maortiz-27-80

JunTaoLuo:
I think there's still 5 points left here since I've done about half of the work but the half remaining seems to be trickier.

BerniXiongA6:
Thanks @JunTaoLuo ! cc: @maortiz-27-80

Issue #14345, opened 2023-07-13
DevOps Engineers Require Secure Access to CMS Github Bot Account


olivereri:
## Description
CMS Platform owns and manages one [GitHub “Bot” account](https://github.com/va-cms-bot) which is used for various procedures in the environments. Generally, this is a Github service account used for automation. The credentials for these accounts are stored in AWS Systems Manager Parameter Store.

To access these accounts, you will need a YubiKey to add to the account.

## Acceptance Criteria
- [x] CMS Team members requiring access to VA-CMS-BOT access identified.
- [x] Yubikey Security Keys purchased.
- [x] Security Keys added to VA-CMS-BOT Account.
- [x] Documentation is updated or created to support this process.

## Implementation Details
- [Recommended Security Key](https://www.yubico.com/product/yubikey-5-nano/)
- [Procedures to add a Security Key to an account](https://docs.github.com/en/authentication/securing-your-account-with-two-factor-authentication-2fa/configuring-two-factor-authentication#configuring-two-factor-authentication-using-a-security-key)
- [General Github Bot Platform Guidance](https://vfs.atlassian.net/wiki/spaces/OT/pages/1824588025)
- VA-CMS-Bot credentials and other related secrets can be found in DSVA AWS Parameter store **/cms/va-cms-bot/**

### Team
Please check the team(s) that will do this work.

- [x] `CMS Team`
- [ ] `Public Websites`
- [ ] `Facilities`
- [ ] `User support`

PR #16279, opened 2023-12-04
Minimize PR overhead for non-code changes

JunTaoLuo:
## Description

Resolves https://github.com/department-of-veterans-affairs/va.gov-cms/issues/13834. Here's a general overview of the design.

### GHA workflows don't happen on documentation-only changes.

- The GHA workflows are disabled using the [`paths-ignore`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) mechanism.
- `.github/workflows/set-tugboat-tests-pending.yml` was renamed to `.github/workflows/set-tests-statuses.yml` and its role is to set the correct status depending on whether the changes contain code or not. For Doc only PRs, it will set all the status for all required checks to success since the CI and Tugboat deployments are disabled. For PRs containing code changes, the current behaviour of setting `va/test/*` status to pending is maintained
- I think it is valuable to keep the `Contextual Advice` workflow enabled since we may want to add advice based on which doc was updated (e.g. communicate to a Slack channel if a doc is updated, etc). As for the `CodeQL` and `Repair PR Title` workflows, I'm ambivalent but kept them for now. 

### Tugboat deploy doesn't happen on documentation-only changes

- Current Tugboat integration uses webhooks to notify Tugboat of PR open, update and close. Since webhooks does not allow for flexibility of these triggers based on the types of file changed, webhooks were replaced with GHA workflows and Tugboat API. See [discussion](https://dsva.slack.com/archives/C01A35JDH88/p1701723014069819).
- Since we need to retain the Preview ID between subsequent workflows (e.g. between creating and deleting the preview), `actions/cache` is used to persist the state.
- On code change PR opened (both regular PRs and Draft PRs), `.github/workflows/tugboat-pr-opened.yml` will be triggered.
  - A call to the Tugboat API is made to create a preview for the PR.
  - The API response is analyzed and the preview ID is extracted to a .tugboat_response.txt.
  - The file is saved to the cache using the PR number as a cache key. The entry is created if new (on PR open) or refreshed if already existent (on PR reopen).
- On code change PR updated or closed .github/workflows/tugboat-pr-{updated|closed}.yml` will be triggered.
  - The preview ID is restored from the cache using the PR number as a cache key.
  - A call to the Tugboat API is made to {rebuild|delete} the preview for the PR

## Testing done

I've tested the changes involving both doc PRs and code PRs in the va.gov-cms-test repo:
- https://github.com/department-of-veterans-affairs/va.gov-cms-test/pull/1458
- https://github.com/department-of-veterans-affairs/va.gov-cms-test/pull/1459

I've confirmed that opening, updating and closing the PRs trigger the behaviour described above. Testing of reopening is completed in:
- https://github.com/department-of-veterans-affairs/va.gov-cms-test/pull/1468

I've also tested the Tugboat Preview ID Cache refresh mechanism using https://github.com/department-of-veterans-affairs/va.gov-cms-test/pull/1462 and confirmed that it ensures the Preview IDs for open PRs are refreshed on a set schedule.

## QA steps

Testing this change requires turning off the webhook so it can't be done in this repo. The changes in this PR are mirrored at https://github.com/department-of-veterans-affairs/va.gov-cms-test/tree/13834-mirror. To test these changes, make a PR to that branch similar to the ones linked in the **Testing done** section.

As any user who can open PRs on this repo
1. For a PR containing code changes (with or without documentation changes)
   - [x] The Github Actions based CI workflows are triggered and completes as before
   - [x] Tugboat deployments are triggered correctly
     - [x] Previews are created on PR open
     - [x] Previews are rebuilt on PR updates, ensure the SHA of the commit used to build the preview is the updated commit.
     - [x] Previews are deleted on PR close
     - [x] Previews are recreated on PR reopen
2. For a PR containing only documentation changes (i.e. without code changes)
   - [x] The Github Actions based CI workflows short circuits and completes quickly
     - Most checks will still run but should complete within 15 seconds compared to >= 1 minute if they didn't short circuit
   - [x] The Tugboat deployments are not triggered
   - [x] The required checks are marked as passed so merging is unblocked

### Definition of Done

- [ ] ~~Documentation has been updated, if applicable.~~
- [ ] ~~Tests have been added if necessary.~~
- [x] Make an announcement in Slack for this change in Tugboat Preview behaviour.
- [ ] ~~If there are field changes, front end output has been thoroughly checked.~~
- [ ] Once this change has been merged to `main`, the Tugboat webhook on this repo can be disabled.
- [ ] Refresh the TUGBOAT_API_TOKEN with an official token.

### Select Team for PR review

- [x] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`


.github/workflows/set-tests-statuses.yml
```diff
@@ -0,0 +1,107 @@
...
+  # To address these two issues, this action sets check statuses directly
+  # to the appropriate states:
+  #  - For docs only changes, all required checks are set to 'success'
+  #  - For code changes, Tugboat tests are set to 'pending' so that we can
+  #    trust our automated code review processes more.
+  set-test-statuses:
+    name: Set Tests Statuses
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
```
JunTaoLuo, line 30: I've kept this from the previous file but I believe this Checkout step is unnecessary. We can skip it and save ~15s.

.github/workflows/set-tests-statuses.yml
```diff
@@ -0,0 +1,107 @@
...
+            const files = await github.paginate(
+              opts,
+              (response) => response.data.map(
+                (file) => file.filename
+              )
+            )
+
+            for (const file of files) {
+              console.log(`Checking PR file: ${file}`)
+              if (!file.endsWith('.md')) {
```
JunTaoLuo, line 51: Currently we use the rule that any non-`.md` file is a code change, which is probably the safest method for now.

.github/workflows/tugboat-pr-closed.yml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Set Preview ID
+      run: |
+        if ! [ -f .tugboat_preview.txt ]; then
+          echo "Preview ID not found, please manually delete Tugboat Preview."
+          exit 1
+        fi
+        PREVIEW_ID=$(cat .tugboat_preview.txt)
+        echo "Preview ID: ${PREVIEW_ID}"
+        echo "PREVIEW_ID=$PREVIEW_ID" >> $GITHUB_ENV
+    - name: Cleanup temporary file
```
JunTaoLuo, line 30: @olivereri when I was testing, I noticed that some files aren't removed when the self-hosted runner has finished running the workflow so I added this cleanup step as a defensive measure. However, files leaking between different workflow runs is ringing some alarm bells in my head. If you think this is a bug, we should file an issue to test/resolve it.

.github/workflows/set-tests-statuses.yml
ndouglas: That might be. I might've had the checkout step because it used to use a script that it no longer uses 🤔

.github/workflows/tugboat-pr-opened.yml
```diff
@@ -0,0 +1,42 @@
...
+          -o .tugboat_response.json \
+          https://api.tugboat.vfs.va.gov/v3/previews
+    - name: Diagnostics
+      run: cat .tugboat_response.json
+    - name: Extract Preview ID
+      run: jq -r .preview .tugboat_response.json > .tugboat_preview.txt
+    - name: Delete Previous Preview ID
+      continue-on-error: true
+      run: |
+        gh extension install actions/gh-actions-cache
```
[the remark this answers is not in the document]
JunTaoLuo: Yes I still need it, which is why this is blocked on https://github.com/department-of-veterans-affairs/va.gov-cms/issues/16358. The PR reopen trigger won't work until we install the Github CLI.

.github/workflows/tugboat-pr-closed.yml
JunTaoLuo: Hmmm, I'm not a fan of trusting all workflows to cleanup properly in general and this is different from GHA runners. But since we control the runners and the workflows in this case, I guess it's not a big deal.

.github/workflows/set-tests-statuses.yml
JunTaoLuo: Will remove and see what (if anything) breaks

.github/workflows/tugboat-pr-opened.yml
```diff
@@ -0,0 +1,42 @@
...
+    name: Create Tugboat Preview
+    steps:
+    - name: Create Tugboat Preview
+      id: tugboat_pr_preview
+      run: |
+        curl --fail \
+          -H "Authorization: Bearer ${{ secrets.TUGBOAT_API_TOKEN }}" \
+          -H "Content-Type: application/json" \
+          -X POST \
+          -d '{ "repo": "${{ secrets.TUGBOAT_REPOSITORY }}", "ref": "${{ github.event.pull_request.number }}", "name": "${{ github.event.pull_request.title }}", "type": "pullrequest" }' \
```
ndouglas, line 24: If this is all of the information that we need to provide, this is fantastic! It seems like it'll resolve an issue we've had where the TIC interprets inline code (e.g. "Fixed a bug in `some_function()`) as an attempt to inject code and silently rejects it, causing the Tugboat build to fail.

JunTaoLuo: Fascinating. That was probably pretty hard to find haha. Though I suppose commit messages and even branch names theoretically can be unsafe if we used it in some shell commands. Hence 
> - [ ] checked for any injection of user content into protected contexts?

ndouglas (approved):
This is fantastic. I verified all of the QA steps on va.gov-cms-test.

Any objections or thoughts @olivereri ?

JunTaoLuo:
@ndouglas two more things about the PR. 

- We need a TUGBOAT_API_TOKEN in GHA for the Tugboat APIs and I've been using my personal token. This works for now but is there a better account to generate this token from? And should the token be saved somewhere (e.g. a team wide 1 Password)?
- I plan to make an announcement of the change since while changing from webhooks to GHA triggers should not have any visible effects, we should still let folks know in case someting bizarre happens that we didn't test for.

ndouglas:
We generally use @va-cms-bot and SSM Parameter Store for generating and storing those credentials, so they're not tied to any particular one of us. 

**TL;DR**: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14345 is the procedure for gaining access to `va-cms-bot`.

JunTaoLuo:
Hmm I suppose the token isn't blocking for now but I'll file a follow up ticket for someone with access to regenerate the token and update it for this repo.

.github/workflows/tugboat-pr-opened.yml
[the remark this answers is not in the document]
JunTaoLuo: I was able to remove the dependency on GH CLI, so this is no longer needed.

olivereri (approved):

ndouglas (approved):
Fantastic work!

JunTaoLuo:
Thanks everyone!

Issue #7673, opened 2022-01-19
Make node count by content type visible in Datadog


timcosgrove:
## Description
This ticket made total published node count visible in Datadog: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/7244

We would like to additionally see node count broken down by content type, visible in Datadog.

## Acceptance Criteria
- [x] CMS Prod is configured to send node count broken down by content type to Datadog
- [x] A widget or dashboard which visualizes these counts is created, such that a viewer can see trends over time.
- [ ] Needs an update hook somewhere so that new content types auto-add themselves to the list in the prometheus settings

## Implementation information
- The CMS can be configured in the UI here: https://prod.cms.va.gov/admin/config/system/prometheus_exporter - then export the config
- One question is whether this configuration can be managed such that any new content types that are created are automatically added to this list.
- Write a hook so new content types get automatically added
- Click together the UI component in Datadog, we don't do this in Terraform anymore

### CMS Team

Please leave only the team that will do this work selected. 

- [x] `Platform CMS Team`
- [ ] `Sitewide CMS Team`
  - [ ] `⭐️ Content ops`
  - [ ] `⭐️ CMS experience`
  - [ ] `⭐️ Offices`
  - [ ] `⭐️ Product support`
  - [ ] `⭐️ User support`

cweagans:
Notes:
* Needs config exported
* Needs an update hook somewhere so that new content types auto-add themselves to the list in the prometheus settings

PR #16390, opened 2023-12-12
VACMS-7673: Export node count by content type via prometheus_exporter.

ndouglas:
## Description

See #7673.

## QA Steps

- [x] https://pr16390-zpu8o82qaowsby6mieux1ev7wj5oahm5.ci.cms.va.gov/metrics contains a boatload of per-content-type stats.


edmund-dunn (approved):
Many, many nodes!! Nice work!

PR #16777, opened 2024-01-09
VACMS-16237: Replace s3 address for assets.

### LINKED ISSUE #16237: Tugboat content releases do not correctly reference some assets (e.g. CSS).
## Description

See [this Slack thread](https://dsva.slack.com/archives/CT4GZBM8F/p1701209446210169).

Content releases on Tugboat use a `buildtype` of `vagovdev` since #1700, following #1630.

[This code](https://github.com/department-of-veterans-affairs/content-build/blob/838acf095462ab786a6b5be4b0c1caaf42409291/src/site/stages/build/plugins/modify-dom/process-entry-names.js#L99) inserts URLs pointing to S3 buckets unless the `buildtype` is set to `localhost`.

Consequently, local changes to these files will not be picked up in the built frontend, as they reference those in the `content.dev.va.gov` S3 bucket.

We should probably change our `buildtype` back to `localhost` or add a new `buildtype`, `tugboat`. The abstraction is leaky but it would probably take more effort than it is worth to fix at that level. This may take some discovery to determine an appropriate approach.

## Acceptance Criteria

- [ ] Tugboat content releases reference local assets.
- [ ] No adverse impact for other values of `buildtype`.

ndouglas:
## Description

See #16237.

This does not fix all of the problem, but I think this fixes the immediate problem. There's a deeper, more significant issue with `vets-website` and `content-build`, I think, that will require considerable more effort (and greater risk) to fix.

## QA Steps
- [x] Navigate to https://web-xdksfw5fi4p35n7uzvs52f3zp306zypa.ci.cms.va.gov/
- [x] Verify that the source of the `content-build.css` file is local, not Dev.
- Screenshot 2024-01-09 at 5 12 03 PM

### DIFF
--- a/scripts/web-build.sh
+++ b/scripts/web-build.sh
@@ -14,7 +14,7 @@ pushd "${repo_root}" > /dev/null
 composer va:web:prepare-dotenv
 touch ./.buildlock
 
-build_type="localhost"
+build_type="vagovdev"
 web_path="./web"
 build_path="${web_path}/build/${build_type}"
 rm -rf "${build_path}"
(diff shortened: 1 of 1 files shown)


edmund-dunn (approved):
Everything worked as expected. Nice work!

Issue #16882, opened 2024-01-17
Error: Call to a member function getValue() on null in Drupal\va_gov_content_types\Entity\Event->addToNationalOutreachCalendar()


dsasser:
## Description
When saving some nodes (eg: node/64741/edit), an error is thrown:

```
Error: Call to a member function getValue() on null in Drupal\va_gov_content_types\Entity\Event->addToNationalOutreachCalendar()
```

Originally reported via [Slack](https://dsva.slack.com/archives/CJT90C0UT/p1705501739844599).

### Engineering notes / background
It appears in some cases the field `field_publish_to_outreach_cal` has no value, and so calling first() on null is returning the error.

## Acceptance criteria
- [ ] Error is no longer thrown on node 64741 (or similar nodes without a field_publish_to_outreach_cal value).

jilladams:
Slack thread with original report / risk assessment for why we should prioritize now: https://dsva.slack.com/archives/CJT90C0UT/p1705501739844599

dsasser:
Verified on Prod:

![Screenshot 2024-01-24 at 8 06 18 AM](https://github.com/department-of-veterans-affairs/va.gov-cms/assets/221539/14487c62-449c-406e-8f1b-3bbae80d5395)

PR #16900, opened 2024-01-18
VACMS-16882: Resolves Error: Call to a member function getValue() on null

dsasser:
## Description
Resolves a bug in addToNationalOutreachCalendar() which can cause a php error.

Relates to #16882

## Testing done
Tested using different user roles, and using new and existing Event nodes.

## QA steps
As a content admin
1. Visit /node/add/event
2. Fill out the required fields
3. In the right sidebar, under "Feature This Content", check the "Featured" checkbox
4. Save the node
   - [x] Validate that there are no errors

### Definition of Done

- [x] If there are field changes, front end output has been thoroughly checked.

### Select Team for PR review

- [ ] `CMS Team`
- [x] `Public websites`
- [ ] `Facilities`
- [ ] `User support`


ndouglas:
Rebuilding because of Tugboat antics.

Becapa (dismissed):
Looks good. I was able to save the node:
Screenshot 2024-01-19 at 3 58 03 PM

Becapa (approved):

edmund-dunn (approved):
Worked as expected. No errors! Nice work!!

docroot/modules/custom/va_gov_content_types/src/Traits/EventOutreachTrait.php
```diff
@@ -46,10 +46,11 @@ public function addToNationalOutreachCalendar(EventInterface $node): void {
     if ($node->hasField(EventOutreachInterface::LISTING_FIELD) &&
       $node->hasField(EventOutreachInterface::PUBLISH_TO_OUTREACH_CAL_FIELD) &&
       $node->hasField(EventOutreachInterface::ADDITIONAL_LISTING_FIELD)) {
-      $addToCalValue = $node->get(EventOutreachInterface::PUBLISH_TO_OUTREACH_CAL_FIELD)->first()->getValue();
-      if (isset($addToCalValue['value'])) {
+      if ($node->get(EventOutreachInterface::PUBLISH_TO_OUTREACH_CAL_FIELD)->first()) {
+        $addToCalValue = $node->get(EventOutreachInterface::PUBLISH_TO_OUTREACH_CAL_FIELD)->first()->getValue();
         $listings = $node->get(EventOutreachInterface::LISTING_FIELD)->getValue();
         $additionalListings = $node->get(EventOutreachInterface::ADDITIONAL_LISTING_FIELD)->getValue();
+        assert(array_key_exists('value', $addToCalValue));
```
swirtSJW, line 53: I am pretty sure I am missing the use case/edge case  here so apologies for the potentially dumb question here.

Is there an edge case where it would get past line 49 and still not have a 'value' here at line 53?  If there is, it might be good to spell out with a comment the situation that would lead to the assertion failing.

swirtSJW (commented):
Just one question from me.

swirtSJW (approved):

docroot/modules/custom/va_gov_content_types/src/Traits/EventOutreachTrait.php
dsasser: Good question Steve. My thinking in this case was that there is actually no guarantee that, even though the field has a value, that the property 'value' exists in the returned array. For entity reference fields for example, the primary field property is 'target_id'. So this assertion was a way for me to protect against any future changes to the target field where it changes to a different type, or for some reason the 'value' property changes to something else.

Generally I've been trying to get away from using magic methods to get to field values, which would actually improve the readability of this code, but this is the outcome of doing that in this situation.

swirtSJW: Makes sense.  Thanks for the explanation @dsasser.

Issue #16348, opened 2023-12-07
Add appointment fields from VAMC facility health service to Service location


omahane:
## User Story or Problem Statement

As an editor, I want to be able to create facility services with service locations specific the appropriate type of facility and service, whether VAMC or VBA.

## Acceptance Criteria
- [ ] The Service location paragraph is able to capture appointment information for facility services
- [ ] Deprecated fields are hidden
- [ ] The user interface for Service Options and Appointments follows the [Figma design](https://www.figma.com/file/aKdT4weNNthDxb3r6MolEh/VBA-Content-Specs?type=design&node-id=218-715&mode=design&t=HgXBnMWY1AGHCHxx-0) 
- [ ] This goes into an integration branch

## Implementation steps

**Service location - Service Options**
- [x] A list field is added to Service location for **Service options: Office visits**, displayed on the form as radio buttons
- [x] A list field is added to Service location for **Service options: Virtual support**, displayed on the form as radio buttons

**Service location - Appointments**
- [x] A list field is added to Service location for **Appointments: Appointment introduction text**, displayed on the form as radio buttons
- [x] An entity reference field is added to Service Location for **Appointments: Phone number for appointments**
- [x] A list field is added to Service location for **Appointments: Is online scheduling available for this service?"**, displayed as a dropdown field

**Default appointment text per Service**
- [x] A markup field `field_vba_fac_serv_appt_default` is added to VBA Facility service for "Appointment lead-in default" to hold the default text (a fraternal twin to the same field on the VAMC facility health service content type)
- [x] The markup field `field_vba_serv_reg_appt_default ` is added to Service region for "Appointment lead-in default" to hold the default text
- [x] The markup field `field_vamc_nonclin_appt_default` is added to VAMC Facility Non-clinical Service for "Appointment lead-in default" to hold the default text
- [x] ~@davidmpickett determines text for all instances of `field_hservices_lead_in_default`~ I can't use this field, as it not from a paragraph, but a content type. Adding new markup for each, instead.
- [ ] Code is refactored from va_gov_backend that handles the logic for the appointment introduction text for the new service location fields and accounts for the VAMC facility health service, VAMC facility non-clinical service, VBA facility service and Service region content types, with the new code going into va_gov_facilities.

xiongjaneg:
@davidmpickett Please review this ticket and share your wisdom

omahane:
@davidmpickett 
[...]

PR #17048, opened 2024-01-26
VACMS-16348: Service location ux enhancements

omahane:
## Description

Relates to #16348 

### When "Use default text" is checked
Screenshot 2024-02-01 at 5 23 34 PM

### When "Customize text" is checked
Screenshot 2024-02-01 at 5 23 19 PM

### When "Remove text" is checked
Screenshot 2024-02-01 at 5 21 55 PM

### What needs to be checked to prove this works?
- [x] Assign QA Content Publisher the following roles:
  - [x] Content creator - VAMC
  - [x] Content creator - VBA
  - [x] Content publisher
- [x] Assign QA Content Publisher to the following Sections:
  - [x] VA Southeast Louisiana health care
  - [x] Des Moines VA Regional Benefit Office

The following spreadsheet identifies the testing scenarios:
https://docs.google.com/spreadsheets/d/1JxdTIbIibkY3UQ3DIix9hr7XcP4CaueL_1sYDHJIKmg/edit#gid=221281300

For each row, 
1. Mark or fill in the field(s) along with any required Service location or node fields.
2. Confirm the appearance or disappearance of the Default text.
3. Confirm the appearance or disappearance of the Custom text field.
4. Save the node.
5. Edit the node to review the Custom text field result.

#### Users

- For Service region, log in as an admin.
- For all other content types, log in as QA Content Publisher.

Note: Each test depends on the presence of a second service location. You may need to add the other.

### Definition of Done

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [x] `Facilities`
- [ ] `User support`


docroot/modules/custom/va_gov_facilities/va_gov_facilities.info.yml
```diff
@@ -3,3 +3,5 @@ type: module
 description: 'Tools to improve user experience for VA Facilities.'
 core_version_requirement: ^9 || ^10
 package: 'Custom'
+dependencies:
+  - va_gov_backend
```
Becapa, line 7: The code in the backend module that this was previously dependent upon (even though it wasn't listed as a dependency then) was moved out and into the facilities module with these changes. Does this need to be listed as a dependency now?

Becapa (approved):
I think this looks good. I had one comment that I added, but it's definitely not blocking. I also noticed most of the functions here have an if statement that wraps the entire function. I wonder if at some point we could talk about the guard clause coding pattern and whether it's something we'd like to use moving forward or not.

docroot/modules/custom/va_gov_facilities/va_gov_facilities.info.yml
omahane: Yes, as I found a method in va_gov_backend code that is being called in this module. I should have explained that.

omahane:
> I think this looks good. I had one comment that I added, but it's definitely not blocking. I also noticed most of the functions here have an if statement that wraps the entire function. I wonder if at some point we could talk about the guard clause coding pattern and whether it's something we'd like to use moving forward or not.

Yes, that's a good idea. I love it and will implement it going forward, unless you feel like it needs it here.

Becapa:
> > I think this looks good. I had one comment that I added, but it's definitely not blocking. I also noticed most of the functions here have an if statement that wraps the entire function. I wonder if at some point we could talk about the guard clause coding pattern and whether it's something we'd like to use moving forward or not.
> 
> Yes, that's a good idea. I love it and will implement it going forward, unless you feel like it needs it here.

No, I don't think it needs it here. It's just more of a preference for readability and none of the code here is so nested that it's unreadable so I think it's good.

davidmpickett:
The ticket says "Deprecated fields are hidden" but they are still showing in the Tugboat. I feel like maybe you and Steve had a conversation about how hiding the fields should wait until after the migration? Will that be a separate PR?

1C5C7E11-0BC9-490A-BD79-F2DF32BC217D_1_201_a

omahane:
> The ticket says "Deprecated fields are hidden" but they are still showing in the Tugboat. I feel like maybe you and Steve had a conversation about how hiding the fields should wait until after the migration? Will that be a separate PR?
> 
Yes, that's going to be separate.

swirtSJW:
Nice work @omahane.   I am merging this into integration now.  So we will need a followup PR to hide the fields.

PR #17154, opened 2024-02-06
VACMS-16853: next build branches ftw

alexfinnarn:
## Description

closes #16853

## Testing done

- Updated some tests for git information/API/frontend. 
- Manually tested locally by running script.
- Need to test on Tugboat with this PR.

## Screenshots

Content release form when you can trigger a build.
Screenshot 2024-02-08 at 1 43 57 PM

Content release form when build is running with a link to the build log.
Screenshot 2024-02-08 at 1 46 02 PM

## QA steps

1. Log in and go to "/admin/content/deploy/next"
   - [x] Confirm you can select versions of next-build and vets-website from git references
   - [x] Confirm there is no build in progress
   - [x] No request or lock files should exist
   - [x] The versions for next-build and vets-website should be "__default"
2. Submit the form with two branches chosen for next-build and vets-website
   - [x] Confirm that a build request file exists. 
   - [x] No build should have started yet so no lock file should exist.
   - [x] The versions for next-build and vets-website should be reported as what you selected.
2. Refresh the page a few times until the background process kicks off from reading the build request file.
   - [x] Confirm that the form fields are locked during the build.
   - [x] Confirm you can view the build log file and it is adding data until the build finishes.
4. Refresh the form page after the build is complete
   - [x] Confirm the disabled fields are no longer disabled.
   - [x] View the log to see that the build finished. It might have an error, but that will be addressed in future tickets.

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.
- [ ] If there are field changes, front end output has been thoroughly checked.

### Select Team for PR review

- [x] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`
- [x] `Accelerated Publishing`

PR #17261, opened 2024-02-16
Revert "VACMS-16853: next build branches ftw"

tjheffner:
Reverts department-of-veterans-affairs/va.gov-cms#17154

there was an error with the autocomplete branch selection on the existing content release form that was not caught by the tests in the original PR environment.


edmund-dunn (approved):
Thank you!

PR #18250, opened 2024-05-30
VACMS-18070: Adding a check for response header for files to pull from s3 buckets.

JakeBapple:
## Description
This work will check for a response header (File-Public-Base-Url-Check) to be present. If it is, it will set the file-base-url for assets to point to the appropriate s3 bucket. If not, it will use webhost variable. This work will initially not have any effect. Once the header has been set in next-build, we should see the assets' paths being directed to the s3 buckets.

Relates to #_18070_. (or closes?)

## Testing done
Tested locally.
Setting variable to both pass and fail to check results in the settings files.

## Screenshots
This is an image of a media asset pulling from the staging s3 bucket because the logic passed:
image

This is the same media asset with the check failing and it pulling from webhost variable:
image

## QA steps

We need to first set the response header in next-build to the specified key/value pair. Then check out assets' url (like an image url) to see if it's pointing to the s3 bucket or to the webhost.
Any image or media assets that were working before, need to still work.

1. Do this
   - [ ] Validate that
2. Then
   - [ ] Validate that
   - [ ] a
   - [ ] b
   - [ ] c

### Definition of Done

### Select Team for PR review

- [x] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/.ddev/mutagen/mutagen.yml
+++ b/.ddev/mutagen/mutagen.yml
@@ -3,8 +3,8 @@
 # and add your own configuration. If you override it you will
 # probably want to check it in.
 # Please do `ddev mutagen reset` after changing the file.
-# See DDEV Mutagen docs at
-# https://ddev.readthedocs.io/en/stable/users/install/performance/#mutagen
+# See ddev mutagen docs at
+# https://ddev.readthedocs.io/en/latest/users/install/performance/
 # For detailed information about mutagen configuration options, see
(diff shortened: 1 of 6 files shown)


edmund-dunn (approved):
Nice work!! This looks great!

Issue #18257, opened 2024-05-31
Unable to upgrade Drupal module with composer: The .git directory is missing from...


dsasser:
## User Story or Problem Statement
I am receiving this error when attempting to upgrade a Drupal module from a dev release to an alpha version.

```
The .git directory is missing from /var/www/html/docroot/modules/contrib/expirable_content
```

This occurs both locally and in Tugboat.
<img width="1145" alt="Screenshot 2024-05-30 at 6 22 18 PM" src="https://github.com/department-of-veterans-affairs/va.gov-cms/assets/221539/cebc79e0-c96a-4f10-9ec5-82a08149b6e2">

The error can be bypassed locally by responding to the prompt, but this isn't an option in CI.

## Description or Additional Context 
This was noted by [Tim](https://dsva.slack.com/archives/CT4GZBM8F/p1678307939948779?thread_ts=1678126053.329369&cid=CT4GZBM8F) last year, and it was solved by manual edits to a lock file

> Edmund was able to bypass this by directly editing composer.lock, but the problem will continue to occur when folks next regenerate composer.lock.

## Steps for Implementation

1. Checkout the main branch
2. Run composer install
3. Checkout branch `VACMS-18005-aging-content-tests`
4. Run composer install
5. Get the error

## Acceptance Criteria
- [ ]

timcosgrove:
As near as I'm able to reconstruct, this removal of .git directories behavior traces back to this PR: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/11893

There was no associated ticket, and not much associated discussion; the only Slack discussion I found is here: https://dsva.slack.com/archives/CT4GZBM8F/p1670544135956989

Based on that discussion and what was written in the PR, this was put in place to alleviate some amount of developer annoyance. This is a worthwhile goal if it doesn't interfere with normal code management.

The problem with this solution is that any package that needs to be managed via source and via a branch rather than a tag becomes a problem here, as .git directories are required in those packages so that they can be updated. In composer, this is any package that is referenced via a `dev-` branch, as we did for expirable_content in this ticket: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15506

No longer removing .git directories should not cause problems with the normal deploy operations of the site, though this will need to be monitored as it rolls out.

PR #11893, opened 2022-12-09
VACMS-000: Remove .git sub-directories after composer install & update

ElijahLynn:
## Description

This fixes that really annoying issue where `composer install` prompts you if you want to overwrite modified changes to certain dependencies/modules. We don't need the .git directories, unless we are working on an upstream module, and in that case we will likely clone it outside of the repo and symlink as needed.

## Testing done
Ran this locally with `composer install`. 

## Screenshots
image

## QA steps
1. `cd docroot/modules/cer`
2. Verify a `git status` returns git info about the `cer` module, not the CMS repo
3. Run `composer install`
4. Verify it finishes with a step that says "Removing .git sub-directories...`
5. Run `git status` again in the `cer` directory, you should see info about this branch instead

### Definition of Done

- [ ] Documentation has been updated, if applicable.
- [ ] Tests have been added if necessary.
- [ ] Automated tests have passed.
- [ ] Code Quality Tests have passed.
- [ ] Acceptance Criteria in related issue are met.
- [ ] Manual Code Review Approved.

### Select Team for PR review

- [ ] `Platform CMS Team`
- [ ] `Sitewide program`
- [ ] `⭐️ Sitewide CMS`
- [ ] `⭐️ Public websites`
- [ ] `⭐️ Facilities`
- [ ] `⭐️ User support`

### Is this PR blocked by another PR?

- [ ] `DO NOT MERGE`

### Does this PR need review from a Product Owner

- [ ] `Needs PO review`

### CMS user-facing annoucement

Is an announcement needed to let editors know of this change?
- [ ] Yes, and it's written in issue ____ and queued for publication.
  - [ ] Merge and ping the UX writer so they are ready to publish after deployment
- [ ] Yes, but it hasn't yet been written
  - [ ] Don't merge yet -- ping the UX writer to write and queue content
- [ ] No announcement is needed for this code change.
  - [ ] Merge & carry on unburdened by announcements

PR #18339, opened 2024-06-17
VACMS-18257: Remove composer git directory removal script.

timcosgrove:
## Description

Relates to #18257

As near as we're able to reconstruct, the removal of .git directories within the composer vendor directory dates back to this ticket: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/11893

Judging by the ticket and discussion around it, this work was done to relieve some annoyance/inconvenience for developers working locally, which is reasonable. 

However, removing the .git directory causes real problems if a composer package is managed via source rather than dist, as the .git directory is required for any updates to the package. There are cases where source is the only option; for example, Drupal contrib modules that do not yet have proper releases are only available to be managed via source. 

Proper functioning of the codebase and its updates must take precedence over developer convenience. 

## Testing done
The Tugboat instance for this PR was refreshed entirely, to build it from scratch and to most closely match the situation on BRD.

This change will need to be closely monitored on BRD when it deploys there, and Tugboat preview instances will need to be monitored when this change rolls out to them to ensure it deploys cleanly.


7hunderbird (approved):
+1 LGTM, let's try it in staging.

Issue #18070, opened 2024-05-08
Allow CMS to use different public file base URL on request


timcosgrove:
## Requirements

We want to allow Drupal to switch its public file base URL on request, so that we can request the appropriate base URL for internal use or for external use.

```[tasklist]
### Acceptance criteria
- [ ] If a request to the CMS contains a custom header with a known value (chosen by the developer), the CMS will use a public-facing file base URL (S3 bucket)
- [ ] In all other circumstances, the file base URL will be the current default
- [ ] If the request contains the custom header, an API response with file URLs will contain URLs that point to the public-facing location
```

## Background & implementation details

[The setting `file_public_base_url` is handled in settings.php](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/main/docroot/sites/default/settings.php#L256). Under normal circumstances, this base url is `{$webhost}/sites/default/files`; for example, for Prod CMS this is https://prod.cms.va.gov/sites/default/files. This is the appropriate base URL to use for CMS-internal use. However, it is behind the firewall with the rest of the CMS, which means file URLs are not accessible to the public. 

We have public-facing S3 buckets that are synced with the CMS filesystem. The base URLs for these are:
- https://dsva-vagov-prod-cms-files.s3.us-gov-west-1.amazonaws.com (Prod; used for production builds)
- https://dsva-vagov-staging-cms-files.s3.us-gov-west-1.amazonaws.com (Staging; use for all other environments)

**Please note that these public-facing URLs do not contain `/sites/default/files`. This is correct; the public facing locations do not use the `sites/default/files` directory. All files start at the root.**

We want to use these base URLs for static files that are built for public use. A good way to do this is to look for a custom request header in requests to Drupal, and switch to the public-facing base URL only if that custom request header is present. 

So, the modification to settings.php (and its environment-specific versions) would be:
- If a custom request header is present with a specific value, use the appropriate public-facing base URL for that environment
- Otherwise, fall back to the default `{$webhost}/sites/default/files`

The system should *not* under any circumstances change the public file base URL to an arbitrary URL specified in the custom header. The header should only be used to indicate that we want to switch - i.e. a value of `S3` - and then the public file base URL switched to a value that we control. 

[...]

PR #18424, opened 2024-06-28
Rework of how the CMS responds to public file base header.

timcosgrove:
## Description

Relates to #18070 , and a follow-on to the work merged in this PR: https://github.com/department-of-veterans-affairs/va.gov-cms/pull/18250

Next-Build will use our public file S3 buckets where Drupal's user files are available to the public via web. We do this by adding a request header when Next Build makes requests to Drupal; and then, on the Drupal side, to choose a different public file base URL if the request header is present. 

On the Next Build side, the change to add the header to the request to Drupal was done in this PR: https://github.com/department-of-veterans-affairs/next-build/pull/560

## Testing done
See QA steps; the same testing was done there. 

## QA steps

Noting for all this testing that the images not appearing for the staging domain dsva-vagov-staging-cms-files.s3.us-gov-west-1.amazonaws.com is a known issue and does not represent a failure of this work. That work is ticketed separately. 

1. Visit https://next-ivzjgpsx0n3uwgkmporajcndsszipus3.ci.cms.va.gov/northern-california-health-care/stories/. You will see images broken. **This is OK and expected here.**
2. Inspect one of the broken images, and find its URL.
    - The URL domain should be dsva-vagov-staging-cms-files.s3.us-gov-west-1.amazonaws.com
    - The URL domain should not be the same as the Tugboat Next Build domain, i.e. next-ivzjgpsx0n3uwgkmporajcndsszipus3.ci.cms.va.gov
3. Load an image URL into its own tab, for example the image next to "Asian American, Native Hawaiian and Pacific Islander Heritage Month". Its URL should be something like https://dsva-vagov-staging-cms-files.s3.us-gov-west-1.amazonaws.com/styles/2_1_large/public/2023-05/Asian%20American%2C%20Native%20Hawaiian%20and%20Pacific%20Islander%20Heritage%20Month%20%281%29.jpg
4. Replace 'staging' in the domain with 'prod': https://dsva-vagov-prod-cms-files.s3.us-gov-west-1.amazonaws.com/styles/2_1_large/public/2023-05/Asian%20American%2C%20Native%20Hawaiian%20and%20Pacific%20Islander%20Heritage%20Month%20%281%29.jpg
    - The image should load for the prod domain.

### DIFF
--- a/docroot/sites/default/settings.php
+++ b/docroot/sites/default/settings.php
@@ -264,7 +264,7 @@ if (!empty($webhost_on_cli)) {
 // Look for an incoming request header to indicate we should use the public
 // asset S3 location for file rather than the Drupal-internal location.
 if (!empty($public_asset_s3_base_url)) {
-  $headersArray = function_exists('apache_request_headers') ? apache_request_headers() : [];
+  $headerArray = function_exists('apache_request_headers') ? apache_request_headers() : [];
   $targetHeader = 'File-Public-Base-Url-Check';
   foreach ($headerArray as $header => $value) {
     if (strtolower($header) == strtolower($targetHeader) && $value === 'true') {
--- a/docroot/sites/default/settings.php
+++ b/docroot/sites/default/settings.php
@@ -265,15 +265,14 @@ if (!empty($webhost_on_cli)) {
 // asset S3 location for file rather than the Drupal-internal location.
 if (!empty($public_asset_s3_base_url)) {
   $headersArray = function_exists('apache_request_headers') ? apache_request_headers() : [];
-  if (
-     array_key_exists('File-Public-Base-Url-Check', $headersArray) && 
-     $headersArray['File-Public-Base-Url-Check'] == 'true'
-   ) {
-   // Point the file base url to the public asset S3 bucket.
-   $settings['file_public_base_url'] = $public_asset_s3_base_url;
+  $targetHeader = 'File-Public-Base-Url-Check';
+  foreach ($headerArray as $header => $value) {
+    if (strtolower($header) == strtolower($targetHeader) && $value === 'true') {
+      // Point the file base url to the public asset S3 bucket.
+      $settings['file_public_base_url'] = $public_asset_s3_base_url;
+    }
+  }
  }
-}
-
 
 // Monolog
 $settings['container_yamls'][] = __DIR__ . '/services/services.monolog.yml';
--- a/docroot/sites/default/settings.php
+++ b/docroot/sites/default/settings.php
@@ -243,6 +243,38 @@ if (!empty($GLOBALS['request']) &&
   $deploy_service->run($GLOBALS['request'], $app_root, $site_path);
 }
 
+// Because Jenkins can't resolve e.g. prod.cms.va.gov DNS, and only the
+// internal*elb addresses. And sometimes the host would return data with
+// "http://default/..." hostnames for files so we set the host here and pass it
+// to the `file_public_base_url` setting to fix that.
+if (!empty($webhost_on_cli)) {
+  if (PHP_SAPI === 'cli') {
+    // This is running from drush so set the webhost.
+    // Var $webhost_on_cli is set in <settings.<environment>.php.
+    $webhost = $webhost_on_cli;
+  }
+  else {
+    // This is running from an HTTP request.
+    $webhost = "{$_SERVER['REQUEST_SCHEME']}://{$_SERVER['HTTP_HOST']}";
+  }
+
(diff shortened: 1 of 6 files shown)


docroot/sites/default/settings.php
```diff
@@ -243,6 +243,38 @@
...
+    // This is running from an HTTP request.
+    $webhost = "{$_SERVER['REQUEST_SCHEME']}://{$_SERVER['HTTP_HOST']}";
+  }
+
+  $settings['file_public_base_url'] = "{$webhost}/sites/default/files";
+}
+
+// Look for an incoming request header to indicate we should use the public
+// asset S3 location for file rather than the Drupal-internal location.
+if (!empty($public_asset_s3_base_url)) {
```
timcosgrove, line 266: Don't bother with any of this if we don't have a URL to send people to.

docroot/sites/default/settings.php
```diff
@@ -243,6 +243,38 @@
   $deploy_service->run($GLOBALS['request'], $app_root, $site_path);
 }
 
+// Because Jenkins can't resolve e.g. prod.cms.va.gov DNS, and only the
```
timcosgrove, line 246: I moved this block back here so that it is in one place. We need the file public base url in each of environment-specific files, but most of the rest of the code is the same, and we'll want it in one place so we can maintain it in one place.

docroot/sites/default/settings.php
```diff
@@ -243,6 +243,38 @@
...
+    $webhost = "{$_SERVER['REQUEST_SCHEME']}://{$_SERVER['HTTP_HOST']}";
+  }
+
+  $settings['file_public_base_url'] = "{$webhost}/sites/default/files";
+}
+
+// Look for an incoming request header to indicate we should use the public
+// asset S3 location for file rather than the Drupal-internal location.
+if (!empty($public_asset_s3_base_url)) {
+  $headersArray = function_exists('apache_request_headers') ? apache_request_headers() : [];
```
timcosgrove, line 267: The previous function used for this - `headers_list()` - is used to retrieve response headers that the PHP process is collecting to send back with a response. `apache_request_headers()` provides us with the incoming request headers. 

Despite its `apache` name, the function does work in some non-apache contexts. However, it does not work in a CLI PHP context, which includes when running drush. If `apache_request_headers()` is called in a CLI context, it causes an error.

Falling back to an empty header just means that we will not find the array key in the next lines, and therefore this won't do anything, which is what we want in that case.

docroot/sites/default/settings.php
```diff
@@ -243,6 +243,38 @@
...
+  if (
+     array_key_exists('File-Public-Base-Url-Check', $headersArray) && 
+     $headersArray['File-Public-Base-Url-Check'] == 'true'
+   ) {
```
timcosgrove, line 271: The previous way of doing this check - `in_array('File-Public-Base-Url-Check: true', $headersArray, true)` - didn't find the header. `apache_request_headers()` returns headers in a header name as key, header value as value format. So, we need to check that the header is set in the first place, and that the value is what we expect it to be.

docroot/sites/default/settings/settings.dev.php
```diff
@@ -46,3 +46,6 @@
 $settings['va_gov_frontend_url'] = 'https://dev.va.gov';
 $settings['va_gov_frontend_build_type'] = 'brd';
 $settings['github_actions_deploy_env'] = 'dev';
+
+// Public asset S3 location
+$public_asset_s3_base_url = "https://dsva-vagov-staging-cms-files.s3.us-gov-west-1.amazonaws.com";
```
timcosgrove, line 51: This is sort of moot at the moment because we don't have an active dev environment, but, covering all bases.

docroot/sites/default/settings.php
```diff
@@ -243,6 +243,37 @@
...
+  $settings['file_public_base_url'] = "{$webhost}/sites/default/files";
+}
+
+// Look for an incoming request header to indicate we should use the public
+// asset S3 location for file rather than the Drupal-internal location.
+if (!empty($public_asset_s3_base_url)) {
+  $headerArray = function_exists('apache_request_headers') ? apache_request_headers() : [];
+  $targetHeader = 'File-Public-Base-Url-Check';
+  foreach ($headerArray as $header => $value) {
+    if (strtolower($header) == strtolower($targetHeader) && $value === 'true') {
```
timcosgrove, line 270: Normalize the header names to lowercase; some web servers do this on their own, and we can't predict what may be changed.

anantais (approved):

PR #18738, opened 2024-07-24
Add Chapters/Steps field to Digital Form content type

derekhouck:
## Description
Adds a "Steps" field (machine name `chapters`) to the Digital Form content type. Also adds a first-pass "Name and Date of Birth" Paragraph type that can be used within the "Steps" field as well as a "Digital Form" Paragraph Browser with a "Patterns" group.

Closes #89100.

## Testing done
Manually created a Digital Form with a Step that included the "Name and Date of Birth" Paragraph type.

## QA steps

1. Do this
   - [ ] Validate that
2. Then
   - [ ] Validate that
   - [ ] a
   - [ ] b
   - [ ] c

### Definition of Done

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


ryguyk (commented):
This looks like a solid starting point.

I think we're going to need to adjust this structure, though, as there is still the concept of pages being a required layer inside chapters. We'd likely never add something like Name/DOB directly to a chapter. This, of course, leads to a host of additional questions around how we're going to ultimately structure the editorial presentation. Perhaps some of these details will be abstracted away from editorial concern. We just don't know yet.

I have some similar questions around the specifics of the Name-and-DOB component itself (naming, options, organization, etc.). I don't know if we know enough to really try to implement something specific. 

I guess this ultimately raises the question of what we're trying to accomplish with this PR. If it's something we think is very preliminary that might simply help us visualize structure a bit, that makes sense, but I don't think we're at a place where we can build something that will closely reflect the final state. From that standpoint, I think we want to make sure we're planning for change. We've already run into the idea that once we do certain things, they are difficult to undo. I want to make sure we're planning for that reality where it applies.

ryguyk (approved):
Approving with understanding that this is mostly placeholder work for trying to get some sample structure in place. We'll be adjusting this heavily as we go.

PR #19025, opened 2024-08-22
88658 digital forms script

derekhouck:
## Description

Relates to #_issueid_. (or closes?)

## QA steps

1. Do this
   - [ ] Validate that
2. Then
   - [ ] Validate that
   - [ ] a
   - [ ] b
   - [ ] c

### Definition of Done

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/scripts/content/script-library.php
+++ b/scripts/content/script-library.php
@@ -19,7 +19,7 @@ use Drupal\taxonomy\Entity\Term;
 use Drupal\user\UserStorageInterface;
 use Psr\Log\LogLevel;
 
-const CMS_MIGRATOR_ID = 1317;
+const CMS_MIGRATOR_ID = 1317; // phpcs:ignore
 
 /**
  * Log a message to stdout.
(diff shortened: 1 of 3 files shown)


scripts/content/script-library.php
```diff
@@ -19,7 +19,7 @@
 use Drupal\user\UserStorageInterface;
 use Psr\Log\LogLevel;
 
-const CMS_MIGRATOR_ID = 1317;
```
derekhouck, line 22: Pre-existing code that PHP_CodeSniffer no longer accepts.

scripts/content/script-library.php
```diff
@@ -46,6 +46,19 @@ function entity_type_manager(): EntityTypeManagerInterface {
...
+ * @param string $env
+ *   The CMS environment.
+ */
+function exit_if_wrong_env(string $env) {
+  if ($env !== 'local' && $env !== 'tugboat') {
+    echo "This script can only be run on local or Tugboat environments.\n";
+    exit();
+  }
+}
+
```
ryguyk, line 61: I wonder what "wrong" is. It's clear what it is for our specific use case, but since this function appears to be implemented as a general-use function, it feels like the function name should be something more specific (e.g. `exit_if_not_local_or_tugboat` - or `exit_if_prod` but with slightly tweaked logic) OR the function should take the list of acceptable environments as a second parameter.

ryguyk (changes requested):

scripts/content/script-library.php
derekhouck: `exit_if_not_local_or_tugboat` is a more clearer name. Updated.

ryguyk (approved):

PR #19078, opened 2024-08-28
VATEAM-90732: Add OMB info fields to Digital Form

derekhouck:
## Description
Adds all fields necessary to render an accurate [OMB info component](https://design.va.gov/components/omb-info) on each form. Specifically, adds the following two fields:

- Expiration date (`string`, will be converted to expected pattern in `content-build`)
- Respondent Burden (`integer`, in minutes)

These two fields are grouped together with OMB Number to form an "OMB Info" field group.

- Closes department-of-veterans-affairs/va.gov-team#90732

## Testing done
Updated the Digital Form bundle class test with the new fields. Tested the form and view displays locally.

## Screenshots
The "OMB info" field group in the form display:
Screenshot 2024-08-28 at 2 27 51 PM

The "OMB info" field group in the node display:
Screenshot 2024-08-28 at 2 29 27 PM

## QA steps

As an administrator
1. Go to Manage > Content > Add content > Digital Form
   - [ ] Verify the "OMB info" field group is visible and the following fields are contained within it:
      - [ ] "Respondent burden"
      - [ ] "OMB Number"
      - [ ] "Expiration date"
2. Verify all three of the above fields are required.
3. Enter data in the "Respondent burden" field
   - [ ] Verify you can only enter numbers.
   - [ ] Verify you can only enter numbers greater than or equal to 0.
4. Select a date in the "Expiration date" field.
   - [ ] Verify it displays the date in MM/DD/YYYY format.
5. Fill out the other required fields on the Digital Form and click "Save" at the bottom of the page.
   - [ ] Verify the Digital Form saves successfully and the view page looks similar to the above screenshot.

### Definition of Done

- [x] If there are field changes, front end output has been thoroughly checked.

### Select Team for PR review

- [x] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/config/sync/core.entity_form_display.node.digital_form.default.yml
+++ b/config/sync/core.entity_form_display.node.digital_form.default.yml
@@ -5,15 +5,18 @@ dependencies:
   config:
     - field.field.node.digital_form.field_administration
     - field.field.node.digital_form.field_chapters
+    - field.field.node.digital_form.field_expiration_date
     - field.field.node.digital_form.field_last_saved_by_an_editor
     - field.field.node.digital_form.field_meta_tags
     - field.field.node.digital_form.field_omb_number
+    - field.field.node.digital_form.field_respondent_burden
     - field.field.node.digital_form.field_va_form_number
     - node.type.digital_form
     - workflows.workflow.editorial
   module:
     - change_labels
     - content_moderation
+    - datetime
     - field_group
     - limited_field_widgets
     - no_table_drag
@@ -29,7 +32,7 @@ third_party_settings:
       label: 'Editorial Workflow'
       region: content
       parent_name: ''
-      weight: 6
+      weight: 7
       format_type: fieldset
       format_settings:
         classes: ''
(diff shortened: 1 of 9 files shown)


ryguyk (approved):
Looks great!

I do have some lingering questions about how configuration like this will be presented/honored/etc. in our custom form experience, but I do think this work will be useful regardless.

derekhouck:
> I do have some lingering questions about how configuration like this will be presented/honored/etc. in our custom form experience, but I do think this work will be useful regardless.

If we can tie our custom display into Drupal's display configuration system, that would be ideal, but I don't know off-hand how much work that would be at this time.

ryguyk:
> If we can tie our custom display into Drupal's display configuration system, that would be ideal, but I don't know off-hand how much work that would be at this time.

We can, at least to some degree. But there's still the question of how things map from the default display to our custom display. For example, does a fieldset become its own page? Or, perhaps to ask it better, since we will have designs for each page of the wizard-like experience, do we define fieldsets accordingly?

In other words, we should be able to access the display configuration, but there are some questions in my head about what we should do with that information or how what that means for how we configure it in the first place.

Issue #19173, opened 2024-09-10
Content audit search: Content types list is not visibly sorted


jilladams:
## User Story or Problem Statement
As a CMS user performing a content audit search, 
I want to be able to select relevant content types easily, 
so that I get relevant results.

## Description or Additional Context 

On the content audit search, the list for content types is not visibly sorted. https://prod.cms.va.gov/admin/content/audit/search

If you inspect, it becomes clear that it's sorted by content type machine name. However, when machine names are no longer used as the popular / common name for a content type, that makes the list weird to navigate. 

![Screenshot 2024-09-10 at 10 21 10 AM](https://github.com/user-attachments/assets/c40a8dd8-172b-4749-ae65-36183a78c409)

## Acceptance Criteria
- [ ] Content type list on content audit search uses a visible sort for better UX

jilladams:
I became annoyed by this, talked to DaveP about it, and he POCed a fix that would alpha sort this list by the value used in the Select, rather than the machine name. SO: I'm cutting the ticket to provide a pathway to merge that improvement.

davidmpickett:
@jilladams PR is ready for review and approval. I'll let you ask CMS team for review

davidmpickett:
Confirmed in prod. The first result is still out of order, but otherwise alpha: https://prod.cms.va.gov/admin/content/audit/search

![Screenshot 2024-09-16 at 4 19 57 PM](https://github.com/user-attachments/assets/64be6ebd-7045-4fe9-95d7-9f3e6757d3d9)


![Screenshot 2024-09-16 at 4 20 05 PM](https://github.com/user-attachments/assets/b38b6d66-9544-4bac-a213-05506314f34a)

PR #19174, opened 2024-09-10
VACMS-19173: Sort content type filter on Content Search view

davidmpickett:
## Description

Relates to #19173

## Testing done

Manual in tugboat

## Screenshots

Screenshot 2024-09-10 at 3 42 45 PM

Screenshot 2024-09-10 at 2 20 57 PM

## QA steps

As a Content Admin or above (e.g. david.pickett@va.gov)
1. Log into the [Tugboat](https://pr19174-cjxiqspnzmgbwxaf2nhket5cmqeykad6.ci.cms.va.gov/)
2. Then go to Content > Content Audit tools > [Content Search](https://pr19174-cjxiqspnzmgbwxaf2nhket5cmqeykad6.ci.cms.va.gov/admin/content/audit/search)
3. Look at the Content type filter and scroll through to see it's in alpha order
4. **SOLO** 
   - [ ] Select a Content type (or ctrl-click to select multiple) and click Filter
   - [ ] Validate that the results only include the selected Content Type(s)
   - [ ] Click the Reset button
5. **NULL** 
   - [ ] Perform a search using another filter (e.g. Title contains "chicago") but no Content Type selected
   - [ ] Validate that the results are not restricted by content types
   - [ ] Click the Reset button
6. **COMBO**
   - [ ] Perform a search using another filter (e.g. Title contains "chicago") & a Content Type selected (e.g. VAMC facility)
   - [ ] Validate that the results match both filter conditions
   - [ ] Click the Reset button

### Definition of Done


jilladams:
QA steps check out, also tested multi-select for multiple content types. Approving!

There is one quirk that I don't think should block this PR bc it may require eng work to resolve and the work in here is a net improvement on the UX even if we don't fix this right now: the list of Content Types is alpha sorted _except_ for the first entry, which starts with M in the user friendly value, and K in the machine name, but appears atop the list:  

Screenshot 2024-09-10 at 2 05 24 PM

jilladams (approved):

davidmpickett:
> QA steps check out, also tested multi-select for multiple content types. Approving!
> 
> There is one quirk that I don't think should block this PR bc it may require eng work to resolve and the work in here is a net improvement on the UX even if we don't fix this right now: the list of Content Types is alpha sorted _except_ for the first entry, which starts with M in the user friendly value, and K in the machine name, but appears atop the list:
> 
> Screenshot 2024-09-10 at 2 05 24 PM

That is a test content type that is created when lower environments are built. I think something about order of operations of how Tugboat builds might be what's bumping it out of alpha order. Agree that it shouldn't block PR given that otherwise we've made a big improvement

jilladams:
Christian signed off: https://dsva.slack.com/archives/C0FQSS30V/p1726072919469939?thread_ts=1726072832.206209&cid=C0FQSS30V.  Enabled auto-merge, so once it passes tests again, it should merge.

omahane (approved):
I agree that this is an improvement, but I also noticed the same issue of the first item being out of order. But it seems to change which one it is. 

Before config sync:
Screenshot 2024-09-12 at 10 03 22 AM

After config sync:
Screenshot 2024-09-12 at 10 18 52 AM

PR #19226, opened 2024-09-17
VAGOV-TEAM-90124: Unique Form Number

ryguyk:
## Description
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/90124

This PR does three main things:
1. Scaffolds our custom module (`va_gov_form_builder`) that will be used extensively for the Form Builder in Drupal, and adds an entry in CODEOWNERS to tie this module to the `@department-of-veterans-affairs/form-engine` team.
2. Implements a `UniqueField` constraint and `UniqueFieldValidator` constraint validator that can ensure uniqueness of entities of a given type based on a field of a given name.
3. Adds the constraint and validator to `field_va_form_number` on the `digital_form` content type.

## Testing done
- Unit and functional tests added.
- Manual testing of node edit/create forms for Digital Forms.

## Screenshots
image

## QA steps

(Ensure Form Builder module is installed/enabled).

### Node Create
1. Navigate to `node/add/digital_form`
   - [ ] Add a new node with a form number that is (likely) unique. 
   - [ ] Validate that the node is successfully saved.
2. Navigate again to `node/add/digital_form`
   - [ ] Add a new node with the same form number from step 1.
   - [ ] Validate that the node is NOT successfully saved and that the error shown in the above screenshot is presented to the user.
   - [ ] Change the form number to a new number that is (likely) unique.
   - [ ] Validate that the node is successfully saved.

### Node Edit
1. Edit the node saved in step 2 by navigating to `admin/content`, locating the node in question, and clicking the `Edit` button.
   - [ ] Enter the same form number from step 1.
   - [ ] Validate that the node is NOT successfully saved and that the error shown in the above screenshot is presented to the user.

### Definition of Done

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`
- [x] `Form Engine`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


derekhouck (commented):
Is there a way to guarantee uniqueness on a database level with Drupal? Particularly because we are ejecting from the standard Drupal UI, I worry about a uniqueness constraint that relies entirely on us adding the validator to every conceivable place a digital form may be created or edited. I think this validator is necessary and beneficial, but is it enough on its own to guarantee uniqueness in form numbers?

tests/phpunit/va_gov_form_builder/functional/Plugin/Validation/Constraint/UniqueFieldValidatorTest.php
```diff
@@ -0,0 +1,187 @@
...
+   * Returns a Digital Form node that has isNew() false.
+   *
+   * @param string $formNumber
+   *   The VA form number to apply to the Digital Form.
+   *
+   * @return \Drupal\node\Entity\Node
+   *   A Digital Form node
+   */
+  private function createExistingDigitalForm($formNumber = self::UNIQUE_VA_FORM_NUMBER) {
+    // Calling $this->createNode saves the entity to the database,
```
derekhouck, line 95: Am I understanding correctly that the unique constraint does not prevent saving to the database? The non-unique node fails validation, but is it still in the table?

derekhouck (approved):
This looks good. I have a small question about whether nodes with violations still get saved to the database or not, but it is not blocking.

It strikes me that this validation is generic enough that it would probably be useful to other Content Types and modules with the CMS. Are we reinventing the wheel here, and if not, is there anything we should do to make sure other CMS devs don't have to reinvent the wheel when they need unique fields in their content types?

ryguyk:
>  Are we reinventing the wheel here, and if not, is there anything we should do to make sure other CMS devs don't have to reinvent the wheel when they need unique fields in their content types?

@derekhouck From what I could find, there were existing bespoke validators/constraints that deal with uniqueness, but they were (too) specific to the context for which they were written (e.g. unique title). I intentionally wrote this constraint so it would be more general-use, so it could potentially benefit another engineer at some point in the future, but I also introduced a sort of namespace that was previously not used widely (`id = "FormBuilder_UniqueField"`) because it felt odd to me that there could be a name collision if, say, two modules tried to implement similar but distinct validators/constraints of the same name. So it's interesting, at this point, to consider another engineer/team/application utilizing this validator/constraint that is named specifically to apply to our Form Builder.

So, while I wrote it to be general, it is still implemented within our module and so I thought it should be prefixed/namespaced accordingly. I do think that this could be used more widely, but, in that case, I think there's a strong argument for it to be moved outside our `va_gov_form_builder` module. Where? I'm not entirely sure. `va_gov_backend`, perhaps, but that already feels like a dumping ground. This needs review from CMS Team, so perhaps @timcosgrove or @edmund-dunn can weigh in.

tests/phpunit/va_gov_form_builder/functional/Plugin/Validation/Constraint/UniqueFieldValidatorTest.php
ryguyk: > As I understand things, that call is always made in normal circumstances when a node is created/attempted to be saved.

Update: $entity->validate() is _not_ called automatically. It _is_ called on the default form-create/form-edit UI, so that experience works "automatically" once we define the constraint/validator and tie it to the node type and field. In order to get things to work in other places, we have at least the following two options:

1. Implement `hook_entity_presave`, in which we'd call `$entity->validate()` explicitly and then not have to do so in individual cases:
   ```
    /**
     * Implements hook_entity_presave().
     */
    function va_gov_form_builder_entity_presave(EntityInterface $entity) {
      // Check if the entity is of the type you want to validate.
      if ($entity->getEntityTypeId() === 'node' && $entity->getType() === 'digital_form') {
        // Validate the entity.
        $violations = $entity->validate();

        // Handle validation errors.
        if ($violations->count() > 0) {
          throw new \Drupal\Core\Entity\EntityStorageException(t('The entity has validation errors.'));
        }
      }
    }
   ```

2. Call `$entity->validate()` anywhere we'd be attempting to create/save nodes programmatically.
   ```

    /**
     * Custom form submit handler.
     */
    function my_custom_form_submit(array &$form, FormStateInterface $form_state) {
      $values = $form_state->getValues();

      $form_name = $form_state->getValue('form_name');
      $form_number = $form_state->getValue('form_number');
      $omb_number = $form_state->getValue('omb_number');

      $digital_form = Node::create([
        'type' => 'digital_form',
        'title' => $form_name,
        'field_va_form_number' => $form_number,
        'field_omb_number' => $omb_number,
      ]);

      // Validate the node entity.
      /** @var \Symfony\Component\Validator\ConstraintViolationListInterface $violations */
      $violations = $node->validate();

      if ($violations->count() > 0) {
        foreach ($violations as $violation) {
          $field_name = $violation->getPropertyPath();
          $message = $violation->getMessage();

          $form_state->setErrorByName($field_name, $message);
        }

        // Prevent saving by exiting the function early.
        return;
      }

      // If there are no violations, save the entity.
      $node->save();

      // Redirect or display a success message.
      \Drupal::messenger()->addMessage(t('Node saved successfully.'));
      $form_state->setRedirect(...); //Go to subsequent page
    }

   ```

It seems that there is a strong argument for making this validation default (option 1), but the biggest drawback seems like it might be the loss of granularity surrounding which field causes the violation, and tying that back to form errors. For that reason, option 2 seems like it might be the better path to pursue first.

derekhouck: After discussing in a huddle, I agree option two seems like the way forward. I appreciate you bringing this to my attention. If we need to implement any kind of CSV or API import moving forward, we will need to make sure we run `$node->validate()` on any nodes created from those imports.

srancour (commented):
I just tested this using the QA steps above and this is not working. I was able to create multiple forms all with the same number and the error never showed up. It didn't make a difference if they were being saved in draft or published. Here is the [content page filtered to the forms with the same numbers that I created in the tugboat instance](https://pr19226-x2w3iqsghtvpfnyxwqoqer3jihq65pee.ci.cms.va.gov/admin/content?title=&type=digital_form&moderation_state=All&owner=All).

ryguyk:
> I just tested this using the QA steps above and this is not working. I was able to create multiple forms all with the same number and the error never showed up. It didn't make a difference if they were being saved in draft or published. Here is the [content page filtered to the forms with the same numbers that I created in the tugboat instance](https://pr19226-x2w3iqsghtvpfnyxwqoqer3jihq65pee.ci.cms.va.gov/admin/content?title=&type=digital_form&moderation_state=All&owner=All).

@srancour Thanks Steve! The module was not installed/enabled on the Tugboat instance. I enabled it and added that to the QA steps.

This is the PR that introduces the module, so I should have explicitly included installing/enabling it in the QA steps. In future PRs, I hope to have this configured to be installed/enabled by default.

ryguyk:
@srancour But this all might be a moot point. @davidmpickett raised attention to the "Allow Only One" module: https://dsva.slack.com/archives/CT4GZBM8F/p1727802046603989?thread_ts=1727797660.036409&cid=CT4GZBM8F

I raised some of my own concerns with that approach, but I think consistency might be a thing here. Steve, I wonder if your team has a strong feeling about this.

ryguyk:
> This is a good callout for me to consider the module installation as part of the PR.
It occurs to me that I'm not sure how this is generally handled. How are modules enabled by default in Tugboat? Are they installed on prod after first introduced, then kept enabled as prod gets copied down? Or what is the best practice here?

srancour:
> I raised some of my own concerns with that approach, but I think consistency might be a thing here. Steve, I wonder if your team has a strong feeling about this.

The thing we are pushing for in our collab cycle is to use modules that already exist, if they can accomplish the thing you're trying to do, over custom code for the exact reasons that @davidmpickett raised in that thread. Custom code is too easily forgotten about and can lead to breaking the experience when we undoubtedly continue to update drupal and modules to new versions. Tagging @gracekretschmer-metrostar, @edmund-dunn, @timcosgrove if any of you want to chime in here as well about the collab cycle and/or module vs custom code use.

ryguyk:
> The thing we are pushing for in our collab cycle is to use modules that already exist

To be very clear, I agree almost entirely with this, in general.

But, this approach runs into problems when the solution that already exists might not be the best solution. So do we double down on it? _Can_ we even double down on it? Specifically, in this case, this existing solution is actually broken:
- https://dsva.slack.com/archives/CT4GZBM8F/p1727818121815619?thread_ts=1727797660.036409&cid=CT4GZBM8F
- https://www.drupal.org/project/allow_only_one/issues/3465556#comment-15737027

So we can't actually use it reliably right now. So, rather than:
> Custom code is too easily forgotten about and can lead to breaking the experience when we undoubtedly continue to update drupal and modules to new versions.

...it's actually the case that the contributed module is the thing that breaks with version updates. A custom solution that is much more bespoke and less general-use will almost always be more resistant to this type of version breakage. This is why I'm always very hesitant to reach for contrib modules in general. 

Now, in this case, this module was developed specifically for this project by a member of the project while he was working on the project, and then became a contrib module because it was contributed back to the community as something that might benefit others. But the author has since moved off this project and seemingly cannot prioritize this fix. So we are stuck waiting (or motivated to contribute a fix, but that requires a lot more time than writing our own solution that covers much less surface area, and time is not something many of us have a lot of 😅).

So I'm not sure what this means. We need this functionality, but I'm not sure the existing solution is really a great option at this point.

Note: There is a workaround, of sorts, but this doesn't feel like it's a good resolution to my concern: https://dsva.slack.com/archives/CT4GZBM8F/p1727818736590729?thread_ts=1727797660.036409&cid=CT4GZBM8F

docroot/modules/custom/va_gov_form_builder/src/Plugin/Validation/Constraint/UniqueFieldValidator.php
```diff
@@ -0,0 +1,147 @@
...
+    // accessCheck FALSE to ensure all nodes of type `$nodeType` are checked,
+    // even if editor does not have access.
+    $query->accessCheck(FALSE)
+      ->condition('type', $this->getBundleType())
+      ->condition($this->getFieldName(), $this->fieldValue);
+
+    // If there's a current node in question,
+    // exclude the current node from the query.
+    if (!$this->entity->isNew()) {
+      $query->condition('nid', $this->entity->id(), '!=');
```
timcosgrove, line 111: What about this constraint on non-nodes?

docroot/modules/custom/va_gov_form_builder/src/Plugin/Validation/Constraint/UniqueFieldValidator.php
```diff
@@ -0,0 +1,147 @@
...
+   */
+  private function getBundleType() {
+    return $this->entity->bundle();
+  }
+
+  /**
+   * Returns the bundle label of the entity being validated.
+   */
+  private function getBundleLabel() {
+    return $this->entityTypeManager->getStorage('node_type')->load($this->getBundleType())->label();
```
timcosgrove, line 76: Is it possible to make this more universally applicable, rather than being specifically for node fields?

timcosgrove (approved):
This passes testing, and the code looks sound. FWIW, I would like to eventually see this generalized, but I think that is @department-of-veterans-affairs/cms-team 's work, not @department-of-veterans-affairs/form-engine team work.

docroot/modules/custom/va_gov_form_builder/src/Plugin/Validation/Constraint/UniqueFieldValidator.php
ryguyk: Same as other comment. This is fairly general, but developed within the context of Digital Form _nodes_. It shouldn't be too difficult to generalize further if needed in the future.

derekhouck (approved):

timcosgrove (approved):

Issue #19255, opened 2024-09-18
[Find Forms] Remove CMS feature toggle and vets-api feature flipper for Find Forms modal


randimays:
## Description
In https://github.com/department-of-veterans-affairs/va.gov-cms/issues/18855, we re-added the download modal to form detail pages. This required implementing a feature toggle for the content-build changes and a feature flipper for the vets-api changes. This ticket is for removing both of those.

#### Feature flipper (vets-api)
Code in vets-api: https://github.com/department-of-veterans-affairs/vets-api/blob/98f206c757b92eca2f062ec4b170f00a487e8c1a/config/features.yml#L513
Feature flipper name: `find_forms_show_pdf_modal`

#### Feature toggle (CMS)
Code in CMS repo: https://github.com/department-of-veterans-affairs/va.gov-cms/blob/a93d0ce7e87cc0b81720fe221f17e7fee306f587/config/sync/feature_toggle.features.yml#L8
Feature toggle name: `FEATURE_FIND_FORMS_MODAL`

#### Feature flipper (vets-api)

- [Flipper admin for staging](https://staging-api.va.gov/flipper/features)
- [Flipper admin for production](https://api.va.gov/flipper/features)
- Usage in vets-website: https://github.com/department-of-veterans-affairs/vets-website/blob/943dd069a3a818c059cf60702b9fc1f2f9493eb4/src/applications/find-forms/widgets/createFindVaFormsPDFDownloadHelper/index.js#L80

#### Feature toggle (CMS)
- Dashboard for production: https://prod.cms.va.gov/admin/config/system/feature_toggle
- Note that the staging CMS feature toggle doesn't affect any FE, so it's not necessary to turn it off here
- Usage in content-build:
  - https://github.com/department-of-veterans-affairs/content-build/blob/9d4593ba7c204b375295a3dae40de7ee50905865/src/site/filters/liquid.js#L2196
  - https://github.com/department-of-veterans-affairs/content-build/blob/9d4593ba7c204b375295a3dae40de7ee50905865/src/site/layouts/va_form.drupal.liquid#L98
  - https://github.com/department-of-veterans-affairs/content-build/blob/9d4593ba7c204b375295a3dae40de7ee50905865/src/site/components/va_related_form.drupal.liquid#L14

### Quality / testing notes
[...]

Issue #18855, opened 2024-08-06
Find Forms: Add download modal on form detail pages


randimays:
## Description
On form detail pages, the download link should open the same modal that is present on the Find Forms search results page. If a PDF is opened in the browser, the user will not be able to save it potentially causing a lot of lost time and work and a frustrating experience. We can't change the user's default settings, but we can guide them to use a PDF reader to open the document (which is the purpose of the modal).

Page examples:
- https://www.va.gov/find-forms/about-form-10-10118/
- https://www.va.gov/find-forms/about-form-10-10072a/
- https://www.va.gov/find-forms/about-form-10-10072c/

#### Form detail pages download link
<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNU0zQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d452901fced9ba3f4a8e2479165b1ad59b3391d6/Screenshot%202024-08-06%20at%2011.51.02%20AM.png" alt="Screenshot 2024-08-06 at 11.51.02 AM.png" />

#### Find Forms search results page
<img width="500" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNVEzQmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--843abe200b6713ff841529327349186e3312f267/Screenshot%202024-08-06%20at%2011.52.29%20AM.png" alt="Screenshot 2024-08-06 at 11.52.29 AM.png" />

### User story

**AS A** Veteran, their caregiver, family member or survivor
**I WANT** to be prompted to use, and/or download Adobe Acrobat to fill our my PDF forms
**SO THAT** I don't open a PDF in my browser, fill it our, then be frustrated when I lose my information.

## Acceptance criteria
- [ ] The modal that currently displays on Find a Form now also displays on the Form Detail page when a PDF link is clicked.

FranECross:
@Agile6MSkinner Just an FYI that I moved this to the top of the backlog for Randi to take care of in Sprint 10.

randimays:
Update: I have added back the modal, but I am blocked in a few ways by the find forms API. In short: I cannot test this locally or in the review instances. I have two PRs open for this work; one for content-build and one for vets-website.

[...]

PR #19611, opened 2024-10-25
VACMS-19255 Remove Find Forms CMS toggle

randimays:
# DO NOT MERGE
This should only be merged after:
- https://github.com/department-of-veterans-affairs/content-build/pull/2328 is merged to production and validated successfully
- this feature toggle is turned off in production

## Description

Ticket: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/19255

In [#18855](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/18855), we added a CMS feature toggle for the forms detail pages to show a `<button>` for a modal trigger (as the form download link) rather than an `<a>`. We will be making the `<button>` the default behavior and removing the conditionals, so we can also remove this CMS feature toggle.


SnowboardTechie (approved):
👍🏻

PR #20113, opened 2024-12-16
Add rule for owenrship of the Manila custom module

jilladams:
## Description
We added a custom module, but not a codeowner rule. Adding that now.


omahane (approved):
Thanks for doing this, @jilladams. Next time I have to add a module, I'll remember to do the same.

PR #20326, opened 2025-01-22
Remove file called 'core' that should not have been added.

timcosgrove:
## Description
This just removes a file that I mistakenly committed to the repo. It appears to be generated by my local ddev instance and it should not be there.


brianseek (approved):
LGTM

Issue #19162, opened 2024-09-09
Recurring Events "Ends" dropdown exists on the page at -1,000px


srancour:
## Statuses
**[2025-01-9] [Fran]** Since this is a defect, and we still own Events, we have refined it

## Describe the defect
In Drupal, when an event is made to be recurring there is a dropdown for "Ends" after the "Repeat every" "Repeat" type dropdown in the DOM. On the page, the "Ends" dropdown sits at -1,000px due to a `left: -1,000px;` on the `.form-item--field-datetime-range-timezone-0-repeat-end`css class. This makes it so anyone on a screen smaller than 2915px can't see it, but it still exists and can be tabbed to with keyboard and a screen reader user still encounters it.

## To Reproduce
1. Edit an event that's recurring like [Music Therapy Jam Session](https://staging.cms.va.gov/central-iowa-health-care/events/53044)
2. Tab though options after "Repeat every" input and dropdown
3. Either see focus onto "Ends" dropdown if screen size is at least 2915px wide, or see focus move onto nothing and then back on second tab.

## Notes 2025-01-09
- See this comment below: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/19162#issuecomment-2339302008

## AC / Expected behavior
- [ ] The value for Ends should be permanently set to On Date
- [ ] When "Make recurring" is selected on the event form, the Ends field should not be readable by screen readers or visible on screen

## Screenshots
![Screenshot of part of the dropdown visible to the left of the screen with a green focus on it](https://github.com/user-attachments/assets/d7b9b788-60f9-4032-95a0-7055ede398d4)
![Screenshot of the dropdown opened and on the 'After' option showing that the option to input text goes in under the 'On these days' option](https://github.com/user-attachments/assets/077c62f1-bba9-4b11-af63-8220bb3ec475)

 ## Additional context
This seems to have been here for 2 years since the [Recurring Events feature](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/6057) was completed at pull #7981. It looks like it came from a [request to 'Remove "Never/After/On date" dropdown'](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/7307#issuecomment-998203897) in a comment on #7037 by @kevwalsh from 2021.

## Desktop (please complete the following information if relevant, or delete)
 - Window width of at least 2915px to visually see dropdown

jilladams:
[...]

PR #20344, opened 2025-01-24
VACMS-19162: Ensure repeat-end dropdown does not exist on the page.

dsasser:
## Description
Hides the repeat-end dropdown on the page to ensure screen readers cannot see it, and editors cannot change the default value from 'On Date'. This prevents editors from inadvertently selecting a recurring pattern that is infinite while keeping the element on the form for functional purposes (otherwise the repeating capabilities would break).

Relates to #19162

## Testing done
Tested in browser

### Before
Screenshot of part of the dropdown visible to the left of the screen with a green focus on it
Screenshot of the dropdown opened and on the 'After' option showing that the option to input text goes in under the 'On these days' option

### After
Screenshot 2025-01-24 at 11 19 31 AM
Screenshot 2025-01-24 at 11 20 22 AM

## QA steps
As a [QA Content Editor](https://pr20344-tvwxs9lbrm4ag2xfonstzxbj2e9qlikl.ci.cms.va.gov/user/4415)

- [ ] Visit the [Create Event](https://pr20344-tvwxs9lbrm4ag2xfonstzxbj2e9qlikl.ci.cms.va.gov/node/add/event) page
- [ ] Fill out the Event date, ensuring to select 'Make Recurring'
- [ ] Select 'weeks' for the 'Repeat every' option
- [ ] Zoom out your screen until it is at least 2915px
- [ ] Verify you cannot see the 'Ends' dropdown (it has 'Never', 'After', and 'On Date') options
- [ ] Tab though options after "Repeat every" input and dropdown and ensure you don't see focus onto "Ends" dropdown if screen size is at least 2915px wide, or see focus move onto nothing and then back on second tab.

### Select Team for PR review

- [ ] `CMS Team`
- [x] `Public websites`
- [ ] `Facilities`
- [ ] `User support`


laflannery (approved):
Looks good!

SnowboardTechie (approved):

PR #20824, opened 2025-03-06
VACMS-00000: fix password issue

edmund-dunn:
## Description

- Adds the force_users_logout module but does not enable it, just to make it an available tool
- resets user passwords for all but a small set of users.


timcosgrove (dismissed):

timcosgrove:
I recommend we hold this until we determine the date to execute.

timcosgrove (dismissed):

timcosgrove (approved):

PR #21051, opened 2025-03-31
VAGOV-TEAM-100805 digital form app url field & render logic

tjheffner:
## Description

Relates to https://github.com/department-of-veterans-affairs/va.gov-team/issues/100805

Adds new field to the Digital Form content type. Updates logic for previewing a Digital Form to use new field.

## Testing done

Verified button rendered (or didn't render) with these 5 cases of field values:
   - `https://staging.va.gov/form-application-url`
   - `staging.va.gov/form-application-url`
   - `/form-application-url`
   - `form-application-url`
   - `empty field`

## Screenshots

Screenshot 2025-03-31 at 1 11 14 PM
Screenshot 2025-03-31 at 1 11 25 PM

## QA steps

1. Do this
   - [ ] Validate that
2. Then
   - [ ] Validate that
   - [ ] a
   - [ ] b
   - [ ] c

### Definition of Done

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
+++ b/docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
@@ -716,18 +716,23 @@ class VaGovFormBuilderController extends ControllerBase {
    *  - /form-application-url
    *  - form-application-url
    * This function normalizes those cases to return a valid preview link.
-   *
-   * @param string $string
-   *   Digital Form node's field_form_application_url value.
    */
-  public function constructLaunchViewLink($string) {
-    $url = parse_url($string);
+  private function getLaunchViewLink() {
+    $application_url_field = $this->digitalForm->get('field_form_application_url');
+
+    // Bail early if no field value.
+    if ($application_url_field->isEmpty()) {
+      return '';
+    }
+
+    // Parse field value into url parts.
+    $url = parse_url($application_url_field->getString());
 
     $scheme = 'https://';
     $preview_host = 'staging.va.gov';
 
     if (str_starts_with($url['path'], 'staging.va.gov')) {
-      return $scheme . $string;
+      return $scheme . $url['path'];
     }
     if (!str_starts_with($url['path'], '/')) {
       return $scheme . $preview_host . '/' . $url['path'];
@@ -747,15 +752,14 @@ class VaGovFormBuilderController extends ControllerBase {
       throw new NotFoundHttpException();
     }
 
-    // To preview, first check that application url for a form exists.
-    $application_url_field = $this->digitalForm->get('field_form_application_url');
-    $isFormViewable = !$application_url_field->isEmpty();
+    $applicationUrl = $this->getLaunchViewLink();
+    $isFormViewable = !empty($applicationUrl);
     $viewFormStatus = $isFormViewable ? 'available' : 'unavailable';
 
     $buttons = $isFormViewable ? [
       'primary' => [
         'label' => 'Launch view',
-        'url' => $this->constructLaunchViewLink($application_url_field->getString()),
+        'url' => $applicationUrl,
         'target' => '_blank',
       ],
       'secondary' => [
--- a/docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
+++ b/docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
@@ -708,6 +708,34 @@ class VaGovFormBuilderController extends ControllerBase {
   }
 
   /**
+   * Uses field value to construct qualified form application preview link.
+   *
+   * Likely to be in one of three or four formats:
+   *  - https://staging.va.gov/form-application-url
+   *  - staging.va.gov/form-application-url
+   *  - /form-application-url
+   *  - form-application-url
+   * This function normalizes those cases to return a valid preview link.
+   *
+   * @param string $string
+   *   Digital Form node's field_form_application_url value.
+   */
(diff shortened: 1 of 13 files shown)


derekhouck:
I just loaded up the deployment environment and I'm not seeing the "App URL" field on the Digital Form edit page. Can you confirm it is showing for you in the deployment environment?

tjheffner:
> I just loaded up the deployment environment and I'm not seeing the "App URL" field on the Digital Form edit page. Can you confirm it is showing for you in the deployment environment?
 
confirmed it wasn't showing up & that was my bad 😅 I hadn't re-imported the config after rebasing locally & fixing conflicts. When the environment rebuilds, they should be available this time. Sorry about that!

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
```diff
@@ -707,6 +707,34 @@ public function reviewAndSign($nid) {
...
+
+    if (str_starts_with($url['path'], 'staging.va.gov')) {
+      return $scheme . $string;
+    }
+    if (!str_starts_with($url['path'], '/')) {
+      return $scheme . $preview_host . '/' . $url['path'];
+    }
+    return $scheme . $preview_host . $url['path'];
+  }
+
```
ryguyk, line 737: Designing `constructLaunchViewLink` to take a parameter is exactly my first inclination as well, so this resonates with me in many ways. And there's certainly room for philosophical discussions around whether that is the best path forward or not, but - for better or worse - I've architected things in such a way where I think this can be made a bit more OOP. As things stand, the `digitalForm` is tracked as internal state in the controller, so I think we can write a method that operates on that state rather than on passed-in parameters.

```
private function getLaunchViewLink() {
  $applicationUrl = $this->digitalForm->get('field_form_application_url');
  if ($applicationUrl->isEmpty()) {
    return ''; 
  }
  .
  .
  .
}
```

To be clear, my first thought was that the method as you have it written might best be declared `static`. But that is a thought I've often had when my first inclination (coming from a more functional paradigm) is to write more functional, pure "utility" methods like this. So, when I had that thought here, it then occurred to me that perhaps we could write this a bit differently to more align with the OOP structure around it.

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
```diff
@@ -719,21 +747,16 @@ public function viewForm($nid) {
...
+    $isFormViewable = !$application_url_field->isEmpty();
     $viewFormStatus = $isFormViewable ? 'available' : 'unavailable';
 
     $buttons = $isFormViewable ? [
       'primary' => [
         'label' => 'Launch view',
-        // Temporary. Send somewhere other than layout page for now.
-        'url' => $this->getPageUrl('home'),
+        'url' => $this->constructLaunchViewLink($application_url_field->getString()),
+        'target' => '_blank',
```
ryguyk, line 759: This logic looks great! 

I have a couple comments about getting this aligned a little more closely with what I have architected:

1. I think variable names in this file should largely be `$camelCase`. I have generally been following a pattern where variables are `$camelCase` in the controller and other OOP lands, and `$snake_case` in more procedural settings like hooks.
2. See comment about the method `constructLaunchViewLink`. If that is changed to read internal state, then this could become something like:
```
$applicationUrl = $this->getLaunchViewLink();
$isFormViewable = !empty($applicationUrl);
.
.
.
    'url' => $applicationUrl,
```

docroot/modules/custom/va_gov_form_builder/templates/page-content/base/single-column-with-buttons.html.twig
```diff
@@ -21,6 +21,9 @@
       href="{{ buttons.primary.url }}"
       role="button"
       class="form-builder-button form-builder-button--primary"
+      {% if buttons.primary.target is defined %}
+      target="{{ buttons.primary.target }}"
+      {% endif %}
```
ryguyk, line 26: Nice!

docroot/modules/custom/va_gov_form_builder/templates/page-content/base/single-column-with-buttons.html.twig
```diff
@@ -21,6 +21,9 @@
       href="{{ buttons.primary.url }}"
       role="button"
       class="form-builder-button form-builder-button--primary"
+      {% if buttons.primary.target is defined %}
+      target="{{ buttons.primary.target }}"
```
ryguyk, line 25: Should line 25 be indented?

ryguyk (changes requested):
Very nice! I requested some changes via a few comments about aligning with some of the more esoteric architecture surrounding this. Let me know if you have any strong feelings against those.

ryguyk:
@tjheffner Where is this coming from?

image

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
tjheffner: Yeah that's fine with me. How do you feel about expanding the method slightly from only constructing the link, e.g. something like `constructFormButtons`, given that the `isFormViewable` logic around the buttons also uses that application url field check?

tjheffner:
> @tjheffner Where is this coming from?
Ah, that was the name of a test form I left the field empty on. I see now how that is not the clearest name with a low-context screenshot 😅

PR #20373, opened 2025-01-30
VAGOV-TEAM-101466: Layout Page

ryguyk:
## Description
This work implements the bulk of the Layout page in Form Builder.
- The general structure is implemented, as shown in the screenshot below.
- Logic to calculate and display step statuses is added.
- A link to the Form Info page is added.
- Other links are left blank for now, as the pages do not exist.

To accomplish this work, some foundational pieces have also been added and/or refactored:
- A `DigitalForm` class is added that serves as a wrapper around a `digital_form` node. This allows business logic around step statuses and other related ideas to be called on the DigitalForm (node-wrapper) itself. Initially during this work, I was adding a lot of this code to the `DigitalFormsService` class, which effectively is a utility class that was requiring a `digital_form` node to be passed in to each call. 
 
This started to feel clunky:
```
'status' => $this->digitalFormsService->getDigitalFormStepStatus($this->digitalFormNode, 'contact_info'),
```
This is much better:
```
'status' => $this->digitalForm->getStepStatus('contact_info'),
```
- The service still exists, and its chief responsibility is fetching `digital_form` nodes and returning `DigitalForm` objects.

Relates to https://github.com/department-of-veterans-affairs/va.gov-team/issues/101466

## Testing done
- Unit/Functional tests added and updated.
- Manual testing done on all Form Builder pages, as this PR refactors some code that touches all pages.

## Screenshots
image

## QA steps
As an admin:
1. Navigate to `/form-builder`.
   - [x] Validate that you are redirected to `/form-builder/home`.
   - [x] On Home page, click on "Build a form".
   - [x] Validate that you are redirected to `/form-builder/form-info` (Form Info page in "create" mode).
1. On the Form Info page:
   - [x] Fill in the fields to create a new form.
   - [x] Click "Save and continue".
   - [x] Validate that you are redirected to the Layout page at `/form-builder/{nid}/layout`.
1. On the Layout page:
   - [x] Validate that the "Form Info" step is "COMPLETE" (green status label).
   - [x] Validate that the Introduction page and Form Steps are all "INCOMPLETE" (red status label). Note: Your personal information will eventually be complete by default, but that work still needs to be done.
   - [x] Validate that the "Review and sign" step is "COMPLETE" (green status label). Note: this page hasn't been completed yet in Form Builder, but this step is automatic. There is no database/content-model component to this.
   - [x] Click on the "View form info" button and validate that you are taken back to the Form Info page, this time in "edit" mode (i.e. fields are populated with the information you just entered to create this form).
[...]

PR #21100, opened 2025-04-08
VAGOV-TEAM-106493: Custom/additional Step Status

dsasser:
## Description

Relates to: department-of-veterans-affairs/va.gov-team#106493

### Generated description

This pull request introduces several changes to enhance the functionality and validation of digital forms in the `va_gov_form_builder` module. The most significant updates include adding a new trait for entity reference revisions operations, updating the logic for determining step status, and modifying the `DigitalForm` class to incorporate these changes.

### Enhancements to Digital Form functionality:

* **Step Status Logic Update:**
  - Updated the logic for determining the status of custom steps using the `getStepStatus` method in the `DigitalForm` class. This now includes recursive entity validation to check for any violations in the custom step paragraph or its descendant paragraphs. [[1]](diffhunk://#diff-9dffa0f8b39e9a7ba0a5c28ef2b99bd9a06ca95b1f77852efd1896d10e9b920cL450-R452) [[2]](diffhunk://#diff-66168129b5c4f6b7bf146b4a0d6471d519f281139944ddd5ec43376c30cf6139R180-R187) [[3]](diffhunk://#diff-66168129b5c4f6b7bf146b4a0d6471d519f281139944ddd5ec43376c30cf6139R215-R225)

* **Entity Reference Revisions Operations Trait:**
  - Introduced a new trait `EntityReferenceRevisionsOperations` to handle recursive validation of entity reference revisions. This trait is used to validate nested entities and ensure comprehensive validation.

### Codebase modifications:

* **DigitalForm Class:**
  - Added the `EntityReferenceRevisionsOperations` trait to the `DigitalForm` class to utilize its recursive validation functionality.
  - Included the `paragraph` field in the `getAllSteps` method to pass the paragraph entity along with the step details.
  - Updated method signatures and docblocks to reflect the changes in validation and step status logic. [[1]](diffhunk://#diff-66168129b5c4f6b7bf146b4a0d6471d519f281139944ddd5ec43376c30cf6139R5-R17) [[2]](diffhunk://#diff-66168129b5c4f6b7bf146b4a0d6471d519f281139944ddd5ec43376c30cf6139R180-R187)

## Testing done
Manual and PHPUnit

## QA steps
As a Form Engine user:
- [ ] Visit /form-builder/76353
- [ ] Validate that the step "Step 4: Veteran's employment history" status is "COMPLETE"
- [ ] Click the 'Add a Step' button
- [ ] Fill in the label and click 'Save and Continue'
- [ ] Select the 'Add single question' button
- [ ] Click the 'Return to steps' button
- [ ] Validate the new step status is "INCOMPLETE"
- [ ] Visit /node/76353/edit
- [ ] Find the new step you added under the 'Steps' section.
- [ ] Fill out the rest of the step by adding a page and any component
- [ ] Save the node
- [ ] Visit /form-builder/76353
- [ ] Validate the new step status is "COMPLETE"

### Definition of Done

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

### DIFF
--- a/docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
+++ b/docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
@@ -447,12 +447,9 @@ class VaGovFormBuilderController extends ControllerBase {
         'steps' => array_map(function ($step) {
           $stepParagraphId = $step['fields']['id'][0]['value'];
           return [
-            // If an additional step exists, it's complete.
-            // @todo This logic needs to be updated to accommodate custom
-            // steps that are not complete by default.
             'type' => $step['type'],
             'title' => $step['fields']['field_title'][0]['value'],
-            'status' => 'complete',
+            'status' => $this->digitalForm->getStepStatus('custom', $step['paragraph']),
             'url' => Url::fromRoute('va_gov_form_builder.step.layout', [
               'nid' => $this->digitalForm->id(),
               'stepParagraphId' => $stepParagraphId,
--- a/docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
+++ b/docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
@@ -19,7 +19,7 @@ use Drupal\va_gov_form_builder\Traits\EntityReferenceRevisionsOperations;
  * The following line is included because the public method `addStep`
  * makes dynamic calls to otherwise uncalled private methods, and those
  * private methods trigger unused-method warnings when they are, in fact,
- * not unused:
+ * unused:
  * @phpcs:disable DrupalPractice.Objects.UnusedPrivateMethod.UnusedMethod
  */
 class DigitalForm {
--- a/docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
+++ b/docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
@@ -14,6 +14,7 @@ use Drupal\va_gov_form_builder\Traits\EntityReferenceRevisionsOperations;
  * @method int id()
  * @method string getTitle()
  * @method \Drupal\Core\Field\FieldItemListInterface get(string $field_name)
+ * @method \Symfony\Component\Validator\ConstraintViolationListInterface validate()
  * @method int save() Saves the entity and returns the save status
  * @method NodeInterface set(string $field_name, mixed $value, bool $notify = TRUE) Sets a field value
  *
@@ -178,14 +179,14 @@ class DigitalForm {
    *
    * @param string $stepName
    *   The step name of the step in question.
-   * @param mixed|\Drupal\paragraphs\ParagraphInterface $paragraph
+   * @param null|\Drupal\paragraphs\ParagraphInterface $paragraph
    *   Optional paragraph entity for this step.
    *
    * @return string
    *   Returns 'complete' if step is complete. Returns 'incomplete' if step is
    *   incomplete or if the step name does not exist.
    */
-  public function getStepStatus(string $stepName, mixed $paragraph = NULL) {
+  public function getStepStatus(string $stepName, null|ParagraphInterface $paragraph = NULL) {
     if ($stepName === 'form_info') {
       // If the node exists, this will necessarily be complete.
       return 'complete';
--- a/docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
+++ b/docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
@@ -4,16 +4,16 @@ namespace Drupal\va_gov_form_builder\EntityWrapper;
 
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\node\NodeInterface;
+use Drupal\paragraphs\ParagraphInterface;
+use Drupal\va_gov_form_builder\Traits\EntityReferenceRevisionsOperations;
 
 /**
  * A wrapper class around Digital Form nodes.
  *
  * @method int id()
  * @method string getTitle()
- * @method \Drupal\Core\Field\FieldItemListInterface get(string $field_name)
- * @method \Symfony\Component\Validator\ConstraintViolationListInterface validate()
  * @method int save() Saves the entity and returns the save status
- * @method \Drupal\node\NodeInterface set(string $field_name, mixed $value, bool $notify = TRUE) Sets a field value
+ * @method NodeInterface set(string $field_name, mixed $value, bool $notify = TRUE) Sets a field value
  *
  * The following line is included because the public method `addStep`
  * makes dynamic calls to otherwise uncalled private methods, and those
@@ -22,6 +22,9 @@ use Drupal\node\NodeInterface;
  * @phpcs:disable DrupalPractice.Objects.UnusedPrivateMethod.UnusedMethod
  */
 class DigitalForm {
+
+  use EntityReferenceRevisionsOperations;
+
   /**
    * The standard steps of a Digital Form.
    */
@@ -105,6 +108,7 @@ class DigitalForm {
         if ($paragraph) {
           $steps[] = [
             'type' => $paragraph->bundle(),
+            'paragraph' => $paragraph,
             'fields' => array_map(function ($field) {
               return $field->getValue();
             }, $paragraph->getFields()),
@@ -172,13 +176,14 @@ class DigitalForm {
    *
    * @param string $stepName
    *   The step name of the step in question.
+   * @param \Drupal\paragraphs\ParagraphInterface|null $paragraph
+   *   Optional paragraph entity for this step.
    *
-   * @return 'complete'|'incomplete'
-   *   Returns 'complete' if step is complete.
-   *   Returns 'incomplete if step is incomplete
-   *   or if the step name does not exist.
+   * @return string
+   *   Returns 'complete' if step is complete. Returns 'incomplete' if step is
+   *   incomplete or if the step name does not exist.
    */
-  public function getStepStatus($stepName) {
+  public function getStepStatus(string $stepName, ParagraphInterface|\NULL $paragraph = \NULL) {
     if ($stepName === 'form_info') {
       // If the node exists, this will necessarily be complete.
       return 'complete';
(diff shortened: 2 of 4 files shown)


docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
```diff
@@ -105,6 +110,7 @@ public function getAllSteps() {
         if ($paragraph) {
           $steps[] = [
             'type' => $paragraph->bundle(),
+            'paragraph' => $paragraph,
```
dsasser, line 113: Including the paragraph in the step gives us access to all entity operations while using the steps.

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
```diff
@@ -447,12 +447,9 @@ public function layout($nid) {
         'steps' => array_map(function ($step) {
           $stepParagraphId = $step['fields']['id'][0]['value'];
           return [
-            // If an additional step exists, it's complete.
-            // @todo This logic needs to be updated to accommodate custom
-            // steps that are not complete by default.
             'type' => $step['type'],
             'title' => $step['fields']['field_title'][0]['value'],
-            'status' => 'complete',
+            'status' => $this->digitalForm->getStepStatus('custom', $step['paragraph']),
```
dsasser, line 452: Necessary in order to get the status for the custom step. Used 'custom' here as that was the language used in the ticket to describe the steps we are targetting.

docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
```diff
@@ -2,26 +2,31 @@
...
 
 /**
  * A wrapper class around Digital Form nodes.
  *
  * @method int id()
  * @method string getTitle()
  * @method \Drupal\Core\Field\FieldItemListInterface get(string $field_name)
- * @method \Symfony\Component\Validator\ConstraintViolationListInterface validate()
  * @method int save() Saves the entity and returns the save status
- * @method \Drupal\node\NodeInterface set(string $field_name, mixed $value, bool $notify = TRUE) Sets a field value
```
dsasser, line 16: Making my IDE happy by simplifying FQDN, as we already have a use statement for this interface.

docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
```diff
@@ -2,26 +2,31 @@
...
+ * @method NodeInterface set(string $field_name, mixed $value, bool $notify = TRUE) Sets a field value
  *
  * The following line is included because the public method `addStep`
  * makes dynamic calls to otherwise uncalled private methods, and those
  * private methods trigger unused-method warnings when they are, in fact,
- * not unused:
+ * used:
```
dsasser, line 23: The addStep methods are in fact being called, despite the PHPCS warning.

docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
```diff
@@ -2,26 +2,31 @@
...
  * @phpcs:disable DrupalPractice.Objects.UnusedPrivateMethod.UnusedMethod
  */
 class DigitalForm {
+
+  use EntityReferenceRevisionsOperations;
```
dsasser, line 28: The trait is intended to allow this and future potential paragraph operations to be encapsulated outside the domain of the DigitalForm class.

docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
```diff
@@ -172,13 +178,14 @@ public function hasChapterOfType($type) {
    *
    * @param string $stepName
    *   The step name of the step in question.
+   * @param mixed|\Drupal\paragraphs\ParagraphInterface $paragraph
```
dsasser, line 181: Added optional paragraph param so the step can make decisions with the entity rather than only the values.

docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
```diff
@@ -206,6 +213,17 @@ public function getStepStatus($stepName) {
         : 'incomplete';
     }
 
+    // Use recursive entity validation to determine custom step status. Any
+    // violation in the custom step paragraph or descendent paragraphs will
+    // cause the status to be incomplete. This method is used to reduce
+    // technical debt if the custom step content model, or the content model of
+    // its children, is updated. This could also support future enhancements to
+    // display _why_ the step is not complete.
+    if ($paragraph instanceof ParagraphInterface && $stepName === 'custom') {
```
dsasser, line 222: The status is dependant on the entity validating, leveraging Drupal's ability to validate an entity, which uses the existing field constraints defined for the field. This would also invoke any custom constraints the entities or fields that are present. This avoids duplicating (or missing) existing or future constraints, while keeping tech debt down.

docroot/modules/custom/va_gov_form_builder/src/Traits/EntityReferenceRevisionsOperations.php
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * @param \Drupal\Core\Entity\ContentEntityInterface $entity
+   *   The entity to validate recursively.
+   * @param \Drupal\Core\Entity\EntityConstraintViolationList $violations
+   *   Violations to add to the list.
+   *
+   * @return \Drupal\Core\Entity\EntityConstraintViolationListInterface
+   *   The comprehensive violations list for referenced the referenced entity.
+   */
+  public function recursiveEntityReferenceRevisionValidator(ContentEntityInterface $entity, EntityConstraintViolationList $violations): EntityConstraintViolationList {
```
dsasser, line 24: This trait exists to perform validation recursively. Validating a node would return no validations even if a custom step (field_chapter->field_digital_form_page) was empty. It is necessary to descend the entity tree in order to perform validation on each referenced entity. In this way, if we add an additional field requirement to the custom step (field_digital_form_page) or one of its components (field_digital_form_page_component), the validations will react accordingly. Keeping parity with the content model design was the intent here.

derekhouck (approved):
This approval will probably get stale very quickly, but mostly doing it to remind myself that it LGTM after our 16th minute conversation about it. Good job!

docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
ryguyk: Warning: Nerd talk forthcoming 🤓 

First and foremost, I like that you used a trait here. I think it is very clean to implement this in a general way that is reusable beyond just the `DigitalForm` class, and I think a trait is a very reasonable way to extract that logic. **I'm not suggesting any changes here**. The conversation that follows is simply for (primarily my own) edification. Please feel free to point out where I might be talking stupid.

Having said that, I find myself wondering about two things:
1. I don't mean to be pedantic, but I think it's worth saying aloud that I don't think traits do anything to enforce encapsulation, per se. I do understand the use of that word ("encapsulation") in a more generic sense to describe pulling the code out into its own "capsule" (file), but in the true OOP sense of the word, I don't think traits have any control over encapsulation.
2. Perhaps more interestingly, I find myself wondering if this needs to be a class method at all. Since you're passing in everything that is needed, it feels to me like this is more of a utility-service function (like a pure functional-programming function).

I don't mean to open a can of worms. I'm happy to not discuss this at all. I tend to ask these sorts of questions in the hopes of learning a few things. Thanks for putting up with them 😅

docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
```diff
@@ -206,6 +213,17 @@ public function getStepStatus($stepName) {
...
+      $violations = $this->recursiveEntityReferenceRevisionValidator($paragraph, new EntityConstraintViolationList($paragraph));
```
ryguyk, line 223: **Not blocking**. More just thinking aloud.

Passing `new EntityConstraintViolationList($paragraph)` as the second parameter initially confused me because it feels like that could be handled inside the method and then just pass `$paragraph` as a single parameter. I do understand that the recursive nature of that method requires being able to pass a second parameter more flexibly, but I find myself wondering if the first call should not require passing in a list of violations, as that feels odd to me.

ryguyk (changes requested):
This looks really good. I requested one small change (with a link to some previous documentation around why), and then added a number of other musings that you can safely ignore unless you're motivated to respond.

Nicely done!

docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
dsasser: 1. The passed in EntityConstraintViolationList() is an empty set, it doesn't perform validation here, it instantiates an empty iterable object. I did try to avoid this, many many times, but PHPCS was complaining any time I tried to set a default value for this parameter, despite mimicking the parent __construct() method signature, as well as other methods. The call to EntityConstraintViolationList() is quick and lightweight so I didn't feel that the tradeoff of time to make PHPCS happy was worth that rabbit hole.
2. This also avoids having a conditional statement in the recursive function that checks if the $violations is empty., and if so applies validation to the incoming entity, which allowed a cleaner approach and less executed code during the recusive process. Though it was more of a code style implementation decision than one of performance.

```php
    /**
     * Creates a new constraint violation list.
     *
     * @param iterable<mixed, ConstraintViolationInterface> $violations The constraint violations to add to the list
     */
    public function __construct(iterable $violations = [])
    {
        foreach ($violations as $violation) {
            $this->add($violation);
        }
    }
```

dsasser:
> Why mixed here? Will it not always be either a paragraph entity or null?

I'll have a look again. PHPCS was giving me fits about `null|ParagraphInterface` for absolutely no clue why.

dsasser:
> Fair enough. I think the intention of the original language was to emphasize that while it is being triggered as "unused", it is not, in fact, unused. The emphasis might not be so necessary as to require the double negative, though, so I'm fine with this.

It didn't make sense to me as I read it. The methods are in fact used, so I figured that was an oversight. If you would like me to change it back let me know.

dsasser:
> This is necessary here and cannot be removed. See [here](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/20373/files#r1937525408). We should perhaps move some of the linked commentary into a code comment.

> TL;DR The annotation here is necessary for calls to validate (and all others listed here) from outside the class (as well as inside).

Not sure I grok the reason actually, but I'm not married to this change. My IDE was letting me know it wasn't used, and I scanned the codebase for the use of validate() (in this class). From your linked PR, it seemed the main reason for this was to satisfy PHPStan? Did I get that right? Anyway, I'll change it back.

docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
ryguyk: Very good insight. Thanks!

ryguyk (approved):
Thanks!

docroot/modules/custom/va_gov_form_builder/src/EntityWrapper/DigitalForm.php
ryguyk: Very thoughtful path.

PR #21212, opened 2025-04-22
VAGOV-TEAM-106480: Date and Date-range flows

ryguyk:
## Description

This work implements the UI for the process of adding/editing date components as part of custom steps. The flow is spread over a number of pages that are implemented here:
1. Date-type selection
2. Page title and body - This page is also used for choice and text components, and should largely be done for those components as well.
3. Response input
   - Single date
   - Date Range

This PR also implements some refactoring and features across the Form Builder UI:
1. Removes the `getFields()` approach to validation, and instead requires classes to override validation methods. This has the benefit of added flexibility in terms of mapping validation errors to specific form fields.
2. Related, but separately, this PR simplifies the naming of fields on each form, as the move away from `getFields()` removes the tight coupling between names on our custom forms and the underlying entities in Drupal. We can now connect a validation error to an arbitrary field (even a nested one, as is necessary with date-range questions).
3. Adds status messages to all forms. This hasn't been strictly necessary up to this point, as all form validation errors can be shown on the individual fields, but this PR adds some form-level error reporting that cannot be tied to a specific field.

Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/106480
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/106481

## Testing done
- Manual testing of the entire Form Builder UI.
- Updated unit tests for refactored base form classes.
- Added unit tests for added themes and libraries.

### Date-type selection
image

### Page title and body
image

#### Single date
image

#### Date range
image

## QA steps
This PR implements some refactoring across Form Builder. We should test all of it.

As a Form Builder user:
1. On the home page, click "Build a form".
   - [x] Validate that the form is created.
1. On the form-layout page, click "View form info". Edit the form info.
   - [x] Validate that the form info is updated.
1. Add a step:
   - On the form-layout page, click "Add step".
   - Enter the step label.
   - Click "Add a single question".
1. Add a question:
   - Click "Add question".
   - Click "Customize".
   - Select "Date", then click "Save and continue".
      - [x] Validate that you are taken to the date-type-selection page.
1. Add a single-date question:
   - Select "Single date", then click "Save and continue".
   - [x] Validate that you are taken to the page-title page for single-date questions.
   - Enter the page title and optional body text, then click "Save and continue".
   - [x] Validate that you are taken to the single-date response-input form.
1. Add a single-date response:
   - Fill in the response information, then click "Save and continue".
   - [x] Validate that you are taken back to the step-layout page.
   - [x] Validate the your new question is shown on the step-layout page.
1. Add another question (this time for date-range):
   - Click "Add question".
   - Click "Customize".
   - Select "Date", then click "Save and continue".
      - [x] Validate that you are taken to the date-type-selection page.
1. Add a date-range question:
   - Select "Date range", then click "Save and continue".
   - [x] Validate that you are taken to the page-title page for date-range questions.
   - Enter the page title and optional body text, then click "Save and continue".
   - [x] Validate that you are taken to the date-range response-input form.
1. Add a date-range response:
   - Fill in the response information, then click "Save and continue".
   - [x] Validate that you are taken back to the step-layout page.
   - [x] Validate the your new question is shown on the step-layout page.

### Definition of Done

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`
- [ ] `Form Engine`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
```diff
@@ -884,6 +1072,304 @@ public function customSingleQuestionResponseKind($nid, $stepParagraphId) {
...
+    try {
+      $pageComponentType = self::getPageComponentType($this->pageParagraph);
+    }
+    catch (\Exception $e) {
+      // This is technically not a page-not-found
+      // error state, but we don't have a better way to
+      // handle this currently.
+      // @todo Implement better way to handle general error states.
+      throw new NotFoundHttpException();
+    }
```
ryguyk, line 1362: This is a case where a page would be found, but the component type cannot be determined. So, it's kind of "not found" in that we wouldn't have a page to show the user. But perhaps there could be a better way to handle error cases, where we can indicate what the problem is rather than just throwing up a "not found" page. Not something to be solved right now.

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
```diff
@@ -884,6 +1072,304 @@ public function customSingleQuestionResponseKind($nid, $stepParagraphId) {
...
+
+    switch ($pageComponentType) {
+      case CustomSingleQuestionPageType::SingleDate:
+        return $this->customSingleQuestionSingleDateResponse($nid, $stepParagraphId, $pageParagraphId);
+
+      case CustomSingleQuestionPageType::DateRange:
+        return $this->customSingleQuestionDateRangeResponse($nid, $stepParagraphId, $pageParagraphId);
+    }
+  }
```
ryguyk, line 1371: This needs to be expanded as we add more component types.

docroot/modules/custom/va_gov_form_builder/src/Form/Base/FormBuilderBase.php
```diff
@@ -76,29 +77,62 @@ public static function create(ContainerInterface $container) {
...
    * @param \Symfony\Component\Validator\ConstraintViolationList $violations
    *   The list of validation violations.
-   * @param string[] $fieldList
-   *   A list of fields to check for validation violations. Violations will
-   *   only trigger form errors if the field to which the violation applies
-   *   is in this list.
+   * @param array $fieldMapping
+   *   An array of field names to check for validation violations.
+   *   Keys are the field names on the entity, and values are the
+   *   corresponding form elements.
```
ryguyk, line 120: This captures a change that was introduced in this PR. We no longer pass just a flat field list to the `setFormErrors` function. Instead, we pass an associative array where each field on the entity that is being validated maps to a field on the form. This leads to the removal of `getFields` on each form class, because now we aren't simply passing in the result of a call to `getFields()`. Instead, each form must define its own validation functions where it can define this mapping.

docroot/modules/custom/va_gov_form_builder/src/Form/FormInfo.php
```diff
@@ -59,30 +46,30 @@ public function buildForm(array $form, FormStateInterface $form_state, $digitalF
...
+    $form['respondent_burden'] = [
       '#type' => 'number',
       '#title' => $this->t('Respondent burden'),
       '#description' => $this->t('Number of minutes as indicated on the form'),
       '#required' => TRUE,
       '#default_value' => $this->getDigitalFormFieldValue('field_respondent_burden'),
     ];
 
-    $form['field_expiration_date'] = [
+    $form['expiration_date'] = [
```
ryguyk, line 72: Another outcome of the removal of `getFields` and the move to a mapping for setting form errors is that our forms no longer need to have field names that match the field name on the entity. This gives us more flexibility, and it also allows more brevity.

docroot/modules/custom/va_gov_form_builder/src/Form/StepLabel.php
```diff
@@ -17,15 +17,6 @@ public function getFormId() {
...
 
-  /**
-   * {@inheritdoc}
-   */
-  protected function getFields() {
-    return [
-      'field_title',
-    ];
-  }
-
```
ryguyk, line 28: Here is an example of removing `getFields`.

docroot/modules/custom/va_gov_form_builder/templates/base/single-column-with-buttons.html.twig
```diff
@@ -1,4 +1,10 @@
 <div id="form-builder-content-{{ page_id }}" class="form-builder-content form-builder-single-column-content form-builder-{{ page_type }}-content form-builder-content--single-column-with-buttons">
+  {% if block('messages') %}
+    <div class="form-builder-single-column-with-buttons__messages">
+      {% block messages %}{% endblock %}
+    </div>
+  {% endif %}
+
```
ryguyk, line 7: Previously, status messages were not showing. This explicitly adds a block to our templates. In our forms, we add a `status_messages` property to the form render arrays, and in the template we render this element in the appropriate block.

docroot/modules/custom/va_gov_form_builder/templates/base/two-column-with-buttons.html.twig
```diff
@@ -0,0 +1,30 @@
+<div id="form-builder-content-{{ page_id }}" class="form-builder-content form-builder-two-column-content form-builder-{{ page_type }}-content form-builder-content--two-colum-with-buttons">
+  {% if block('messages') %}
+    <div class="form-builder-two-column-with-buttons__messages">
+      {% block messages %}{% endblock %}
+    </div>
+  {% endif %}
```
ryguyk, line 6: Same comment about status messages. This implements a block in the two-column base template.

docroot/modules/custom/va_gov_form_builder/templates/form/base/single-column-with-buttons.html.twig
```diff
@@ -2,6 +2,10 @@
 
 {% set page_type = 'form' %}
 
+{% block messages %}
+  {{ form.status_messages }}
+{% endblock %}
```
ryguyk, line 7: Here is where the status-messages form element is actually added to the block (single-column).

docroot/modules/custom/va_gov_form_builder/templates/form/base/two-column-with-buttons.html.twig
```diff
@@ -0,0 +1,11 @@
+{% extends '@va_gov_form_builder/base/two-column-with-buttons.html.twig' %}
+
+{% set page_type = 'form' %}
+
+{% block messages %}
+  {{ form.status_messages }}
+{% endblock %}
```
ryguyk, line 7: Here is where the status-messages form element is actually added to the block (two-column).

docroot/modules/custom/va_gov_form_builder/templates/form/custom-single-question-single-date-response.html.twig
```diff
@@ -0,0 +1,63 @@
+{#
```
ryguyk, line 1: There's a lot of room for some DRYing up of shared layout between single-date and date-range pages. We'll get even more of this with the addition of the other four types (radio, checkbox, text, text area). But it's a grain of sand on the beach of finding abstractions in this UI, and for now we need to move on.

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
dsasser: Agreed. The 404 will prevent acting on unknown component types which is enough safety at the moment to prevent any data persisting to the entity that would otherwise break the form.

dsasser (changes requested):
Overall this works great and is well organized. There could be opportunities for DRYing up as you say, or making abstractions that support additional flexibility, but as it stands this is well written code that is working (save the problem I had below).

There are a few sprinkled notes, but nothing is blocking other than the 404 I'm getting.

Blocking:
I'm getting a 404 when viewing a single date question from the step layout page. Date range components are working, but when I click the label of a single date question to view it, I get a 404.

Screenshot 2025-04-23 at 1 33 22 PM

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
```diff
@@ -149,6 +158,94 @@ protected function loadStepParagraph($paragraphId) {
...
+
+    // Ensure the page paragraph belongs to the step paragraph.
+    if (!$this->stepParagraph) {
+      throw new NotFoundHttpException();
+    }
+    $parentId = $paragraph->get('parent_id') && is_object($paragraph->get('parent_id'))
+      ? $paragraph->get('parent_id')->value
+      : '';
+
+    if ($parentId !== $this->stepParagraph->id()) {
```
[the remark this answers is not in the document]
ryguyk: I made a comment on your PR about the block of code you're referencing. I'm not certain I'm understanding what you're saying here. We might need to chat about this one.

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
```diff
@@ -149,6 +158,94 @@ protected function loadStepParagraph($paragraphId) {
...
+   * Loads and sets the page-paragraph object.
+   *
+   * @param string $paragraphId
+   *   The paragraph id to load.
+   *
+   * @throws \Symfony\Component\HttpKernel\Exception\NotFoundHttpException
+   *   If the paragraph is not found, or if the paragraph
+   *   does not belong to $this->stepParagraph.
+   */
+  protected function loadPageParagraph($paragraphId) {
```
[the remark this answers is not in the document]
ryguyk: I think this comment also relates to the confusion I mention elsewhere. Probably worth chatting about it.

docroot/modules/custom/va_gov_form_builder/src/Form/Base/FormBuilderBase.php
```diff
@@ -76,29 +77,62 @@ public static function create(ContainerInterface $container) {
...
    */
-  protected static function setFormErrors(FormStateInterface $form_state, ConstraintViolationList $violations, array $fieldList) {
+  protected static function setFormErrors(
```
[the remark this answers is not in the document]
ryguyk: I have been making this decision based on readability on the screen. I think this formatting has happened automatically for me in other repos, so perhaps I don't have things set up perfectly here. But this is a good callout. I prefer consistency. I'm not making a change here, specifically, as I don't think that's going to solve the problem, but this is very good to consider.

dsasser (approved):
The [previous 404](https://github.com/department-of-veterans-affairs/va.gov-cms/pull/21212#issuecomment-2828007083) when viewing a single date is now resolved, so this LGTM!

tjheffner (approved):
ran through the QA steps and everything checks out for me. Admittedly I didn't examine the code with a fine tooth comb, but nothing jumped out as a particularly unpleasant code smell

PR #21224, opened 2025-04-25
VAGOV-TEAM-106494 repeatable fields

tjheffner:
## Description

Relates to https://github.com/department-of-veterans-affairs/va.gov-team/issues/106494

## Testing done

Add this Trait to the `FormInfo.php` for testing:
```php
use Drupal\va_gov_form_builder\Traits\RepeatableFieldGroup;

class FormInfo extends FormBuilderFormBase {

use RepeatableFieldGroup;
```

Added this code block to `FormInfo.php` for testing:
```php
    $field_definitions = [
      'label' => ['#type' => 'textfield', '#title' => 'Label for radio item'],
      'description' => ['#type' => 'textfield', '#title' => 'Description for radio item'],
    ];

    $this->addRepeatableFieldGroup($form, $form_state, 'dynamic_radio', $field_definitions, 1, 5);
```

Visit `/form-builder/76353/form-info` to see the repeatable fields + button

## Screenshots
Screenshot 2025-04-28 at 11 52 53 AM
Screenshot 2025-04-28 at 11 53 12 AM
Screenshot 2025-04-28 at 11 53 29 AM

## QA steps

- [ ] Check out branch locally
- [ ] Add the code block from above
- [ ] Visit the FormInfo page to see the fields
- [ ] Enter value in fields
- [ ] Hit "add another"
- [ ] See that a new set of fields was added and the previous field values are unchanged
- [ ] Repeat until the max of 5 is hit
- [ ] See the button disappear when max items are added


docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
```diff
@@ -0,0 +1,89 @@
...
+    // Get the number of items in the form already.
+    $num_items = $form_state->get($element_name . '_count');
+    // We have to ensure that there is at least one item field.
+    if ($num_items === NULL) {
+      $form_state->set($element_name . '_count', 1);
+      $num_items = $form_state->get($element_name . '_count');
+    }
+
+    // Create the container to hold our dynamic items.
+    $form[$element_name . '_fieldset'] = [
```
ryguyk, line 25: I'm wondering if we might need to return a render array element rather than pass in `$form` and directly work with it. This might make it more possible for the caller to put it where it needs/wants it in the form structure. 
Example:
```
$form['radios'] = addRepeatableFieldGroup(...);
```

I'm not sure if this will be necessary or not. I'm mostly interested in this because of the conversation we had around how we might handle default values. Consider the approach we discussed (using radios as example):
1. Render all the existing radio fields by iterating over the components.
2. Then, after that, call this `addRepeatableFieldGroup` _only for the new ones_.

If we do this, will it be problematic that some of the radios (the existing ones) are under one parent, while the others (the new ones) are under a different parent? Will that mess us up on the form processing? Or might it actually be useful in the form editing, since new ones will possibly have to be treated differently than existing ones in terms of persisting to the database. I'm not sure, but this is one thing we'll have to have an eye out for as we try to incorporate this into the forms that need it.

ryguyk (commented):
I'm neither approving nor requesting changes. I left some thoughts. We can chat about how to best proceed. I think one very viable path forward is merging this and then seeing if we need to tweak once we incorporate it in our forms that need it.

docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
```diff
@@ -0,0 +1,89 @@
...
+      '#type' => 'container',
+      '#attributes' => ['id' => $element_name . '-wrapper'],
+    ];
+    // Helper for ajax callbacks to easily grab element name.
+    $form['#repeatableFieldName'] = $element_name;
```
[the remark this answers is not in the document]
tjheffner: Happy to make that change instead, this approach just felt straightforward to me

Are there going to be cases where there would need to be multiples of this feature per form? Given the current structure directs you to a single type of element to add per step, it didn't cross my mind that there may be multiple types of repeated fields on a single page step

e.g. these would all be cases of "one type" of repeating fields on a given step
Screenshot 2025-04-28 at 1 22 17 PM

docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
tjheffner: I need to think about this approach more. On the surface, I think it makes sense to pass in the full `$form` object, given that there are several pieces around the render array being constructed:

- the wrapping container that ajax targets
- the individual items inside said container [that will be updated/replaced when ajax rebuilds]
- the conditional form action for the ajax trigger button


I guess I'm still not quite grasping the importance of the default values here, they're all text fields right? So the default value is what was entered, or it's a blank string?

ryguyk (approved):
I think getting this merged makes the most sense. We'll see if it works for our use cases once we get it in and try to incorporate it in some real forms.

PR #21237, opened 2025-04-29
VAGOV-TEAM-106494 index iteration

tjheffner:
## Description

Relates to https://github.com/department-of-veterans-affairs/va.gov-team/issues/106494

Updates the $min parameter to be a startingIndex instead. Updates logic to use this value instead of always starting at 0. This should allow the edit page to define how many objects exist already and then add ajax items from there.

Also updated the form validation to not trigger on the ajax add button, only the actual form submit for the page.

## Testing done
Add this Trait to the `FormInfo.php` for testing:
```php
use Drupal\va_gov_form_builder\Traits\RepeatableFieldGroup;

class FormInfo extends FormBuilderFormBase {

use RepeatableFieldGroup;
```

Added this code block to `FormInfo.php` for testing:
```php
    $field_definitions = [
      'label' => ['#type' => 'textfield', '#title' => 'Label for radio item', '#required' => TRUE],
      'description' => ['#type' => 'textfield', '#title' => 'Description for radio item'],
      'required' => [
        '#type' => 'checkbox',
        '#title' => $this->t('Required for the submitter'),
        '#displayIndexCount' => FALSE
      ],
    ];

    $this->addRepeatableFieldGroup($form, $form_state, 'dynamic_radio', $field_definitions, 1, 5, 'Radio Item');
```

Visit `/form-builder/76353/form-info` to see the repeatable fields + button

## Screenshots
Count starting at a higher index:
Screenshot 2025-04-29 at 12 02 50 PM

Button disappearing when max items reached:
Screenshot 2025-04-29 at 12 02 53 PM

Validation only triggered when save & continue hit:
Screenshot 2025-04-29 at 12 02 56 PM

## QA steps

- [ ] Checkout branch locally
- [ ] Add the code from above to the necessary file
- [ ] Verify functionality works as expected for 5 new items
- [ ] Update startIndex to `3` and `ddev drush cr`
- [ ] Refresh the page, see the item count start at 3 and add items.
- [ ] Verify form validation does not trigger on the ajax button, but on the `Save & Continue` button.
- [ ] Updated startIndex to `5` and `ddev drush cr`
- [ ] Verify no button appears on the page as the max limit has been reached already.

### DIFF
--- a/docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
+++ b/docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
@@ -17,7 +17,7 @@ trait RepeatableFieldGroup {
    */
   public function addRepeatableFieldGroup(array &$form, FormStateInterface $form_state, string $element_name, array $field_definitions, int $startIndex = 1, int $max = 10, $button_name = 'Add another') {
     // If an additional item would be over the limit, bail early.
-    if ($startIndex >= $max) {
+    if ($startIndex > $max) {
       return;
     }
 
@@ -57,6 +57,9 @@ trait RepeatableFieldGroup {
     if ($num_items < $max) {
       $form[$element_name . '_fieldset']['actions'] = [
         '#type' => 'actions',
+        '#attributes' => [
+          'class' => ['form-builder-add-item-wrapper'],
+        ],
       ];
 
       $form[$element_name . '_fieldset']['actions']['add_item'] = [
@@ -67,6 +70,9 @@ trait RepeatableFieldGroup {
         '#ajax' => [
           'callback' => '::addMoreCallback',
           'wrapper' => $element_name . '-wrapper',
+        ],
+        '#attributes' => [
+          'class' => ['form-builder-add-item'],
         ],
         // Don't validate on ajax add, only on full page submission.
         '#limit_validation_errors' => [],
--- a/docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
+++ b/docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
@@ -63,6 +63,8 @@ trait RepeatableFieldGroup {
           'callback' => '::addMoreCallback',
           'wrapper' => $element_name . '-wrapper',
         ],
+        // Don't validate on ajax add, only on full page submission.
+        '#limit_validation_errors' => [],
       ];
     }
   }
--- a/docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
+++ b/docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
@@ -15,12 +15,17 @@ trait RepeatableFieldGroup {
   /**
    * Creates repeatable fields dynamically using AJAX.
    */
-  public function addRepeatableFieldGroup(array &$form, FormStateInterface $form_state, string $element_name, array $field_definitions, int $min = 1, int $max = 10) {
-    // Get the number of items in the form already.
+  public function addRepeatableFieldGroup(array &$form, FormStateInterface $form_state, string $element_name, array $field_definitions, int $startIndex = 1, int $max = 10) {
+    // If an additional item would be over the limit, bail early.
+    if ($startIndex >= $max) {
+      return;
+    }
+
+    // Get the number of ajax items in the form already.
     $num_items = $form_state->get($element_name . '_count');
     // We have to ensure that there is at least one item field.
     if ($num_items === NULL) {
-      $form_state->set($element_name . '_count', 1);
+      $form_state->set($element_name . '_count', $startIndex);
       $num_items = $form_state->get($element_name . '_count');
     }
 
@@ -33,7 +38,7 @@ trait RepeatableFieldGroup {
     // Fields get their own values instead of copying the first group.
     $form[$element_name . '_fieldset'][$element_name]['#tree'] = TRUE;
 
-    for ($i = 0; $i < $num_items; $i++) {
+    for ($i = $startIndex - 1; $i < $num_items; $i++) {
       // Create a new fields for each item.
       foreach ($field_definitions as $field_key => $definition) {
         $form[$element_name . '_fieldset'][$element_name][$i][$field_key] = $definition;
--- a/docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
+++ b/docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
@@ -15,7 +15,11 @@ trait RepeatableFieldGroup {
   /**
    * Creates repeatable fields dynamically using AJAX.
    */
-  public function addRepeatableFieldGroup(array &$form, FormStateInterface $form_state, string $element_name, array $field_definitions, int $startIndex = 1, int $max = 10, $button_name = 'Add another') {
+  public function addRepeatableFieldGroup(array &$form, FormStateInterface $form_state, string $element_name, array $field_definitions, int $min = 1, int $max = 10, $startIndex = 1, $button_name = 'Add another') {
+    // This should never happen, but you never know.
+    if ($min > $max) {
+      return;
+    }
     // If an additional item would be over the limit, bail early.
     if ($startIndex > $max) {
       return;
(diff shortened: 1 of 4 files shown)


derekhouck (approved):
LGTM, but I'm not in the weeds, so maybe wait for one more approval before merge.

docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
```diff
@@ -15,12 +15,17 @@ trait RepeatableFieldGroup {
   /**
    * Creates repeatable fields dynamically using AJAX.
    */
-  public function addRepeatableFieldGroup(array &$form, FormStateInterface $form_state, string $element_name, array $field_definitions, int $min = 1, int $max = 10) {
-    // Get the number of items in the form already.
+  public function addRepeatableFieldGroup(array &$form, FormStateInterface $form_state, string $element_name, array $field_definitions, int $startIndex = 1, int $max = 10) {
+    // If an additional item would be over the limit, bail early.
+    if ($startIndex >= $max) {
```
ryguyk, line 20: Just a note that callers will _not_ need to do any subtraction when calling, as I previously asserted. I was envisioning this in the caller:
```
    $this->addRepeatableFieldGroup(
      $form,
      $form_state,
      'dynamic_text_input_fields',
      $repeatableTextInputFieldDefinitions,
      2,
      self::MAX_REPEATABLE_FIELDS - $existingComponentCount,  // HERE
    );
```

But we should not need to do this, as this "plugin" is treating things in a way that effectively says, "Give me the _total_ number of inputs you want, *including the ones you have rendered manually*. Also, tell me how many you've rendered there (`$startingIndex`)."

I think this works well.

ryguyk, line 20: Just a note that callers will _not_ need to do any subtraction when calling, as I previously asserted. I was envisioning this in the caller:
```
    $this->addRepeatableFieldGroup(
      $form,
      $form_state,
      'dynamic_text_input_fields',
      $repeatableTextInputFieldDefinitions,
      2,
      self::MAX_REPEATABLE_FIELDS - $existingComponentCount,  // HERE
    );
```

But we should not need to do this, as this "plugin" is treating things in a way that effectively says, "Give me the _total_ number of inputs you want, *including the ones you have rendered manually*. Also, tell me how many you've rendered there (`$startingIndex`)."

I think this works well.

docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
```diff
@@ -15,12 +15,17 @@ trait RepeatableFieldGroup {
...
+      return;
+    }
```
ryguyk, line 22: Just a note that callers will _not_ need to do any subtraction when calling, as I previously asserted. I was envisioning this in the caller:
```
    $this->addRepeatableFieldGroup(
      $form,
      $form_state,
      'dynamic_text_input_fields',
      $repeatableTextInputFieldDefinitions,
      2,
      self::MAX_REPEATABLE_FIELDS - $existingComponentCount,  // HERE
    );
```

But we should not need to do this, as this "plugin" is treating things in a way that effectively says, "Give me the _total_ number of inputs you want, *including the ones you have rendered manually*. Also, tell me how many you've rendered there (`$startingIndex`)."

I think this works well.

ryguyk, line 22: Just a note that callers will _not_ need to do any subtraction when calling, as I previously asserted. I was envisioning this in the caller:
```
    $this->addRepeatableFieldGroup(
      $form,
      $form_state,
      'dynamic_text_input_fields',
      $repeatableTextInputFieldDefinitions,
      2,
      self::MAX_REPEATABLE_FIELDS - $existingComponentCount,  // HERE
    );
```

But we should not need to do this, as this "plugin" is treating things in a way that effectively says, "Give me the _total_ number of inputs you want, *including the ones you have rendered manually*. Also, tell me how many you've rendered there."

I think this works well.

ryguyk (changes requested):
This looks great! Nothing here should need changing.

I do think we need one more feature, though.

image

I'm thinking we can update things in one of two ways:

1.
```
    $repeatableTextInputFieldDefinitions = [
      'label' => [
        '#type' => 'textfield',
        '#title' => $this->t($labelText),
        '#required' => TRUE,
      ],
      'hint_text' => [
        '#type' => 'textfield',
        '#title' => $this->t($hintText),
      ],
      'required' => [
        '#type' => 'checkbox',
        '#title' => $this->t('Required for the submitter'),
        // ACCEPT THIS CUSTOM KEY.
        // DEFAULTS TO TRUE.
        '#displayIndexCount' => FALSE
      ],
    ];
```
2.
```
    $this->addRepeatableFieldGroup(
      $form,
      $form_state,
      'dynamic_text_input_fields',
      $repeatableTextInputFieldDefinitions,
      2,
      self::MAX_REPEATABLE_FIELDS - $existingComponentCount,
      // ACCEPT THIS PARAMETER THAT DEFINES
      // WHICH KEYS SHOULD HAVE THE COUNTER
      // APPENDED.
      [
        'label',
        'hint_text',
      ]
    );
```

docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
ryguyk, line 20: I think `$startIndex` can be equal to `$max`, just not greater than it. So:
```
if ($startIndex > $max)
```
That said, I think we need to update a few other things to accommodate some other requirements, and those might play into this logic. So perhaps not useful to think of this in isolation right now. In either case, I think this general idea should hold, where the index can equal the max, but it's an error once it goes beyond that.

docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
```diff
@@ -33,7 +38,7 @@ public function addRepeatableFieldGroup(array &$form, FormStateInterface $form_s
     // Fields get their own values instead of copying the first group.
     $form[$element_name . '_fieldset'][$element_name]['#tree'] = TRUE;
 
-    for ($i = 0; $i < $num_items; $i++) {
+    for ($i = $startIndex - 1; $i < $num_items; $i++) {
       // Create a new fields for each item.
       foreach ($field_definitions as $field_key => $definition) {
         $form[$element_name . '_fieldset'][$element_name][$i][$field_key] = $definition;
```
ryguyk, line 44: I'm wondering if `$i` here is still ok:
```
$form[$element_name . '_fieldset'][$element_name][$i][$field_key]
```

We will start our count at something greater than zero.

docroot/modules/custom/va_gov_form_builder/src/Traits/RepeatableFieldGroup.php
ryguyk: I've put some thought into how we can perhaps most easily accommodate the additional requirements. First, here's the brainstorming:

```
// Our API
addRepeatableFieldGroup(
  $form,
  $form_state,
  $element_name,
  $field_definitions,
  $min,
  $max,
  $start,
);

$start
  1-based index that should be the index displayed for the first element rendered by this call.
  Importantly, this element will only be rendered on the first render if $start <= $min.

$min
  The greatest 1-based index that *must* be displayed on the page. 

$max
  The greatest 1-based index that *can* be displayed on the page.

$numItems (not passed in, but calculated within)
  The number of elements that are displayed currently (including the manually rendered existing components).

Examples:
=========
1. If $min >= $max, exit early.
This is nonsense.

2. If $start > $max, exit early.
This marks the edge case where additional elements are not allowed.

3. If $start == $max, this call

3. If $start = 1 and $min = 2, this call must render TWO (2) elements.
This critical example captures the initial case for a create-mode form for radio/checkbox inputs.

4. If $start = 3 and $min = 2 and $max = 7, this call must not render any elements on initial form render,
but will allow up to FIVE (5) button clicks. This example captures the case for an edit-mode form for
radio/checkbox inputs that had exactly two items previously added, and now the user has returned to edit.
Again, in this case, there should be no empty fields displayed unless the user clicks the "Add" button.

```
Then, here is the code I think we'd need in order to kick things off correctly:
```
    if ($min >= $max) {
      return;
    }

    if ($start > $max) {
      return;
    }

    // Get the number of items in the form already.
    $num_items = $form_state->get($element_name . '_count');

    // If null, this is the first render.
    // We need to render
    if ($num_items === NULL) {
      $num_items = max($min, $startIndex - 1);
      $form_state->set($element_name . '_count', $num_items);
    }
```

ryguyk (approved):
This looks great! I think (read: hope) this covers all of the requirements now. If any others pop up, we can address in another follow-up PR, but that feels unlikely now. I think we can get this merged.

PR #21298, opened 2025-05-06
VAGOV-TEAM-106487: Employment History

ryguyk:
## Description

Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/106487

## Testing done
- Unit tests added/updated for themes
- Manual testing done throughout Form Builder

### Repeating-set selection page
image

### Employment History page (partial)
image

## QA steps
As a Form Builder user:

1. Select an existing form or create a new one.
1. On the form-layout page, click "Add a step"
   - [x] Validate that you are taken to the step-label page.
1. Enter step label "Employment history" and click "Save and continue". 
   - [x] Validate that you are taken to the step-style page.
1. Click on "Add a repeating set".
   - [x] Validate that you are taken to the custom-or-predefined-repeating-set page.
1. Click on "Add Employment history".
   - [x] Validate that you are taken to the employment-history non-editable-pattern page.
1. At the bottom, click on "Save and continue".
   - [x] Validate that you are taken back to the form-layout page.
   - [x] Validate that your new step appears on the form-layout page.
   - [x] Validate that your new step has the "Delete" button available. If there are additional non-default steps, validate that "Move up" and "Move down" buttons are also displayed.
   - [x] Validate that the step label for your newly created step is "Your employers". Note that this is different from the label you entered. The work in this PR should overwrite the previously entered label for this step type.

### Definition of Done

### Select Team for PR review

- [ ] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`
- [x] `Form Engine`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


docroot/modules/custom/va_gov_form_builder/css/non-editable-pattern.css
```diff
@@ -14,3 +14,24 @@
...
 }
+
+.form-builder-non-editable-pattern__default-appearance-intro {
+  border-bottom: 1px solid var(--vads-color-base-light);
+  margin: var(--units-6) 0;
+  padding-bottom: var(--units-3);
+}
+
+.form-builder-non-editable-pattern__default-appearance
+  .form-builder-non-editable-pattern__default-appearance-heading {
```
ryguyk, line 25: Needed higher selector specificity here.

docroot/modules/custom/va_gov_form_builder/css/non-editable-pattern.css
```diff
@@ -14,3 +14,24 @@
...
+  font-family: var(--font-family-sans);
+}
+
+.form-builder-non-editable-pattern__default-appearance
+  .form-builder-non-editable-pattern__preview-title {
```
ryguyk, line 30: Needed higher selector specificity here.

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
```diff
@@ -318,9 +326,8 @@ protected function getPageUrl($page) {
     // Pages that relate to a step within a form.
     // Require a nid and stepParagraphId.
     $stepPages = [
-      'step.layout',
+      'step.home',
```
ryguyk, line 329: This PR includes a lot of substitutions like this. We aren't getting rid of the step-layout page, nor are we replacing it in any way. Instead, it's just no longer the entry point into the controller. Here's why:

Previously, the step-layout concept was something that we applied to every step. It definitely applied to custom single-question steps, and that page showed all the questions added to the step. It _sort of_ applied to custom list-and-loop steps, though that version of the step-layout page just indicated that the step type is not supported in Form Builder.

Now, this PR introduces another step type, `list_loop_employment_history`. Importantly, this step type does not need a step-layout page. Also, importantly, every step "home" page shares the same route, which maps to a url that is `/form-builder/{nid}/step/{stepParagraphId}`. So we need that route to map to a controller method that can determine which type of paragraph is in play, and then call the appropriate method.

So this PR adds a `stepHome` method (and concept), which figures out what kind of paragraph is being referenced, and then calls either `stepLayout` (the existing method, which handles custom single-question and custom list-and-loop steps in the same way as previously) or `employemtHistory`.

Conceptually, it's a routing method that routes a generic step URL to its appropriate handler. In fact, I initially called this `stepRouter`, but the concept of a route `step.router` felt a little odd, and `step.home` started to feel a lot better. We can now keep the concept in mind of a step having a "home" page, and that differs depending on the type of step. For a custom step, there's still the step-layout page. For a predefined step like Employment history, the home page is the page that shows what that step entails.

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
```diff
@@ -1127,83 +1306,88 @@ public function stepStyle($nid) {
...
-    //
-    // However, since the step paragraph doesn't actually exist
-    // at this point (this is part of the process of creating it),
-    // the previously entered step label is stored in session storage.
-    // So, we can't do what we'd typically do and, instead, we need
-    // to generate the breadcrumb trail without it and then splice
-    // it in.
-    $breadcrumbs = $this->generateBreadcrumbs('layout', 'Step style');
-    array_splice($breadcrumbs, 2, 0, [
-      [
```
ryguyk, line 1151: We still want the same breadcrumb, but this PR refactors how this is accomplished. In short, it moves away from splicing and towards pulling from session storage if it's there, and defaulting if it's not there, all in the `generateBreadcrumbs` method. This is much shorter, cleaner, and now more DRY as we have another method that would have needed to do this same thing.

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
```diff
@@ -1713,8 +1897,12 @@ public function reviewAndSign($nid) {
...
       '#theme' => self::PAGE_CONTENT_THEME_PREFIX . 'review_and_sign',
       '#preview' => [
-        'alt_text' => 'Statement-of-truth preview',
-        'url' => '/modules/custom/va_gov_form_builder/images/statement-of-truth.png',
+        [
+          'image' => [
+            'alt_text' => 'Statement-of-truth preview',
+            'url' => '/modules/custom/va_gov_form_builder/images/statement-of-truth.png',
+          ],
+        ],
```
ryguyk, line 1905: We see this update in a few places in this PR. In order to try to align the employment-history page with the other non-editable-pattern pages, and to make that template reusable among all of them, we needed to change the structure of the `#preview` theme variable. What used to be just `alt_text` and `url` for a single image is now an array of things that includes the image and also optional `title` and `description` keys. These are not passed here, as the existing non-editable-pattern pages were only using an image, but we can see the full structure being utilized in the new code in the `employmentHistory` method.

docroot/modules/custom/va_gov_form_builder/src/Form/Base/FormBuilderStepCustomOrPredefinedBase.php
```diff
@@ -0,0 +1,135 @@
+<?php
+
+namespace Drupal\va_gov_form_builder\Form\Base;
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Abstract base class for custom-or-predefined step pages.
+ */
+abstract class FormBuilderStepCustomOrPredefinedBase extends FormBuilderStepBase {
```
ryguyk, line 10: Much of the same code was needed for the single-question and repeating-set versions of the form. I considered using only a single form with conditional logic within, but it started to feel a lot more clean separating them. Then, with that approach, there was a lot of repeated code. So this abstract base class is the result. It allows some clean shared logic while allowing the child classes to differentiate where needed.

docroot/modules/custom/va_gov_form_builder/src/Form/StepStyle.php
```diff
@@ -1,233 +0,0 @@
-<?php
```
ryguyk, line 1: This PR changes a few things regarding where and how certain things happen, and this file deletion is a result of that. Previously, this `StepStyle` form was where a step paragraph was actually created. Now, as this work helped us discover that the paragraph creation should actually happen after the _next_ screen, this form is no longer needed. This step-style page still is needed, but this PR moves it to a page-content page rather than a form page, as it doesn't really do anything substantial any longer. 

So some of this logic (i.e. the creation of the paragraph) is now moved into the new custom-or-predefined-* forms, and this file is removed.

docroot/modules/custom/va_gov_form_builder/va_gov_form_builder.routing.yml
```diff
@@ -46,10 +46,10 @@ va_gov_form_builder.contact_info:
     _controller: '\Drupal\va_gov_form_builder\Controller\VaGovFormBuilderController::contactInfo'
   requirements:
     nid: \d*
-va_gov_form_builder.step.layout:
+va_gov_form_builder.step.home:
```
ryguyk, line 49: See previous commentary about `step.layout` vs. `step.home`. This is the route change in accordance with that.

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
```diff
@@ -1053,6 +1201,37 @@ public function stepLayout($nid, $stepParagraphId) {
...
+    // Custom list & loop.
+    if ($stepType === 'digital_form_custom_step' || $stepType === 'digital_form_list_loop') {
+      return $this->stepLayout($nid, $stepParagraphId);
+    }
+
+    // Employment history.
+    if ($stepType === 'list_loop_employment_history') {
+      return $this->employmentHistory($nid, $stepParagraphId);
+    }
+  }
```
dsasser, line 1233: Unlikely (impossible) to occur I know, but there is nothing returned if the step type isn't one of the three in the above conditionals. Should we have a fallback response if that should happen in the future?

docroot/modules/custom/va_gov_form_builder/src/Form/CustomOrPredefinedRepeatingSet.php
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setStepParagraphFromFormState(FormStateInterface $form_state) {
+    // Predefined employment-history step always gets a hard-code
+    // label "Your employers". This should override the label previously
+    // entered by the user.
+    // Note: the user is notified of this in the UI.
```
dsasser, line 77: I have to admit to missing this text the first time through.

docroot/modules/custom/va_gov_form_builder/templates/page-content/step-style.html.twig
```diff
@@ -4,18 +4,25 @@
  * Template for Form Builder step page: Step Label
```
dsasser, line 4: > Step Label

Should be 'Step Style'?

dsasser (approved):
QA steps passed, abstraction decisions make sense and code quality is great (especially all the helpful comments). LGTM!

I did have a couple questions/comments but nothing blocking.

docroot/modules/custom/va_gov_form_builder/src/Controller/VaGovFormBuilderController.php
ryguyk: Good call! 

For now, I'm going to throw a `LogicException`. I thought about a 404 (`NotFoundHttpException`), as we have in some other places, but I've found during development that sometimes I'll hit a 404 and it's really because of something like this where it would have been better if I had thrown an exception with an explanation. In this case, this will likely only ever be raised during new feature development when an engineer perhaps forgets to wire this up with the new paragraph type. In that case, it'll be much more useful to have this throw an exception that explains the issue.

docroot/modules/custom/va_gov_form_builder/src/Form/CustomOrPredefinedRepeatingSet.php
ryguyk: In the UI? Or this comment?

I don't think it's obvious in the UI at all. Perhaps that's something we can raise with Design in the audit.

dsasser: Yeah in the UI I didn't read fully the help text so it took me a second to figure out why the label changed.

Issue #21301, opened 2025-05-06
Cannot Add VAMC Facility Health Services to Academy VA Clinic or Evans VA Clinic (VISN 19, VA Eastern Colorado)


stefaniefgray:
## Describe the defect
New VAMC Facility Health Services cannot be added to either the Academy VA Clinic or Evans VA Clinic (VISN 19, VA Eastern Colorado) homepages.

Both entries show up accurately on the Facility Locator, but something may have happened when these facilities were removed then re-published two years later (without the CMS Support team being informed of this) that affected them in particular.

## To Reproduce
1. Use the Content menu to navigate to the "Add VAMC Facility Health Service" page: https://prod.cms.va.gov/node/add/health_care_local_health_service
2. For "Section" (located on the right side menu) select "VA Eastern Colorado" (VISN 19)
3. Next, find the "Health service and facility basic info" menu. 
4. For "System Health Service" select any health service listed.
5. For "Facility" select Academy VA Clinic or Evans VA Clinic.
6. You will not be able to create any new VAMC Facility Health Services for either of these locations.

## Screenshots

**If you follow the above steps in order, you will encounter the following error message:** 

(Note that node 7415 is the VAMC System Health Service for "Addiction and substance use care at VA Eastern Colorado health care")

<img width="850" alt="Image" src="https://github.com/user-attachments/assets/d41db1e6-aded-49f6-b7e7-a0a7575788c7" />

**However, if you select an alternate clinic such as Garden of the Gods or Golden -- and THEN try to select Academy or Evans, you will see the following error message:**

<img width="817" alt="Image" src="https://github.com/user-attachments/assets/d4a435a2-a130-44a6-bca1-138bde6cecd7" />

## Context
The Academy VA Clinic and Evans VA Clinic VAMC Facility homepages were automatically created by the CMS in June 2022 after being pulled in from the Facilities API.

We followed up with the VISN 19 team in 2022 about adding facility health services before each clinic homepage could be published to the live site. (Veteran impact: Our team does not want new facility homepages to be published without any health services information listed.)

Their team specifically instructed us not to publish either clinic homepage, and to remove both of them from the live site. Jira ticket: https://va-gov.atlassian.net/browse/VAHELP-4965

However, in 2024, both facility homepages were published by members of the VISN 19 team. They did not notify the VA Drupal CMS Help Desk Support team, and therefore, the new VAMC facility publishing process was not vetted or followed.

[...]

PR #21310, opened 2025-05-06
VACMS-21301: VAMC Facility Health Service bug for facilities with no existing services

brianseek:
## Description
There was a bug in the SQL queries to reduce the list of services for a given facility. One query was fetching all of the existing services and another was fetching the relevant services. The second query was relying on a value from the first query which could be empty. Fortunately the value in question (selected facility nid) was already being passed into the function.

Closes to #21301

### Generated description

This pull request includes a small but important change to the `function _va_gov_facilities_get_vamc_system_service_names` in the `va_gov_facilities.module` file. The change updates the query parameter to use the `selected_facility_id` instead of the `existing_facility_services` array, ensuring the correct facility ID is referenced in the query.

## Testing done
Tested locally

## QA steps

 - [x] Log in on staging as [Tyler.Moore3@va.gov](mailto:Tyler.Moore3@va.gov) , password drupal8
 - [x] Go to https://pr21310-bp6z95boirhla2ikttiasdopenwribsh.ci.cms.va.gov/node/add/health_care_local_health_service
 - [x] Select section "VA Eastern Colorado" (listed under VISN 19)
 - [x] Go to the "Health service and facility basic info" dropdown
 - [x] Try to create any new health services for the Academy VA Clinic or the Evans VA Clinic.
 - [x] Verify that you can create new health services
 - [x] Verify that those new services are properly excluded form the drop down if you go to add another new service

### Definition of Done

### Select Team for PR review

- [x] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

### DIFF
--- a/docroot/modules/custom/va_gov_facilities/va_gov_facilities.module
+++ b/docroot/modules/custom/va_gov_facilities/va_gov_facilities.module
@@ -156,9 +156,9 @@ function _va_gov_facilities_get_vamc_system_service_names(int $selected_facility
     WHERE nfa.bundle = 'regional_health_care_service_des'
     AND nfa.field_administration_target_id =
       (SELECT nfa.field_administration_target_id FROM node__field_administration as nfa
-      WHERE nfa.entity_id = :existing_service_id)",
+      WHERE nfa.entity_id = :selected_facility_id)",
     [
-      ':existing_service_id' => $existing_facility_services[0]->field_regional_health_service_target_id,
+      ':selected_facility_id' => $selected_facility_id,
     ]
     );
   $none_option = [


docroot/modules/custom/va_gov_facilities/va_gov_facilities.module
```diff
@@ -156,9 +156,9 @@ function _va_gov_facilities_get_vamc_system_service_names(int $selected_facility
     WHERE nfa.bundle = 'regional_health_care_service_des'
     AND nfa.field_administration_target_id =
       (SELECT nfa.field_administration_target_id FROM node__field_administration as nfa
-      WHERE nfa.entity_id = :existing_service_id)",
+      WHERE nfa.entity_id = :selected_facility_id)",
     [
-      ':existing_service_id' => $existing_facility_services[0]->field_regional_health_service_target_id,
```
brianseek, line 161: This was failing because $existing_facility_services was empty since the facility in question had no existing services. But also we don't need to fetch this value off of that return...because we already know the value. 

It's been a minute since I've done heavy Drupal, but do we not like entityQuery anymore?

timcosgrove: entityQuery doesn't really allow for joins, so you'd need to do at least two queries here. That doesn't necessarily mean it's the wrong thing to do, but, that might be why someone would reach for straight sql.

brianseek: Oh that makes sense. I'm not positive the join is really needed anyway but I don't understand all the ins and outs of these rules.

brianseek: oh I think I get it...jeez I really hate reading SQL

edmund-dunn (approved):
Everything works as expected. Nice work and sleuthing!

PR #21319, opened 2025-05-07
Disable workflow-on-push for build docs.

timcosgrove:
## Description

This just disables automatic run of the build docs workflow for the time being, since the version on `main` is not ready.


brianseek (approved):
👍

PR #21317, opened 2025-05-07
Vagov team 108941 intro page

tjheffner:
## Description
Redo of https://github.com/department-of-veterans-affairs/va.gov-cms/pull/21305, I butchered the rebase.

Relates to https://github.com/department-of-veterans-affairs/va.gov-team/issues/108941

## Testing done

- Verified field data persisted on existing forms.
- Verified data updated on save and persisted. 
- Verified access control.
- Verified status on form layout page tracked properly

## Screenshots
Screenshot 2025-05-06 at 12 09 58 PM

## QA steps
- visit form builder, employment questionnaire. see status as complete
- open intro page, see field values present. 
- update values, save. should be kicked to form layout page
- return to intro page, see updated values

- create a new form via form builder ui
- intro page status should be `incomplete`
- click on intro page, fill values and save
- status should now show complete on form layout page

PR #21345, opened 2025-05-12
Revert "Vagov team 108941 intro page"

edmund-dunn:
Reverts department-of-veterans-affairs/va.gov-cms#21317

Reerting this because of failing phpunit tests.


omahane (approved):
Looks like a complete reversion. Go PHPUnit and pass once more.

PR #22662, opened 2025-10-31
Revert "Bump @babel/runtime from 7.16.7 to 7.28.4 in the npm_and_yarn group across 1 directory"

omahane:
Reverts department-of-veterans-affairs/va.gov-cms#22367


omahane:
This is not the way.

PR #22686, opened 2025-11-03
[docs] Documentation fork update

timcosgrove:
## Description

Quick update to docs. 
- slightly more clarity on not using forks
- remove link to Codespaces (actual instructions are kept but marked historical)


omahane (approved):
Good idea to deprecate and/or these instructions.

Issue #22672, opened 2025-11-03
va.gov-cms: Prep repo for opening back up for production releases


laflannery:
## Description
We are opening up the va.gov-cms repo and releasing to production again. This is process to set this up and allow production releases.

### Steps for Implementation

1. Merge the integration branch into main
   - This has not yet been started and there may be issues or conflicts to work out between the 2 branches
2. Modify CMS Prod deploy Jenkins job to only fire on Mondays and Thursdays
   - This is a code change
   - We can make and merge this now, ahead of the first scheduled Production release on Thursday 11/6 
      - The Jenkins job is disabled so doing this ahead of time will not make this fire immediately

### Acceptance Criteria
- [x] The integration branch is fully merged into main
- [x] CMS Prod deploy Jenkins job has been changed to only fire on Mondays and Thursdays

PR #22688, opened 2025-11-03
VACMS-22672: merge integration into main

timcosgrove:
## Description

https://github.com/department-of-veterans-affairs/va.gov-cms/issues/22672

This merges the current state of the integration branch into main with the intention of shifting back to `main` as our target branch for merge.


omahane (changes requested):
Thanks to Copilot, I can confirm we should remove the buildForm method signature change.

Issue #22938, opened 2025-12-10
Update Vet Center Flagged Facilities runbooks with current steps and team assignments


kellylein:
## User Story or Problem Statement
The following Vet Center facility runbook templates may contain outdated or incorrect content, and need to be reviewed and potentially revised:

1. - [ ] [New Vet Center runbook](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/abb85447c64e1bbb10e92097807a556cb2ca182c/.github/ISSUE_TEMPLATE/runbook-vet-center-new.md) - Priority, comes in via Flagged Facilities listing
2. - [ ] [Vet Center Name Change runbook](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/abb85447c64e1bbb10e92097807a556cb2ca182c/.github/ISSUE_TEMPLATE/runbook-vet-center-name-change.md) - Priority, comes in via Flagged Facilities listing
3. - [ ] [Closed Vet Center runbook](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/abb85447c64e1bbb10e92097807a556cb2ca182c/.github/ISSUE_TEMPLATE/runbook-vet-venter-closed.md) - Priority, comes in via Flagged Facilities listing
4.  - [ ] [runbook-vet-center-cap-to-outstation.md](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/abb85447c64e1bbb10e92097807a556cb2ca182c/.github/ISSUE_TEMPLATE/runbook-vet-center-cap-to-outstation.md) - Does not come in via Flagged Facilities listing, CAPs are not listed and do not have a CMS help desk/editor/Facilities team comms process. Added link to Mural.
5.  - [ ] [runbook-vet-center-outstation-to-vet-center.md](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/abb85447c64e1bbb10e92097807a556cb2ca182c/.github/ISSUE_TEMPLATE/runbook-vet-center-outstation-to-vet-center.md) - Does not come in via Flagged Facilities listing, CAPs are not listed and do not have a CMS help desk/editor/Facilities team comms process. Also added link to Mural.

**Edit:** Removed extraneous runbooks that do not come in as Flagged Facilities and pertain to extremely rare situations. All VAMC Flagged Facilities runbooks are complete.

## Description or Additional Context 
[...]

PR #23294, opened 2026-01-22
Create runbook-vet-center-new-outstation.md

stefaniefgray:
Added per https://github.com/department-of-veterans-affairs/va.gov-cms/issues/22938

Old "New Vet Center" runbook had MVC and Outstation steps mixed in with regular Vet Center steps -- this breaks out new Outstation runbook


pamelaabanda6-va (approved):
Looks all good to me!

PR #23297, opened 2026-01-22
Create runbook-vet-center-closed-mvc-or-outstation.md

stefaniefgray:
Added per https://github.com/department-of-veterans-affairs/va.gov-cms/issues/22938

Old "Closed Vet Center" runbook had MVC and Outstation steps mixed in with regular Vet Center steps -- this breaks out MVC and Outstation steps


omahane (changes requested):
One wee change.

PR #23480, opened 2026-02-13
Remove unneeded `FEATURE_NEXT_BUILD_CONTENT_OUTREACH_ASSET` flag

pwolfert:
## Description

This content type doesn't actually represent its own page. The only place where this node is used is in the `/outreach-and-events/outreach-materials/` page, and that has its own template with its own feature flag.

## Testing done

I went to the https://va-gov-cms.ddev.site/admin/config/system/feature_toggle on my local server (after running `ddev drush cim` to re-import configs (thanks, Tim!)) and looked for it and verified that it's not there anymore.

## QA steps

Just make sure the flag isn't there anymore. It doesn't hurt anything. It's just confusing to us and unnecessary.


brianseek (approved):
🪓

Issue #21619, opened 2025-06-26
Disable Contextual Links from Taxonomy Template


TroyCMSSupport:
## Describe the defect
Content Admins are currently able to remove national service, but if misused, can cause problems with publishing content.

## To Reproduce
1. Log in to CMS as a Content Admin
2. Go to [Lakeland Vet Center page](https://prod.cms.va.gov/lakeland-vet-center)
3. Scroll down to any of the Services listed on the page
4. Hover the mouse cursor just above the "?" on the right side of the service
5. See the Contextual Edit button (pencil)
6. Click on the 'Pencil' to see the "edit" and "delete" options

## AC / Expected behavior
Content Admins are able to edit and remove nationally-managed content BUT SHOULD NOT BE ABLE TO.

## Screenshots
<img width="1039" alt="Image" src="https://github.com/user-attachments/assets/395a533a-cff8-47ee-b6f9-0fb163ac5799" />

 ## Additional context
Note that CMS Editors do not have the ability to see/use this tool - this applies only to administrator level accounts.

 ## Acceptance Criteria
- [ ] Contextual links are removed from taxonomy template

PR #23663, opened 2026-03-06
Conditionally remove edit button

zachbutton:
## Description

Closes https://github.com/department-of-veterans-affairs/va.gov-cms/issues/21619

Removes the edit button depending on taxonomy type.

This change is scoped to the taxonomy term view displays that are marked as nationally managed (they include the “Why can’t I edit this?” national-service tooltip text), to avoid affecting unrelated taxonomy contexts.

- [`core.entity_view_display.taxonomy_term.health_care_service_taxonomy.vamc_facility_service.yml`](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/main/config/sync/core.entity_view_display.taxonomy_term.health_care_service_taxonomy.vamc_facility_service.yml)
- [`core.entity_view_display.taxonomy_term.health_care_service_taxonomy.vba_facility_service.yml`](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/main/config/sync/core.entity_view_display.taxonomy_term.health_care_service_taxonomy.vba_facility_service.yml)
- [`core.entity_view_display.taxonomy_term.health_care_service_taxonomy.vet_center_service.yml`](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/main/config/sync/core.entity_view_display.taxonomy_term.health_care_service_taxonomy.vet_center_service.yml)

## Testing done

https://va-gov-cms.ddev.site/lakeland-vet-center
- Tested that edit button is removed on the services mentioned in the ticket

https://va-gov-cms.ddev.site/help/how-to-use-the-pdf-media-library
- Tested that edit button is still present

### DIFF
--- a/docroot/themes/custom/vagovclaro/templates/layout/taxonomy-term--products.html.twig
+++ b/docroot/themes/custom/vagovclaro/templates/layout/taxonomy-term--products.html.twig
@@ -29,9 +29,6 @@
     'vocabulary-' ~ term.bundle|clean_class,
   ]
 %}
-{%
-  set hide_contextual_links = term.bundle == 'health_care_service_taxonomy' and view_mode in ['vamc_facility_service', 'vba_facility_service', 'vet_center_service']
-%}
 <div{{ attributes.setAttribute('id', 'taxonomy-term-' ~ term.id).addClass(classes) }}>
   {% if not hide_contextual_links %}
     {{ title_prefix }}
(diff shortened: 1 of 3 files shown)


docroot/themes/custom/vagovclaro/templates/layout/taxonomy-term--products.html.twig
```diff
@@ -29,9 +29,14 @@
     'vocabulary-' ~ term.bundle|clean_class,
   ]
 %}
+{%
```
omahane: CONSIDER: A possibly better place to do this would be in the .theme file where we already have some alteration happening. 

One thing about this right now is that it doesn't account for Drupal admins, for whom we are also hiding the edit button, whether we should or not.

omahane (approved):
Works great and is future proofed for when we want to show either VAMC or VBA services in the same way. (I believe we are only doing so currently with Vet Centers.)

I had one comment to consider, but it's very minor.

PR #23992, opened 2026-04-10
VACMS-23750: tmgmt update

edmund-dunn:
## Description

Relates to #23750 

## QA steps

If this passes automated testing, it should be good

### Select Team for PR review

- [x] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


omahane (approved):
MGMT would be proud!

PR #24168, opened 2026-04-27
Update dependencies to fix cypress on eks

edmund-dunn:
## Description

```
Oops...we found an error preparing this test file:
tests/cypress/integration/features/api/news_story.feature
The error was:
TypeError: Cannot read properties of undefined (reading 'sync')
at /opt/drupal/node_modules/@badeball/cypress-configuration/dist/cypress-configuration/cypress-post10-configuration.js:239:65
at Array.flatMap (<anonymous>:null:null)
at resolvePost10TestFiles (/opt/drupal/node_modules/@badeball/cypress-configuration/dist/cypress-configuration/cypress-post10-configuration.js:239:25)
at resolveTestFiles (/opt/drupal/node_modules/@badeball/cypress-configuration/dist/cypress-configuration/index.js:79:74)
at compile (/opt/drupal/node_modules/@badeball/cypress-cucumber-preprocessor/dist/template.js:36:153)
at Transform.flush [as _flush] (/opt/drupal/node_modules/@badeball/cypress-cucumber-preprocessor/dist/subpath-entrypoints/browserify.js:25:57)
at Transform.final [as _final] (node:internal/streams/transform:132:10)
at callFinal (node:internal/streams/writable:698:12)
at prefinish (node:internal/streams/writable:710:7)
at finishMaybe (node:internal/streams/writable:720:5)
at Transform.Writable.end (node:internal/streams/writable:634:5)
at DuplexWrapper.<anonymous> (/opt/drupal/node_modules/duplexer2/index.js:29:14)
at Object.onceWrapper (node:events:631:28)
at DuplexWrapper.emit (node:events:529:35)
at finishMaybe (/opt/drupal/node_modules/readable-stream/lib/_stream_writable.js:630:14)
at afterWrite (/opt/drupal/node_modules/readable-stream/lib/_stream_writable.js:492:3)
at onwrite (/opt/drupal/node_modules/readable-stream/lib/_stream_writable.js:483:7)
at WritableState.onwrite (/opt/drupal/node_modules/readable-stream/lib/_stream_writable.js:180:5)
at afterWrite (node:internal/streams/writable:500:5)
at afterWriteTick (node:internal/streams/writable:487:10)
at processTicksAndRejections (node:internal/process/task_queues:81:21)
This occurred while Cypress was compiling and bundling your test code. This is usually caused by:</anonymous></anonymous>
A missing file or dependency
A syntax error in the file or one of its dependencies
Fix the error in your code and re-run your tests.
```

The error is caused by the glob override in package.json. The override forces glob to ^10.5.0, but glob v10 changed its API — glob.sync no longer exists as a property on the default export (it was replaced with a named globSync export). The @badeball/cypress-configuration package (a dependency of @badeball/cypress-cucumber-preprocessor) relies on the older glob.sync() API, so the override breaks it.

Fix: Remove the glob override from package.json.

The fix is straightforward: I removed the "glob": "^10.5.0" override from the overrides section in package.json.

Why this works: The @badeball/cypress-configuration package (used internally by @badeball/cypress-cucumber-preprocessor) calls glob.sync(), which is the API from glob v7–v9. The override was forcing all nested glob dependencies to resolve to v10+, where glob.sync was removed from the default export (replaced by a standalone globSync function). By removing the override, the @badeball/cypress-cucumber-preprocessor dependency tree can resolve to the glob version it actually needs.

### Generated description

This pull request makes a minor update to the `package.json` dependencies by removing the unused `glob` package. This helps keep the dependency list clean and reduces potential security risks.

- Removed the `glob` dependency from `package.json` as it is no longer needed.

## QA steps

If testing passes we should be good.

### Select Team for PR review

- [x] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`


pamelaabanda6-va (approved):
lgtm!

PR #24193, opened 2026-04-30
VACMS-24184: Refactor tag resolution logic in build-drupal-tag-container.yml for improved clarity and validation

edmund-dunn:
## Description

Relates to #24184 

### Generated description

This pull request updates the tag selection logic in the `build-drupal-tag-container` GitHub Actions workflow to improve how tags are chosen and validated when building containers. The main improvement is a clearer and more robust priority order for determining which tag to use, as well as more explicit handling of tags pushed directly to the repository.

**Tag selection and validation improvements:**

* Refactored the tag selection logic to prioritize: (1) explicit workflow input, (2) a pushed tag ref, and (3) the latest semver tag, making the process more predictable and maintainable.
* Added explicit handling for cases where a tag is pushed directly (`GITHUB_REF_TYPE=tag`), ensuring the pushed tag is validated and used if it matches the semver pattern.
* Removed redundant and now-unreachable code for tag validation, simplifying the workflow logic and reducing duplication.

## QA steps

If this passes automated testing it should be good.

### Select Team for PR review

- [x] `CMS Team`
- [ ] `Public websites`
- [ ] `Facilities`
- [ ] `User support`

- [ ] `DO NOT MERGE`

- [ ] `Needs PO review`

### DIFF
--- a/.github/workflows/build-drupal-tag-container.yml
+++ b/.github/workflows/build-drupal-tag-container.yml
@@ -76,50 +76,30 @@ jobs:
             } | sort -V -r | head -n1
           }
 
-          EVENT='${{ github.event_name }}'
           INPUT_TAG='${{ github.event.inputs.tag || inputs.tag }}'
-          if [ "$EVENT" = "workflow_dispatch" ] || [ "$EVENT" = "workflow_call" ]; then
-            if [ -n "$INPUT_TAG" ]; then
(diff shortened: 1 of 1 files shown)


omahane (approved):
I don't see any issue with this. Looks like a good refactor.

